PR #110, opened 2015-07-04
broke apart smoothness and smallness terms

lheagy:
@sgkang, @grosenkj : is this more clear?


grosenkj:
This is a cleaner way to implement this. Nice job.

PR #118, opened 2015-10-28
Add Cython dependency

bsmithyman:
This should really be dependent on Cython, if we're going to call `cythonize` directly in the build script, and if Cython is needed for good performance. Or, if not, we would need to make the import of `cythonize` conditional, and then skip building the extension if Cython isn't installed. Alternatively, don't build the .pyx file, and check in a compiled .c file; then make the following changes to setup.py:

- replace `from Cython.Build import cythonize` on Line 10 with `from setuptools.extension import Extension`
- replace `ext_modules = cythonize('SimPEG/Utils/interputils_cython.pyx')` on Line 53 with `ext_modules = [Extension('interputils_cython', sources=['SimPEG/Utils/interputils_cython.c'])]`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,8 @@ setup(
     version = "0.1.3",
     packages = find_packages(),
     install_requires = ['numpy>=1.7',
-                        'scipy>=0.13'
+                        'scipy>=0.13',
+                        'Cython'
                        ],
     author = "Rowan Cockett",
     author_email = "rowan@3ptscience.com",


rowanc1:
Do you know what the standard in the community is wrt checking in the compiled code?

bsmithyman:
Yeah, good question, I'm not sure. To be fair, it's not compiled to binary code, it's compiled/templated to C, so it should still sit in the Git repository cleanly, just as a derived file. Kind of like checking in compiled documentation, vs. doxygen source. In principle, it's all cpython calls, so it should be portable across cpython installations (but not, for example, pypy or jython). It would add a step to generating the source if you change the pyx file, but other than that I suppose there's no real disadvantage. Perhaps there's a way to have it auto-generated with a Git hook? Also, it could be done only on releases (to lose the Cython dependency w/ pip installations), and keep the original in the Git tree. We could write a "make_release" script to handle duplicating things to a .gitignore'd subdirectory and changing the setup file. All just thoughts, though.

PR #165, opened 2015-11-10
Be compatible with outside metaclasses, for now

bsmithyman:
Making this change would make SimPEG more compatible with multiple inheritance in outside code. Metaclass inheritance is tricky, and for the moment the SimPEG metaclass doesn't really do anything. Setting `SimPEGMetaClass = type` preserves all the existing functionality in SimPEG code, and plays a bit better with outside multiple inheritance involving metaclasses. In the future, this may need to be addressed in more detail.

### DIFF
--- a/SimPEG/Utils/codeutils.py
+++ b/SimPEG/Utils/codeutils.py
@@ -3,10 +3,7 @@ import time
 import numpy as np
 from functools import wraps
 
-
-class SimPEGMetaClass(type):
-    def __new__(cls, name, bases, attrs):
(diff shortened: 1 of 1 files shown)


lheagy:
The travis fail is probably a poor choice in random seed for the coordinate test, I fix that!

bsmithyman:
Sure, appreciate it. I talked with Rowan about this a bit, and we figured this is the easiest way to reserve the right for SimPEG to use this later, but avoid breaking things in the meantime. Longish-term, if we manage to have the seismic stuff grow together with this it should be straightforward to resolve any conflicts, anyway.

PR #176, opened 2015-11-24
Layered Model

rowanc1:
added model builder to create layered model. 
untested and no error checking yet


rowanc1:
This is a single layered model. The model builder needs to be more rigourously tested, but that can be done later.

PR #179, opened 2015-11-25
Em/dev

lheagy:
zero, identity updates in simpegEM


rowanc1:
@lheagy I think that we should merge into dev?

The pull request before this will bring dev up to speed with master, can we close this and wait till the dev branch is ready?

RTD now builds the EM/DEV branch.

Issue #185, opened 2015-11-25
SimPEG import error


grosenkj:
I updated to the current version of master. 

```python
import SimPEG 
```
Fails with an error
```python
ImportError: No module named TreeUtils
```
Looking at the code, I found TreeUtils.pyx. Quick googling and I found that .pyx are Cython related files. I ran 
```python
import pyximport
pyximport.install()
```
and import SimPEG works without an ImportError. 

The additional Cython import should be added to the SimPEG code base.

rowanc1:
There is a pull request for this. That will fix the no import symptoms. But you should really be building the extensions ahead of time:

```
python setup.py build_ext --inplace
```

grosenkj:
Imports without problems after building the extensions.

Issue #183, opened 2015-11-25
need warning / error handling for TreeUtils


lheagy:
If setup.py not run, the current warnings are unhelpful

(do we also have a python version that can be called as a backup if calling the cython stuff did not work?)

rowanc1:
#185

PR #187, opened 2015-11-25
Addresses #183: Import errors for tree mesh

rowanc1:
#185 

This will yell at you if you try to create the TreeMesh without first running the `setup.py`


grosenkj:
Good to have SimPEG yell at you if you haven't run setup.py. The error is helpful.

rowanc1:
:) novel

PR #191, opened 2015-11-28
Meshless Identity Map, Regularization for Active Cell models 

### LINKED ISSUE #78: Active Cell Mapping bug.
If we have a mapping of:

```
S*(P*m + v)
```

Where S is some weighting, P is a projection up to the full domain, and v is the background this will NOT work for the regularization. Here we are multiplying by the difference: (m - mref) and the v vector is in the wrong place.

This could either be fixed by changing the regularization to map both m and mref up separately and then subtract them, but I think that is confusing.

Also, really the active cell mapping should have a background of **zero** rather than the background conductivity... which circumvents this issue all together.

At any rate, the active cell mapping is not really a linear operator...and could lead to some difficulties in usage.

lheagy:
- Meshlesses Identity Map (takes nP instead of a mesh)
- Tikhonov regularization if active cells are used (don't take derivs across interfaces between active cells and not)
- testing improvements: test 1D, 2D, 3D on a random tensor mesh , also test that for a constant mref, phi_m(ref) = 0

(closes #78)

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -118,6 +118,78 @@ class IdentityMap(object):
     def __str__(self):
         return "%s(%s,%s)" % (self.__class__.__name__, self.shape[0], self.shape[1])
 
+
+class IdentityMap_Meshless(IdentityMap):
+
+    def __init__(self, nP=None, **kwargs):
+        IdentityMap.__init__(self, None, **kwargs)
+        self._nP = nP
+
+    @property
+    def nP(self):
+        """
+            :rtype: int
+            :return: number of parameters in the model
+        """
+        if self._nP is None:
+            return '*'
+        return self._nP
+
+    @property
+    def shape(self):
+        """
+            The default shape is (mesh.nC, nP).
+
+            :rtype: (int,int)
+            :return: shape of the operator as a tuple
+        """
+        if self._nP is None:
+            return ('*', '*')
+        return (self.nP, self.nP)
+
+
+    def _transform(self, m):
+        """
+            Changes the model into the physical property.
+
+            .. note::
+
+                This can be called by the __mul__ property against a numpy.ndarray.
+
+            :param numpy.array m: model
+            :rtype: numpy.array
+            :return: transformed model
+
+        """
+        return m
+
+    def inverse(self, D):
+        """
+            Changes the physical property into the model.
+
+            .. note::
+
+                The *transformInverse* may not be easy to create in general.
+
+            :param numpy.array D: physical property
+            :rtype: numpy.array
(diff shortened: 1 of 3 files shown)


SimPEG/Maps.py
```diff
@@ -118,6 +118,78 @@ def __mul__(self, val):
     def __str__(self):
         return "%s(%s,%s)" % (self.__class__.__name__, self.shape[0], self.shape[1])
 
+
+class IdentityMap_Meshless(IdentityMap):
```
lheagy: or IdentityMap_NoMesh ?

lheagy:
@grosenkj : what do you think of this?

grosenkj:
@lheagy : I think this looks very nice, I am for it being merged in.

lheagy:
@rowanc1 , @sgkang : good to go??

SimPEG/Maps.py
rowanc1: Can we put three extra if statements instead of a new class?

```
if type(self.mesh) is int:
    self._nP = self.mesh
    self.mesh = None
```

Could change the input to `mesh_or_nP`?

rowanc1:
Can we get rid of the _Meshless class and just make that default behaviour in the Identity?

```
eg = IdentityMap(5)
```

instead of _meshless?

rowanc1:
@lheagy I have updated this functionality. Let me know what you think, after the tests are run we can merge?

lheagy:
Nice, I think this looks good.

Is there any reason that we should only take the mesh or nP (ie. not allow you to give both?). I can't think of a case off-hand where you would need / want both, but maybe there is no harm in specifying both?

rowanc1:
I was just thinking of this. It actually might need some tweaking to error
if you specify both, because we don't actually support this functionality.

lheagy:
would you want both in the case of an active cell model?

lheagy:
TODO: add a test to look at the gradient when there is topography for a half-space (should be zero!)

rowanc1:
This will conflict with regularization in PR #214. @lheagy is this good to merge in?

lheagy:
I think it is good to go. (still need to give some thought to taking both or only one of nP, mesh. there may be cases where we do want to give both --> for dealing with topo? but this can be handled in an issue)

lheagy:
It looks like these changes got over-written somewhere down the line...

rowanc1:
It looks as if they got to dev?

https://github.com/simpeg/simpeg/blame/dev/SimPEG/Maps.py#L15

lheagy:
ah, I was confused about what should be in regularization.py

PR #192, opened 2015-12-04
SimPEG examples

rowanc1:
I have created a number of examples that get run everytime. This should be helpful for on-boarding people a bit faster. Fee free to contribute more examples!

I have not tested these heavily on the RTFD servers, so I am not sure if they will pass. We might have to do some work to make the images on travis and then send them over to RTFD.

I have also updated the plotting code in the tree mesh slightly.


lheagy:
can we merge this one?

rowanc1:
Probably!

rowanc1:
The tests were not being found on the examples, hence the drop in coverage. There were also problems in the rtfd.

rowanc1:
The examples now seem to be being tested, and also work on RTFD.

http://simpeg.readthedocs.org/en/feat-examples/examples/Mesh_Operators_CahnHilliard.html

rowanc1:
Test coverage has fell because of the examples (where much of the code is visualization, and is not covered with `plotIt=False`). I will look into this testing in the future, but will merge this for now.

Issue #162, opened 2015-11-05
pymatsolver on travis


lheagy:
to speed up solves on travis

bsmithyman:
This would be great, and it's also essentially the environment my JupyterHub container runs, so I'd be happy to help. I've had mixed luck getting it to work on Linux, though.

bsmithyman:
Regarding this: I'm able to compile on my Ubuntu-based infrastructure, which is roughly what Travis is running, but by default it's trying to link in the MPI version of MUMPS. Do we have an idea as to which packages should (or should not) be installed to make sure that the serial MUMPS version is linked when building pymatsolver.MumpsSolver?

lheagy:
@grosenkj figured a lot of this out a couple months ago to get serial mumps working, he may be able to give us a hand!?

PR #193, opened 2016-01-04
Speed up the tests with pymatsolver.

rowanc1:
This should address #162

I have put a bunch of work into pymatsolver to get it working on ubuntu and mac.

https://github.com/rowanc1/pymatsolver


rowanc1:
This has sped up the test time from 1:33 to 30 min. Not bad!

PR #205, opened 2016-01-15
Update and rename EM_FDEM_1D_Inversion.py to EM_TDEM_1D_Inversion.py

lheagy:
- this is a TDEM inversion (not FDEM)


lheagy:
mind if I merge this directly into master? or should it go onto dev first?

Issue #212, opened 2016-01-28
Mesh IO


rowanc1:
I would propose the file structure be replaced:

``` python
#----- replace ------
Utils.readUBCTensorMesh('myMesh.msh')
#----- with ------
M = SimPEG.Mesh.TensorMesh.readUBC('myMesh.msh')
v  = M.readModelUBC('model.dat')
```

This works well with the various types of meshes as well.

``` python
M.writeUBC('myNewMesh.msh')
M.writeModelUBC('myModel.dat', v)
```

I think that would clean up the syntax a bit, and allow these IO formats to live in the correct mesh file, but be used the same.

Similar:
`writeVTK`

rowanc1:
@grosenkj the VTK IO depends on vtk libs? I will try to update travis and merge in those from treeMeshCounting.

grosenkj:
I like this, makes sense. 

We should maybe consider having:

```python
M.writeUBC('myNewMesh.msh',model=None)
```
when model isn't given it just writes the mesh. Model should be a dict of ```{'model.dat':v}```. Same would go for writeVTK. 

What do you think?

grosenkj:
Yes, vtkIO depends on VTK. 

Awesome...

rowanc1:
I quite like that.
I would propose a wrapper function:

```python
def writeModelUBC(self):
    pass

@meshWriter(writeModelUBC)
def writeFORMAT(self, meshName):
    pass
```

Where the meshWriter:

```python
def meshWriter(modelWriter):
    def meshWrapper(func):
        @wraps(func)
        def wrapper(self, meshName, models=None):
            func(self, meshName)
            if models is None: return
            assert type(models) is dict, 'models must be a dict'
            for key in models:
                assert type(key) is str, 'The dict key is a file name'
                modelWriter(self, key, models[key])
        return wrapper
    return meshWrapper
```

rowanc1:
Decided to not go with the meshWriter wrapper.

rowanc1:
@grosenkj wants the feature of utils depreciated:

Please put this back in the Utils class.

rowanc1:
```python
readUBCocTreeFiles(meshFile,modelFiles=None)
readUBCTensorMesh(fileName)
readUBCTensorModel(fileName, mesh)
writeUBCTensorMesh(fileName, mesh)
writeUBCTensorModel(fileName, mesh, model)
writeUBCocTreeFiles(fileName,mesh,modelDict=None)
[...]

PR #213, opened 2016-01-28
Tree Mesh Updates

rowanc1:
Includes access to the treemesh through `__getitem__`.
Adds an inversion directive.


rowanc1:
We may want to address #212 before pushing to dev.

rowanc1:
@grosenkj can you please check the mesh IO before we merge this in?

Issue #202, opened 2016-01-15
Using the DataClass


rowanc1:
We should be more about the data class.

```
from SimPEG import DC
dobs = DC.DCData.fromFile('myData.dat')
survey = dobs.survey
dobs.plotSection()
dobs.std[5] = notTrustWorthy
```

The Std should live with the data specifically (not the survey). The plotting functions for the data should be associated with the data class.

lheagy:
also `dobs.eps` or `dobs.floor` and maybe `dobs.flag` (ie the flag would be `notTrustWorth` and `std` a number

grosenkj:
@lheagy, I like your idea. The idea of having the `notTrustWorth` flag could come in handy. It could be an independent flag or be addressed by setting `std` to ``NaN` or `False` to indicate it's untrustworthyness. I could vote either way, but at the end of the day it won't matter how it will be implemented but keeping things consistent will be nice.

rowanc1:
Data is created by:

```python
d = Data(survey, v=vectorData, std=stdVector)
# You can access the data by source and receiver pairs:
assert isinstance(d[Src,Rx], np.ndarray)
# it will also have a std property
d.std[Src, Rx]
# which is really just another data object
d.floor # is the epsilon value used to not go crazy in the regularization.
#for any of these you can call the 
Utils.mkvc(d)
#to get back to a vector.
```

```python
dobs[Src,Rx] == np.ndarray
dobs.std[Src, Rx] == np.ndarray
dobs.floor[Src, Rx] == np.ndarray
```

lheagy:
Wd should not be making assumptions about the noise model - it should ask the data class what the uncertainties are and simply do `Wd = 1/data.uncertainty`

Instead of calling the uncertainty percentage `std` (which is protected by numpy and generally confusing) how about having uncertainty as a simple class that is then attached to the data class
(Base uncertainty class) 
- `percentage` (or maybe `standard_deviation`) 
- `floor` 
- `eval` : `return self.standard_deviation * np.absolute(data) + self.floor`

both `percentage` and `floor` have default values 

(I don't really like `percentage`, maybe `standard_deviation` is better)

grosenkj:
Very similar to issue #184.

I support this change, I be nice to have better access to the uncertainties via the `Data` class.

lheagy:
This is closely related to the changes being made in #562. I have sketched out a first pass implementation below. It lives on the `ref/simulation` branch at the moment. Some important changes:
[...]

PR #217, opened 2016-01-29
Em/inversioncleanup

lheagy:
- let survey have both std and eps (percentage and a floor, respectively)
- if survey has std and eps, use these in the definition of the data misfit
- use `u` instead of `f` for fields in Jvec, Jtvec to be consistent with the rest of SimPEG
- add a 1D FDEM example inversion
- use survey.getSyntheticData in both FDEM and TDEM examples
- added a transpose property to the zero class (with test) 
- from SimPEG.EM import mu_0


SimPEG/DataMisfit.py
```diff
@@ -103,10 +89,18 @@ def Wd(self):
...
             survey = self.survey
-            eps = np.linalg.norm(Utils.mkvc(survey.dobs),2)*1e-5
-            self._Wd = Utils.sdiag(1/(abs(survey.dobs)*survey.std+eps))
+
+            if getattr(survey,'std', None) is None:
+                print 'SimPEG.DataMisfit.l2_DataMisfit assigning default std of 5%'
+                survey.std = 0.05
+
+            if getattr(survey, 'eps', None) is None:
+                print 'SimPEG.DataMisfit.l2_DataMisfit assigning default eps of 1e-5 * ||dobs||'
```
rowanc1: yay comments!

SimPEG/Survey.py
```diff
@@ -205,6 +205,7 @@ class BaseSurvey(object):
     __metaclass__ = Utils.SimPEGMetaClass
 
     std = None       #: Estimated Standard Deviations
+    eps = None       #: Estimated Noise Floor
```
rowanc1: I suppose these would be better in the data class? But here is a good place for them now.

lheagy:
@grosenkj , @rowanc1 , @sgkang : can you guys have a look at what I did, in particular wrt the survey having a std and eps. I think eventually this should be tied to the data, but for now, at least it will be used when we define Wd.

grosenkj:
It makes more sense to me to have std and eps in the Data class, rather then survey. I like the suggestions in issue #202, since it is both the floor and uncert has to do with the data not survey per say. 

Would it make sense to restrict survey.dobs to be a SimPEG data class, rather than a numpy array?

lheagy:
@grosenkj : agreed. I think we want to address that in a separate pr though? That one will take a little bit of thought

rowanc1:
This looks good to me.

rowanc1:
(as in, there is a lot more thought involved in the data/dobs/std/eps conversation, outside the scope of this PR)

PR #227, opened 2016-02-02
Dev

lheagy:
- mappings can take nP or a mesh
- meshIO

- data misfit looks at survey for std, eps
- FDEM.py uses u instead of f for fields
- added 1D FDEM inversion example  
- use survey.makeSyntheticData for 1D FDEM, TDEM examples
- Zero has a transpose
- mu_0, epsilon_0 accessible from EM (EM.mu_0, EM.epsilon_0)


SimPEG/Utils/matutils.py
```diff
@@ -37,6 +40,9 @@ def mkvc(x, numDims=1):
 
 def sdiag(h):
     """Sparse diagonal matrix"""
+    if isinstance(h, Zero):
+        return h
```
lheagy: this should return a new instance of Zero(), not h?

rowanc1: I guess it only matters if zero has a shape to it, which I don't think we have yet..?

lheagy: right, it purposely doesn't have a shape, but is this potentially confusing to read?

SimPEG/Examples/EM_FDEM_1D_Inversion.py
```diff
@@ -0,0 +1,116 @@
+from SimPEG import *
```
grosenkj: We shouldn't do this. Much rather
```python
import SimPEG [as something]
or
from SimPEG import Mesh, Maps, ...etc
```
Makes things easier to follow.

grosenkj:
I think this looks good. Just made a general comment on the import of SimPEG in examples, see Issue #235

PR #228, opened 2016-02-03
Example: DC PseudoSection Simulation

fourndo:
Add DC_Pseudo_Section example.

### DIFF
--- a/SimPEG/Examples/DC_PseudoSection_Simulation.py
+++ b/SimPEG/Examples/DC_PseudoSection_Simulation.py
@@ -16,6 +16,43 @@ def run(loc=np.c_[[-50.,0.,-50.],[50.,0.,-50.]], sig=np.r_[1e-2,1e-1,1e-3], radi
     @fourndo
     """
     
+    def getIndicesSphere(center,radius,ccMesh):
+        """
+            Creates a vector containing the sphere indices in the cell centers mesh.
+            Returns a tuple
+    
+            The sphere is defined by the points
+    
+            p0, describe the position of the center of the cell
+    
+            r, describe the radius of the sphere.
+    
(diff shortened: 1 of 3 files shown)


lheagy:
for this example to run natively in SimPEG, we need to merge in simpegDCIP into simpeg. @rowanc1 @sgkang, can we do this soon!?

SimPEG/Examples/DC_PseudoSection_Simulation.py
```diff
@@ -0,0 +1,179 @@
...
+    hzind = [(dx,15,-1.3),(dx, 15)]
+    
+    mesh = Mesh.TensorMesh([hxind, hyind, hzind], 'CCN')
+    
+    
+    # Set background conductivity
+    model = np.ones(mesh.nC) * sig[0]
+    
+    # First anomaly
+    ind = Utils.ModelBuilder.getIndicesSphere(loc[:,0],radi[0],mesh.gridCC)
```
lheagy: It doesn't look like this Util is on this branch, could you add it??

fourndo:
Fixed the broken link to Utils and tested locally. Should be good to go.

lheagy:
Thanks! I will add the auto-docs for it then we can merge it in. We can move the getIndicesSphere into the base utils if you think it will be a handy function

rowanc1:
@lheagy I am looking through DCIP branch. I am not quite sure whether @sgkang's updates have made it into the most recent PR. I will investigate a bit more and then pull it into a branch over here.

lheagy:
kk, it looks to me like the 'IP' branch is essentially master in simpegDC, is that right @sgkang ?

rowanc1:
@fourndo @lheagy this is included in the DCIP/dev PR. #236

PR #229, opened 2016-02-04
add test to ensure examples are up to date

lheagy:


lheagy:
not entirely sure why the pr fails... I will try to write some more useful output to the test.

lheagy:
now it passes! @rowanc1 : can you take a quick look when you get a chance?

rowanc1:
No idea. Seems to work now though?

lheagy:
in the first pass through, I used relpath instead of abspath, and I think it was a bit touchy. abspath seems to be better behaved

PR #243, opened 2016-02-13
Update EM_FDEM_SusEffects.py

lheagy:
if mumps import fails use LU


lheagy:
this is failing because the DC example does not have the right dependencies (DCIP has not been merged in)

PR #246, opened 2016-02-15
Naming conventions in Maps (re #231)

### LINKED ISSUE #231: Naming Convention in Maps
We are inconsistent in naming conventions for maps
```
Maps.ExpMap
Maps.ActiveCells
Maps.Vertical1DMap
```

Do we want the trailing `Map` or not? (I vote not). Also, active cells to me sounds like we are grabbing a subset of the model (opposite). How about:
```
Maps.Exp
Maps.SurjActiveCells
Maps.SurjVertical1D
```
or replace `Surj` with `Surject`. To me, these are a bit more explicit in describing the action they are taking. Thoughts?

lheagy:
- FullMap --> SurjectFull
- Vertical1DMap --> SurjectVertical1D
- Map2Dto3D --> Surject2Dto3D
- ActiveCells --> InjectActiveCells
- ActiveCellsTopo --> InjectActiveCellsTopo

to close #231 

@rowanc1 , @sgkang , @grosenkj : what do you guys think of this?

(I left ExpMap, IdentityMap, #231 talks about dropping the `Map` at the end, but this is a bit of a pain, and I do like having a verb in the name --> each of the maps is doing something, Mapping, Surjecting, Injecting)

### DIFF
--- a/SimPEG/Examples/EM_FDEM_1D_Inversion.py
+++ b/SimPEG/Examples/EM_FDEM_1D_Inversion.py
@@ -21,8 +21,8 @@ def run(plotIt=True):
 
     active = mesh.vectorCCz<0.
     layer = (mesh.vectorCCz<0.) & (mesh.vectorCCz>=layerz)
-    actMap = Maps.ActiveCells(mesh, active, np.log(1e-8), nC=mesh.nCz)
-    mapping = Maps.ExpMap(mesh) * Maps.Vertical1DMap(mesh) * actMap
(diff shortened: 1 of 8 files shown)


lheagy:
give a depreciation warning on old names

lheagy:
@rowanc1, @sgkang : what do you guys think of this? (I added Future Warnings and tested backwards compatibility) we can phase out the old names before removing them

sgkang:
As long as backward compatible, I am happy, so go for it! Thx @lheagy

PR #249, opened 2016-02-15
MT1D_analytic_Example

thast:
MT1D analytic example .py file  for EM_examples repository


SimPEG/Examples/DC_PseudoSection_Simulation.py
```diff
@@ -0,0 +1,220 @@
+from SimPEG import *
+# import simpegDCIP as DC
+from SimPEG import DCIP as DC
```
lheagy: right now DCIP isn't on the Examples or master branch, so this import is failing, we will need to get DCIP into master first

lheagy:
To make it more clear which examples are being added (and making sure that they all work on master), I am going to close this pr and open a new one without the DC example from the branch ex/mt1d. the DC examples are on the dcip/dev branch and I will leave the examples branch as is

PR #250, opened 2016-02-16
Ex/mt1d

lheagy:
Examples for 1DMT and electrostatic sphere

For the electrostatic sphere, a method `run` needs to be added: @thast


thast:
@lheagy : Travis passes but I do not understand why coverall did not (actually, I do not how coveralls works, and how to interpret its error reports...). From what I get it seems to be the SimPEG/Optimization.py and SimPEG/Examples/__init__.py 

Any thought about it?

lheagy:
Coveralls looks to see how much of the code base is tested. What this is basically saying is there is code being added that is not tested (which is ~ok in the example if there is lots of plotting functions. The plotting is not tested.) But, we should first remove the changes in Optimization.py from this pull request (it should be done in a separate pr... I can take care of this)

lheagy:
Can we get this picture:

cyyqaol5zh

thast:
You mean for simpeg website? sure. Do you also need the code that generate the movie? (functions are in the .py, but not the animation stuff)

rowanc1:
@thast you are on twitter. :) One of the nicer matplotlib examples I have
seen. Nice work!
https://twitter.com/simpegpy/status/699741704346165248

SimPEG/Optimization.py
```diff
@@ -990,4 +990,18 @@ def findSearchDirection(self):
                     cgFlag = 1
                 # End CG Iterations
 
+            # Take a gradient step on the active cells if exist
```
rowanc1: This part is being taken care of in another PR.

thast:
@rowanc1  Great! I'm glad you like it!

SimPEG/Examples/MT_1D_analytic_nlayer_Earth.py
```diff
@@ -0,0 +1,428 @@
...
+from ipywidgets import *
+from SimPEG.EM.Utils import k, omega
+
+"""
+MT1D: n layered earth problem
+*****************************
+
+Author: Thibaut Astic
+Contact: thast@eos.ubc.ca
+Date: January 2016
```
lheagy: git keeps track of all of this: odds are the example will be updated - so I would suggest removing this so that we don't have to manually keep track of it

lheagy: Also, @thast : I am not sure if you made updates to this code that you would like to put in this pull request before it gets onto dev. Up to you!

SimPEG/Examples/sphereElectrostatic_example.py
```diff
@@ -0,0 +1,785 @@
+from scipy.constants import epsilon_0
+import matplotlib.pyplot as plt
+import matplotlib.colors as colors
+import numpy as np
+from SimPEG.Utils import ndgrid, mkvc
+
+'''
+Authors: Thibaut Astic, Lindsey Heagy, Sanna Tyrvainen, Ronghua Peng
+Date: December 2015
```
lheagy: similar comment on docs as above. up to you

thast:
@lheagy : I removed the dates and the .py files are up to date.

lheagy:
this example is pretty involved - I actually think the code is more well suited in the em_examples repository (it is more an example of some of the impressive things that can be done, rather than an example of how to use the code - which is fine, but in my opinion better suited for the em_examples repo). From the SimPEG docs, we can point to the em_examples running in the binders. Thoughts @thast , @rowanc1 ?

thast:
I'm fine with that solution. It's true it is a stand-alone piece of code.

rowanc1:
:+1:

lheagy:
sounds good. I will close this pr, and if all of the code is already in the em_examples repo, we can delete this branch. @thast : can you confirm?

PR #253, opened 2016-02-17
Allow moving off bounds in projected gradient

rowanc1:
The current implementation does not allow you to move off the
bounds (lower/upper) once you have gotten on them, this allows you
to move off of the bound.

Please note that more testing should be done to ensure that this does
not introduce oscillations into the optimization routine.

### DIFF
--- a/SimPEG/Optimization.py
+++ b/SimPEG/Optimization.py
@@ -888,6 +888,8 @@ class ProjectedGNCG(BFGS, Minimize, Remember):
     maxIterCG = 5
     tolCG = 1e-1
 
+    stepOffBoundsFact = 0.1 # perturbation of the inactive set off the bounds
+
     lower = -np.inf
     upper = np.inf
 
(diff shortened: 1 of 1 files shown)


rowanc1:
@fourndo wrote:

Ok so terminology is really annoying, but let's say that cells that are outside or on the bounds are called 'active', we needed a strategy to bring them back in the inversion. So the extra snippet simply takes the gradient descent of the active set and scales it so it small compares to the GN direction. 
Than after on line 1004, it checks if the gradient direction is going the right way. i.e. If a cell is below the lower bound, then make sure the step direction is going to increase it. If yes, then good it is back in the inactive set. If not, then bad and zero gradient for you. Same idea but opposite for the upper bound.

rowanc1:
@sgkang you mentioned that you were having some problems with the optimization? Can you demonstrate?

Possibly we can get an example into here as well?!

SimPEG/Optimization.py
```diff
@@ -990,4 +990,18 @@ def findSearchDirection(self):
...
 
+            # Take a gradient step on the active cells if exist
+            if temp != self.xc.size:
+
+                rhs_a = (Active) * -self.g
+
+                dm_i = max( abs( delx ) )
+                dm_a = max( abs(rhs_a) )
+
+                delx = delx + rhs_a * dm_i / dm_a /10.
```
lheagy: where does the 10 come from? is it an arbitrary parameter?

fourndo: Yep, totally arbitrary. It's like taking a step 10% of the maximum step in the active set.

lheagy: is that potentially quite large? 

We might also want to make it a key word argument so that it can be tuned easily if necessary

lheagy:
before merging this, I would suggest we turn the factor of 0.1 into a kwarg, maybe: `stepOffBounds` or `stepOffBoundsFact` ?

fourndo:
I vote for stepOffBoundsFact

lheagy:
@rowanc1 : any idea what is going on with coveralls here?

rowanc1:
Probably didn't collect all the info from travis. could rerun it?

Issue #168, opened 2015-11-13
Renaming smoothModel in the Tikhonov regularization.


grosenkj:
This is very confusing name for using to control the use of m_ref in the smoothness weights.
Can this be change to mrefInWsmooth or something more descriptive of that this flag intended use.

sgkang:
True. That was mostly following UBC terminology. We need to put some documentation about that as well. I like mrefInWsmooth

lheagy:
This one got me again today! is everyone agreed on mrefInSmooth ? we can add this regularization changes in pr #254

grosenkj:
I agree, I like mrefInSmooth. Thanks for addressing this.

Issue #251, opened 2016-02-16
indActive in Regularization


lheagy:
store the projection matrices in the base regularization

lheagy:
This was addressed in #277 and #295

PR #254, opened 2016-02-17
Sparse Regularization

rowanc1:


rowanc1:
We were thinking about renaming unitCellGradx to cellGradStencilx this has the following advantages:

* tab completion from the cellGrad starting point
* unit refers to the unit domain [0-1] where as this is unitless, or a stencil of +/-1's

rowanc1:
Can we deal with #251 here? @lheagy

lheagy:
I think this is a sensible place to take care of it

lheagy:
Before we merge this, could a couple people (@rowanc1 @sgkang @grosenkj ??)  take a look at the regmesh @fourndo and I wrote. We should also add:
- [ ] at least some minimal documentation 
- [ ] an example using the active cell reg to simpeg examples

SimPEG/Regularization.py
```diff
@@ -1,5 +1,179 @@
...
+        return self._aveFz2CC
+
+    @property
+    def aveCC2Fz(self):
+        if getattr(self, '_aveCC2Fz', None) is None:
+            self._aveCC2Fz =  Utils.sdiag(1./(self.aveFz2CC.T).sum(1)) * self.aveFz2CC.T
+        return self._aveCC2Fz
+
+    @property
+    def cellGradx(self):
```
lheagy: naming question: do we want to call this `cellDiffx` (same for y,z). When we reduce to active space, it isn't really a gradient operator per-se ?

SimPEG/Regularization.py
```diff
@@ -1,5 +1,179 @@
...
+        return self._cellGrady
+
+    @property
+    def cellGradz(self):
+        if getattr(self, '_cellGradz', None) is None:
+            self._cellGradz = self._Pafz.T * self.mesh.cellGradz * self._Pac
+        return self._cellGradz
+    
+    @property
+    def faceDivx(self):
```
lheagy: same naming question: `faceDiffx` ?

SimPEG/Regularization.py
```diff
@@ -1,5 +1,179 @@
...
+        return self._faceDivy
+    
+    @property
+    def faceDivz(self):
+        if getattr(self, '_faceDivz', None) is None:
+            self._faceDivz = self._Pac.T * self.mesh.faceDivz * self._Pafz
+        return self._faceDivz
+
+    @property
+    def cellGradxStencil(self):
```
lheagy: cellDiffxStencil?

SimPEG/Regularization.py
```diff
@@ -282,3 +529,197 @@ def evalDeriv(self, m):
...
+            out2 = mD2.T * ( self.Ws.T * r2 )
+            out = out1+out2
+        elif self.smoothModel == False:
+            mD = self.mapping.deriv(m - self.mref)
+            r = self.W * ( self.mapping * (m - self.mref) )
+            out = mD.T * ( self.W.T * r )
+        return out
+
+
+class SparseRegularization(Simple):
```
lheagy: namespace question: do we want to call this `Sparse` (so it is `Regularization.Sparse` )? Thoughts @rowanc1 , @fourndo ?

fourndo: I would vote for Regularization.Sparse . In case if later start doing sparsity in order domain (i.e. data)

lheagy: kk, so we can update the name from class `SparseRegularization` to `Sparse`. Do you want to do this? (I don't want to break all of your current, personal code :) )

lheagy:
this pr will now close #168

lheagy:
@rowanc1, @fourndo : are you guys good if we merge this one?

SimPEG/Mesh/DiffOperators.py
```diff
@@ -401,14 +415,19 @@ def fget(self):
         return locals()
     cellGrady = property(**cellGrady())
 
+    def _cellGradzStencil(self):
+        if self.dim < 3: return None
+        BC = ['neumann', 'neumann']
```
rowanc1: just a note. boundary conditions are different here vs in the main cellGrad call?

SimPEG/Regularization.py
```diff
@@ -276,249 +631,95 @@ def evalDeriv(self, m):
...
-#             self.Rx = Utils.speye(self.mesh.unitCellGradx.shape[0])
 
-#         else:
-#             f_m = self.mesh.unitCellGradx * self.m
-#             self.rx = self.R( f_m , self.qx, self.eps)
-#             self.Rx = Utils.sdiag( self.rx )
+class Sparse(Simple):
+
+    eps   = 1e-1
+    m     = None
```
rowanc1: what is m?

SimPEG/Regularization.py
```diff
@@ -276,249 +631,95 @@ def evalDeriv(self, m):
...
+    qz    = 2.
+
+    def __init__(self, mesh, mapping=None, indActive=None, **kwargs):
+        Simple.__init__(self, mesh, mapping=mapping, indActive=indActive, **kwargs)
+
+
+    @property
+    def Ws(self):
+        """Regularization matrix Ws"""
+        if getattr(self, 'm', None) is None:
```
lheagy: should this be looking at the mapping?
(not m?)

rowanc1:
I think that `Sparse.W`'s  should not be properties. They should be evaluated every time because they depend on the model??

So the eval function of sparse needs to change? unless this is lagged? In which case can we call it something other than `m`?

SimPEG/Regularization.py
fourndo: model

SimPEG/Directives.py
```diff
@@ -283,3 +282,52 @@ def endIter(self):
...
+
+             if getattr(self, 'eps_min', None) is not None:
+                 self.reg.eps = np.max([self.eps_min,eps])
+             else:
+                 self.reg.eps = eps
+
+
+         # Update the model used for the IRLS weights
+         if getattr(self, 'm', None) is None:
+             self.reg.m = self.invProb.curModel
```
rowanc1: can we call this something else? reg._____

SimPEG/Regularization.py
rowanc1: not mref?

SimPEG/Regularization.py
```diff
@@ -276,249 +631,95 @@ def evalDeriv(self, m):
...
+            self.Rs = Utils.speye(self.regmesh.nC)
+
+        else:
+            f_m = self.m
+            self.rs = self.R(f_m , self.p, self.eps)
+            #print "Min rs: " + str(np.max(self.rs)) + "Max rs: " + str(np.min(self.rs))
+            self.Rs = Utils.sdiag( self.rs )
+
+        self._Ws = Utils.sdiag((self.regmesh.vol*self.alpha_s*self.gamma)**0.5)*self.Rs
```
rowanc1: these are computed on every call, so I would prefer to not store links to them.

rowanc1: just `return Utils.sdiag((self.regmesh.vol*self.alpha_s*self.gamma)**0.5)*self.Rs`

SimPEG/Regularization.py
```diff
@@ -276,249 +631,95 @@ def evalDeriv(self, m):
...
+
+        if getattr(self, 'm', None) is None:
+            self.Rx = Utils.speye(self.regmesh.cellDiffxStencil.shape[0])
+
+        else:
+            f_m = self.regmesh.cellDiffxStencil * self.m
+            self.rx = self.R( f_m , self.qx, self.eps)
+            self.Rx = Utils.sdiag( self.rx )
+
+        if getattr(self, '_Wx', None) is None:
```
rowanc1: I don't believe we should ever use a cached version?

SimPEG/Regularization.py
fourndo: I bit dangerous to deal with mapped m, for the reasons we talked about the other day. But I guess if the inversion is done entirely in a different space, than maybe you are right. Or actually the weights R should be mapped. Not sure on that one.

SimPEG/Regularization.py
```diff
@@ -276,249 +631,95 @@ def evalDeriv(self, m):
...
+
+        if getattr(self, 'm', None) is None:
+            self.Rz = Utils.speye(self.regmesh.cellDiffzStencil.shape[0])
+
+        else:
+            f_m = self.regmesh.cellDiffzStencil * self.m
+            self.rz = self.R( f_m , self.qz, self.eps)
+            self.Rz = Utils.sdiag( self.rz )
+
+        if getattr(self, '_Wz', None) is None:
```
rowanc1: should we cache?

fourndo: Nan probably not. I just don't know what I'm doing. Want me to switch it back to def only?

SimPEG/Regularization.py
fourndo: Never had succes with mref in. Going to sparse norms is a bit change of philosophy over the l2. We are encouraging the model to either 0 or large. So if mref is in, it's like implying that you want a model that either resemble mref or completely different. So if mref is slightly wrong, than you end up with super large jumps. We could definitely put the option in though, a bit like SMOOTH_MOD vs SMOOTH_MOD_DIFF

SimPEG/Mesh/DiffOperators.py
lheagy: this is cut and pasted from inside cellGradz (previously lines 409-411). Should this be calling `self.setCellGradBC(self._cellGradBC_list)`

SimPEG/Regularization.py
```diff
@@ -276,249 +523,223 @@ def evalDeriv(self, m):
...
+    """
+        Only for tensor mesh
+    """
+
+    mrefInSmooth = True  #: SMOOTH and SMOOTH_MOD_DIF options
+    alpha_s     = Utils.dependentProperty('_alpha_s', 1.0, ['_W', '_Ws'], "Smallness weight")
+    alpha_x     = Utils.dependentProperty('_alpha_x', 1.0, ['_W', '_Wx'], "Weight for the first derivative in the x direction")
+    alpha_y     = Utils.dependentProperty('_alpha_y', 1.0, ['_W', '_Wy'], "Weight for the first derivative in the y direction")
+    alpha_z     = Utils.dependentProperty('_alpha_z', 1.0, ['_W', '_Wz'], "Weight for the first derivative in the z direction")
+    wght        = 1.
```
lheagy: `cellweights` ?

lheagy: we can make this a part of base regularization - I imagine this will be useful for all regularizations!

SimPEG/Regularization.py
```diff
@@ -126,98 +413,58 @@ class Tikhonov(BaseRegularization):
     alpha_zz = Utils.dependentProperty('_alpha_zz', 0.0, ['_W', '_Wzz'], "Weight for the second derivative in the z direction")
 
     def __init__(self, mesh, mapping=None, indActive = None, **kwargs):
-        BaseRegularization.__init__(self, mesh, mapping=mapping, **kwargs)
-        self.indActive = indActive
+        BaseRegularization.__init__(self, mesh, mapping=mapping, indActive=indActive, **kwargs)
 
     @property
     def Ws(self):
```
lheagy: would ppl be game if we renamed this to Wsmall? 

(s could mean small or smooth, so it is ambiguous...)

rowanc1: I like Wsmall

lheagy:
The fail is in MT: @grosenkj - can you have a look? It looks like an issue of parameter choice to me.

SimPEG/Directives.py
```diff
@@ -216,7 +216,7 @@ def endIter(self):
         # Save the data.
         ms = self.reg.Ws * ( self.reg.mapping * (self.invProb.curModel - self.reg.mref) )
         phi_ms = 0.5*ms.dot(ms)
-        if self.reg.smoothModel == True:
+        if self.reg.mrefInSmooth == True:
```
lheagy: default should be false

rowanc1:
@lheagy do you know the state of this?

lheagy:
yes, we need to add
- [ ] docs 
- [ ] example
for the sparse regularization

fourndo:
Cool. 
@lheagy: Can we sit down today to look at specifics.? Where to put what type of thing.

lheagy:
Hey @fourndo: I am downtown today but could give you a call at some point today. For docs, basically enough to get the gist of what the sparse regularization does (the mesh class docs are a great example: http://docs.simpeg.xyz/en/latest/api_InnerProducts.html#module-SimPEG.Mesh.InnerProducts with associated code in Mesh/InnerProducts.py). The example doesn't need to be anything major, just something to show how it works (similar to scripts in Examples folder). If you have one with the PF stuff, we can include it when we merge that in instead.

fourndo:
Oh yea of course, now I remember. Ok i ll put something together.

lheagy:
Perfect, Feel free to ping me on slack if you want to chat!

SimPEG/Examples/DC_PseudoSection_Simulation.py
```diff
@@ -0,0 +1,220 @@
+from SimPEG import *
```
lheagy: This is the same as the one on the Examples Branch - @fourndo ? If so, I will remove it from this PR and we can take care of it on the Examples one

SimPEG/Mesh/DiffOperators.py
lheagy: @rowanc1 : is there something you think should be different here?

SimPEG/Directives.py
```diff
@@ -356,10 +356,29 @@ def endIter(self):
...
+                self.k = int(self.survey.nD/10)
+                
+            def JtJv(v):
+
+                Jv = self.prob.Jvec(m, v)
+            
+                return self.prob.Jtvec(m,Jv)
+            
+            JtJdiag = Utils.diagEst(JtJv,6,k=100)
+            JtJdiag = JtJdiag
```
lheagy: you don't need this line

SimPEG/Directives.py
```diff
@@ -356,10 +356,29 @@ def endIter(self):
             PC     = Utils.sdiag(diagA**-1.)
             self.opt.approxHinv = PC
             print 'Updated pre-cond'
+
+class update_Wj(InversionDirective):
```
lheagy: nice!

SimPEG/Directives.py
```diff
@@ -356,10 +356,29 @@ def endIter(self):
...
+            if self.k is None:
+                self.k = int(self.survey.nD/10)
+                
+            def JtJv(v):
+
+                Jv = self.prob.Jvec(m, v)
+            
+                return self.prob.Jtvec(m,Jv)
+            
+            JtJdiag = Utils.diagEst(JtJv,6,k=100)
```
lheagy: `k = self.k`

lheagy: and instead of `6` you can use `len(m)`

lheagy:
This branch is breaking because there are examples that are out of date. To handle this, we should clean up the Examples branch and merge #279 onto dev and then merge this pr

rowanc1:
@fourndo Can we use the convention of uppercase class names?

SimPEG/Directives.py
```diff
@@ -356,10 +356,29 @@ def endIter(self):
...
     
-#==============================================================================
-#          import pylab as plt
-#          plt.figure()
-#          ax = plt.subplot(221)
-#          self.prob.mesh.plotSlice(self.invProb.curModel, ax = ax, normal = 'Z', ind=-5, clim = (0, 0.005))
-#==============================================================================
+    def endIter(self):
+        
+        if self.opt.iter == 2:
```
rowanc1: Only on the second iteration?

SimPEG/Maps.py
```diff
@@ -759,15 +759,27 @@ class PolyMap(IdentityMap):
 
             m = [\sigma_1, \sigma_2, c]
 
-    """
-    def __init__(self, mesh, order, logSigma=True, normal='X'):
+    """
```
rowanc1: @fourndo  any chance you can set your editor to delete trailing spaces?

SimPEG/Optimization.py
```diff
@@ -990,4 +990,18 @@ def findSearchDirection(self):
                     cgFlag = 1
                 # End CG Iterations
 
+            # Take a gradient step on the active cells if exist
+            if temp != self.xc.size:
```
rowanc1: This I am pretty sure has been dealt with on the other PR? #253

lheagy: yes, the changes in opt and reg in this pr are behind the sparse reg branch and pr #253. We can do them one after another or combine??

lheagy:
this will be taken care of in #289

SimPEG/Directives.py
fourndo: I will add a variable to control this

SimPEG/Maps.py
fourndo: Will do

PR #256, opened 2016-02-21
rx.projectFields --> rx.eval

### LINKED ISSUE #247: Rx.eval?
Instead of `rx.projectFields`?

This would be more inline with the rest of simpeg.

lheagy:
Evaluate receivers:  
- rx.projectFields --> rx.eval
- rx.projectFieldsDeriv --> rx.evalDeriv

This change is not backwards compatible (but should be internal??) 

Closes #247


lheagy:
no problem! Any reason we should make this backwards compatible? or is the hard change ok? ( I am fine with making it a hard change)

PR #257, opened 2016-02-28
Em/dev

lheagy:
- e,b,h,j from all formulations 
    - improved testing
    - clean up of source integrations when using different formulations 
- notation clean up in sensitivity calc --> keeping track of where the vector has been multiplied
- notation clean up for problems: "Deriv_m" not necessary to state in FDEM (all derivs once they get to the base problem are wrt m


lheagy:
@sgkang , @grosenkj : would you guys mind taking a look through these changes when you have a min?

lheagy:
merge time?

PR #265, opened 2016-03-13
Dev

lheagy:
Major Additions
- MT: 1D, 3D
- DCIP: 3D 

Updates
- FDEM (all fields fluxes from all formulations) 
- rx --> eval (no longer project fields)
- mapping names updated (backwards compatible, with depreciation warnings)


lheagy:
should we just go for it?

PR #267, opened 2016-03-21
Em/ref/notation

lheagy:
notational clean up 
- use `f` for fields
- use `u` for solution vector

### DIFF
--- a/SimPEG/DataMisfit.py
+++ b/SimPEG/DataMisfit.py
@@ -89,7 +89,7 @@ class l2_DataMisfit(BaseDataMisfit):
         """
 
         if getattr(self, '_Wd', None) is None:
-            
+
             survey = self.survey
 
             if getattr(survey,'std', None) is None:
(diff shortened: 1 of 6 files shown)


SimPEG/DataMisfit.py
```diff
@@ -121,11 +121,11 @@ def evalDeriv(self, m, u=None):
         prob   = self.prob
         survey = self.survey
         if u is None: u = prob.fields(m)
-        return prob.Jtvec(m, self.Wd * (self.Wd * survey.residual(m, u=u)), u=u)
+        return prob.Jtvec(m, self.Wd * (self.Wd * survey.residual(m, u)), u)
```
lheagy: do we want to change this to `f` throughout the code-base?

lheagy: what do you think, @rowanc1 ?

lheagy: this is the topic of #271, and can be addressed in a separate pr so it is consistent across the code base

PR #273, opened 2016-03-30
make integrate = False default for all sources

lheagy:
Make source integration default false for Raw Vec sources (assuming you are providing something that is already integrated), can be changed with the kwarg `integrate`. @grosenkj

### DIFF
--- a/SimPEG/EM/FDEM/SrcFDEM.py
+++ b/SimPEG/EM/FDEM/SrcFDEM.py
@@ -9,8 +9,10 @@ class BaseSrc(Survey.BaseSrc):
     """
 
     freq = None
-    # rxPair = RxFDEM
-    integrate = True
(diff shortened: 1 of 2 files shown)


lheagy:
@grosenkj : the fail is with MT, and none of these changes should have affected it. If there is a random number generator used here, can you fix it to one that we expect to work every time? (`np.random.seed(#)`)  Thanks! 

job 1149 6 - simpeg simpeg - travis ci 2016-03-30 07-14-27

lheagy:
@grosenkj : are you good if I merge this?

grosenkj:
Yes, I am good with this. I not sure why that derivative check is failing, because it is actually using a half-space value so there is no inherent randomness to the check. Travis might have been having a difficult day.

PR #274, opened 2016-03-31
use `f` for fields and `u` for a solution vector

### LINKED ISSUE #271: naming: f is for fields
through much of the codebase, `u` is used to denote fields, but in the EM paper, and EM code, `f` is used (see pr #267). `f` should be used throughout the code for fields, `u` is a solution vector from which we can construct `f`.

lheagy:
Problem.Jvec, Problem.Jtvec, Problem.fields, DataMisfit,  survey.dpred all take a fields object, `f` not a solution vector `u`. DCIP will take some more clean-up, but this will be handled in the refactor of DC. These changes will close #271. This touches a lot of the repo, so a quick review by a few people would be much appreciated!

### DIFF
--- a/SimPEG/DataMisfit.py
+++ b/SimPEG/DataMisfit.py
@@ -123,5 +123,5 @@ class l2_DataMisfit(BaseDataMisfit):
     @Utils.timeIt
     def eval2Deriv(self, m, v, f=None):
         "eval2Deriv(m, v, f=None)"
-        if f is None: f = prob.fields(m)
-        return self.prob.Jtvec_approx(m, self.Wd * (self.Wd * prob.Jvec_approx(m, v, f=f)), f=f)
+        if f is None: f = self.prob.fields(m)
+        return self.prob.Jtvec_approx(m, self.Wd * (self.Wd * self.prob.Jvec_approx(m, v, f=f)), f=f)
--- a/SimPEG/DataMisfit.py
+++ b/SimPEG/DataMisfit.py
@@ -22,11 +22,11 @@ class BaseDataMisfit(object):
         Utils.setKwargs(self,**kwargs)
 
     @Utils.timeIt
-    def eval(self, m, u=None):
-        """eval(m, u=None)
+    def eval(self, m, f=None):
+        """eval(m, f=None)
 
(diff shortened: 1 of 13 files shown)


SimPEG/DataMisfit.py
```diff
@@ -47,12 +47,12 @@ def evalDeriv(self, m, u=None):
...
     @Utils.timeIt
-    def eval2Deriv(self, m, v, u=None):
-        """eval2Deriv(m, v, u=None)
+    def eval2Deriv(self, m, v, f=None):
+        """eval2Deriv(m, v, f=None)
 
             :param numpy.array m: geophysical model
             :param numpy.array v: vector to multiply
-            :param numpy.array u: fields
+            :param Fields f: fields
```
rowanc1: does `Fields` show up in the docs? It might be better to be `SimPEG....Fields`? (where ever it actually is)

SimPEG/FLOW/Richards/RichardsProblem.py
```diff
@@ -45,19 +45,19 @@ def nD(self):
 
     @Utils.count
     @Utils.requires('prob')
-    def dpred(self, m, u=None):
```
rowanc1: It is wonderful to have other people come into this code. Thanks!

rowanc1:
I don't see any problems here. Changes to code should be pretty simple if people outside are dependent on this, but I doubt that they would be.
I am good to merge this in.

rowanc1:
(any idea why coverage increased?)

SimPEG/FLOW/Richards/RichardsProblem.py
lheagy: Tests are helpful!

SimPEG/DataMisfit.py
lheagy: This does link up properly (SimPEG.Fields doesn't) 
data misfit simpeg 0 1 10 documentation 2016-03-31 23-56-35 png 2016-03-31 23-58-21

lheagy:
perfect, thanks for taking a look!
(coverage probably increased b/c I deleted more lines than added?)

PR #276, opened 2016-04-01
Testing the Docs

rowanc1:
This is a really good idea, I will try to take a look at this later on in the week.

I think that it might also be good to check the examples, and make sure that the images are run every time on travis, but read the docs just gets the images. Some of the computations are way to heavy for RTFD, and we shouldn't be doing this there!

I will give this some thought.


rowanc1:
Close in favour of #324.

PR #285, opened 2016-04-06
Dcip/dev

lheagy:
- updates to examples and IO Utils in DCIP 
- the sparse reg stuff will be updated in #254


lheagy:
This will be taken care of in #289

Issue #290, opened 2016-04-07
import of osr, gdal on MT import


lheagy:
On master, right now `osr` and `gdal` are imported with MT, can we only import these if necessary?
<img width="503" alt="mt" src="https://cloud.githubusercontent.com/assets/6361812/14343619/137ba2b6-fc57-11e5-962f-ce7d38a7549b.png">

grosenkj:
Done, fixed in branch mt/iss290. 

Moved the import into the projection function, so it should happen up on the function call.

grosenkj:
Made a pull request #291.

PR #291, opened 2016-04-07
Fixing Issue #290

grosenkj:
Fixed osr import in ediFilesUtils.
Moved the import into projection functions which should only import on calling of the function.

Also fixed a small boolean error in Directives.


lheagy:
so great Thanks @grosenkj

SimPEG/Directives.py
```diff
@@ -255,7 +255,7 @@ def endIter(self):
             mref = 0
```
lheagy: This function has the exact same name as the one on 208. Which one is the one we want to keep? (first pass they look like duplicates?) @grosenkj

lheagy:
looks good @grosenkj : I will let you resolve this (it is likely only in Directives.py) If elsewhere too, I am happy to help out

lheagy:
@grosenkj : mind if I delete this branch (mt/iss290) ?

PR #294, opened 2016-04-07
remove mutable arguments.

rowanc1:


lheagy:
why are mutable arguments poor practice?

bsmithyman:
If I recall correctly, they're instantiated at class definition time, so all instances of the class point at the same object. So, if one of them changes it, it changes the default for other instances.

PR #302, opened 2016-04-24
Installation (i.e., setup.py) is no longer dependent on Numpy

bsmithyman:
When we go to install SimPEG (or packages w/ SimPEG dependencies) on architecture that doesn't have `numpy` present already, it will fail to install. This happens because `setup.py` imports `numpy` before it's guaranteed to be setup on the system. This change implements a workaround that lets `distutils` get up and going before `numpy` is imported.

See [StackOverflow](http://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,15 +5,16 @@ SimPEG is a python package for simulation and gradient based
 parameter estimation in the context of geophysical applications.
 """
 
-import numpy as np
-
 import os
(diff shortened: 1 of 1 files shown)


lheagy:
Thanks @bsmithyman!

PR #308, opened 2016-05-03
if mapping is none, create an identity map that is size indactive.nonzero

lheagy:
If no mapping provided and indActive provided, make sure the mapping is the size of number of active cells


rowanc1:
:)

PR #310, opened 2016-05-05
modularizing regularization

lheagy:
- have Tikhonov inherit simple (simple is simpler than Tikhonov!)
- break apart each of the components contributing to the smoothness (ie. x,y,z) and evaluate independently, then sum

@rowanc1, @grosenkj : can you guys take a look at this and see what you think? Thanks!


SimPEG/Regularization.py
```diff
@@ -403,105 +403,61 @@ def eval2Deriv(self, m, v=None):
...
+
+    mrefInSmooth = False  #: SMOOTH and SMOOTH_MOD_DIF options
     alpha_s      = Utils.dependentProperty('_alpha_s', 1e-6, ['_W', '_Wsmall'], "Smallness weight")
     alpha_x      = Utils.dependentProperty('_alpha_x',  1.0, ['_W', '_Wx'],     "Weight for the first derivative in the x direction")
     alpha_y      = Utils.dependentProperty('_alpha_y',  1.0, ['_W', '_Wy'],     "Weight for the first derivative in the y direction")
     alpha_z      = Utils.dependentProperty('_alpha_z',  1.0, ['_W', '_Wz'],     "Weight for the first derivative in the z direction")
-    alpha_xx     = Utils.dependentProperty('_alpha_xx', 0.0, ['_W', '_Wxx'],    "Weight for the second derivative in the x direction")
-    alpha_yy     = Utils.dependentProperty('_alpha_yy', 0.0, ['_W', '_Wyy'],    "Weight for the second derivative in the y direction")
-    alpha_zz     = Utils.dependentProperty('_alpha_zz', 0.0, ['_W', '_Wzz'],    "Weight for the second derivative in the z direction")
+    wght        = 1.
```
rowanc1: can we rename to `weight`?

rowanc1: Can it be replaced by `alpha`'s? @fourndo?

lheagy: `cell_weights`?

SimPEG/Regularization.py
```diff
@@ -403,105 +403,61 @@ def eval2Deriv(self, m, v=None):
...
 
+        if isinstance(self.wght,float):
+            self.wght = np.ones(self.regmesh.nC) * self.wght
+
     @property
     def Wsmall(self):
         """Regularization matrix Wsmall"""
         if getattr(self,'_Wsmall', None) is None:
-            self._Wsmall = Utils.sdiag((self.regmesh.vol*self.alpha_s)**0.5)
+            self._Wsmall = Utils.sdiag((self.regmesh.vol*self.alpha_s*self.wght)**0.5)
```
lheagy: remove vol

SimPEG/Regularization.py
lheagy: `@property` : in base regularization

SimPEG/Directives.py
```diff
@@ -303,26 +303,27 @@ def endIter(self):
         # Set the weighting matrix to None so that it is recomputed next time
         # it is called in the inversion
         self.reg._W = None
+        self.reg._Wsmooth = None
```
lheagy: we should just not store Wsmooth for the sparse reg?

lheagy: this is because I did not look at the 2Deriv... it was still looking to Wsmooth (which it should no longer be doing). I have pushed a change where we now compute the 2Deriv for all of the regularization components. This whole implementation will be cleaner when we break it apart further...

lheagy: so... you should be able to remove these two lines that remove any caching and it should work!?

fourndo: Needed to update the def W to include the list of gradients for the inversion to run. Testing the directive now.

SimPEG/Directives.py
```diff
@@ -144,6 +144,35 @@ def endIter(self):
             if self.debug: print 'BetaSchedule is cooling Beta. Iteration: %d' % self.opt.iter
             self.invProb.beta /= self.coolingFactor
 
+#class BetaSchedule_PGN_CG(InversionDirective):
```
lheagy: can this inherit the BetaSchedule directive? Then it will have the cooling factor and rate by default - it is an extension on the BetaSchedule?

SimPEG/Directives.py
```diff
@@ -144,6 +144,35 @@ def endIter(self):
...
+#    coolingRate = 1
+#    GN_step_last = None
+#    GN_step_c = None
+#    
+#    def endIter(self):
+#        
+#        """ Compute the change in GN step, and proceed with cooling if below tol"""
+#        if self.opt.iter == 1:
+#            self.GN_step_last = np.linalg.norm(self.opt.xc - self.opt.x_last)
+#            d_GN_step = 1.
```
lheagy: what is the 'd' indicating? is it the scaling?

SimPEG/Optimization.py
```diff
@@ -893,6 +893,10 @@ def __init__(self, **kwargs):
     lower = -np.inf
     upper = np.inf
 
+    # Variables to control inner GN iterations
+    rdm_tol         = 1e-2 # Tolerance for largest change in step (Default 1%)
```
lheagy: rdm?

fourndo: relative dm?
Measuring the change in step length

lheagy: ah, k thanks. Maybe `scale_dm`? ie. it is how much we scale back dm

SimPEG/Directives.py
fourndo: Yea this was my first kick at the cat.
I commented it out so I can steal from it. If you still have strong feelings for a directive, we can go back to it. Otherwise, I would just kill it.

SimPEG/Optimization.py
```diff
@@ -939,6 +943,72 @@ def approxHinv(self, value):
         self._approxHinv = value
 
     @Utils.timeIt
+    def minimize(self, evalFunction, x0):
```
lheagy: cool, I think this looks good! is there a test that runs it? 
@rowanc1 : what do you think?

fourndo: still have to write.
can you point to a test for the other solvers?
Still have to write a test for the compact norms, but I have an idea for that

lheagy: the ones in test/base/optimize.py are closest to what you might want to include here: https://github.com/simpeg/simpeg/blob/master/tests/base/test_optimizers.py

sgkang:
That looks good to me. If everyone is happy then can we merge to dev?

fourndo:
It s mostly my fault if not merged in yet. Haven't taken the time to do the last mods and write the test. Next on my list.

SimPEG/Directives.py
```diff
@@ -288,6 +327,13 @@ def endIter(self):
...
 
+        # Reset the regularization matrices so that it is
+        # recalculated for current model
+        self.reg._Wsmall = None
+        self.reg._Wx = None
+        self.reg._Wy = None
+        self.reg._Wz = None
+
          # Update the model used for the IRLS weights
         self.reg.curModel = self.invProb.curModel
```
rowanc1: Copy the beta cooling code, and only update the `reg.curModel` on a schedule?

lheagy:
I am good with merging this. @fourndo : would you mind making an issue with the things you still have in mind that need to be taken care of here? then feel free to merge it in!

PR #319, opened 2016-05-22
Target Misfit

### LINKED ISSUE #266: Target Misfit value should not be hard-coded
Right now, there is no way to change what we choose to use as a target misfit (there should be a default, which is what it currently is `0.5*survey.nD`, but... I would like to see it defined with kwargs
- `chifact` with default value of 1
- `phid_star` (maybe?) with default value `0.5*survey.nD`

ie. so `self._target = chifact * phid_star`

lheagy:
- use a `chifact` and a `phi_d_star` to define target misfit. ie
```
target = chifact * phi_d_star
```
to close #266

### DIFF
--- a/SimPEG/Directives.py
+++ b/SimPEG/Directives.py
@@ -146,10 +146,15 @@ class BetaSchedule(InversionDirective):
 
 class TargetMisfit(InversionDirective):
 
+    chifact = 1.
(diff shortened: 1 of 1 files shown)


rowanc1:
This looks solid. Thanks @lheagy.

Issue #321, opened 2016-05-27
PropMap Bug - linked property derivatives


lheagy:
The linked map is failing the derivative test. Just `rho` on its own passes, but not getting sigma from rho

```
from SimPEG import *
from scipy.constants import mu_0
from SimPEG import Tests
import matplotlib.pyplot as plt

class MyReciprocalPropMap(Maps.PropMap):
    sigma  = Maps.Property("Electrical Conductivity", defaultInvProp=True, propertyLink=('rho',   Maps.ReciprocalMap))
    rho    = Maps.Property("Electrical Resistivity",                       propertyLink=('sigma', Maps.ReciprocalMap))
    mu     = Maps.Property("Mu", defaultVal=mu_0,    propertyLink=('mui',  Maps.ReciprocalMap))
    mui    = Maps.Property("Mu", defaultVal=1./mu_0, propertyLink=('mu',   Maps.ReciprocalMap))

mesh = Mesh.TensorMesh([4,5], x0='CC')

mapping = Maps.ExpMap(mesh)
propmap = MyReciprocalPropMap([('rho', mapping)])

x0 = np.random.rand(mesh.nC) 
m  = propmap(x0) 

# test Sigma
def testme(v):
    return 1./(m.rhoMap*v), m.sigmaDeriv

Tests.checkDerivative(testme, x0, dx=0.01*x0, num=5)
```
![image](https://cloud.githubusercontent.com/assets/6361812/15594321/671001fe-2369-11e6-8e5a-c2746c2ccb41.png)

PR #322, opened 2016-05-27
Bug/propmap

lheagy:
- Fixed bug in propmap derivatives for linked properties as per issue #321. 
- Use fixed propmap in EM problem (cleans up mass matrices and ensures that derivs of linked properties are tested) 

image


sgkang:
Nice @lheagy. I had to change `MfRhoDeriv` because of this, but now we can directly use curmodel.rhoDeriv

`   
    def MfRhoDeriv(self,u):
        """
            Derivative of :code:`MfRho` with respect to the model.
        """
        return self.mesh.getFaceInnerProductDeriv(self.curModel.rho)(u) * (-Utils.sdiag(self.curModel.rho**2) * self.curModel.sigmaDeriv)
        # self.curModel.rhoDeriv`

PR #323, opened 2016-05-29
plotImage for curvilinear mesh

rowanc1:
Updates to the example to use the new code.


rowanc1:
This should be an easy merge. Let me know if you have any questions!

lheagy:
Beautiful!

figure 1 2016-05-29 13-51-39

For the Example name - what do you think of calling it `Mesh_BasicForwardDC` or something like that? It is really an example for the mesh class

PR #328, opened 2016-05-31
Refactor the Namespace of MT.

rowanc1:


grosenkj:
MT namespace updated to be NSEM (Natural Source EM).
Structure and naming of methods and components reflect recent changes to FDEM namespace.

PR #331, opened 2016-06-06
Automate the epsilon picking based on percentile of model values for …

fourndo:
Automate the epsilon picking based on percentile of model values for DEFAULT mode. Fix example.

Fix bug with Maps using array of values
Use the PR to look at coverage... need to write tests


lheagy:
nice! Would you mind doing this pr onto dev first? then we can merge from dev into master (http://nvie.com/posts/a-successful-git-branching-model/)

PR #332, opened 2016-06-06
Automate the epsilon picking based on percentile of model values for …

fourndo:
Automate the epsilon picking based on percentile of model values for DEFAULT mode. Fix example.

Fix bug with Maps using array of values.

### DIFF
--- a/SimPEG/Directives.py
+++ b/SimPEG/Directives.py
@@ -253,8 +253,7 @@ class SaveOutputDictEveryIteration(_SaveEveryIteration):
 class Update_IRLS(InversionDirective):
 
     eps_min = None
-    eps_p = None
-    eps_q = None
+    eps = None
     norms = [2.,2.,2.,2.]
     factor = None
     gamma = None
@@ -263,6 +262,7 @@ class Update_IRLS(InversionDirective):
     f_old = None
     f_min_change = 1e-2
     beta_tol = 5e-2
+    prctile = 95
 
     # Solving parameter for IRLS (mode:2)
     IRLSiter   = 0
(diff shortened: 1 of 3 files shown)


SimPEG/Directives.py
```diff
@@ -343,14 +356,14 @@ def endIter(self):
...
 
-            # Cool the threshold parameter if required
-            if getattr(self, 'factor', None) is not None:
-                eps = self.reg.eps / self.factor
-
-                if getattr(self, 'eps_min', None) is not None:
-                    self.reg.eps = np.max([self.eps_min,eps])
-                else:
-                    self.reg.eps = eps
+#            # Cool the threshold parameter if required
```
lheagy: do you want to remove the commented-out code? or is this in progress?

SimPEG/Maps.py
```diff
@@ -502,7 +502,9 @@ def __init__(self, mesh, indActive, valInactive, nC=None):
         if Utils.isScalar(valInactive):
             self.valInactive = np.ones(self.nC)*float(valInactive)
         else:
-            self.valInactive = valInactive.copy()
+            self.valInactive = np.ones(self.nC)
```
lheagy: what was the bug here?

fourndo: if the number of valInactive is different than nC, we were getting an error.

SimPEG/Directives.py
fourndo: We might want to explore the cooling schedule later when doing the non-linear stuff. So lets keep it for now.

lheagy: sounds good!

SimPEG/Maps.py
lheagy: ah, ok.

lheagy:
looks good to me!

lheagy:
@fourndo : feel free to merge if you are happy with this

PR #337, opened 2016-06-11
Feat/docs gae travis deploy

lheagy:
deploy the docs! https://simpegdocs.appspot.com/ (we can migrate this to http://docs.simpeg.xyz if ppl are happy with the docs)

- docs on google app engine
- On commits to master, the docs are deployed to master after a successful build 
- testing on docs 
- typo fix in MT_3D_Forward example when not using mumps solver
- update dc example to use refactored dc code
- cleanup of docstrings


lheagy:
@rowanc1 : what do you think of this?

rowanc1:
This deploys as soon as it gets into master? Have you tested the deploy process?

lheagy:
I have tested the deploy- http://simpegdocs.appspot.com was launched from
Travis (though I updated the version number assignment after that testing)

rowanc1:
Wonderful. Good to go then?

lheagy:
I think so! There will be a few improvements to make, but we can do that in
a separate pr

PR #338, opened 2016-06-14
Dev

lheagy:
deploy the docs! 

docs are deployed from travis to google app engine.


.travis.yml
```diff
@@ -46,12 +53,28 @@ install:
...
 # Run test
 script:
+  # test docs
   - nosetests $TEST_DIR --with-cov --cov SimPEG --cov-config .coveragerc -v -s
 
 # Calculate coverage
 after_success:
   - coveralls --config_file .coveragerc
 
+  - if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" ]; then
```
lheagy: we should do this in a makefile instead (it would clean up the travis file a lot)

SimPEG/MT/Utils/dataUtils.py
```diff
@@ -44,19 +44,19 @@ def rotateData(MTdata,rotAngle):
     return MT.Data.fromRecArray(outRec)
 
 
-def appResPhs(freq,z):
+def appResPhs(freq, z):
```
lheagy: could you add docstrings to these @grosenkj  :)

SimPEG/MT/Utils/dataUtils.py
```diff
@@ -44,19 +44,19 @@ def rotateData(MTdata,rotAngle):
...
     app_res = ((1./(8e-7*np.pi**2))/freq)*np.abs(z)**2
```
lheagy: where does the 8e-7 come from?

grosenkj: I used the values for mu_0 and omega and simplified. I need to clean out the Utils and fix things like this. 

Should use the constants here, would be much clearer.

PR #340, opened 2016-06-21
Analytics

sgkang:
Needs be merged to master to deploy on binder examples (dipole)


lheagy:
could you do this pr onto dev first? then we can do dev onto master

lheagy:
I am going to close this and instead we will merge #341 and then we can generate a pr from dev to master

PR #345, opened 2016-06-22
Feat/mappingderivs

### LINKED ISSUE #342: mapping.deriv
we should allow `maps.deriv` to take a vector and return a vector. Right now, we are passing around matrices (mostly sparse, so not really a big deal, but still it is unnecessary). If `v=None` we can simply return the matrix

lheagy:
mapping derivs can take a vector and return a vector or pass around matrices if `v=None`

This closes #342


lheagy:
@rowanc1 : what do you think of this?

SimPEG/Maps.py
```diff
@@ -90,13 +90,15 @@ def deriv(self, m):
...
+        if v is not None:
+            return v
         return sp.identity(self.nP)
 
     def test(self, m=None, **kwargs):
         """Test the derivative of the mapping.
 
             :param numpy.array m: model
-            :param kwargs: key word arguments of :meth:`SimPEG.Tests.checkDerivative`
+            :param kwargs: key word arguments of :math:`SimPEG.Tests.checkDerivative`
```
rowanc1: Is this math?

lheagy: ha, no it is not `meth` or `math`. This should be `ref` ?

rowanc1:
This looks like a good addition!

lheagy:
- [ ] `ParametricSplineMap` and 
- [ ] `ParametricPolyMap` 

are not currently tested. 
@sgkang : could you add tests for these?

lheagy:
@sgkang : could you add tests for 
- [ ] parametric spline and
- [ ] parametric poly 
maps?

Also: is the 'logSigma' flag necessary? could this be accomplished with a combo map?

PR #361, opened 2016-07-17
Tracking PR for Python 3 diffs

bsmithyman:
...not yet ready...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - 2.7
+  - 3.4
 
 sudo: false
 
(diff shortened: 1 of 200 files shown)


rowanc1:
@bsmithyman we are doing this mostly manually instead of introducing the `old_div` all over the place.

@jcapriot has put a lot of work into this, and I think it is pretty close!

bsmithyman:
@rowanc1: No I completely agree. `old_div` is bad news; I spent a couple of hours on the old `feat/Python3` branch manually dealing with those instances. Sounds like this is exactly the right direction, and nice work @jcapriot!

PR #365, opened 2016-07-22
fixed install on python3

wassname:
I tried install the Python3 branch but it had errors. These changes fixe them:
- 2to3 on setup.py
- removed uneeded circular import in Maps.py
- added missing old_div in mathutils (created by 2to3)

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -11,7 +11,6 @@ from . import Utils
 import numpy as np, scipy.sparse as sp
 from scipy.sparse.linalg import LinearOperator
 from .Tests import checkDerivative
-from .PropMaps import PropMap, Property
 from numpy.polynomial import polynomial
 from scipy.interpolate import UnivariateSpline
 import warnings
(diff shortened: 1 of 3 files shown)


lheagy:
Thanks @wassname! We only started the python3 upgrade last weekend - none of the testing is back up and running, so beware that a lot of things might be broken on that branch. Currently, master and dev rely on Python 2.7

wassname:
No problems, I understand it's a work in progress. But I'm finding simpeg useful so I wanted to push some fixes back in case they are useful when you tackle python3. Thanks for the work on it.

bsmithyman:
Yep, the changes are not really very close to complete on that branch. Debating if there is a better way to automate these things (currently it is in a post- `futurize --stage1; futurize --stage2; pasteurize` state).

rowanc1:
@wassname @bsmithyman @lheagy we are updating this on #401 thanks @wassname for your contributions here. @jcapriot has done a lot of manual work that I think is fixing most of the issues! We will hopefully get this into dev and then master in the next few days.

The python3 tests are super slow because we have not yet setup the solver in 3.

PR #369, opened 2016-07-28
Add comment lines removal in UBCoctree read-in function 

grosenkj:
Remove text that is behind  (! comment symbol) in the UBCoctree files that are being read in.


rowanc1:
This is closed in favour of #387.

PR #379, opened 2016-08-03
bug fix in cyl mesh face inner products

lheagy:
- use sympy to generate a test showing the bug in the cyl mesh inner product
- more robust testing on cyl mesh
- bug fix in cyl mesh inner products


SimPEG/Mesh/CylMesh.py
```diff
@@ -358,17 +358,15 @@ def _fastInnerProduct(self, projType, prop=None, invProp=False, invMat=False):
         if prop.size == self.nC:
             Av = getattr(self, 'ave'+projType+'2CC')
             Vprop = self.vol * Utils.mkvc(prop)
-            M = self.dim * Utils.sdiag(Av.T * Vprop)
+            nelts = np.sum(getattr(self, 'vn'+projType).nonzero())
+            M =  nelts * Utils.sdiag(Av.T * Vprop)
```
lheagy: @rowanc1 : can you help me sort through this conceptually?

lheagy:
@rowanc1, see what you think!

.travis.yml
```diff
@@ -42,7 +42,7 @@ before_install:
   - conda update --yes conda
 
 install:
-  - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib cython ipython nose vtk sphinx
+  - conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib cython ipython nose vtk sphinx sympy
```
lheagy: are you good with sympy being a dependency? (makes writing those types of tests a bit easier)

SimPEG/Mesh/CylMesh.py
```diff
@@ -383,46 +428,67 @@ def _fastInnerProductDeriv(self, projType, prop, invProp=False, invMat=False):
...
+                dMdprop = nelts * Utils.sdiag(- MI.diagonal()**2) * Av.T * V
 
         if tensorType == 1:
             Av = getattr(self, 'ave'+projType+'2CC')
             V = Utils.sdiag(self.vol)
+            nelts = np.sum(getattr(self, 'vn'+projType).nonzero())
+
             if not invMat and not invProp:
-                dMdprop = self.dim * Av.T * V
+                dMdprop = nelts * Av.T * V
```
lheagy: is `nelts` (number of elements, 1 for edges, 2 for faces on cyl mesh) a clear enough variable name? any other suggestions

.travis.yml
rowanc1: We are not putting this in the req.txt?

Here it is completely fine!

lheagy: I did put it in req.txt, but it is not necessary, could be removed and only put in .travis

requirements.txt
```diff
@@ -3,3 +3,4 @@ scipy
 ipython
 matplotlib
 cython
+sympy
```
rowanc1: I would prefer this not to be a direct dep? Just in Travis.

SimPEG/Mesh/CylMesh.py
rowanc1: `n_elements`?

Does this generalize to the innerproducts file directly?

lheagy: I have generalized it now. There are a couple if statements, which isn't great, but overall most of the code is repeated, so nice to have it back in the tensor mesh.

PR #380, opened 2016-08-04
Em/srcdipoleorientations

lheagy:
- define a dipole with a moment (scalar) and orientation (vector)
- allow 'x', 'y','z' orientations for FDEM dipole source


lheagy:
These same changes are on #382 (and #382 does a better job). So to get those updates, you can merge em/dev into analytics once #382 has gone in.

PR #381, opened 2016-08-05
Em/ref/tdem

sgkang:
Refactoring of TDEM code.

- Fixed memory issues in TDEM code
- Still have some todos:

- [x] General waveform (includes fwd and sensitivity)
- [ ] Line sources
- [ ] def fields_nostore (For forward modelling we do not require storing all fields)


sgkang:
Not sure why this keep failing ... in test_TDEM_crosscheck.py 
(no problem in local machine) any idea? @lheagy

lheagy:
lets keep the scope of this pr to replacing the functionality of the old tdem code which it currently should do. adding `fields no store` and a line source can be addressed in a separate pr

PR #384, opened 2016-08-07
Em/dev onto analytics

lheagy:
- allow multiple source orientations
- added testing for sources
- allow secondary fields to be pulled out of fields object


lheagy:
The merge conflict should be isolated to a the test_FDEM_analytics. @micmitch if you want to resolve this, merge em/dev into analytics locally. I am happy to help if you run into any issues!

PR #387, opened 2016-08-08
Add comment lines removal in UBC Octree

rowanc1:


rowanc1:
This is a direct pr onto dev and should replace/address:

#368 and #369

lheagy:
looks good!

PR #326, opened 2016-05-31
Delete DCIP folder.

rowanc1:
I have deleted the DCIP folder.

Can @sgkang and @fourndo confirm that all functions have been moved to EM/Statics?

PR #388, opened 2016-08-08
DC example 

lheagy:
DCIP pseudosection example removed from pr #326 to be addressed separately
@fourndo, @sgkang


rowanc1:
This has been merged in and kept updated. Closing this PR.

PR #391, opened 2016-08-11
build docs for dc, some pep8 styling while at it

lheagy:
image

@sgkang, see what you think!


sgkang:
That looks good to me, I'll start adding some doc for DC and IP stuff. 
Nice work ! @lheagy

PR #392, opened 2016-08-14
Dev

lheagy:
- build docs for dc
- light pep8 formatting in DC code 
(from #391)


SimPEG/EM/Static/DC/FieldsDC.py
```diff
@@ -4,7 +4,7 @@
 from scipy.constants import epsilon_0
 
 
-class Fields(SimPEG.Problem.Fields):
+class FieldsDC(SimPEG.Problem.Fields):
```
sgkang: That's more explicit. Nice.

lheagy: perfect - this also makes docs easier :)

PR #399, opened 2016-08-21
Py3

jcapriot:

2 python3 tests fail:


bsmithyman:
Fantastic. This has been needed for a while, and will really help!

PR #400, opened 2016-08-21
Py3

jcapriot:

2 python3 tests fail:


jcapriot:
Passes.
So I still might have missed some integer divisions "//". But I can vouch for the 75% of the code base that the examples touch.

SimPEG/DataMisfit.py
```diff
@@ -1,11 +1,14 @@
...
-import Utils
-import Survey
-import Problem
+from SimPEG import Utils
+from SimPEG import Survey
+from SimPEG import Problem
 
+from six import add_metaclass
 
+@add_metaclass(Utils.SimPEGMetaClass)
```
rowanc1: Is there a reason that you did `add_metaclass` rather than `with`?

SimPEG/Examples/PF_Gravity_Inversion_Linear.py
```diff
@@ -145,7 +145,7 @@ def run(plotIt=True):
         mesh.plotSlice(m_l2, ax=ax, normal='Z', ind=zpanel,
                        grid=True, clim=(model.min(), model.max()))
         plt.plot(([mesh.vectorCCx[0], mesh.vectorCCx[-1]]),
-                 ([mesh.vectorCCy[ypanel], mesh.vectorCCy[ypanel]]), Color='w')
+                 ([mesh.vectorCCy[ypanel], mesh.vectorCCy[ypanel]]), color='w')
```
rowanc1: this was failing to start with? interesting..

SimPEG/Examples/__init__.py
```diff
@@ -1,30 +1,31 @@
...
+from SimPEG.Examples import Mesh_QuadTree_Creation
+from SimPEG.Examples import Mesh_QuadTree_FaceDiv
+from SimPEG.Examples import Mesh_QuadTree_HangingNodes
+from SimPEG.Examples import Mesh_Tensor_Creation
+from SimPEG.Examples import MT_1D_ForwardAndInversion
+from SimPEG.Examples import MT_3D_Foward
+from SimPEG.Examples import PF_Gravity_Inversion_Linear
+from SimPEG.Examples import PF_Magnetics_Analytics
+from SimPEG.Examples import PF_Magnetics_Inversion_Linear
+from SimPEG.Examples import Utils_surface2ind_topo
```
rowanc1: A lot of this is done through auto. Did you end up fixing that as well?

SimPEG/FLOW/Richards/Empirical.py
```diff
@@ -1,13 +1,15 @@
...
 
+@add_metaclass(Utils.SimPEGMetaClass)
 class NonLinearMap(object):
     """
     SimPEG NonLinearMap
 
     """
 
-    __metaclass__ = Utils.SimPEGMetaClass
+    #__metaclass__ = Utils.SimPEGMetaClass
```
rowanc1: We can probably kill these lines. In fact the whole `SimPEGMetaClass` actually needs to be removed.

image

Rather silly.

SimPEG/DataMisfit.py
jcapriot: Both of them worked, I just thought that the decorator form using <add_metaclass> was a little clearer than the <with_metaclass>, because it explicitly states which thing you're adding is the metaclass.
Also the codecy check likes it better.

SimPEG/Examples/__init__.py
```diff
@@ -61,12 +62,12 @@
         if line == "##### AUTOIMPORTS #####\n":
             inimports = not inimports
             if inimports:
-                out += '\n'.join(["import {0!s}".format(_) for _ in exfiles])
+                out += '\n'.join(["from SimPEG.Examples import {0!s}".format(_) for _ in exfiles])
```
jcapriot: Yep, this should have been the fix here for the autogenerated imports

rowanc1:
Taken over in #401.

PR #401, opened 2016-08-22
Py3k

rowanc1:
Clone of #400 

2 python3 tests fail:


.travis.yml
```diff
@@ -57,8 +57,7 @@ install:
       git clone https://github.com/rowanc1/pymatsolver.git;
       cd pymatsolver; python setup.py install; cd ..;
     else
-      git clone https://github.com/dwfmarchant/pyMKL.git;
-      cd pyMKL; python setup.py install; cd ..;
+      echo "pymatsolver not updated for python3";
```
rowanc1: @jcapriot did you completely remove all the mkl/paradiso stuff from this PR?

jcapriot: I reverted back to the previous so it should have undone everything.

rowanc1:
image


The tests seem to be working on my computer. There does seem to be some issues however with the FDEM deriv tests? Does anyone have an idea of why this might not work?

jcapriot:
Can we do a check in the pardiso solver interface to ensure that the sparse matrix is in a CSR format, and that it's "has_sorted_indices" property is true? I don't see that last check in the pyMKL package.

rowanc1:
@bsmithyman do you have any off the cuff ideas?!

rowanc1:
At least the fail is consistent now.

SimPEG/EM/Static/Utils/StaticUtils.py
```diff
@@ -312,7 +313,7 @@ def xy_2_r(x1,x2,y1,y2):
             srcClass = DC.Src.Dipole([rxClass], M[0,:], N[-1,:])
         SrcList.append(srcClass)
     else:
-        print """stype must be either 'pole-dipole', 'dipole-dipole' or 'gradient'. """
+        print("""stype must be either 'pole-dipole', 'dipole-dipole' or 'gradient'. """)
```
lheagy: Question: why 3 " ?

SimPEG/EM/Utils/testingUtils.py
```diff
@@ -82,10 +83,11 @@ def getFDEMProblem(fdemType, comp, SrcList, freq, useMu=False, verbose=False):
     prb.pair(survey)
 
     try:
-        from pymatsolver import MumpsSolver
-        prb.Solver = MumpsSolver
-    except ImportError, e:
+        from pymatsolver import PardisoSolver
```
lheagy: likely better to do the import at the top
```python
try: 
    from pymatsolver import PardisoSolver as Solver
except ImportError:
    from SimPEG import SolverLU as Solver
```

SimPEG/Examples/DC_Analytic_Dipole.py
```diff
@@ -27,9 +30,9 @@ def run(plotIt=True):
     problem = DC.Problem3D_CC(mesh)
     problem.pair(survey)
     try:
-        from pymatsolver import MumpsSolver
-        problem.Solver = MumpsSolver
-    except Exception, e:
+        from pymatsolver import PardisoSolver
+        problem.Solver = PardisoSolver
```
lheagy: same here, we should move imports to top of file

SimPEG/Mesh/MeshIO.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np, os
 from SimPEG import Utils
+import six
```
lheagy: has this been added to requirements.txt? (should it be?)

SimPEG/Mesh/View.py
```diff
@@ -675,41 +679,6 @@ def plotImage(self, I, ax=None, showIt=False, grid=False, clim=None):
...
         ax.set_xlabel('x')
         ax.set_ylabel('y')
-        if showIt: plt.show()
+        if showIt:
+            plt.show()
         return [scalarMap]
-
-
-if __name__ == '__main__':
-    import numpy as np
```
lheagy: did this get added back somewhere as an example? @rowanc1

SimPEG/EM/Static/Utils/StaticUtils.py
rowanc1: Could be fixed, might have been multiline before?

tests/em/fdem/forward/test_FDEM_analytics.py
```diff
@@ -48,9 +49,9 @@ def setUp(self):
         prb.pair(survey)
 
         try:
-            from pymatsolver import MumpsSolver
-            prb.Solver = MumpsSolver
-        except ImportError, e:
+            from pymatsolver import PardisoSolver
```
lheagy: same here - to top of file

lheagy:
@jcapriot , @rowanc1 : do you have some recommendations for tutorials on python 3, relative imports etc? It would be great if we could include these in the [simpeg/tutorials](https://github.com/simpeg/tutorials) to support the development team!

SimPEG/Mesh/View.py
rowanc1: #403

PR #406, opened 2016-08-23
Em/removedatapair

sgkang:
Remove datapair in Jvec for FDEM and MT
(Should handle memory leackage)

PR #407, opened 2016-08-23
Remove the datapair

rowanc1:
Merging into the correct branch. See #406 for original comments.


rowanc1:
@lheagy @sgkang good to go?

lheagy:
looks good to me!

PR #415, opened 2016-09-02
Py3k to dev

lheagy:
- resolve merge conflicts between Py3k, dev in #401 
- update re-factored TDEM and tests so they are python 3 compatible


rowanc1:
:+1:

PR #420, opened 2016-09-03
Dev

lheagy:
- python 3!


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: python
 python:
+  - 3.4
```
lheagy: do we want to test both on all commits? or should we just do this on dev, master (and pr's to dev, master)?

PR #434, opened 2016-09-17
Dev

lheagy:
travis typo fixes


lheagy:
we will want to remove the version restrictions on matplotlib once #435 has been resolved

PR #438, opened 2016-09-20
mt/refactRx into mt/dev

grosenkj:
Moving refactoring of 
- Receivers to classes
- Derivatives return Zero instead of None
- Namespace updated and directory moved to EM.NSEM


lheagy:
could you please do this pr onto mt/dev first? Thanks!

grosenkj:
I changed the base to mt/dev, since that was to intended target of this pr. I don't know that got mixed up, but time for some sleep now. 
It be nice to try to get this sorted tomorrow and move some of this code down.

PR #439, opened 2016-09-20
build the nsem docs

lheagy:
Build the NSEM docs!

image


tests/docs/test_docs.py
```diff
@@ -13,11 +13,13 @@ def test_html(self):
...
-            "{0!s}".format((self.path_to_docs)),
-            "{0!s}".format((html_path))])
-        assert check == 0
+        check = subprocess.Popen(("sphinx-build", "-nW", "-b", "html", "-d",
+            "{0!s}".format(doctrees_path),
+            "{0!s}".format(self.path_to_docs),
+            "{0!s}".format(html_path)),
+            stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, err = check.communicate()
+        assert check == 0, 'Test failed with stdout:\n{}\n and stderr:\n{}\n'.format(out,err)
```
lheagy: personally, I find this to be more difficult to sort through 
image

The previous version is syntax highlighted in travis, so I would advocate for that. Th

image

Thoughts @grosenkj , @rowanc1 ?

rowanc1: Do both fail? One is inside of a nosetest?

lheagy: both fail. This is the log for the first example (https://travis-ci.org/simpeg/simpeg/jobs/161787181) this is the second (https://travis-ci.org/simpeg/simpeg/jobs/161414352, or similarly from em.geosci: https://travis-ci.org/ubcgif/em/builds/124879386) .

grosenkj: I agree, the second formatted one is much easier to read.  
I was running the test locally and I wasn't getting good error reporting. It looks like travis grabs them and prints them out and by redirecting them it doesn't happen. 
I vote for the old version.

PR #442, opened 2016-09-22
blanket ignore scipy.org in the linkcheck - it can be slow to load

lheagy:
blanket ignore linkcheck errors on:
- http://wiki.python.org/
- http://docs.scipy.org/

These can be slow to load, but do exist.


lheagy:
These links seem to have stabilized, so this is not needed

PR #446, opened 2016-09-26
Dev

lheagy:
- typo fixes in travis for auto-deploys
- update paper example to start with citation


lheagy:
@rowanc1 : thoughts on getting this in? bumpversion and go??

PR #447, opened 2016-09-26
array * Zero() == Zero()!!

rowanc1:
Seems that it the `__numpy_ufunc__` flag it is there this is good enough  …

I am actually a bit unsure of why this works.
to see more:

http://docs.scipy.org/doc/numpy-dev/reference/arrays.classes.html#numpy.class.__numpy_ufunc__

Since we have all other functions implemented:

	"your __numpy_ufunc__ method will not be called"

I figure if it is a boolean, not a callable, we will figure this out
pretty fast if it starts to fail in the future.

This adds a test to assert that we get a zero back on multiplication.

### DIFF
--- a/SimPEG/Utils/matutils.py
+++ b/SimPEG/Utils/matutils.py
@@ -3,6 +3,7 @@ import numpy as np
 import scipy.sparse as sp
 from .codeutils import isScalar
 
+
 def mkvc(x, numDims=1):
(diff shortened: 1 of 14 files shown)


lheagy:
image

rowanc1:
@lheagy can we merge this in? It looks like the links are not passing. but that was solved in #449

PR #449, opened 2016-09-26
patch linkchecking 

lheagy:
allow retrys for pinging links, increase timeout time


rowanc1:
So great. Looking forward to this getting in!

PR #467, opened 2016-10-04
Projection Map

lheagy:
Simple projection map to select indices of a model 

cc @fourndo

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -709,6 +709,51 @@ class ComplexMap(IdentityMap):
     # inverse = deriv
 
 
+class Projection(IdentityMap):
+    """
+        A map to rearrange / select parameters
+
+        :param int nP: number of model parameters
+        :param numpy.array index: indices to select
+    """
+
+    def __init__(self, nP, index, **kwargs):
+        assert isinstance(index, (np.ndarray, slice, list)), (
+            'index must be a np.ndarray or slice, not {}'.format(type(index)))
+        super(Projection, self).__init__(nP=nP, **kwargs)
+
+        if isinstance(index, slice):
+            index = list(range(*index.indices(self.nP)))
+        self.index = index
+        self._shape = nI, nP = len(self.index), self.nP
+
+        # sparse projection matrix
+        self.P = sp.csr_matrix(
+            (np.ones(nI), (range(nI), self.index)), shape=(nI, nP)
+        )
+
+    def _transform(self, m):
+        return m[self.index]
+
+    @property
+    def shape(self):
+        """
+        Shape of the matrix operation (number of indices x nP)
+        """
+        return (len(self.index), self.nP)
+
+    def deriv(self, m, v=None):
+        """
+            :param numpy.array m: model
+            :rtype: scipy.sparse.csr_matrix
+            :return: derivative of transformed model
+        """
+
+        if v is not None:
+            return self.P * v
+        return self.P
+
+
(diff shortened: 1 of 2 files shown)


SimPEG/Maps.py
```diff
@@ -709,6 +709,51 @@ def adj(v):
...
+    """
+
+    def __init__(self, nP, index, **kwargs):
+        assert isinstance(index, (np.ndarray, slice, list)), (
+            'index must be a np.ndarray or slice, not {}'.format(type(index)))
+        super(Projection, self).__init__(nP=nP, **kwargs)
+
+        if isinstance(index, slice):
+            index = list(range(*index.indices(self.nP)))
+        self.index = index
```
rowanc1: does it matter here that index is either a list or a numpy array?

rowanc1: Do we want an assertion that `max(index) < nP` just for sanity?

SimPEG/Maps.py
```diff
@@ -709,6 +709,51 @@ def adj(v):
...
+        self._shape = nI, nP = len(self.index), self.nP
+
+        # sparse projection matrix
+        self.P = sp.csr_matrix(
+            (np.ones(nI), (range(nI), self.index)), shape=(nI, nP)
+        )
+
+    def _transform(self, m):
+        return m[self.index]
```
rowanc1: `assert len(m) == self.nP, "Ensure that the model has {} elements.".format(self.nP)`

SimPEG/Maps.py
```diff
@@ -709,6 +709,51 @@ def adj(v):
...
+
+    @property
+    def shape(self):
+        """
+        Shape of the matrix operation (number of indices x nP)
+        """
+        return (len(self.index), self.nP)
```
rowanc1: `return self._shape` from above?

tests/base/test_maps.py
```diff
@@ -246,5 +246,17 @@ def test_ParametricSplineMap(self):
...
         self.assertTrue(mParamSpline.testVec())
 
+    def test_Projection(self):
+        nP = 10
+        m = np.arange(nP)
+        self.assertTrue(np.all(Maps.Projection(nP, slice(5))*m == m[:5]))
+        self.assertTrue(np.all(Maps.Projection(nP, slice(5, None))*m == m[5:]))
+        self.assertTrue(np.all(Maps.Projection(nP, np.r_[1, 5, 3, 2, 9, 9])*m ==
+                        np.r_[1, 5, 3, 2, 9, 9]))
+        self.assertRaises(Exception, Maps.Projection(nP, [1, 5, 3, 2, 9, 9]))
```
rowanc1: why does this raise?

SimPEG/Maps.py
lheagy: not necessarily -> you may want to repeat indices.

SimPEG/Maps.py
lheagy: perhaps that should be implemented in the base maps. We do not currently perform this check on other maps

tests/base/test_maps.py
lheagy: it is a list

lheagy: do we want to support lists?

SimPEG/Maps.py
lheagy: yep, good call

tests/base/test_maps.py
lheagy: we should support lists though

PR #470, opened 2016-10-12
ref/JvecRx

lheagy:
Move call of the fields deriv into the receivers. This is one step to using the same Jvec, Jtvec code between fdem, nsem

cc: @grosenkj

### DIFF
--- a/SimPEG/EM/FDEM/RxFDEM.py
+++ b/SimPEG/EM/FDEM/RxFDEM.py
@@ -40,7 +40,7 @@ class BaseRx(SimPEG.Survey.BaseRx):
 
         return P*f_part
 
-    def evalDeriv(self, src, mesh, f, v, adjoint=False):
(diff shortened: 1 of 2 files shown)


lheagy:
@rowanc1 : are you good if I merge this?

SimPEG/EM/FDEM/RxFDEM.py
```diff
@@ -40,7 +40,7 @@ def eval(self, src, mesh, f):
 
         return P*f_part
 
-    def evalDeriv(self, src, mesh, f, v, adjoint=False):
+    def evalDeriv(self, src, mesh, f, du_dm_v=None, v=None, adjoint=False):
```
rowanc1: Just moving this deriv into the Rx? Cool.

rowanc1:
That looks solid.

SimPEG/EM/FDEM/RxFDEM.py
lheagy: yep!

PR #471, opened 2016-10-23
properties

rowanc1:
# Properties

In this version `PropMaps` and indiscriminate `mappings` have been removed. A `Depreciate` module has been provided to help with the upgrade process.

Most details are in #444, also includes a lot of whitespace cleaning.

### Features
- WireMap
- Props
- Depreciate

### Happy changes
- tests are now on images and plotting in the examples
- lots of whitespace and PEP8 updates
- Props can be pickled
- Forward simulations without mappings (e.g. set `simga` directly)

### Changes
- dependency on the `properties` library (`v0.2.1`) http://propertiespy.readthedocs.io/en/latest/
- reference to `curModel` has been replaced with `model`

# Examples

Below are a few examples for the `Wires`, `Props` and `Depreciate`.

## Properties

The `properties` library allows for documentation and change `observer`s on things like the `model` change. It also can be serialized and provides auto documentation on the classes. Below is an example contrasting the changes before and after the `PropMap` removal. All classes have been updated internally, however, because of the brittleness of this class, there are some definitions outside of the `SimPEG` codebase.

#### Old code
```python
class EMPropMap(Maps.PropMap):

    sigma = Maps.Property(
        "Electrical Conductivity",
        defaultInvProp=True, 
        propertyLink=('rho', Maps.ReciprocalMap)
    )

    mu = Maps.Property(
        "Inverse Magnetic Permeability",
        defaultVal=mu_0,
        propertyLink=('mui', Maps.ReciprocalMap)
    )

    rho = Maps.Property(
        "Electrical Resistivity",
        propertyLink=('sigma', Maps.ReciprocalMap)
    )
    
    mui = Maps.Property(
        "Inverse Magnetic Permeability",
        defaultVal=1./mu_0,
        propertyLink=('mu', Maps.ReciprocalMap)
    )
```

#### New code

The new `Props` are bound directly on the `Problem` object and there is no default `mapping`. This means that you **must** specify the mapping that you intend to use explicitly.

```python
class BaseEMProblem(Problem.BaseProblem):

    sigma, sigmaMap, sigmaDeriv = Props.Invertible(
        "Electrical conductivity (S/m)"
    )

    rho, rhoMap, rhoDeriv = Props.Invertible(
        "Electrical resistivity (Ohm m)"
    )

    Props.Reciprocal(sigma, rho)

    mu = Props.PhysicalProperty(
        "Magnetic Permeability (H/m)",
        default=mu_0
    )
    mui = Props.PhysicalProperty(
[...]

PR #473, opened 2016-10-24
Em/dev

rowanc1:


lheagy:
@fourndo , @sgkang: can you guys take a look at this pr? There is some white-space fixing, this is minor... The major item is the removal of the PropMap and use of Properties. These changes touch the EM, PF, Static codes, including IP, so your thoughts would be much appreciated! A more detailed overview of the changes is in #471.

lheagy:
- [ ] can we throw an error if you try to set curModel? (now `model` needs to be set).

rowanc1:
Good point. Add a `@property` stub that just raises an `Exception`?

rowanc1:
@sgkang did you merge all of the em/dev branch back into itself? What happened on the last commit? there are 108 changes?

sgkang:
I am pretty sure that is pulling the latest em/dev. I was way behind when I comitted changes, then pulled.

lheagy:
I will reopen a pull request from https://github.com/simpeg/simpeg/pull/473/commits/0c6e9e3d1922d6dc6a94994631ad990d1c340e7f

PR #475, opened 2016-10-24
curModel property stub 

lheagy:
add property stub for the curModel so that an exception is raised if you try to set it


lheagy:
I will re-open this onto the new em/dev

PR #320, opened 2016-05-26
add parametrized block in a layer map

lheagy:
Added a parametrized block in a layer map defined with arctans 

image

TODO:
- [x] docs
- [x] example

PR #477, opened 2016-10-24
Maps/feat parametrized block

lheagy:
content from #320

Added Maps
------------
- parametrized layer map
- parametrized block in a layer map
- parametrized  casing and layer (cyl mesh only) 
and associated documentation and tests for these maps

Added Sources
---------------
- primary-secondary mapped sigma for primary-secondary problems that include derivatives in the primary fields 

Added Examples
-----------------
- Primary-Secondary Casing example from the EM paper
- added example for Layer in Half-space map
- added example for Block in Layer map


lheagy:
I will re-open this onto the updated em/dev

PR #478, opened 2016-10-24
Em/dev

lheagy:
Pr content from #473 with the implementation of properties. The full description of changes is outlined in #471


rowanc1:
@fourndo I think if we can get this onto dev first, before your branch, that would be best.
I will take a look this eve.

PR #479, opened 2016-10-24
curModel property stub

lheagy:
add property stub for the curModel so that an exception is raised if you try to set it

(originally from #475)


rowanc1:
Minor change to actually test the set callback. The Exception was raised, and second line does not test. hence 75% test coverage.

PR #485, opened 2016-10-25
Ref/maps

lheagy:
- add comment to break up Maps.py file into 
    - Meshless Maps (eg. `ExpMap`, `LogMap`...) that do not require a mesh
    - Surjection, Injection and Interpolation Maps
    - Parametric Maps
- cleaned up init'ing so that meshless maps do not require a mesh. 
- moved the location of the `Weighting` map and `ComplexMap` so that they are under the Meshless Map header

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -361,7 +361,7 @@ class Wires(object):
 
 ###############################################################################
 #                                                                             #
-#                    Mesh Independent Maps                                    #
+#                          Mesh Independent Maps                              #
 #                                                                             #
 ###############################################################################
 
@@ -499,6 +499,7 @@ class LogMap(IdentityMap):
     def inverse(self, m):
         return np.exp(Utils.mkvc(m))
 
+
 class Weighting(IdentityMap):
     """
         Model weight parameters.
@@ -579,7 +580,7 @@ class ComplexMap(IdentityMap):
 
 ###############################################################################
 #                                                                             #
-#            Surjection, Injection and Interpolation Maps                     #
+#                 Surjection, Injection and Interpolation Maps                #
 #                                                                             #
 ###############################################################################
 
@@ -841,7 +842,7 @@ class InjectActiveCells(IdentityMap):
 
 ###############################################################################
 #                                                                             #
-#                         Parametric Maps                                     #
+#                             Parametric Maps                                 #
 #                                                                             #
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


SimPEG/Maps.py
```diff
@@ -359,6 +359,12 @@ def nP(self):
         return self._nP
 
 
+###############################################################################
+#                                                                             #
+#                    Mesh Independent Maps                                    #
```
rowanc1: this isn't centered at all. sigh.

rowanc1: :)

lheagy: you need to tilt your head a bit :)

SimPEG/Maps.py
```diff
@@ -490,6 +499,90 @@ def deriv(self, m, v=None):
...
+                'parameters'
+            )
+
+        super(Weighting, self).__init__(mesh=mesh, nP=nP, **kwargs)
+
+        if weights is None:
+            weights = np.ones(self.nP)
+
+        self.weights = np.array(weights, dtype=float)
+        self.P = Utils.sdiag(self.weights)
```
rowanc1: I wonder if we need to create this even? or maybe lazily?

because we could just multiply `m` directly.

SimPEG/Maps.py
```diff
@@ -361,7 +361,7 @@ def nP(self):
 
 ###############################################################################
 #                                                                             #
-#                    Mesh Independent Maps                                    #
+#                          Mesh Independent Maps                              #
```
rowanc1: much better!

rowanc1:
This looks great. feel free to merge if it passes!

SimPEG/Maps.py
```diff
@@ -491,6 +500,93 @@ def inverse(self, m):
...
+    def _transform(self, m):
+        return self.weights*m
+
+    def inverse(self, D):
+        return self.weights**(-1.) * D
+
+    def deriv(self, m, v=None):
+        if v is not None:
+            return self.weights * v
+        return self.P
```
rowanc1: Here. need an `@property` for `P` and lazy load it.

SimPEG/Maps.py
```diff
@@ -491,6 +500,93 @@ def inverse(self, m):
...
+
+        self.weights = np.array(weights, dtype=float)
+
+    @property
+    def shape(self):
+        return (self.nP, self.nP)
+
+    @property
+    def P(self):
+        return Utils.sdiag(self.weights)
```
lheagy: P is still formed, just not stored. There isn't really any reason to store it

PR #489, opened 2016-10-27
Em/dev

lheagy:
content from #320, #477, #480 

Added Maps
------------
- parametrized layer map
- parametrized block in a layer map
- parametrized  casing and layer (cyl mesh only) 
and associated documentation and tests for these maps

Added Sources
---------------
- primary-secondary mapped sigma for primary-secondary problems that include derivatives in the primary fields 

Added Examples
-----------------
- Primary-Secondary Casing example from the EM paper
- added example for Layer in Half-space map
- added example for Block in Layer map


lheagy:
@rowanc1 : should we do another code review on this before bringing it in?

.travis.yml
```diff
@@ -49,9 +49,9 @@ before_install:
 install:
 
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then
-      conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib cython ipython nose sphinx sympy vtk;
+      conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib cython ipython h5py nose sphinx sympy vtk;
```
rowanc1: can we add these to a `requirements_dev.txt`

rowanc1: maybe an issue

rowanc1:
I did not review throughly. There may be some conflicts with the changes i made on `dcip/dev` but we can deal with that later.

.travis.yml
lheagy: good call, I have started #491. We can take care of this in a separate pr

PR #493, opened 2016-11-01
Change input file in Gravity Driver

fourndo:
Small changes to Grav Driver for Craig's example to be published


SimPEG/PF/GravityDriver.py
```diff
@@ -190,7 +190,8 @@ def activeCells(self):
             self._activeCells = inds
 
             # Reduce m0 to active space
-            self._m0 = self.m0[self._activeCells]
+            if len(self.m0) > len(self._activeCells):
```
lheagy: do you want to throw a warning here?

PR #496, opened 2016-11-02
Add Craig's Laguna grav inversion example

fourndo:


SimPEG/Examples/PF_Gravity_Laguna_del_Maule_Inversion.py
```diff
@@ -0,0 +1,247 @@
+# coding: utf-8
+'''
```
rowanc1: Can this be moved to document the run function (that way we pick it up in the docs). 

After, run `python __init__.py` again!

Can you also please use `"""` doc quotes rather than `'''`. :)

rowanc1:
@fwkoch it looks like there was some unintended consequences of the `properties` update today. I think just the `_get` changed to not taking the default.

rowanc1:
Getting PR #498 into dev and doing an update to this branch will hopefully fix the issues.

PR #501, opened 2016-11-08
Update Maps.py

sgkang:
Fix indexing bug for Wire map

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -354,7 +354,7 @@ class Wires(object):
             wire = Projection(self.nP, slice(start, start + arg[1]))
             setattr(self, arg[0], wire)
             maps += [(arg[0], wire)]
-            start = arg[1]
+            start += arg[1]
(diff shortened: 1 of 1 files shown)


rowanc1:
Oh boy. Thanks @sgkang.

SimPEG/Maps.py
```diff
@@ -354,7 +354,7 @@ def __init__(self, *args):
             wire = Projection(self.nP, slice(start, start + arg[1]))
             setattr(self, arg[0], wire)
             maps += [(arg[0], wire)]
-            start = arg[1]
+            start += arg[1]
```
lheagy: nice catch!

sgkang:
No worries:0 
Here is what we could do those wires:0 Actually having fun with wires

image

(Current source (vector) inversion for SP problem)

rowanc1:
Can you give a presentation soon to update people with what is going on!?

Looks awesome.

PR #503, opened 2016-11-09
Em/waveform ex

sgkang:
It seems that I don't have access to push this branch. 

Anyway, here is an example for 1D TDEM inversion with VTEM waveform @lheagy


lheagy:
Thanks @sgkang! em/dev is now like dev - only pr's onto it, no direct pushing  

image

rowanc1:
I think init py needs to run again to make the rst?

Issue #465, opened 2016-10-01
rx.projComp should be rx.orientation


lheagy:
This is confusing

![image](https://cloud.githubusercontent.com/assets/6361812/19016983/1d628024-87df-11e6-8960-eefca9dc9418.png)

rowanc1:
![image](https://cloud.githubusercontent.com/assets/913249/21528264/2e55559a-cce6-11e6-921e-6447fd9d4ae1.png)

lheagy:
right now, we are using "orientation" for `x`, `y`, `z` and still using `component` for `real` `imag`. I think this is working okay. Feel free to re-open if you would like to continue the discussion.

Issue #464, opened 2016-10-01
Forward Simulation .validate()


lheagy:
Prior to computing anything, there are some sanity checks that would be useful to run first. For example: 
- check that the end-product of a mapping is the same as `mesh.nC`
- check that receiver locations are inside the mesh 

as well as problem specific sanity checks 
- FDEM: check that mesh beyond skin depth
- TDEM: check that mesh beyond diffusion distance, check that time, space discretization is stable

rowanc1:
There is some work that we just did on validators:
3ptscience/properties#45

rowanc1:
From @lheagy:

I think that is everything for the problem. For a forward simulation, we could also validate that 
- [ ] mesh size > skin depth of lowest freq
- [ ] smallest cells are < skin depth of highest freq
  (perhaps not error on these but warn)
- [ ] check that src loc, rx locs are in the domain

lheagy:
I have started playing around with some of these ideas in my research repo: https://github.com/lheagy/casingResearch/blob/master/CasingSimulations/Run.py#L101. I would be curious to hear thoughts, feedback

PR #508, opened 2016-11-18
Em/feat/muinversion

lheagy:
Machinery to invert for magnetic permeability from FDEM data
- derivatives and tests for all 4 FDEM formulations
- example of inverting for mu_r (relative permeability)
- addition of a Mu_relative map, ChiMap
- Cyl-mesh specific interpolation (important for CCV interpolation) 

image

@rowanc1 : could you take a look?


SimPEG/EM/Base.py
```diff
@@ -126,6 +126,19 @@ def MfMui(self):
             self._MfMui = self.mesh.getFaceInnerProduct(self.mui)
         return self._MfMui
 
+    def MfMuiDeriv(self, u):
+        """
+        Derivative of :code:`MfMui` with respect to the model.
+        """
+        if (
```
rowanc1: minor comment on style:

I would return Zero early. It is the easy case to exclude and allows you to whittle down assumptions and get to the hard things last.

SimPEG/EM/Base.py
```diff
@@ -135,6 +148,23 @@ def MfMuiI(self):
             self._MfMuiI = self.mesh.getFaceInnerProduct(self.mui, invMat=True)
         return self._MfMuiI
 
+    # TODO: This should take a vector
+    def MfMuiIDeriv(self, u):
+        """
+        Derivative of :code:`MeSigma` with respect to the model
+        """
+        # TODO: only works for diagonal tensors. getEdgeInnerProductDeriv,
```
rowanc1: can we raise an exception instead? There are tensor utils available to figure this out.

SimPEG/EM/Base.py
```diff
@@ -135,6 +148,23 @@ def MfMuiI(self):
             self._MfMuiI = self.mesh.getFaceInnerProduct(self.mui, invMat=True)
         return self._MfMuiI
 
+    # TODO: This should take a vector
+    def MfMuiIDeriv(self, u):
+        """
+        Derivative of :code:`MeSigma` with respect to the model
```
rowanc1: Incorrect doc string.

SimPEG/EM/Base.py
```diff
@@ -145,6 +175,19 @@ def MeMu(self):
             self._MeMu = self.mesh.getEdgeInnerProduct(self.mu)
         return self._MeMu
 
+    def MeMuDeriv(self, u):
+        """
+        Derivative of :code:`MfMui` with respect to the model.
```
rowanc1: `MfMu` not `MfMui`

SimPEG/EM/Base.py
```diff
@@ -154,6 +197,23 @@ def MeMuI(self):
             self._MeMuI = self.mesh.getEdgeInnerProduct(self.mu, invMat=True)
         return self._MeMuI
 
+    # TODO: This should take a vector
+    def MeMuIDeriv(self, u):
+        """
+        Derivative of :code:`MeSigma` with respect to the model
```
rowanc1: doc string.

SimPEG/EM/Base.py
```diff
@@ -154,6 +197,23 @@ def MeMuI(self):
...
+        """
+        # TODO: only works for diagonal tensors. getEdgeInnerProductDeriv,
```
rowanc1: I would again raise a `NotImplementedError`

SimPEG/EM/Base.py
```diff
@@ -154,6 +197,23 @@ def MeMuI(self):
...
+        #       invMat=True should be implemented in SimPEG
+
+        if (
+            getattr(self, 'muMap', None) is not None or
```
rowanc1: Do we need to be careful with the attr not existing?
Do we have to check both mu and mui maps? If one is set the other should be too?

```python
if self.muMap is None:
    return Utils.Zero()
dMeMuI_dI = -self.MeMuI**2
dMe_dmu = self.mesh.getEdgeInnerProductDeriv(self.mu)(u)
return dMeMuI_dI * (dMe_dmu * self.muDeriv)
```

I think that is much cleaner. Would need to validate that `muiMap` not being `None` would still work.

SimPEG/EM/Base.py
```diff
@@ -170,7 +230,15 @@ def MeSigmaDeriv(self, u):
         """
         Derivative of MeSigma with respect to the model
         """
-        return self.mesh.getEdgeInnerProductDeriv(self.sigma)(u) * self.sigmaDeriv
+        if (
```
rowanc1: same comment here. I think we can just use `sigmaMap`?

SimPEG/EM/Base.py
```diff
@@ -189,9 +257,14 @@ def MeSigmaIDeriv(self, u):
         # TODO: only works for diagonal tensors. getEdgeInnerProductDeriv,
```
rowanc1: raise exception?

SimPEG/EM/FDEM/FieldsFDEM.py
```diff
@@ -344,11 +345,12 @@ def startup(self):
         self._MeSigma = self.survey.prob.MeSigma
         self._MeSigmaDeriv = self.survey.prob.MeSigmaDeriv
         self._MfMui = self.survey.prob.MfMui
+        self._MfMuiDeriv = self.survey.prob.MfMuiDeriv
 
     def _GLoc(self, fieldType):
         if fieldType == 'e':
```
rowanc1:
```python
in ['e', 'eSecondary']
```

??

SimPEG/EM/FDEM/ProblemFDEM.py
```diff
@@ -361,20 +441,41 @@ def getADeriv(self, freq, u, v, adjoint=False):
...
+            return MfMuiDeriv.T * (C * (MeSigmaI.T * (C.T * v)))
+        return C * (MeSigmaI * (C.T * (MfMuiDeriv * v)))
+
+    def getADeriv(self, freq, u, v, adjoint=False):
+        if adjoint is True and self._makeASymmetric:
+            v = self.MfMui * v
+
+        ADeriv =  (
+            self.getADeriv_sigma(freq, u, v, adjoint) +
+            self.getADeriv_mui(freq, u, v, adjoint)
```
rowanc1: ha. so easy.

SimPEG/EM/FDEM/ProblemFDEM.py
```diff
@@ -361,20 +441,41 @@ def getADeriv(self, freq, u, v, adjoint=False):
...
+        MfMuiDeriv = self.MfMuiDeriv(u)
+        MeSigmaI = self.MeSigmaI
+        C = self.mesh.edgeCurl
+
+        if adjoint:
+            return MfMuiDeriv.T * (C * (MeSigmaI.T * (C.T * v)))
+        return C * (MeSigmaI * (C.T * (MfMuiDeriv * v)))
+
+    def getADeriv(self, freq, u, v, adjoint=False):
+        if adjoint is True and self._makeASymmetric:
```
rowanc1: perhaps we need some more validation thoughts.

If all properties are not invertible. Etc. This should be introduced as a slightly more general concept. Could you help me think through some ideas of things to check:

- [ ] has a model (to do an early exit of zero)
- [ ] all physical properties are satisfied or have mappings
- [ ] the mappings are of the right size

The model being set should do some of these checks as well. and the `m=None` on fields and data calculation.

SimPEG/EM/FDEM/RxFDEM.py
```diff
@@ -11,17 +12,27 @@ class BaseRx(SimPEG.Survey.BaseRx):
     """
 
     def __init__(self, locs, orientation=None, component=None):
-        assert(orientation in ['x','y','z']), "Orientation {0!s} not known. Orientation must be in 'x', 'y', 'z'. Arbitrary orientations have not yet been implemented.".format(orientation)
-        assert(component in ['real', 'imag']), "'component' must be 'real' or 'imag', not {0!s}".format(component)
+        assert(
+            orientation in ['x', 'y', 'z']
```
rowanc1: seriously need some `properties` stuff in here.

SimPEG/EM/FDEM/RxFDEM.py
```diff
@@ -11,17 +12,27 @@ class BaseRx(SimPEG.Survey.BaseRx):
...
+        ), "Orientation {0!s} not known. Orientation must be in 'x', 'y', 'z'."
+        " Arbitrary orientations have not yet been implemented.".format(
+            orientation
+        )
+        assert(
+            component in ['real', 'imag']
```
rowanc1:
```python
component = properties.StringChoice(
    'Component of the fields object',
    choices=('real', 'imag')
)
```

SimPEG/Examples/EM_FDEM_Mu_Inversion.py
```diff
@@ -0,0 +1,181 @@
...
+
+    mtrue = mur[active]
+
+    # Maps
+    actMap = Maps.InjectActiveCells(mesh, active, mur_air, nC=mesh.nCz)
+    surj1Dmap = Maps.SurjectVertical1D(mesh)
+    murMap = Maps.Mu_relative(mesh)
+
+    # Mapping
+    muMap =  murMap * surj1Dmap * actMap
```
rowanc1: should test with a `mui` inversion as well?

rowanc1: I suppose this is an example. Could have it as an option?

SimPEG/Maps.py
```diff
@@ -519,6 +520,56 @@ def inverse(self, m):
         return np.exp(Utils.mkvc(m))
 
 
+class ChiMap(IdentityMap):
```
rowanc1: @fourndo are you aware that this is coming?

SimPEG/Maps.py
```diff
@@ -519,6 +520,56 @@ def inverse(self, m):
...
+    def deriv(self, m, v=None):
+        if v is not None:
+            return mu_0 * v
+        return mu_0 * sp.eye(self.nP)
+
+    def inverse(self, m):
+        return m / mu_0 - 1
+
+
+class Mu_relative(IdentityMap):
```
rowanc1: cool map. We might want to change `Maps.py` into a folder?

`Maps.EM.Mu_relative` etc.

SimPEG/Mesh/CylMesh.py
```diff
@@ -367,6 +367,52 @@ def aveF2CCV(self):
                                           'yet implemented')
         return self._aveF2CCV
 
+
+    def getInterpolationMat(self, loc, locType='CC', zerosOutside=False):
```
rowanc1: we didn't have this?

SimPEG/Mesh/TensorMesh.py
```diff
@@ -217,7 +229,7 @@ def isInside(self, pts, locType='N'):
             inside = inside & (pts[:,i] >= tensor.min()-TOL) & (pts[:,i] <= tensor.max()+TOL)
         return inside
 
-    def getInterpolationMat(self, loc, locType='CC', zerosOutside=False):
+    def _getInterpolationMat(self, loc, locType='CC', zerosOutside=False):
```
rowanc1: what is the rename here?

SimPEG/Mesh/CylMesh.py
rowanc1: I guess you answered it in your notes above. CCv needs specific treatment in the Cyl mesh.

SimPEG/EM/FDEM/FieldsFDEM.py
lheagy: we don't currently have an eSecondary receiver implemented, but I guess there is no harm in allowing those fields to be retrieved.

SimPEG/EM/Base.py
lheagy: right now, the mu inversion is only implemented in the frequency domain, so if you were to set up a time domain problem, the attribute currently does not exist. 
Perhaps though, getting a 'muMap' does not exsist error would be ok in the mean-time?

SimPEG/EM/Base.py
lheagy: Thanks for catching this

SimPEG/EM/FDEM/RxFDEM.py
lheagy: I am still not sure if 'component' is the right word. I had a hard time explaining this - generally when referring to 'component' the first association is with the spatial component. eg. "we are measuring the x-component of the electric field". Perhaps `part in ["real", "imag"]` ?

SimPEG/Examples/EM_FDEM_Mu_Inversion.py
lheagy: the tests are testing all formulations, so we will hit examples where the mui map is used. The tests could be expanded to include setting both mu and mui maps. I think generally, one would invert for 'mu', thinking in terms of mu-inverse is pretty unnatural.

SimPEG/Mesh/CylMesh.py
lheagy: yep

SimPEG/Mesh/TensorMesh.py
lheagy: This is so I can call it from the CylMesh Interpolation. The Edge, CC and Face interpolation from Tensor Mesh are all valid for Cyl

image

SimPEG/Examples/EM_FDEM_Mu_Inversion.py
lheagy: I have added mui inversions to the testing

SimPEG/EM/Base.py
lheagy: good call, thanks!

SimPEG/EM/Base.py
lheagy: yep, good call

SimPEG/EM/FDEM/RxFDEM.py
rowanc1: This would also become a boolean `real_component = True`

SimPEG/Mesh/TensorMesh.py
rowanc1: Could alternatively call `super` or `BaseTensorMesh.getInterpolationMat(self, ...)`  for future reference. I think this is fine.

SimPEG/EM/FDEM/ProblemFDEM.py
lheagy: I think that is everything for the problem. For a forward simulation, we could also validate that - [ ] mesh size > skin depth of lowest freq
- [ ] smallest cells are < skin depth of highest freq
(perhaps not error on these but warn)
- [ ] check that src loc, rx locs are in the domain

SimPEG/EM/FDEM/ProblemFDEM.py
lheagy: :)

SimPEG/EM/FDEM/RxFDEM.py
lheagy: yes we do. I think this should be addressed in a separate pr though?

SimPEG/EM/FDEM/RxFDEM.py
lheagy: it could be - I think it is odd to treat them unequally though. Keeping it as a string, I could see us allowing:
`["real", "in-phase"]` and `["imag", "imaginary", "quadrature", "out-of-phase" ]`

@dougoldenburg: any suggestions on what to label `real` and `imag` data (does "component" make sense, or do you have another suggestion?)

SimPEG/Mesh/TensorMesh.py
lheagy: ah, I didn't know super could do that. clever!

SimPEG/EM/FDEM/RxFDEM.py
lheagy: I am currently working on implementing this on the em/ref/tdem-hj branch

lheagy:
@rowanc1. It looks like the default value for mu is now not being set for the FDEM problem. Another issue currently is that I wanted to inject properties into Maps (by inheriting BaseSimPEG from Props), but since Props imports Maps, it was circular. Perhaps we move BaseSimPEG elsewhere? 

cc @fwkoch

lheagy:
@rowanc1: what are your thoughts on bringing this in?

SimPEG/EM/FDEM/RxFDEM.py
rowanc1: I like `part`. Measure the x-`component` and take the real `part`.

SimPEG/EM/Base.py
```diff
@@ -132,6 +145,26 @@ def MfMuiI(self):
             self._MfMuiI = self.mesh.getFaceInnerProduct(self.mui, invMat=True)
```
rowanc1: Is this used anywhere?

SimPEG/EM/FDEM/ProblemFDEM.py
```diff
@@ -15,28 +20,46 @@ class BaseFDEMProblem(BaseEMProblem):
...
+        \\\(\\\mathbf{h}\\\)
 
     """
 
     surveyPair = SurveyFDEM
     fieldsPair = FieldsFDEM
 
+    mu, muMap, muDeriv = Props.Invertible(
+        "Magnetic Permeability (H/m)",
+        default=mu_0
```
rowanc1: did the default thing get fixed?

SimPEG/EM/FDEM/RxFDEM.py
rowanc1: See #465

SimPEG/EM/FDEM/ProblemFDEM.py
rowanc1: I added a bunch of the checks/summary to PR #519 
Copied over the other comment to #464

SimPEG/EM/FDEM/RxFDEM.py
```diff
@@ -11,17 +12,27 @@ class BaseRx(SimPEG.Survey.BaseRx):
...
+            ), "'component' must be 'real' or 'imag', not {0!s}".format(
+                component
+            )
 
         self.projComp = orientation
         self.component = component
 
-        SimPEG.Survey.BaseRx.__init__(self, locs, rxType=None) #TODO: remove rxType from baseRx
+        # TODO: remove rxType from baseRx
```
rowanc1: #524

SimPEG/EM/FDEM/SrcFDEM.py
```diff
@@ -488,6 +488,33 @@ def s_e(self, prob):
 
             return -C.T * (MMui_s * self.bPrimary(prob))
 
+    def s_eDeriv(self, prob, v, adjoint=False):
+        if not hasattr(prob, 'muMap') or not hasattr(prob, 'muiMap'):
+            return Zero()
+        else:
+            formulation = prob._formulation
+
+            if formulation is 'EB':
```
rowanc1: this should be changed to `==` rather than `is`

SimPEG/EM/FDEM/SrcFDEM.py
```diff
@@ -488,6 +488,33 @@ def s_e(self, prob):
...
+                    self.bPrimary(prob)
+                ) * prob.muiDeriv
+                C = prob.mesh.edgeCurl
+
+                if adjoint:
+                    return -MMui_sDeriv.T * (C * v)
+
+                return -C.T * (MMui_sDeriv * v)
+
+            elif formulation is 'HJ':
```
rowanc1: `==`

SimPEG/Examples/EM_FDEM_Mu_Inversion.py
```diff
@@ -0,0 +1,181 @@
...
+    )
+
+    srcList = [
+        FDEM.Src.MagDipole([bzr], freq, srcLoc, orientation='Z')
+        for freq in freqs
+    ]
+
+    surveyFD = FDEM.Survey(srcList)
+    prbFD = FDEM.Problem3D_b(
+        mesh, sigma=surj1Dmap * sigma, muMap=muMap, Solver=Solver
```
rowanc1: This is cool that you use a map to set sigma, but don't put it in the model location.

SimPEG/EM/Base.py
lheagy: not at the moment.

lheagy: it might be used in the TDEM implementation later on

SimPEG/EM/FDEM/RxFDEM.py
lheagy: sounds good. we can address this in a separate pr

SimPEG/EM/FDEM/ProblemFDEM.py
lheagy: yes, the default bug was fixed in #513

SimPEG/Examples/EM_FDEM_Mu_Inversion.py
lheagy: :)

PR #510, opened 2016-11-21
Ref/treemesh

lheagy:
- modularize faceDiv calculation further so that the stencil can be accessed
- have regularization construct a cell diff from a face div if no cell differencing operators are available 
- whitespace cleanup in TreeMesh.py
- make `_cellGradStencil` behave as a property rather than a function for differential operators

cc @micmitch


SimPEG/Regularization.py
```diff
@@ -193,7 +193,19 @@ def cellDiffx(self):
         :return: differencing matrix for active cells in the x-direction
         """
         if getattr(self, '_cellDiffx', None) is None:
-            self._cellDiffx = self._Pafx.T * self.mesh.cellGradx * self._Pac
+            if getattr(self.mesh, 'cellGradx', None) is not None:
+                cellDiffx = self.mesh.cellGradx
+            elif getattr(self.mesh, 'faceDivx', None) is not None:
```
lheagy: @rowanc1 : do you have thoughts on a more elegant implementation for this (getting the interior cells and dealing with hanging faces)? This is a bit of a trick, but does do the job

lheagy:
@micmitch: which branch is this content now on in discretize?

micmitch:
I moved this content to discretize/dev/TreeTesting. I'm reasonably certain that all the changes made it over there. DC inversions on a tree mesh current run on this branch but we have tearing where cell sizes change. After speaking with @lheagy it seems as though we need to go back and build up the cellGrad function more carefully to account for hanging faces.

lheagy:
This is now on the TreeTesting branch of discretize

Issue #474, opened 2016-10-24
Composable Objective Functions


rowanc1:
# State of the SimPEG

Currently we have a number of `Regularization` and `DataMisfit` classes which implement parts of the objective function. These are kept separate and explicitly combined in the `InvProblem`.

``` python
dmis = DataMisfit.l2_DataMisfit(survey)
reg = Regularization.Tikhonov(mesh)
invProb = InvProblem.BaseInvProblem(dmis, reg, opt)
```

In the `Tikhonov` regularization there are a number of fitting parameters:
`alpha_s`, `alpha_x`, `alpha_y`, `alpha_z`, `alpha_xx`, `alpha_yy`, and `alpha_zz`
these in turn combined with a linear operator (derivative or volume term):
`Wx`, `Wy`, `Wz`, `Wsmall`, `Wxx`, `Wyy`, `Wzz`, and `Wsmooth2`
finally these are put together in a single operator `W`, which allows you combine in a single call:

``` python
weighted_residual = W * (mapping * (m - mref))
obj_function = 0.5 * r.dot(r)
```

Not only is this repeated seven times in a single class, the idea is also repeated throughout the `DataMisfit` classes as well! This means if we implement something like total variation, we have to duplicate that code in a lot of places.

Yikes.

What I would like to talk about in this issue is that we can break up the `Regularization` classes into a bunch of composable `ObjectiveFunction`s which can be added together and multiplied by scalars (the `alpha`). These should also be used by the `DataMisfit` classes.
# Bugs in the current idea
1. The `alpha`s and `W`s are super repetitive
2. We cannot really do joint inversions (multiple physics, or properties)
3. Higher order regularizations should be composable (e.g. `Tikhonov`)
4. There is only a single `beta` applied in the inversion class, this is not extensible
### Current Workarounds

There is a `MultiRegularization` class made by @sgkang in the SIP folder which allows multiple models to be regularized at the same time.
You can get around a lot of it by doing `InjectActiveCell` maps or writing other custom `Regularization` classes. e.g. in joint inversion for  volume and slowness:
http://gif.eos.ubc.ca/sites/default/files/HeaglySEG2014.pdf
### Other discussions

This was also brought up in the SimPEG meeting by @fourndo on October 4th, 2016
https://youtu.be/A_oy9w8J0q4?t=24m18s
This goes through some of the group discussions on composable objective functions.
# Some constraints
- Keep the `Tikhonov` idea, although this may turn into a function rather than a class
[...]

PR #511, opened 2016-11-23
Ref/objectivefunctions

lheagy:
# Composable Objective Functions
First pass implementation to address #474  
- wrote a base objective function class that allows for composable objective functions, so you can now say
```python
phi = alpha_s * Regularization.Smallness(mesh) + alpha_x * Regularization.Smoothness(mesh) 
```
- regularization and datamisfit now inherit from this. This results in several namespace changes
    - `reg.eval(m)` --> `reg(m)`, `dmisfit.eval(m)` --> `dmisfit(m)`
    - `reg.evalDeriv(m)` --> `reg.deriv(m)`, `dmisfit.evalDerivl(m)` --> `dmisfit.deriv(m)`
    -  `reg.eval2Deriv(m)` --> `reg.deriv(m)`, `dmisfit.eval2Derivl(m)` --> `dmisfit.deriv(m)`
- all regularizations and datamisfit now each have their own individual `W` matrix. There are no longer differently named W's, eg
    - `dmistfit.Wd = Utils.sdiag(1/wd)` --> `dmisfit.W = Utils.sdiag(1/wd)`

## Regularization
- regularization refactor to be more modular. 
    - For example, `Regularization.Tikhonov` is now a composite regularization that has an objective function list (`reg.objfcts`) consisting of `Regularization.Smallness`, `Regularization.Smooth_x`, `Regularization.Smooth_y`, `Regularization.Smooth_z`, `Regularization.Smooth_xx`, `Regularization.Smooth_yy`, `Regularization.Smooth_zz`.
- implemented properties in regularization (requires dev right now) for 
    - `indActive`
    - `mref`
    - `cell_weights`
    - `mrefInSmooth`
    - `alpha_?` (x, y, ...) 
- any time properites are updated on a `CompositeRegularization`, those changes are propagated to the objective functions that make it up. 

## Other impacts
- update IRLS directive to hold the `l2model` (the directive should keep track of that model, it does not need to be stored on the regularization) cc @fourndo 
- updated `Props` to include a `SimPEGArray` and `SimPEGFloat` that allow `Utils.Zero` and `Utils.Identity` @rowanc1 
- update examples to set data weights on `dmisfit.W` instead of `dmisfit.Wd` as that will depreciate on these changes cc @fourndo 

## Prior to merging
- [x] @rowanc1 : could you please have a look at this and we can chat through it
- [x] @fourndo: once I have gone through this with Rowan and we make sure the implementation is what we are after, I would like to get your thoughts on the Sparse Regularization changes and do a thorough review to make sure I haven't broken anything! 
- [x] we should wait to merge until properties/dev is on master 

# Straight Ray Tomography
- implement `SEIS.StraightRay`, `Problem`, `Survey`, `Src` and `Rx`
- example: SEIS_StraightRayTomo.py

image 

# InvProb Compatibility with Composable Objective functions
- allow both `dmisfit` to be either a base objective function or a base data misfit
- extend invProb in cases where `dmisfit` is a combo objective function so that it can stash multiple fields from different problems 

# Example Joint Inversion
- include an example "joint" inversion (consists of 2 data misfit terms and a regularization) based on the Heagy et. al., 2014 SEG abstract http://library.seg.org/doi/abs/10.1190/segam2014-1639.1: SEIS_VolumeRegularization.py

image


fourndo:
Ok sweet. Doesn't seem like too much work on my end. I think it is for the best

lheagy:
only the dev branch of properties will work at the moment

leouieda:
@lheagy @fourndo @rowanc1 this is very similar to what I ended up doing in Fatiando a while ago. I just wanted to share some thoughts I've been having. This is what it looks like in Fatiando (from this [gallery example](http://www.fatiando.org/dev/gallery/seismic/srtomo_regularized.html#sphx-glr-gallery-seismic-srtomo-regularized-py)):

    solver = srtomo.SRTomo(tts, srcs, recs, mesh)
    smooth = solver + 1e8*Smoothness2D(mesh.shape)  
    smooth.fit()

But I ended up changing my mind about this implementation because it places a lot of work on the end user of the inversion. At least in my implementation, there was no way for the developer of an inversion to specify what the default/recommended regularization should be (aside from a comment on the docstring) and the user needs to import the regularization classes all the time. 

It is really cool to be able to write code that looks like the equations that we see all the time in papers. But that only serves the ones that have a deep understanding of this. Someone who just wants to run an inversion on their data might not appreciate this as much as we do.

I'm still a bit conflicted about this and just wanted to share another perspective.

rowanc1:
I remember talking through some of these with you at scipy a few years ago.

I think how we want to deal with each audience in the geophysical inversion community is through a layered/nested approach.

The API we are describing here is targeted at advanced users who are developing new inversion algorithms - a few people at CSM are working on clustering algorithms - joint inversions - coordinate descent etc. As you note, I think we also need a higher level API that probably does not go into details **at all** about how the various objective functions are composed, but instead choose sensible (hopefully robust) defaults and algorithms.

So instead of sacrificing and finding one api that suites everyones needs, create a nested structure that serves multiple audiences.

It would be really interesting to hear your thoughts on the design of the higher level one as well.

lheagy:
Thanks @leouieda, I agree with @rowanc1, that our first target user is the advanced user and subsequent layers can be designed to support more applied users. Perhaps it is worth drafting a "design philosophy" document?

leouieda:
@rowanc1 and @lheagy it's great that you are planning for a higher level interface. I think that was my main blunder in Fatiando. I completely agree that things should be infinitely configurable to advanced users. It would be great if the lower level interface could be accessed from the higher level one. For example, an expert won't want to manually configure every single inversion, so they would use the high API but then when some customization is required they could plug in a different regularization/optimization into the high level API.

SimPEG/Maps.py
```diff
@@ -29,11 +29,19 @@ def __init__(self, mesh=None, nP=None, **kwargs):
         Utils.setKwargs(self, **kwargs)
 
         if nP is not None:
+            if isinstance(nP, str):
```
rowanc1: this should probably be `string_types` from `six`

SimPEG/Regularization.py
```diff
@@ -61,320 +87,439 @@ def dim(self):
...
+        elif getattr(self.regmesh, 'nC') != '*':
+            return self.regmesh.nC
+        else:
+            return '*'
 
-        if mesh is not None and nP is None:
-            nP = self.regmesh.nC
-        self.nP = nP
+    @property
+    def mesh(self):
```
rowanc1: `Props.Mesh`? This could also have `allowed_mesh_types` or something. Which would be helpful...

SimPEG/Regularization.py
```diff
@@ -416,655 +563,1030 @@ def eval2Deriv(self, m, v=None):
...
 
-    @Utils.timeIt
-    def _evalSmoothx(self, m):
-        if self.mrefInSmooth == True:
-            r = self.Wx * ( self.mapping * (m - self.mref) )
-        elif self.mrefInSmooth == False:
-            r = self.Wx * ( self.mapping * (m) )
-        return 0.5 * r.dot(r)
+        super(BaseComboRegularization, self).__init__(
+            objfcts=[], multipliers=None
```
rowanc1: here we should pass on `objfcts` instead of another empty array?

SimPEG/Regularization.py
```diff
@@ -416,655 +563,1030 @@ def eval2Deriv(self, m, v=None):
...
-    def _evalSmooth(self, m):
-        phiSmooth = self._evalSmoothx(m)
-        if self.regmesh.dim > 1:
-            phiSmooth += self._evalSmoothy(m)
-        if self.regmesh.dim > 2:
-            phiSmooth += self._evalSmoothz(m)
-        return phiSmooth
+    # Properties
+    alpha_s = Props.SimPEGFloat("smallness weight")
+    alpha_x = Props.SimPEGFloat("weight for the first x-derivative")
```
rowanc1: these alphas are around for *all* combo regularizations? this seems a bit strange ... much more like tikhonov only?

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+        return self._eval(x, **kwargs)
+
+    @property
+    def nP(self):
+        if getattr(self, '_nP', None) is None:
+            self._nP = '*'
+        return self._nP
+
+    @Utils.timeIt
+    def _eval(self, x, **kwargs):
```
lheagy: My only thought with keeping `_eval` is that it can still be found with tab complete. I could be convinced either way though

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+
+        return ComboObjectiveFunction(
+            objfcts=objfctlist, multipliers=multipliers
+        )
+
+    def __radd__(self, objfct2):
+        return self+objfct2
+
+    def __mul__(self, multiplier):
+        return ComboObjectiveFunction([self], [multiplier])
```
lheagy: there is a check in the `__init__` of the `ComboObjectiveFunction`

image

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+            assert isinstance(fct, BaseObjectiveFunction), (
+                "Unrecognized objective function type {} in objfcts. All "
+                "entries in objfcts must inherit from "
+                "ObjectiveFunction".format(fct.__class__.__name__)
+            )
+
+            # ensure all objective functions have the same nP
+            if fct.nP != '*':
+                if nP != '*':
+                    assert nP == fct.nP, (
```
lheagy: - the objective function does need to keep track of `nP`. If we are adding two objective functions together, they need to take the same size of input. 
- currently, the base objective function does not have a mapping. Currently only the `Regularization` has a mapping, but for joint inversions, a mapping might be necessary. 

So prior to merging, 
- [ ] the BaseObjective function should have a mapping

SimPEG/Maps.py
lheagy: nice catch, thanks!

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+                        "Objective Functions must all have the same nP, not "
+                        "{}".format([f.nP for f in objfcts])
+                    )
+                else:
+                    nP = fct.nP
+            self.objfcts.append(fct)
+
+        if multipliers is None:
```
lheagy: When a single objective function is evaluated, we want to be able to see the output prior to taking a product with a multiplier (e.g. in the iteration printers). As a simple example, we want to print `phi_d`, `phi_m` and `beta` independently, which seemed much less transparent to me if beta is attached to `phi_m`: what should `__call__` do? 

Alternatively, it can be useful to print the smallness, and smoothness_x, smoothness_y, ... independently to ensure that alphas are chosen correctly. 

So we do need a way to evaluate the objective function and later take a product with a scalar in a transparent way. My thinking in putting it on the `ComboObjectiveFunction` is that it really only makes sense to use a multiplier if we are combining objective functions - otherwise it really has no impact on the inversion.

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+                "Length of multipliers ({}) must be the same as the length of "
+                "objfcts ({})".format(len(multipliers), len(self.objfcts))
+            )
+        self._multipliers = multipliers
+
+        super(ComboObjectiveFunction, self).__init__(**kwargs)
+        self._nP = nP
+
+    @property
+    def multipliers(self):
```
lheagy: or perhaps this is solved with properties? 

we do want to be able to change the multipliers - treating them as directives, but I agree there needs to be further thought into how we do this in a safe way

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+            elif isinstance(objfct_H, Utils.Identity) and self.nP != '*':
+                H += multpliter * sp.eye(self.nP)  # if we need a shape
+            else:
+                H += multpliter * objfct_H
+        return H
+
+    # This assumes all objective functions have a W.
+    # The base class currently does not.
+    @property
+    def W(self):
```
lheagy: no. I thought it might be handy to be able to interact with, but it is not actually used anywhere in the code

SimPEG/Regularization.py
```diff
@@ -416,655 +563,1030 @@ def eval2Deriv(self, m, v=None):
...
-    @Utils.timeIt
-    def _evalSmooth(self, m):
-        phiSmooth = self._evalSmoothx(m)
-        if self.regmesh.dim > 1:
-            phiSmooth += self._evalSmoothy(m)
-        if self.regmesh.dim > 2:
-            phiSmooth += self._evalSmoothz(m)
-        return phiSmooth
+    # Properties
+    alpha_s = Props.SimPEGFloat("smallness weight")
```
lheagy: `Props.Float` will fail if a `Utils.Zero()` is used. Currently `Utils.Zero()` is being used to indicate that a function should not be evaluated.

rowanc1: I was suggesting to rename this *and* keep the functionality.

SimPEG/Regularization.py
lheagy: Right now, all the combo regularizations use these.
`Tikhonov`, `Simple` and `Sparse` use them. 
We could instead move them to `Simple` and have `Tikhonov` and `Sparse` inherit `Simple`? (This choice is a bit arbitrary though). what do you think @rowanc1 ?

SimPEG/Regularization.py
```diff
@@ -416,655 +563,1030 @@ def eval2Deriv(self, m, v=None):
...
+    :param float alpha_xx: (default 1) smoothness weight for second derivative in the x-direction
+    :param float alpha_yy: (default 1) smoothness weight for second derivative in the y-direction
+    :param float alpha_zz: (default 1) smoothness weight for second derivative in the z-direction
+    """
 
-        #     assert np.mod(nmod, 1) > 1e-8, 'Mismatch between model and mesh'
+    def __init__(
+        self, mesh,
+        alpha_s=1e-6, alpha_x=1.0, alpha_y=1.0, alpha_z=1.0,
+        alpha_xx=Utils.Zero(), alpha_yy=Utils.Zero(), alpha_zz=Utils.Zero(),
```
lheagy: right now yes, but not in a safe way... 
- [ ] The combo regularization just doesn't include them in the list. it would be better to keep them in the `objfcts` list, but simply not evaluate. This can be fixed prior to merging

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+
+        super(ComboObjectiveFunction, self).__init__(**kwargs)
+        self._nP = nP
+
+    @property
+    def multipliers(self):
+        return self._multipliers
+
+    def _eval(self, x, **kwargs):
+        f = Utils.Zero()
```
lheagy: yeah, that is a good call

SimPEG/Regularization.py
lheagy: ah, I see

SimPEG/Regularization.py
```diff
@@ -61,320 +87,439 @@ def dim(self):
...
 
 
-class BaseRegularization(object):
+###############################################################################
+#                                                                             #
+#                          Single Regularization                              #
+#                                                                             #
+###############################################################################
+
+class BaseRegularization(ObjectiveFunction.L2ObjectiveFunction):
```
lheagy: yeah, thats fair

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,258 @@
...
+            else:
+                assert(W.shape[0]) == self.nP, (
+                    'nP must be the same as W.shape[0], not {}'.format(self.nP)
+                )
+        self._W = W
+
+    @property
+    def W(self):
+        if getattr(self, '_W', None) is None:
+            if self._nP != '*' and self._nP is not None:
```
lheagy: right now, nP isn't a `property` so we need to set the private, but we should implement a property for this. 
`Props.Integer` ?

SimPEG/Regularization.py
```diff
@@ -61,320 +87,439 @@ def dim(self):
...
+    :param int nP: number of parameters
+    :param IdentityMap mapping: regularization mapping, takes the model from model space to the space you want to regularize in
+    :param numpy.ndarray mref: reference model
+    :param numpy.ndarray indActive: active cell indices for reducing the size
+    of differential operators in the definition of a regularization mesh
 
     """
 
     counter = None
+    mapPair = Maps.IdentityMap  #: Base class of expected maps
```
lheagy: `Props.Mapping` has a few things baked-in that are not strictly necessary here (eg. the reciprocal) so I think it is worth teasing that apart before having objective functions rely on it

SimPEG/DataMisfit.py
```diff
@@ -69,60 +33,108 @@ class l2_DataMisfit(BaseDataMisfit):
...
+        return self._W
 
-            self._Wd = Utils.sdiag(1/(abs(survey.dobs)*survey.std+survey.eps))
-        return self._Wd
-
-    @Wd.setter
-    def Wd(self, value):
-        self._Wd = value
+    @W.setter
+    def W(self, value):
```
lheagy: we should add a check that allows the user to input a vector and we cast it to a sparse diag matrix

SimPEG/Directives.py
```diff
@@ -313,6 +314,9 @@ class Update_IRLS(InversionDirective):
     coolingFactor = 2.
     coolingRate = 1
 
+    # store the l2model
+    l2model = None
```
lheagy: @fourndo: I have moved the storing of the `l2model` onto the IRLS directive instead of the regularization (eg to fetch the `l2model`, you now need to use 
```
IRLS = Directives.Update_IRLS(norms=([0, 1, 1, 1]),  eps=None, f_min_change=1e-2, minGNiter=3)
update_Jacobi = Directives.Update_lin_PreCond()
inv = Inversion.BaseInversion(invProb, directiveList=[IRLS, betaest, update_Jacobi])

inv.run(m)
l2model = IRLS.l2model
```
The `l2model` is not used by the regularization, and is quite specific to the use of this directive.

SimPEG/Directives.py
```diff
@@ -416,12 +415,10 @@ def endIter(self):
 
             # Reset the regularization matrices so that it is
             # recalculated for current model
-            self.reg._Wsmall = None
-            self.reg._Wx = None
-            self.reg._Wy = None
-            self.reg._Wz = None
-            self.reg._W = None
-            self.reg._Wsmooth = None
+            # self.reg._Wsmall = None
```
lheagy: these are no longer stored, so they do not need to be cleared

SimPEG/Maps.py
```diff
@@ -61,8 +68,6 @@ def shape(self):
             :rtype: tuple
             :return: shape of the operator as a tuple (int,int)
         """
-        if self._nP is not None:
-            return (self.nP, self.nP)
```
lheagy: this was a bug - we should not assume that the matrix is square.

SimPEG/Optimization.py
```diff
@@ -7,80 +7,158 @@
 norm = np.linalg.norm
 
 
-__all__ = ['Minimize', 'Remember', 'SteepestDescent', 'BFGS', 'GaussNewton', 'InexactGaussNewton', 'ProjectedGradient', 'NewtonRoot', 'StoppingCriteria', 'IterationPrinters']
```
lheagy: only whitespace changes in this file, no code changes.

SimPEG/DataMisfit.py
```diff
@@ -1,65 +1,45 @@
...
+    @property
+    def nP(self):
+        if self._mapping is not None:
+            return self.mapping.nP
+        elif self.prob.model is not None:
+            return len(self.prob.model)
+        else:
+            return '*'
 
+    @property
```
lheagy: it might be better to do a soft depreciation on this - allow there to be a setter that throws a warning and sets W instead of Wd. thoughts @rowanc1 ?

SimPEG/DataMisfit.py
lheagy: done.

rowanc1:
In conversation with @lheagy:
- [ ] Depreciate the `Smooth_xx(mesh)` classes in favour of `Smooth2(mesh, orientation='x')`

This creates less classes and will be a bit more future friendly (if we have random orientations etc.). All of these classes should have an `alpha` parameter. They gain meaning when they are attached to a `Tikhonov` class as `alpha_s`,`alpha_x`, etc.

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,326 @@
...
+        )
+
+    def test(self, x=None, num=4, plotIt=False, **kwargs):
+        deriv = self._test_deriv(x=x, num=num, **kwargs)
+        deriv2 = self._test_deriv2(x=x, num=num, plotIt=False, **kwargs)
+        return (deriv & deriv2)
+
+    __numpy_ufunc__ = True
+
+    def __add__(self, objfct2):
```
lheagy: adding these in a reasonable way is a bit subtle... so far, this is the best I have come up with

image

SimPEG/Utils/matutils.py
```diff
@@ -536,10 +536,15 @@ def __gt__(self, v):
         return 0 > v
 
     @property
-    def transpose(self): return Zero()
+    def transpose(self):
+        return Zero()
```
rowanc1: we have T and transpose.

Also I think this could be `return self`

tests/base/test_Objectivefct.py
```diff
@@ -0,0 +1,136 @@
...
+        self.assertTrue(phi2.test())
+
+    def test_3sum(self):
+        nP = 90
+        phi1 = 0.3 * ObjectiveFunction.L2ObjectiveFunction(W=sp.eye(nP))
+        phi2 = 0.6 * ObjectiveFunction.L2ObjectiveFunction(W=sp.eye(nP))
+        phi3 = ObjectiveFunction.L2ObjectiveFunction(W=sp.eye(nP)) / 9.
+
+        phi = phi1 + phi2 + phi3
+        print(phi.multipliers)
```
rowanc1: `asssert len(multipliers)`?

SimPEG/Regularization.py
```diff
@@ -1,12 +1,33 @@
...
+from . import Utils
+from . import Maps
+from . import Mesh
+from . import ObjectiveFunction
+from . import Props
+
+__all__ = [
+    'Smallness',
+    'Smooth_x', 'Smooth_y', 'Smooth_z',
+    'Smooth_xx', 'Smooth_yy', 'Smooth_zz',
```
rowanc1: I think these should be turned from classes into inputs. `Smooth(orientation='x')` `Smooth2` etc.

SimPEG/Regularization.py
```diff
@@ -416,655 +577,1031 @@ def eval2Deriv(self, m, v=None):
...
-            self.cell_weights = np.ones(self.regmesh.nC) * self.cell_weights
+class Smooth_zz(BaseSmooth2):
+    """
+    Second-order smoothness along z. Regularizes on the second spatial
+    derivative in the z-direction
+    """
 
-        # if getattr(self, 'model', None) is None:
-        #             self.model = np.ones(self.regmesh.nC)
+    _multiplier_pair = 'alpha_zz'
```
rowanc1: I think ditching these names and using `properties.link()` might be better? That way changes on each side will be reflected? Right now I think that it is down (i.e. from tik-->smooth) only.

tests/base/test_Objectivefct.py
lheagy: good call

jcapriot:
It would probably be useful to assert the length of the `objfcts` and `multipliers` are equal upon initialization of a `ComboObjectiveFunction`.
Also would it be helpful to attach a `__getitem__` and a `__len__` function for the `ComboObjectiveFunction`? The `__getitem__` would allow it to be iterated over, so can replace `for multiplier, objfct in zip(self.multipliers,self.objfcts):` with just `for multiplier, objfct in self:`.

SimPEG/ObjectiveFunction.py
```diff
@@ -221,68 +221,41 @@ def multipliers(self, value):
...
+                pass
+            else:
+                objfct_H = objfct.deriv2(x, **kwargs)
+                if isinstance(objfct_H, Utils.Identity):
+                    if self.nP != '*':
+                        objfct_H = sp.eye(self.nP)  # if we need a shape
+                    elif objfct.nP != '*':
+                        objfct_H = sp.eye(objfct.nP)  # if we need a shape
+                    else:
+                        raise Exception('need to set nP on objective functions to get H')
```
jcapriot: I think lines 249 to 255 here are all side effects of the fact that `deriv2` can return a matrix if it is called without a vector, `v`  (See [`BaseRegularization`](https://github.com/simpeg/simpeg/blob/ref/objectivefunctions/SimPEG/Regularization.py#L550-L577) ). However, Considering a `BaseRegularization` class will almost always be added to a `BaseDataMisfit` like class whose implementations define a `deriv2` which **does** require a vector, `v`. (See [`l2_DataMisfit`](https://github.com/simpeg/simpeg/blob/ref/objectivefunctions/SimPEG/DataMisfit.py#L147-L163
)). Should we support only calling `deriv2` with a `v` (essentially only providing the Hessian times a vector operation)? That would simplify this code a bit :).

lheagy: Agreed - I think if provided a vector, we should simply return the sum. We may want to allow the case where `v=None` depending on the optimization scheme chosen.

SimPEG/Utils/matutils.py
lheagy: good call

lheagy: I will leave this here though, and we can make that change in discretize

SimPEG/Regularization.py
```diff
@@ -61,320 +87,439 @@ def dim(self):
...
-    def invProb(self): return self.parent
-    @property
-    def reg(self): return self
-    @property
-    def opt(self): return self.parent.opt
-    @property
-    def prob(self): return self.parent.prob
-    @property
-    def survey(self): return self.parent.survey
+    def mapping(self):
```
lheagy: `Props.Mapping` has a few additional complexities that aren't needed here (eg. the reciprocal).  I would like to give that some more thought before we bring it in

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,314 @@
...
+        return g
+
+    def deriv2(self, m, v=None, **kwargs):
+
+        H = Utils.Zero()
+        for multiplier, objfct in self:
+            if isinstance(multiplier, Utils.Zero) or multiplier == 0.: # don't evaluate the fct
+                pass
+            else:
+                objfct_H = objfct.deriv2(m, v, **kwargs)
```
lheagy: good call @jcapriot :)

rowanc1:
Following up from the meeting today:
https://youtu.be/P_Q38W6sXro?t=11m15s

There were problems coming out of multiple ways to set the multipliers.

```python
reg1.alpha_xx = 2.0
reg1.multipliers[2] = 3.0
print(reg1.multipliers[2])   # 2.0, weird....
```

image

The reason why this fails (I think!) is that the multipliers list is being created on demand from the alphas on each class. Using the `Tikhonov` class, we are already a few layers away from what is going on, with the alphas having to be linked up through on change calls. This means that there are already three ways to set the multiplier value (alpha_xx, objs[2].alpha?, multipliers[2]), and that the data could be stored in a few places.

I don't think that exposing the multipliers directly is a good idea. Instead they should just be set through the `alpha` perhaps. Moving `multipliers` to a private var `_multipliers`. This would limit the interface of how people are expected to interact with this class.

This would mean if you need to change it, you should use a directive (e.g. `beta`), set the alpha or create a new objective (by doing a new addition).

For example:

```
beta_est = Directives.BetaEstimate([data, reg], ratio=10.)
obj = data + beta * reg
```

Or a cooling `beta` etc.

@lheagy @jcapriot

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,314 @@
...
+    @multipliers.setter
+    def multipliers(self, value):
+        self._multipliers = value
+
+
+    def __call__(self, m, **kwargs):
+
+        f = 0.
+        for multiplier, objfct in self:
+            if isinstance(multiplier, Utils.Zero) or multiplier == 0.: # don't evaluate the fct
```
rowanc1: This should just be `== 0` the zero class has a comprehension for that!

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,314 @@
...
+                pass
```
rowanc1: This could be a continue statement:

```python
for mult, objfct in self:
    if mult == 0:
        continue
    f += mult * objfct(m, **kwargs)
return f
```

I find the pass a bit strange to read is all..

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def __getitem__(self,key):
+        return self.multipliers[key],self.objfcts[key]
+
+    @property
+    def __len__(self):
+        return self.objfcts.__len__
+
+    @property
+    def multipliers(self):
```
rowanc1: Thoughts on making this private?

lheagy: I think that is a good call, if you would like to have something that changes over the course of the inversion, it should be a directive, not going in and tinkering with the multipliers

SimPEG/ObjectiveFunction.py
lheagy: yep :) issue in discretize??

SimPEG/ObjectiveFunction.py
lheagy: sure

SimPEG/ObjectiveFunction.py
rowanc1: I am pretty sure that this works now actually.

rowanc1: image

lheagy: huh, cool!

SimPEG/InvProblem.py
```diff
@@ -104,7 +129,15 @@ def getFields(self, m, store=False, deleteWarmstart=True):
                 break
 
         if f is None:
-            f = self.prob.fields(m)
+            if isinstance(self.dmisfit, DataMisfit.BaseDataMisfit):
+                f = self.dmisfit.prob.fields(m)
+            elif isinstance(self.dmisfit, ObjectiveFunction.BaseObjectiveFunction):
```
lheagy: I see this more as a patch for now, but some of these ideas can stick around when we redefine the InvProb (or remove it). We do want to be able to store the fields, so I have added a `_hasFields` flag to the `BaseObjectiveFunction`, and the fields will be created and stored in a list on the `invProb`. Thoughts @rowanc1?

SimPEG/Maps.py
```diff
@@ -373,6 +383,254 @@ def nP(self):
         return self._nP
 
 
+class SelfConsistentEffectiveMedium(IdentityMap, properties.HasProperties):
```
lheagy: I am wondering if we want to start a `SimPEG.Emperical` for these sorts of things (and create a mapping based on that)

SimPEG/SEIS/__init__.py
```diff
@@ -0,0 +1 @@
+from . import StraightRay
```
lheagy: do we want to call this `StraightRayTomo` or `StraightRayTomography` or is `StraightRay` descriptive enough?

SimPEG/Examples/SEIS_TomoJointWithVolume.py
```diff
@@ -0,0 +1,203 @@
...
+        '\nTotal recovered volume (no vol misfit term in inversion): '
+        '{}'.format(
+            dmisVol(mopt1)
+        )
+    )
+
+    # with the volume regularization
+    vol_multiplier = 9e4
+    reg2 = reg
+    dmis2 = dmis + vol_multiplier * dmisVol
```
lheagy: joint inversion!

SimPEG/Regularization.py
```diff
@@ -416,655 +575,802 @@ def eval2Deriv(self, m, v=None):
...
+        )
 
-                mats.append(Wy)
+        # Utils.setKwargs(self, **kwargs)
 
-            self._Wy = sp.block_diag(mats)
+    # Properties
+    norms = properties.Array(
+        "Norms used to create the sparse regularization",
+        default=[0., 2., 2., 2.]
```
lheagy: @fourndo: Do we want the default to be `[0., 2., 2., 2.]` or should it be `[2., 2., 2., 2.]`?

lheagy:
@fourndo: I believe that the current failure with potential fields is due to the way that we are stashing R. I think a cleaner way to do this is to make `f_m` a settable property, (and is set to None, it is re-calculated). This way R is always computed using the correct norm. Thoughts?

SimPEG/DataMisfit.py
```diff
@@ -1,65 +1,49 @@
...
         .. note::
 
-            You should inherit from this class to create your own data misfit term.
+            You should inherit from this class to create your own data misfit
+            term.
     """
 
     debug   = False  #: Print debugging information
-    counter = None   #: Set this to a SimPEG.Utils.Counter() if you want to count things
+    counter = None  #: Set this to a SimPEG.Utils.Counter() if you want to count things
```
lheagy: @rowanc1: I am not actually sure what this does. I left it there as it was in the original Regularization.

SimPEG/Directives.py
```diff
@@ -327,127 +325,171 @@ def target(self, val):
...
+            if self.ComboObjFun:
 
-                indl, indu = imodel*self.reg.regmesh.nC, (imodel+1)*self.reg.regmesh.nC
+                for reg in self.reg.objfcts:
 
-                if getattr(self, 'eps', None) is None:
-                    self.reg.eps_p[imodel] = np.percentile(np.abs(mmap[indl:indu]),self.prctile)
-                else:
-                    self.reg.eps_p[imodel] = self.eps[0]
+                    ## NEED TO CHANGE THE DEFAULT VALUE TO SOMETHING ELSE
```
lheagy: @fourndo : do you want to change this in the Sparse Reg?

SimPEG/EM/Static/SIP/Regularization.py
```diff
@@ -9,7 +9,7 @@
 from SimPEG.Regularization import BaseRegularization, Simple
 
 
-class MultiRegularization(Simple):
+class MultiRegularization(BaseRegularization):
```
lheagy: @sgkang: we can pull this out in a separate pr and use the combo objective functions. There is a small example of how you can do that in the tests: https://github.com/simpeg/simpeg/blob/ref/objectivefunctions/tests/base/test_regularization.py#L227

SimPEG/InvProblem.py
```diff
@@ -113,6 +146,18 @@ def getFields(self, m, store=False, deleteWarmstart=True):
...
         return f
 
+    def get_dpred(self, m, f):
+        if isinstance(self.dmisfit, DataMisfit.BaseDataMisfit):
+            return self.dmisfit.survey.dpred(m, f=f)
+        elif isinstance(self.dmisfit, ObjectiveFunction.BaseObjectiveFunction):
+            dpred = []
+            for i, objfct in enumerate(self.dmisfit.objfcts):
+                if hasattr(objfct, 'survey'):
+                    dpred += [objfct.survey.dpred(m, f=f[i])]
```
lheagy: if you give multiple data misfits, `invProb.dpred` will return you a list the same length as the number of data misfits. 

- [ ] this needs to be tested

SimPEG/DataMisfit.py
```diff
@@ -1,65 +1,49 @@
...
+
+    _hasFields = True  #: Data Misfits take fields, handy to store them
```
lheagy: this is needed by the `BaseObjectiveFunction` so that it knows whether or not to expect fields when computing a residual and if fields should be passed (eg. regularization should not take in fields, but data misfits can)

SimPEG/Maps.py
lheagy: this should eventually be pulled out, but at this point, I would rather address that in a separate pr. thoughts @rowanc1?

SimPEG/Maps.py
```diff
@@ -372,6 +383,254 @@ def nP(self):
...
+            Prolate spheroids (\alpha > 1, needle-like) have not been
+            implemented yet
+
+        For reference, see
+        `Torquato (2002), Random Heterogeneous Materials <https://link.springer.com/book/10.1007/978-1-4757-6355-3>`_
+
+
+    """
+
+    sigma0 = properties.Float(
```
lheagy: properties!

SimPEG/ObjectiveFunction.py
lheagy: now this is tested.

SimPEG/Optimization.py
```diff
@@ -7,80 +7,158 @@
...
-                      "stopType": "critical"}
-
-    phi_d_target_Minimize = { "str": "%d : phi_d  = %1.4e <= phi_d_target  = %1.4e ",
-                              "left": lambda M: M.parent.phi_d, "right": lambda M: M.parent.phi_d_target,
-                              "stopType": "critical"}
-
-    phi_d_target_Inversion = { "str": "%d : phi_d  = %1.4e <= phi_d_target  = %1.4e ",
-                               "left": lambda I: I.phi_d, "right": lambda I: I.phi_d_target,
-                               "stopType": "critical"}
+    iteration = {
```
lheagy: these are just whitespace changes

SimPEG/Props.py
```diff
@@ -11,6 +11,39 @@
 from . import Utils
 
 
+class Array(properties.Array):
```
lheagy: these should likely be moved over to discretize

SimPEG/SEIS/StraightRay/StraightRayProblem.py
```diff
@@ -0,0 +1,110 @@
+import numpy as np
```
lheagy: - [ ] add test on straight ray problem

examples/04-grav/plot_laguna_del_maule_inversion.py
```diff
@@ -12,6 +12,7 @@
 Craig Miller
 """
 import os
+import shutil
```
lheagy: @fourndo: can you check that these results are what are expected?

fourndo: Looks fine. Great job on the example structure, loving it.

fourndo:
Oh nice, test joint eh. Such a trendsetter. I ll copy you tomorrow...

lheagy:
#559 is an outstanding issue on this pr, but I think it warrants further discussion, so perhaps it should be addressed in a separate pr?

lheagy:
The travis fail is in the docs. 
image

I am trying to bug-hunt this and am wondering if it is due to an update in how documentation is constructed in properties @fwkoch, @rowanc1? 

I know that `properties.Model` does not exist, and I did a search through SimPEG looking for any reference to that, but didn't find anything. We do however have `SimPEG.Props.Model` which inherits `properties.Array`. Is it possible that some wires are getting crossed in the building of the docs from properties? Thanks for your help!!

SimPEG/DataMisfit.py
lheagy: an adequate message is shown and it should be clear what action needs to be taken, so I am fine with doing a hard depreciation here.

SimPEG/DataMisfit.py
```diff
@@ -69,60 +53,118 @@ class l2_DataMisfit(BaseDataMisfit):
...
     .. math::
 
-        \mu_\\text{data} = {1\over 2}\left| \mathbf{W}_d (\mathbf{d}_\\text{pred} - \mathbf{d}_\\text{obs}) \\right|_2^2
+        \mu_\\text{data} = {1\over 2}\left|
+        \mathbf{W}_d (\mathbf{d}_\\text{pred} -
+        \mathbf{d}_\\text{obs}) \\right|_2^2
 
     """
 
+    std = 0.05  #: default standard deviation if not provided by survey
```
lheagy: we should consider removing these defaults. These are important things to consider in an inversion. However that is outside the scope of this pr.

SimPEG/EM/Static/SIP/Regularization.py
lheagy: see what you think. We can do this in a separate pr

SimPEG/Maps.py
```diff
@@ -108,7 +115,7 @@ def deriv(self, m, v=None):
         """
         if v is not None:
             return v
-        return sp.identity(self.nP)
+        return Utils.Identity()
```
lheagy: todo: this change should be reverted.

SimPEG/ObjectiveFunction.py
```diff
@@ -0,0 +1,432 @@
...
+
+    def __getitem__(self, key):
+        return self._multipliers[key], self.objfcts[key]
+
+    @property
+    def __len__(self):
+        return self.objfcts.__len__
+
+    @property
+    def _multipliers(self):
```
lheagy: I think multipliers should be exposed at the top level.

lheagy:
I think this is basically ready to go - thoughts @fourndo, @sgkang, @rowanc1 ? The main todo that I think is still within the scope of this pr is bringing the multipliers up to top level name-space (`.multipliers` instead of `._multipliers`)

SimPEG/Regularization.py
```diff
@@ -410,661 +588,807 @@ def eval2Deriv(self, m, v=None):
...
-        return 0.5 * r.dot(r)
+        """
+        Weighting matrix
+        """
+        if self.cell_weights is not None:
+            return Utils.sdiag(self.cell_weights)
+        elif self._nC_residual != '*':
+            return sp.eye(self._nC_residual)
+        else:
+            return Utils.Identity()
```
fourndo: Where did the volumes go?
```
if self.cell_weights is not None:
            return Utils.sdiag((self.regmesh.vol*self.cell_weights)**0.5)
        elif self._nC_residual != '*':
             return sp.eye(self._nC_residual)
        else:
            return Utils.sdiag(self.regmesh.vol**0.5)
```

Should have a SimpleSmall class. @

fourndo: @lheagy : What's the ```self._nC_residual``` for?

SimPEG/Regularization.py
```diff
@@ -410,661 +588,807 @@ def eval2Deriv(self, m, v=None):
...
+        """
+        W = getattr(
+            self.regmesh,
+            "cellDiff{orientation}".format(
+                orientation=self.orientation
+            )
+        )
+        if self.cell_weights is not None:
+            Ave = getattr(self.regmesh, 'aveCC2F{}'.format(self.orientation))
+            W = (
```
fourndo: @lheagy : Missing volumes here too.

SimPEG/Regularization.py
lheagy: TODO: also check if we need _nC_residual or if nP is enough

lheagy: @fourndo `_nC_residual` gives you the shape of `W * mapping(m-mref)` (so it is the other side of the mapping

lheagy: @fourndo: do we want all cell weights under a sqrt? currently they are not

SimPEG/Regularization.py
lheagy: agreed. I think that is out of scope here for now

SimPEG/Regularization.py
lheagy: This has been addressed in #592. The diff is a bit confusing as I created a `Small` (with vols) and a `SimpleSmall` (without vols)

SimPEG/EM/Static/SIP/Regularization.py
sgkang: @lheagy I am not using this anymore, so we can deprecate this.

lheagy: kk, so are you good if I remove this whole file?

PR #513, opened 2016-11-25
Patch/properties

lheagy:
- updates to work with beta version of properties
- `scope.help` --> `scope.doc`
- set values to undefined to clear rather than None

TODO:
- the reciprocal default is not being set correctly, so I have set a default for `mui` as well as `mu`. This should be corrected prior to merging. cc @rowanc1, @fwkoch

### DIFF
--- a/SimPEG/EM/Base.py
+++ b/SimPEG/EM/Base.py
@@ -34,7 +34,8 @@ class BaseEMProblem(Problem.BaseProblem):
         default=mu_0
     )
     mui = Props.PhysicalProperty(
-        "Inverse Magnetic Permeability (m/H)"
+        "Inverse Magnetic Permeability (m/H)",
+        default=1./mu_0
     )
 
     Props.Reciprocal(mu, mui)
--- a/SimPEG/EM/Base.py
+++ b/SimPEG/EM/Base.py
@@ -35,7 +35,6 @@ class BaseEMProblem(Problem.BaseProblem):
     )
     mui = Props.PhysicalProperty(
         "Inverse Magnetic Permeability (m/H)"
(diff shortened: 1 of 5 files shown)


SimPEG/EM/Base.py
```diff
@@ -34,7 +34,8 @@ class BaseEMProblem(Problem.BaseProblem):
         default=mu_0
     )
     mui = Props.PhysicalProperty(
-        "Inverse Magnetic Permeability (m/H)"
+        "Inverse Magnetic Permeability (m/H)",
+        default=1./mu_0
```
lheagy: this default should not be necessary

SimPEG/Props.py
```diff
@@ -41,11 +41,11 @@ def reciprocal_prop(self):
 
     def clear_props(self, instance):
         if self.prop:
-            instance._set(self.prop.name, None)
+            instance._set(self.prop.name, properties.utils.undefined)
```
lheagy: @fwkoch, @rowanc1 : could you sanity-check me on this?

SimPEG/Props.py
```diff
@@ -157,11 +156,16 @@ def fget(self):
             return mapping * self.model
 
         def fset(self, value):
-            value = scope.validate(self, value)
+            if value is not properties.utils.undefined:
+                value = scope.validate(self, value)
+                if scope.reciprocal:
```
lheagy: the reciprocal should only be cleared if we are not setting it to an undefined (otherwise we run into issues with defaults... by setting a reciprocal with a default, we previously cleared the default. There is now a test to check this )

PR #514, opened 2016-11-28
Add fdel to Props, take None out of property validation.

fwkoch:
Physical property default is only set if the input is not None.
This commit also removes _exclusive_kwargs, which was removed from
properties


lheagy:
thanks @fwkoch !

PR #517, opened 2016-11-30
Dev

lheagy:
# Properties
- SimPEG now depends on `properties[math]` (https://github.com/3ptscience/properties)
```
pip install properties[math]
```

# Mappings and Models
- depreciate `mapping` in the problem definition. Now the physical property to which the mapping is related must be explicitly stated. eg.
```SimPEG.EM.FDEM.Problem3D_e(mesh, mapping=sigmaMap)```
becomes 
```SimPEG.EM.FDEM.Problem3D_e(mesh, sigmaMap=sigmaMap)```
This change is not backwards compatible 
- depreciate `problem.curModel` in favour of `problem.model` 
- implement `properties` for the physical properties in all SimPEG problems 

# Added Maps
- Parametrized block in layer
- Parametrized layer 

# EM
- break `SrcFDEM` out of survey and put into a separate file
- implement primary-secondary source for FDEM problem
- testing of sources for FDEM 

# Static
- tests dipole-dipole, dipole-pole, pole-dipole, and pole-pole
- corresponding pole source and receiver classes are implemented
- option for boundary conditions for Problem3D_CC (dirichlet or neumann)
- consistent handling of null-space for both Problem3D_CC and Problem3D_N
- 2.5D DC code is implemented (`Problem2D_CC` and `Problem2D_N`), and tested
- Useful utiilities
   - `plot_pseudoSection`: plotting pseudo-section for 2D data
   - `gen_DCIPsurvey`: generating simple surveys
   - read/writeUBC_DC3Dobs
   - readUBC_DC2D model
   - convertObs_DC3D_to_2D: Collapse 3D survey to along line 2D survey
   - Analytic functions are added:  `DCAnalytic_Pole_Dipole`, `DCAnalytic_Dipole_Pole`,  `DCAnalytic_Pole_Pole`, `DCAnalytic_Dipole_Dipole`, `DCAnalyticSphere`

# Mesh View
- added the ability to do `mirror=True` on the cyl mesh and have it plot a mirror image so corss sections can be easily viewed
- added the ability to down-sample the area being viewed using `range_x`, `range_y` to define the bounds of the region you want to view. `sample_grid = [dx, dy]` can be used to re-sample the data (handy for the casing-type meshes)

# Examples
- so much whitespace cleanup :) 
- add Craig's example: PF_Gravity_Laguna_del_Maule_Inversion
- add examples from SimPEG.EM paper

# Docs
- deploy: http://dev-docs.simpeg.xyz

# Prior to merging:
- [x] ensure dev, master tests are run on both python 2.7 and python 3
- [x] improve error handling and messages in the mappings (started in #519)
- [x] clean up the linear example as per #509 (seems to work on mine: dom)
- [x] bumpversion, tag, release

### CODE REVIEW

**Comment by rowanc1** [comment]
I have been working a bit on more sensible errors and suggestions for the HasModel instances:

image

image

image

**Comment by rowanc1** [comment]
@lheagy I think that I would rather do a hard depreciation on the `mapping` instead of *trying* to upgrade to the new version implicitly.

This would mean that a sensible error would be raised if you try to set `mapping` and with #519 there is a lot more information about what to do next that will be shown to the user on validation.

File: SimPEG/Problem.py [hunk 1]
```diff
@@ -6,45 +6,96 @@
...
+
+    #: Solver options as a kwarg dict
+    solverOpts = {}
+
+    #: A SimPEG.Mesh instance.
+    mesh = None
+
+    def __init__(self, mesh, **kwargs):
+        if 'mapping' in kwargs:
+            import warnings
```
lheagy: if we are going to hard-depreciate, would it be best it this were an error rather than a warning - the code will fail later on regardless, so it might be best to fail here. Thoughts @rowanc1?

rowanc1:
Yes, I would say an exception would be better. There is a bit of a problem in the LinearProblem as it still sort of uses a mapping -- perhaps we should rename this as well and just be very specific with referring to things that have a map. even `modelMap` would be better.

So:
- [ ] raise exception here if `mapping` is in the kwargs. 
- [ ] delete the `Depreciate.py`
- [ ] remove the various stubs for saying what the previous main mapping was
- [ ] put `.validate()` at the top of the main problem/survey calls (giving a summary), test this.

Other thoughts, because it is now annoying to create `Identity` maps.
- [ ] maybe have a validation shortcut for identity maps `Problem(mesh, sigmaMap=1)`? or `exp`?

SimPEG/Depreciate.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
```
rowanc1: This will be done in #536

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
 python:
-  - 3.4
-  - 2.7
+  - 3.5
```
lheagy: @rowanc1 : thoughts on testing on both 3.5 and 2.7? is just 3.5 good enough?

rowanc1: I have updated it to test on 2.7 as well.

grosenkj:
I am happy with this, doesn't have much new NSEM development. Most of the updates are in regards to the `mapping` changes.

Issue #432, opened 2016-09-16
pip install simpeg in windows


sgkang:
I was installing simpeg using pip install through git repo 

so, I did
`pip install git+git://github.com/simpeg/simpeg.git`

and the below error is that I have got. I guess that is because of C ++ compiler, and possibly 
just because of the path. Any ideas?

C:\Users\administrator\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\link.exe /DLL /nologo /INCREMENTAL:NO /LIBPATH:S:\EOSC350\anaconda\libs /LIBPATH:S:\EOSC350\anaconda\PCbuild\amd64 /LIBPATH:S:\EOSC350\anaconda\PC\VS9.0\amd64 /EXPORT:initSimPEG/Utils/interputils_cython build\temp.win-amd64-2.7\Release\SimPEG/Utils/interputils_cython.obj /OUT:S:\EOSC350\pythonprograms\simpeg\SimPEG\Utils\interputils_cython.pyd /IMPLIB:build\temp.win-amd64-2.7\Release\SimPEG/Utils\interputils_cython.lib /MANIFESTFILE:build\temp.win-amd64-2.7\Release\SimPEG/Utils\interputils_cython.pyd.manifest
LINK : error LNK2001: unresolved external symbol initSimPEG/Utils/interputils_cython
build\temp.win-amd64-2.7\Release\SimPEG/Utils\interputils_cython.lib : fatal error LNK1120: 1 unresolved externals
error: command 'C:\Users\administrator\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\link.exe' failed with exit status 1120

lheagy:
Brandon encountered a linker error with windows at BIRS. Here are his notes: 

**Notes on SimPEG Installation** 
RE: Microsoft Visual C++ Compiler Issues
B. Karchewski
20 August 2016

System:
Windows 7 Enterprise x64 PC
Intel Core i7-5600 2.60GHz CPU
msys MinGW for Unix-like commands
Anaconda python 2.7 build

Steps:

1. Clone SimPEG from git repo, git clone https://github.com/simpeg/simpeg
2. Install SimPEG using setup.py, python setup.py install
	-Error encountered: Missing linker, needs MSC v.1500 (Microsoft Visual C++ 2008) Runtime Library
	-Available: https://www.microsoft.com/en-ca/download/details.aspx?id=29
[...]

PR #520, opened 2016-12-19
Make setup.py windows friendly

rowanc1:
I do not have a windows machine, this should help either way - but I am not sure if it will solve all of the windows issues. Can @sgkang or @micmitch try this out?

See #432


rowanc1:
I would rather push on getting `discretize` package better tested and out there than spend time on the `setup.py` in simpeg. We have been updating that repository to make it more windows friendly.

PR #523, opened 2016-12-28
Feat/plot with mirroring

lheagy:
Add a `mirror` kwarg for the cyl mesh plotting so that when plotting a model on a cylindrically symmetric mesh we can look at a symmetric plot 

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-  - 3.4
+  - 3.5
   # - 2.7
 
 sudo: false
--- a/SimPEG/Examples/Mesh_Plot_Cyl.py
+++ b/SimPEG/Examples/Mesh_Plot_Cyl.py
@@ -30,7 +30,7 @@ def run(plotIt=True):
     h = Utils.meshTensor([(cs, nc)])
 
     # define a mesh
-    mesh = Mesh.CylMesh([h, 1, h], x0 = np.r_[0., 0., -h.sum()/2.])
+    mesh = Mesh.CylMesh([h, 1, h], x0='00C')
 
     # Put the model on the mesh
     sigma = sig_air*np.ones(mesh.nC)  # start with air cells
--- a/SimPEG/Examples/Mesh_Plot_Cyl.py
+++ b/SimPEG/Examples/Mesh_Plot_Cyl.py
@@ -1,0 +1,66 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from SimPEG import Mesh, Utils, Maps
+
+# Set a nice colormap!
+plt.set_cmap(plt.get_cmap('viridis'))
+
+
+def run(plotIt=True):
+    """
+        Plot Mirrored Cylindrically Symmetric Model
+        ===========================================
+
+        Here, we demonstrate plotting a model on a cylindrically
+        symmetric mesh with the plotting symmetric about x=0.
+    """
+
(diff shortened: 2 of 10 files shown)


SimPEG/Examples/Mesh_Plot_Cyl.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # x-direction
+    cs = 1
+    nc = np.ceil(2.5*(- (sphere_z-sphere_radius))/cs)
+
+    # cell spacings in the x and z directions
+    h = Utils.meshTensor([(cs, nc)])
+
+    # define a mesh
+    mesh = Mesh.CylMesh([h, 1, h], x0 = np.r_[0., 0., -h.sum()/2.])
```
rowanc1: `x0='00C'` works by the way.

SimPEG/Examples/Mesh_Plot_Cyl.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # Put the model on the mesh
+    sigma = sig_air*np.ones(mesh.nC)  # start with air cells
+    sigma[mesh.gridCC[:, 2] < 0.] = sig_halfspace  # cells below the earth
+
+    # indices of the sphere
+    sphere_ind = (
+        (mesh.gridCC[:, 0]**2 + (mesh.gridCC[:, 2] - sphere_z)**2) <=
```
rowanc1: we should make this into a `VectorArray` which would allow us to do `mesh.gridCC.x` saving at least 4 characters every time...

SimPEG/Mesh/View.py
```diff
@@ -513,7 +513,28 @@ def _plotCylTensorMesh(self, plotType, *args, **kwargs):
         # Hackity Hack:
         # Just create a TM and use its view.
         from SimPEG.Mesh import TensorMesh
-        M = TensorMesh([self.hx, self.hz], x0=[self.x0[0], self.x0[2]])
+
+        mirror = kwargs.pop('mirror', None)
+        if mirror is True:
```
rowanc1: looks good!

SimPEG/Examples/Mesh_Plot_Cyl.py
lheagy: right (this was adapted from a notebook where I was shifting things around more... I can update)

lheagy:
I am not entirely sure why this is failing here... https://travis-ci.org/simpeg/simpeg/jobs/187259357 only the fdem forward tests are failing (which shouldn't have been affected??)

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 python:
-  - 3.4
+  - 3.5
```
lheagy: anaconda is now distributing 3.5 as the standard python 3 package

rowanc1:
When I just setup a headless linux machine I had to explicitly tell it to use backend `agg` for matplotlib. Otherwise it would try to connect to a QT backend by default.

SimPEG/Examples/Mesh_Plot_Cyl.py
lheagy: that would be nice - I have started playing around with bringing `properties` into the `Mesh` on the `ref/Props` branch https://github.com/simpeg/simpeg/tree/ref/Props

lheagy:
Thanks @rowanc1!

PR #534, opened 2017-01-12
Feat/mesh plotting

lheagy:
Add `range_x` and `range_y` kwargs to `Mesh.plotImage` that creates a smaller mesh and samples your data only in that region. 

This is essential for working with large meshes and plotting vector plots as the `view='vec'` previously interpolated the entire domain onto a mesh with cell widths equal to that of the minimum cell width ... big problem if you have lots of padding cells. 

Added an example demonstrating its use (with a fine core cyl mesh and a load of padding b/c we want to look at 4 decades of frequency)s
figure_2
using a sphere model
figure_3
and we can look at the fields! (if you try without the range - it will run into a memory error)
figure_4

(I will add this to discretize in a separate pr, but I need it working with SimPEG dev soon for some of the em-examples)

### DIFF
--- a/SimPEG/Mesh/View.py
+++ b/SimPEG/Mesh/View.py
@@ -42,13 +42,15 @@ class TensorView(object):
     #         pltNum +=1
     #     if showIt: plt.show()
 
-    def plotImage(self, v, vType='CC', grid=False, view='real',
-                  ax=None, clim=None, showIt=False,
-                  pcolorOpts=None,
-                  streamOpts=None,
-                  gridOpts=None,
-                  numbering=True, annotationColor='w'
-                  ):
+    def plotImage(
+        self, v, vType='CC', grid=False, view='real',
+        ax=None, clim=None, showIt=False,
+        pcolorOpts=None,
+        streamOpts=None,
+        gridOpts=None,
+        numbering=True, annotationColor='w',
+        range_x=None, range_y=None
+    ):
         """
         Mesh.plotImage(v)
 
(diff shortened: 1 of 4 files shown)


SimPEG/Mesh/View.py
```diff
@@ -358,22 +365,55 @@ def _plotImage2D(self, v, vType='CC', grid=False, view='real',
...
+
+            ind_CCx = np.ones(self.vnC, dtype=bool)
+            ind_CCy = np.ones(self.vnC, dtype=bool)
+            if range_x is not None:
+                x = tMi.vectorNx
+
+            if range_y is not None:
+                y = tMi.vectorNy
+
+            if range_x is not None or range_y is not None:  # use interpolated values
```
lheagy: I am not sure we want to be doing this. If not, we just need a robust way to select cell centers for (U, V) and node locations for vector node locations. thoughts @rowanc1 ?

SimPEG/Mesh/View.py
```diff
@@ -42,13 +42,15 @@ def __init__(self):
...
-                  numbering=True, annotationColor='w'
-                  ):
+    def plotImage(
+        self, v, vType='CC', grid=False, view='real',
+        ax=None, clim=None, showIt=False,
+        pcolorOpts=None,
+        streamOpts=None,
+        gridOpts=None,
+        numbering=True, annotationColor='w',
+        range_x=None, range_y=None
```
rowanc1: would these be better as `xlim` and `ylim`?

lheagy: image

PR #535, opened 2017-01-13
Dcip/dev

sgkang:
- fix bugs in Pole rxs
- fix bug in ExtractCoreMesh


sgkang:
@rowanc1 @lheagy those are small changes, but used in `em_examples` can we merge this to dev so that we can use simpeg/dev rather than simpeg/dcip/dev?

SimPEG/Utils/meshutils.py
```diff
@@ -137,8 +137,8 @@ def ExtractCoreMesh(xyzlim, mesh, meshType='tensor'):
         xmin, xmax = xyzlim[0,0], xyzlim[0,1]
         ymin, ymax = xyzlim[1,0], xyzlim[1,1]
 
+        xind = np.logical_and(mesh.vectorCCx>xmin, mesh.vectorCCx<xmax)
```
rowanc1: it looks like 1d is still broken (you are making a 2d mesh)

PR #536, opened 2017-01-15
fully depreciate old mappings

lheagy:
- add SimPEG version number to depreciation messages
- remove Depreciate.py
- remove the `_depreciate_main_map` stub
- add test for depreciation of mappings 

image

- [ ] @rowanc1: can you confirm that we will bumpversion to 0.4.0 on dev (I have added version numbers to the depreciation messages).


SimPEG/Examples/FLOW_Richards_1D_Celia1990.py
```diff
@@ -60,7 +60,7 @@ def run(plotIt=True):
     def getFields(timeStep, method):
         timeSteps = np.ones(360/timeStep)*timeStep
         prob = Richards.RichardsProblem(
-            M, mapping=E,
+            M, modelMap=E,
```
rowanc1: this will be completely changed in a future release.

SimPEG/Problem.py
```diff
@@ -40,24 +37,15 @@ class BaseProblem(Props.HasModel):
...
         if 'mapping' in kwargs:
-            import warnings
-            warnings.warn(
-                'The `mapping` property has been depreciated, '
-                'please use:\n\n\n'
-                '\tfrom SimPEG import Depreciate\n'
-                '\tDepreciate.use_old_mappings()\n\n\n'
-                'To bring back old functionality.'
+            raise Exception(
+                'Depreciated (in 0.4.0): use one of {}'.format(
```
rowanc1: cool.

SimPEG/Problem.py
```diff
@@ -77,7 +65,17 @@ def mapping(self):
 
         """
```
rowanc1: Can we remove the comment from the mapping class?

SimPEG/Examples/FLOW_Richards_1D_Celia1990.py
lheagy: fair (just needed it to pass :) )

rowanc1:
:+1:

PR #538, opened 2017-01-17
fix title in EM Cyl example, updated linear inversion

### LINKED ISSUE #509: Bug: Linear Problem
I re-ran the linear problem (http://docs.simpeg.xyz/content/examples/Inversion_Linear.html) on the dev branch and now get 
image

while on master we recover something much more reasonable

image

a random seed is set, so it not just a one-off. This needs to be resolved prior to putting any pr from dev to master!

lheagy:
- Fix the linear example to close #509
- fix title for the cyl inversion 
image


rowanc1:
image

PR #543, opened 2017-02-15
Fixing up 3D NSEM example

grosenkj:
Fixing NSEM 3D forward example by:

- A better mesh/model with adequate padding and visulazation
- Assigning uncertainties to the data
- Plot the data

### DIFF
--- a/SimPEG/Examples/EM_NSEM_3D_Foward.py
+++ b/SimPEG/Examples/EM_NSEM_3D_Foward.py
@@ -111,7 +111,13 @@ def run(plotIt=True):
             components=['xx', 'xy', 'yx', 'yy'], ax=ax_p, errorbars=True)
         ax_p.legend(bbox_to_anchor=(1.05, 1), loc=2)
 
-        plt.show()
+
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


SimPEG/Examples/EM_NSEM_3D_Foward.py
```diff
@@ -69,12 +75,43 @@ def run(plotIt=True, nFreq=1):
...
+            components=['xy', 'yx'], ax=ax_r, errorbars=True)
+        ax_r_on = dataVec.plot_app_res(
+            np.array([-200, 0]),
+            components=['xx', 'yy'], ax=ax_r_on, errorbars=True)
+        ax_p = dataVec.plot_app_phs(
+            np.array([-200, 0]),
+            components=['xx', 'xy', 'yx', 'yy'], ax=ax_p, errorbars=True)
+        ax_p.legend(bbox_to_anchor=(1.05, 1), loc=2)
+
+        plt.show()
```
lheagy: Right now the testing is running this, can you put the `plt.show()` beneath `if name == '__main__'` - that should do the trick!

PR #547, opened 2017-02-28
Feat/straightray

lheagy:
# Straight Ray Tomography
- implement `SEIS.StraightRay`, `Problem`, `Survey`, `Src` and `Rx`
- example: SEIS_StraightRayTomo.py

image 

# InvProb Compatibility with Composable Objective functions
- allow both `dmisfit` to be either a base objective function or a base data misfit
- extend invProb in cases where `dmisfit` is a combo objective function so that it can stash multiple fields from different problems 

# Example Joint Inversion
- include an example "joint" inversion (consists of 2 data misfit terms and a regularization) based on the Heagy et. al., 2014 SEG abstract http://library.seg.org/doi/abs/10.1190/segam2014-1639.1: SEIS_VolumeRegularization.py

image


SimPEG/Examples/SEIS_TomoJointWithVolume.py
```diff
@@ -0,0 +1,203 @@
...
+        '\nTotal recovered volume (no vol misfit term in inversion): '
+        '{}'.format(
+            dmisVol(mopt1)
+        )
+    )
+
+    # with the volume regularization
+    vol_multiplier = 9e4
+    reg2 = reg
+    dmis2 = dmis + vol_multiplier * dmisVol
```
rowanc1: Wouldn't it be nice if you could just easily create data misfits by adding them together?

Oh wait!

SimPEG/Maps.py
```diff
@@ -382,6 +383,254 @@ def nP(self):
         return self._nP
 
 
+class SelfConsistentEffectiveMedium(IdentityMap, properties.HasProperties):
```
rowanc1: We need to start thinking about pulling these out into their own package.

SimPEG/Maps.py
```diff
@@ -382,6 +383,254 @@ def nP(self):
...
+
+    alpha0 = properties.Float(
+        "aspect ratio of the phase-0 ellipsoids", default=1.
+    )
+
+    alpha1 = properties.Float(
+        "aspect ratio of the phase-1 ellipsoids", default=1.
+    )
+
+    rel_tol = properties.Float(
```
rowanc1: Thanks for starting these with properties, really cool!

SimPEG/Optimization.py
```diff
@@ -823,6 +823,10 @@ def bfgsH0(self):
             Must be a SimPEG.Solver
         """
         if getattr(self, '_bfgsH0', None) is None:
+            print("""
+                Default solver: SolverDiag is being used in bfgsH0
```
rowanc1: Diag? really? That does not seem like a good idea....

SimPEG/SEIS/StraightRay/StraightRayProblem.py
```diff
@@ -0,0 +1,110 @@
...
+import scipy.sparse as sp
+import matplotlib.pyplot as plt
+
+from SimPEG import Problem
+from SimPEG import Utils
+from SimPEG import Mesh
+from SimPEG import Props
+
+
+def lengthInCell(O, D, x, y, plotIt=False):
```
rowanc1: Calling @jcapriot .... :)

PR #568, opened 2017-04-03
Update nsem/dev with recent SimPEG developments

grosenkj:
Updating nsem/dev and pulling in recent SimPEG changes.


grosenkj:
The travis build fails on the doc testing  with the following error. 

image

@lheagy Any idea's what this could be? I have had issue passing this test for a while though this is a different error.

grosenkj:
Was able to track down the bug causing the earlier fail. But I am still not passing the last test, giving this errer

image

Anyone. Idea's what this is? (cc. @lheagy )

Issue #560, opened 2017-03-09
deploy from python 3


lheagy:
docs and pypi deployments should be done from the python 3 builds on travis ci. currently they are done on 2.7.

lheagy:
this is causing pypi installation failures on python 3

rowanc1:
Would this change anything on the 2.7 release on PyPI? Probably should be encouraging people to use Python3 anyways...

lheagy:
(not sure... we can test on a --pre deployment)

lheagy:
right now, the docs need to remain on python2 (we are using webapp2, which currently depends on python2). https://webapp2.readthedocs.io/en/latest/

The pypi deploy, however, should still be on python 3

lheagy:
@rowanc1, @bsmithyman : do you have any recommendations for running webapp2 (or something similar??) with python3?

rowanc1:
Because we are deploying the docs to appengine, that environment only has py2.7 support. Webapp2 is in pre release for python 3, but we only need that for docs - which has to be in py2.7 anyways.

lheagy:
kk, thanks! the docs are still not deploying properly :( (They haven't since the addition of the gallery...) https://travis-ci.org/simpeg/simpeg/jobs/216589157 Any ideas? I can try hitting on this later today if nothing jumps out.

rowanc1:
`gcloud -q preview app deploy ./docs/app.yaml --version ${TRAVIS_COMMIT} --promote;`

I think that the `preview` is no longer necessary.

I also think that we need to specify the project in there as well? (in addition to the `set project` that is already there)

Something like:

```bash
if [ "$TRAVIS_BRANCH" = "master" ]; then
  gcloud auth activate-service-account --key-file credentials/client-secret.json;
  export GAE_PROJECT=simpegdocs;
else
  gcloud auth activate-service-account --key-file credentials/client-secret-dev.json;
  export GAE_PROJECT=simpeg-dev-docs;
fi;
gcloud config set project $GAE_PROJECT;
gcloud app deploy ./docs/app.yaml --project $GAE_PROJECT --version ${TRAVIS_COMMIT} --promote;
```

Notice that I added the: `--project $GAE_PROJECT`, this was me looking at:
http://stackoverflow.com/questions/33493750/error-response-13-an-internal-error-occurred-when-deploying-project-on-googl

I am not sure that this is necessary anymore either:
`gcloud -q components update gae-python;`
travis seems to be complaining.

[...]

Issue #554, opened 2017-03-05
Gallery of Examples


rowanc1:
We could move to a gallery of examples, upgrading a few of them as we go. By default this means that they are a bit harder to interact with from the package (i.e. they are more part of the docs than the code). But I think we can get away with that by doing a make script etc.

Things:
- Better visibility on examples
- Can download them individually as `.ipynb` or as `.py`
- Captures output
- Need to copy from package to docs
- Would like to introduce properties into some of the examples to help document them.
- Need to think about how to show examples from sub packages.

![image](https://cloud.githubusercontent.com/assets/913249/23589734/e3ff82c0-018f-11e7-9356-bb5c31db3f5d.png)

lheagy:
I think this would be great! The main way of interacting with an example (at least my take) is to tinker with the code. Just importing an example and watching it run doesn’t teach much other that “this is possible”
 
Also, there are some cool plug-ins for building these types of examples from a notebook (something to think about)
- http://sphinx-gallery.readthedocs.io/en/latest/tutorials/plot_notebook.html 
- http://nbsphinx.readthedocs.io/en/0.2.13/

rowanc1:
A PR for this should also address #553 #544 #491

I will do a first pass and see what people think

rowanc1:
Some progress:

![image](https://cloud.githubusercontent.com/assets/913249/23591697/d729631e-01b1-11e7-835e-21354c239817.png)

rowanc1:
I also brought in the notebooks from the `notebooks` repo. Not really sure if this is the best way to do that.

![image](https://cloud.githubusercontent.com/assets/913249/23592610/db1e4228-01c0-11e7-9cb4-4f6cc91ce8f1.png)

lheagy:
The dev-docs are not currently deploying. We likely need to update what is uploaded to google app engine. https://travis-ci.org/simpeg/simpeg/jobs/209059459

![image](https://cloud.githubusercontent.com/assets/6361812/23726279/8c9de03c-0408-11e7-931b-c166fca4e920.png)

lheagy:
I am hoping #563 resolves this

lheagy:
It looks like there is now a conflict with python 3.6 and deploying gae. https://travis-ci.org/simpeg/simpeg/jobs/210143591
![image](https://cloud.githubusercontent.com/assets/6361812/23833135/d1ed2bea-06fe-11e7-8a37-ef20cc40eebb.png)

[...]

PR #570, opened 2017-04-04
Deploy the docs!

lheagy:
- update .travis.yml and app.yaml so that docs with examples deploy
image

- add an `edit_on_github` link to every page to replace the "page source" as the ability to edit is more useful.
image


lheagy:
This addresses #554 and #560

docs/_ext/edit_on_github.py
```diff
@@ -0,0 +1,42 @@
...
+sidebar.
+
+Loosely based on https://github.com/astropy/astropy/pull/347
+"""
+
+import os
+import warnings
+
+
+def get_github_url(app, view, path):
```
rowanc1: this is cool!

docs/app.yaml
```diff
@@ -1,5 +1,5 @@
-# application: simpegdocs
-# version: 1
+# application: simpeg-dev-docs
```
rowanc1: is there a reason for this?

lheagy: ah, no I was testing deployment locally and wanted to do that on dev-docs. Would you prefer I update this back to simpegdoc before merging?

Issue #578, opened 2017-04-07
Validating the directives list


lheagy:
## The problem
In some cases, the order of the directives matters - eg `update_IRLS` needs to be called before `Update_lin_PreCond` and right now, there are no checks in place to ensure that they are provided in the correct order. This is subtle because the inversion will still run without issue. 

## An Approach
I suggest adding a 
```python
def validate(self, directivesList):
    assert("directives list makes sense for this directive")
```
to each directive (Base class just passes) and on the directivesList setter in the inversion we do a 
```python
directivesList.validate()
``` 
which just goes through the directives List and calls validate for each directive, eg
```python
def validate(self):
    [directive.validate(self) for directive in self.dList]
```

cc @fourndo, @rowanc1

PR #580, opened 2017-04-07
Feat/directives validation

lheagy:
## Validate Directives

Add validation on directives and directives List as per #578. 
- `BaseDirective` has a method `validate(self, directiveList=None)` which returns `True` by default
-  `DirectiveList` has a method `validate(self)` that goes through the directive list and validates each one
- `InvProb` runs the directive list validation on setter


SimPEG/Directives.py
```diff
@@ -588,6 +597,28 @@ def endIter(self):
...
+    def validate(self, directiveList):
+        # check if a linear preconditioner is in the list, if not warn else
+        # assert that it is listed after the IRLS directive
+        dList = directiveList.dList
+        self_ind = dList.index(self)
+        lin_precond_ind = [
+            isinstance(d, Update_lin_PreCond) for d in dList
+        ]
+
+        if any(lin_precond_ind):
```
lheagy: @fourndo: can you take a look at this? Here I check if the there is a `Update_lin_PreCond` directive in the directive list and then make sure it is after the IRLS directive or throw a warning if not in the list (not an error - see what you think is more appropriate)

SimPEG/Inversion.py
```diff
@@ -21,6 +21,7 @@ class BaseInversion(object):
     def directiveList(self):
         if getattr(self, '_directiveList', None) is None:
             self._directiveList = Directives.DirectiveList(inversion=self)
+            self._directiveList.validate()  # validate if we skip setter
```
lheagy: @rowanc1: what do you think of these locations for the validation?

examples/05-mag/plot_MVI_Cartesian.py
```diff
@@ -174,7 +174,7 @@ def run(plotIt=True):
     update_Jacobi = Directives.Update_lin_PreCond()
 
     inv = Inversion.BaseInversion(invProb,
-                                  directiveList=[betaest, update_Jacobi, IRLS, ])
+                                  directiveList=[betaest, IRLS, update_Jacobi])
```
lheagy: @fourndo: I updated your example :)

SimPEG/Directives.py
lheagy: you can check out the test file that I added to see it in action

requirements_dev.txt
```diff
@@ -14,3 +14,4 @@ ipython
 matplotlib
 properties[math]
 wheel
+pytest
```
lheagy: @rowanc1: this is to test that we are raising warnings

lheagy:
@fourndo, @rowanc1: this is a fairly light one. Please have a look when you have a chance and let me know what you think! @fourndo: this is coming in on the mag amplitude branch, so I will let you merge once both you and @rowanc1 are happy :)

lheagy:
@fourndo: I went through and resolved the merge conflicts here. If you are happy with these changes, can you please merge this pr. Thanks!

lheagy:
This will also update MagAmplitude to the most recent ref/objectivefunctions

fourndo:
Yup sounds great @lheagy , I ll keep an eye out for the end of the test.

lheagy:
@fourndo: it looks like there are quite a number of other changes on the MagAmplitde branch that are making getting this pr to pass a bit tricky. So I am going to close this and re-open a pr onto the objective function branch with the directives changes. You can then merge that upstream to MagAmplitude.

Issue #577, opened 2017-04-07
Directives + Combo Objective Functions


lheagy:
## Problem: Combo objective functions + Directives 
@fourndo and I have been chatting about how to deal with combo objective functions and directives (as per #511, #492) . For example, `targetMisfit` currently needs to evaluate `dmisfit.survey.nD`, but if you have multiple data sets, what does this mean?? 

## Approaches 

There are a couple things we could do...
- `targetMisfit` then becomes a list of target misfits. This goes through the combo data misfit and produces a target misfit for each. (this limits flexibility - if you want to fit one data set to a chi fact of 1 and another to a chifact of 2, this starts to get complicated) 
- `targetMisfit` takes an objective function and knows to evaluate that. In the case of a joint inversion, you would simply give 2 target misfit directives in the `directivesList`. This then needs to be handled carefully in the inversion (we want both target misfits to be satisfied before exiting). 

I would vote for option 2 (and in general, providing objective functions to directives to that the directive always knows what it should be operating on).  

To handle the need to satisfy multiple "stopping criteria", I would suggest that directives have a `stopping_criteria_satisfied` (or something like that, name open to discussion) that is `None` if the directive doesn't trigger a reason to exit the inversion, is `False` by default and set to `True` once satisfied. The inversion then looks at the `directivesList` and sees if all stopping criteria are `True` or None and can exit based on that. 

Thoughts?

fourndo:
Option 2 sounds great to me.

re: stopping_criteria -> That's also a good idea. I case of joint inversion, there could be a scenario where one of the misfit function could be triggered on/off (in terms of gradient step) after converging, but we still want to update it's measure of data fit as the other function continues to update.

lheagy:
This is similar to the discussion currently happening in #637 - worth comparing notes here. cc @fourndo

Issue #559, opened 2017-03-07
Combo Objective function


fourndo:
The new regularization structure proposed by @lheagy is a two-level nesting of lists. The first level is the standard Combo Regularization object containing a list of base regularizations:

`reg1 = [reg.Small, reg.Derivx, reg.Derivy, reg.Derivz]`

Then there is the option of a higher level if combining regularizations (i.e. for different portions of the model using `Maps.Wires`)

`reg = reg1 + reg2 `

In which case we get a list of ComboObj

`reg = [reg1, reg2]`

So the issue is if I pass **reg** OR **reg1** to a function/directive, the function needs to adapt in order to change the variables on the first or second level of the list. Here is my current work around:

```python
def initialize(self):

        # Check if it is a ComboObjective
        if not isinstance(self.reg, Regularization.BaseComboRegularization):

            # It is a Combo objective, so will have to loop
            self.ComboObjFun = True

        # Don't worry about this line
        if self.mode == 1:

           # Here is the hack, needs an inner loop to touch the elements of each combo
            if self.ComboObjFun:

                self.norms = []
                for reg in self.reg.objfcts:
                    self.norms.append(reg.norms)
                    reg.norms = [2., 2., 2., 2.]

            else:
                # Just act on the first level
                self.norms = self.reg.norms
                self.reg.norms = [2., 2., 2., 2.]
```

rowanc1:
In this case do you just want to update all of the norms?

Maybe we could have a property in there that returns the regularizations as if it were `flatten`ed?

In which case you could have a single piece of code that says:

```python
for reg in self.reg.flatten():
    reg.norm = 2.
```

This would also work if perhaps your code was used with something that was more 2D in nature (e.g. symmetric cyl mesh?) and only has three sub-regularizations.

I think this would only fix this issue, and we will run into this again in quite a few different ways. Not quite sure of the way out yet!

cc @lheagy

lheagy:
I like the idea behind `flatten`, but it is a little subtle. The behaviour @rowanc1 is describing would (in the current implementation) turn `reg` into a comboRegularization... 

In @fourndo's example case, `reg1` and `reg2` each potentially have their own `mref` and potentially different `cell_weights`. A comboRegularization would try to share these. My first idea around this would be to create an `expose('norms')` which (first checks that both `reg1` and `reg2` have a property called `norms`) and then adds that property to the top level `reg` so that when you do 
```
reg = reg1 + reg2
reg.expose('norms')
[...]

PR #581, opened 2017-04-09
Feat/combo directives

lheagy:
## Directives and Combo Objective Functions for Joint inversions 

- first pass adding an `expose()` method on a combo objective function that gives you hooks to properties of the objective functions lower in the list (as per issues #577, #559)
- update directives to work with `expose` where appropriate
- allow `dmisfit` and `reg` to be set in the init of directives (so that for example we can have two target misfits) 
- give directives a `stopping_criteria_satisfied` property. If all of the directives in the directives list that have a stopping criteria are satisfied, the inversion will stop

This is a first pass, and definitely needs a bit of cleanup, but I would appreciate feedback before going any further.

### DIFF
--- a/SimPEG/ObjectiveFunction.py
+++ b/SimPEG/ObjectiveFunction.py
@@ -429,7 +429,7 @@ class ComboObjectiveFunction(BaseObjectiveFunction):
             properties = dict(zip(properties, len(properties)*[None]))
 
         # go through the properties list and expose them
-        for prop, val in properties.iteritems(): # skip if already in self._exposed
+        for prop, val in properties.items(): # skip if already in self._exposed
             # if prop not in self._exposed:
             if getattr(type(self), prop, None) is not None:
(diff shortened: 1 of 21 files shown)


SimPEG/ObjectiveFunction.py
```diff
@@ -371,6 +400,64 @@ def W(self):
                 W.append(curW)
         return sp.vstack(W)
 
+    def expose(self, properties):
```
lheagy: I would appreciate a quick review on this by someone well-versed in python (@rowanc1, @bsmithyman, @jcapriot :) ). In particular - is messing with `__getattribute__` and `__setattr__` as in lines 444 a terrible idea??

lheagy:
after chatting with @rowanc1, the current implied behaviour of the get is unclear for an exposed property
image


instead, we should be explicitly linking properites
image

PR #583, opened 2017-04-14
Patch/survey pairs

lheagy:
## survey.pair(prob)
- have the survey check if it is already paired to problem and raise an error if it is (cc @fourndo: I updated your examples to do `survey.unpair()` if you are re-using the same survey more than once.  
- whitespace cleanup in Survey.py


lheagy:
@fourndo, @rowanc1: can you guys take a quick look at this when you get a chance? It is a small pr, but will hopefully prevent odd issues arising in joint inversions down the road in the case of a survey being paired to more than one problem

PR #592, opened 2017-04-26
Volumes in Regularization

lheagy:
- add back the volume term in the small, smooth contributions
- use the square root of the cell weights as the diag of the W matrix cc @fourndo

### DIFF
--- a/SimPEG/Regularization.py
+++ b/SimPEG/Regularization.py
@@ -770,9 +770,9 @@ class SimpleSmall(BaseRegularization):
         \\mathbf{W} (\\mathbf{m} - \\mathbf{m_{ref}})
 
     where :math:`\\mathbf{m}` is the model, :math:`\\mathbf{m_{ref}}` is a
-    reference model (default Zero) and :math:`\\mathbf{W}` is a weighting
-    matrix (default Identity. If cell weights are provided, then it is
-    :code:`diag(cell_weights)`)
+    reference model and :math:`\\mathbf{W}` is a weighting
+    matrix (default Identity). If cell weights are provided, then it is
+    :code:`diag(np.sqrt(cell_weights))`)
 
 
     **Optional Inputs**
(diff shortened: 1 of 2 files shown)


SimPEG/Regularization.py
```diff
@@ -801,6 +751,62 @@ def mapping(self, val):
...
+            mesh=mesh, **kwargs
+        )
+
+    @property
+    def W(self):
+        """
+        Weighting matrix
+        """
+        if self.cell_weights is not None:
+            return Utils.sdiag(np.sqrt(self.cell_weights))
```
lheagy: `SimpleSmall` uses the sqrt of the cell weights as the diag if they are included. There is no volume contribution.

SimPEG/Regularization.py
```diff
@@ -951,6 +957,58 @@ def __init__(
...
+
+        super(Small, self).__init__(
+            mesh=mesh, **kwargs
+        )
+
+    @property
+    def W(self):
+        """
+        Weighting matrix
+        """
```
lheagy: Small uses the sqrt of the cell_weights times the volume as the diag

SimPEG/Regularization.py
```diff
@@ -1011,20 +1069,20 @@ def W(self):
         Weighting matrix that constructs the first spatial derivative stencil
         in the specified orientation
         """
-        W = getattr(
+        vol = self.regmesh.vol.copy()
+        if self.cell_weights is not None:
+            vol *= self.cell_weights
+
```
lheagy: for the SmoothDeriv, we multiply be `sdiag(sqrt(Ave * (vol * cell_weights))`

fourndo: Confirmed

SimPEG/Regularization.py
fourndo: Confirmed

PR #492, opened 2016-11-01
Mag amplitude

fourndo:
Here is a great example of a PR that went out of control.
Don't try this at home kids, I don't know what I am doing :\

Features in the making:
------------------------
- [x] Implement Magnetic Vector Inversion in Cartesian (MVI-C)
- [x] Implement Magnetic Amplitude Inversion (MAI)
- [x] Added functionality for Equivalent Source inversion
- [x] Create tests for MVI and MAI
- [x] Publish example for Cooperative Method (CMI) - 
   Motivated by paper with Jeff/USGS in the making ... say whaaat!

Gravity - PDE
--------------
- [x] Forward gravity and gravity gradiometry in PDE
- [x] Analytic tests for G and GG

- [ ] Document everything

PR #593, opened 2017-04-26
directive validation

lheagy:
- changes from #580 onto the objective function branch (as per #578). 
- `BaseDirective` has a method `validate(self, directiveList=None)` which returns `True` by default
-  `DirectiveList` has a method `validate(self)` that goes through the directive list and validates each one
- `InvProb` runs the directive list validation on setter


fourndo:
Yup looks pretty clean to me. Go for it. The changes would be my last updates to #492

Issue #544, opened 2017-02-18
Getting Started


rowanc1:
The docs say that we should install Python 2.7. This should be changed to 3!

lheagy:
👍  We should also consider combining the "Getting started" that is currently in the tutorials

rowanc1:
@lheagy do you have a plan on how we keep `tutorials` and `simpeg` docs separate?

lheagy:
This is not entirely clear at the moment, but I think that they should have distinct purposes:

`simpeg` docs examples: each example demonstrates how to accomplish a task within SimPEG. I would love to see the addition of a `Purpose` at the top of each example - what are we going to accomplish in this example? The target audience for these types of examples I would say is people who are familiar with the field and with running numerical simulations, but are new to SimPEG

`tutorials`: these contain much more background information and theory. The `stories` are work through a problem and explain underlying concepts (eg.  http://tutorials.simpeg.xyz/content/stories/TKC_DC_story.html). The intent of the other pages (Pixels and their neighbours, ...) each explain a fundamental concept of inversions. We might want to consider moving the `tutorials` to http://inversion.geosci.xyz. 

There are some points of overlap that should be addressed:
- Getting started: http://tutorials.simpeg.xyz/content/gettingstarted.html
- practices: http://tutorials.simpeg.xyz/content/practices.html 
Both of these I think may be better suited in the SimPEG docs. I would be happy to move them over.

rowanc1:
I like the `inversion.geosci.xyz` idea. There is quite a lot of content that we could also pull from the SimPEG paper (CC-BY 👍). I will add a `purpose` to the examples and get people to look over them.

I agree with moving the `Getting started` and `Practices` over. I can tackle that to start with.

lheagy:
Yeah, I think `inversion.geosci.xyz` clarifies the scope of that body of work and how it is distinct from the SimPEG docs. Perfect, let me know when you are done and I can transfer the tutorial repo to `geoscixyz/inversion`

rowanc1:
Talking through this with @lheagy @dougoldenburg and @sgkang we were thinking `computation` rather than `inversion`. We have started a first pass here:

https://github.com/geoscixyz/computation

lheagy:
This closed with #595

PR #595, opened 2017-05-03
Feat/docs

lheagy:
Bring in the documentation pages from the previous SimPEG tutorials (as per #544):
- "Getting Started for Developers"
- "Practices" 

image

- Clean up use of brackets in an assert statement in Directives


docs/content/api_core/api_getting_started_developers.rst
```diff
@@ -0,0 +1,226 @@
...
+
+
+.. _getting_started_if_all_is_well:
+
+If all is well ...
+------------------
+
+You should be able to open a terminal within SimPEG/examples and run an example, ie.::
+
+    python 01-basic/plot_inversion_linear.py
```
rowanc1: I don't think this one works with the new example system?

lheagy: image
(we added more data, so I think it should be fine)

rowanc1: Oh, I see you updated the plot. I thought this still was `Examples.Linear.run()` or something. Weird. Looks good. :)

Issue #596, opened 2017-05-03
remote download default


lheagy:
Right now, the remote download defaults to the current working directory. Do we want it here? or should it default to your Downloads folder?

cc @fourndo, @rowanc1

lheagy:
or at the very least, we should have an option that allows you to specify that you want it going to your downloads folder so that we don't need to be doing this in scripts / notebooks (it is an intimidating way to start a tutorial!) 
```
basePath = os.path.sep.join(
    os.path.abspath(os.getenv('HOME')).split(
        os.path.sep
    )['Downloads']+['SimPEGtemp']
)
```

fourndo:
I say go for it like you proposed
```remoteDownload(url, cloudfiles, downloads_folder=True)``

**Comment by lheagy** [comment]
cool, for naming of variables, does the following work for you? I think it is a little more standard
```
def remoteDownload(url, remoteFiles, path='.', directory='SimPEGtemp', rm_previous=False)
```

(where `rm_previous` is either `True` or `False` and if a directory with the specified name already exists and `rm_previous` is `True`, that directory will be deleted and then written over)

ahartikainen:
(You can ignore this comment)

Why there needs to be a default path?

lheagy:
@ahartikainen: we have changed it so that the default path is the local directory (we need to specify a place to download the file to)

lheagy:
the default `folder` is not `.` (the current working directory). These changes were made in #598

PR #598, opened 2017-05-04
Ref/remote download

lheagy:
modify the `remoteDownload` utils so that more of the string handling is done by the util as per #596 

Before
```
basePath = os.path.sep.join(os.path.abspath(os.getenv('HOME')).split(os.path.sep)+['Downloads']+['SimPEGtemp'])
basePath = os.path.abspath(remoteDownload(url, cloudfiles, basePath=basePath+os.path.sep))
```

After
```
basePath = remoteDownload(url, path='~/Downloads', overwrite=True)
```

I have also updated the examples and tests that rely on this function

### DIFF
--- a/SimPEG/EM/Base.py
+++ b/SimPEG/EM/Base.py
@@ -256,7 +256,9 @@ class BaseEMProblem(Problem.BaseProblem):
         Inverse of the edge inner product matrix for \\(\\sigma\\).
         """
         if getattr(self, '_MeSigmaI', None) is None:
-            self._MeSigmaI = self.mesh.getEdgeInnerProduct(self.sigma, invMat=True)
+            self._MeSigmaI = self.mesh.getEdgeInnerProduct(
+                self.sigma, invMat=True
+            )
         return self._MeSigmaI
 
     # TODO: This should take a vector
@@ -425,7 +427,7 @@ class BaseEMSrc(Survey.BaseSrc):
         """
         return Utils.Zero()
 
-    def s_mDeriv(self, prob, v, adjoint = False):
+    def s_mDeriv(self, prob, v, adjoint=False):
         """
         Derivative of magnetic source term with respect to the inversion model
 
@@ -438,7 +440,7 @@ class BaseEMSrc(Survey.BaseSrc):
 
         return Utils.Zero()
 
-    def s_eDeriv(self, prob, v, adjoint = False):
+    def s_eDeriv(self, prob, v, adjoint=False):
         """
         Derivative of electric source term with respect to the inversion model
 
--- a/SimPEG/Utils/io_utils.py
+++ b/SimPEG/Utils/io_utils.py
@@ -3,6 +3,7 @@ import numpy as np
 from SimPEG import Mesh
 import time as tm
 import re
+import warnings
 
 
 def read_GOCAD_ts(tsfile):
@@ -146,11 +147,17 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
     return insideGrid
 
 
-def remoteDownload(url, remoteFiles, basePath=None):
+def remoteDownload(
+    url, remoteFiles, path='.', directory='SimPEGtemp', rm_previous=False
+):
     """
     Function to download all files stored in a cloud directory
-    var:    url ("http:\\...")
-    list:   List of file names to download
+
+    :param str url: url of the storage bucket ("http://...")
+    :param list remoteFiles: List of file names to download from the storate bucket
+    :param str path: path to where the directory is created and files downloaded (default is the current directory)
+    :param str directory: name of the directory to be created and have content downloaded to
+    :param bool rm_previous: remove file and contents if a directory with the specified name already exists
     """
 
     # Download from cloud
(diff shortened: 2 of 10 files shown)


SimPEG/EM/Base.py
```diff
@@ -256,7 +256,9 @@ def MeSigmaI(self):
         Inverse of the edge inner product matrix for \\(\\sigma\\).
         """
         if getattr(self, '_MeSigmaI', None) is None:
-            self._MeSigmaI = self.mesh.getEdgeInnerProduct(self.sigma, invMat=True)
+            self._MeSigmaI = self.mesh.getEdgeInnerProduct(
```
lheagy: this is just white-space

SimPEG/Utils/io_utils.py
```diff
@@ -146,35 +147,72 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    :param bool rm_previous: remove file and contents if a directory with the specified name already exists
     """
 
     # Download from cloud
     import urllib
     import shutil
     import os
     import sys
+
+    def rename_path(downloadpath):
```
lheagy: if we start to make heavy use of this function, I can see a number of scenarios where we do not necessarily want to over-write the contents of `SimPEGtemp`, so here it will add `SimPEGtemp(1)` (or up the integer value until we hit a unique name, similar to what you would see with a web browser)

SimPEG/Utils/io_utils.py
```diff
@@ -146,35 +147,72 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
     return insideGrid
 
 
-def remoteDownload(url, remoteFiles, basePath=None):
+def remoteDownload(
+    url, remoteFiles, path='.', directory='SimPEGtemp', rm_previous=False
```
lheagy: see what you think of these input names @fourndo, @rowanc1

SimPEG/Utils/io_utils.py
```diff
@@ -146,35 +147,72 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
-
-    if os.path.exists(basePath):
-        shutil.rmtree(basePath)
-
-    os.makedirs(basePath)
-
-    print("Download files from URL...")
+    # ensure we are working with absolute paths
+    path = os.path.abspath(path)
+    downloadpath = os.path.sep.join([path]+[directory])
```
rowanc1: `[path, directory]`? Same difference.

examples/04-grav/plot_laguna_del_maule_inversion.py
```diff
@@ -25,14 +25,18 @@ def run(plotIt=True, cleanAfterRun=True):
...
-                                (os.path.sep)+['Downloads']+['SimPEGtemp'])
-    basePath = os.path.abspath(remoteDownload(url,
-                                              cloudfiles,
-                                              basePath=basePath+os.path.sep))
+    cloudfiles = [
+        'LdM_grav_obs.grv', 'LdM_mesh.mesh',
+        'LdM_topo.topo', 'LdM_input_file.inp'
+    ]
+
+    # Download to Downloads/SimPEGtemp
```
rowanc1: Here we are not specifying where things get put? I think they should just go directly into `pwd`? Unless you specify a `folder` or something?

Can we just rename this to `download`? seems a bit redundant.

```python

file_names = download([url1, url2], folder='~/Downloads/mag_stuff', overwrite=True)
# or 
file_name = download(url1)
# where
assert isinstance(file_names, list)
assert len(file_names) == 2
assert isinstance(file_name, str)
```

I think that the default should the the pwd of where the python file is run from, appending a `SimPEGtemp` seems odd to me. Downloads are downloads.

rowanc1: This also would mean that you don't have to reconstruct the file name after the fact, you just loop through the list:

```python
with open(file_name, 'r') as f:
    f.read()

with open(file_names[0], 'r') as f2:
    f2.read()
```

rowanc1:
Does this work on windows as well? 

`C:\Users\Myname\Downloads`?

SimPEG/Utils/io_utils.py
lheagy: Yeah, this is likely confusing. `path` is the path to your working directory, `directory` is the name of the new folder we are adding where we download things. 

Thoughts on alternatives?

lheagy:
This `~` as a path does not work on windows, but `os.getenv('HOME')` seems to work for most systems.

examples/04-grav/plot_laguna_del_maule_inversion.py
lheagy: - `download` works for me. @fourndo ?
- SimPEGtemp was mainly motivated because we make heavy use of this in examples, so for someone unfamiliar it is easy to find (we can still do this and specify it in the example script rather than the util). However, I agree that can be cleared up by specifying the folder (I am tempted to call it `path` or `directory` instead as it is not just a name of a folder but does include path information)

rowanc1:
You can use `expanduser`:
https://docs.python.org/2/library/os.path.html#os.path.expanduser

examples/04-grav/plot_laguna_del_maule_inversion.py
rowanc1: I think that it could be both? You can call `abspath` on the thing and it should expand it out regardless.

```python
folder='here'
folder='or/here'
folder='/Users/rowan/downloads/or/even/here'
```

I see all of these as folders?

lheagy:
I think most of this works - the main subtlety here is the behaviour of `overwrite`. Previously it deleted the old directory and replaced it with a new one, now, we are not forcing the creation of a new directory, so I am hesitant to have overwrite remove the directory - it should only overwrite the files (otherwise I can just see someone trying and rapidly regretting `download(url, `~`, overwrite=True)`)

rowanc1:
Yeah, I think that removing things should never happen. :) Only overwriting files, never folders.

lheagy:
- [ ] @fourndo 
- [ ] @rowanc1 

please take a look when you get a chance and let me know what you think. Thanks!

tests/base/test_utils.py
```diff
@@ -303,5 +307,30 @@ def testProbing(self):
...
+
+        file_names = download(
+            [url1, url2], path='./test_urls', overwrite=True
+        )
+        # or
+        file_name = download(url1, path='./test_url', overwrite=True)
+        # where
+        assert isinstance(file_names, list)
+        assert len(file_names) == 2
+        assert isinstance(file_name, str)
```
rowanc1: Ha, it's like I coded...

rowanc1:
This looks great! thanks @lheagy  :)

examples/04-grav/plot_laguna_del_maule_inversion.py
lheagy: ok, fair enough, folder can work

PR #602, opened 2017-05-16
Patch for BCGS talk

sgkang:
- add static utils that we can drape electrode locations below topography
- VTEM waveform


sgkang:
Not sure where to fix on my end. Any ideas? @lheagy @rowanc1 
image

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -74,6 +74,30 @@ def eval(self, time):
...
 
 
+class VTEMWaveform(BaseWaveform):
+
+    eps = 1e-9
+    offTime = 4.2e-3
+    peakTime = 2.73e-3
+    a = 3.
+
+    def __init__(self, offTime=4.2e-3, peakTime=2.73e-3, a=3.):
```
lheagy: re-stating the offTime, peak time and a here is redundant (stating it in lines 79 - 82) is enough. You can replace these with `**kwargs` instead and in the init below just do
```
BaseWaveform.__init__(self, hasInitialFields=False, **kwargs) 
```
and that will take care of setting the parameters

sgkang: agree. did remove unnecessary statements

PR #606, opened 2017-05-30
download the file for testing

lheagy:
@thast: I have uploaded the test file to https://storage.googleapis.com/simpeg/tests/utils/vancouver_topo.xyz and in the testing script, we can download it. If you are good with this, we can remove the file from the repo and just access it from the cloud. see what you think!


thast:
This is a lot cleaner. Go for it.

PR #611, opened 2017-06-07
NSEM layered earth analytic

thast:
Adding a wrapped version of the MT layered Earth app to use as a analytic Utils.


thast:
@lheagy : here is my proposal for the Layered Earth NSEM Utils. Let me know what you think.
- I tried to stick with the format of FDEMCasing.py
- At the end of the file, you will find some examples on how to use it (also serve as a test)

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,128 @@
...
+        UD[:, i] = _Tinv(H[i+1], K[i])*_Pinv(Z[i])*_P(Z[i+1])*UD[:, i+1]
+        UD = UD/((_np.abs(UD[0, :]+UD[1, :])).max())
+
+    for j in range(0, n+1):
+        EH[:, j] = _np.matrix([[1., 1, ], [-1./Z[j], 1./Z[j]]])*UD[:, j]
+
+    return UD, EH, Z, K
+
+# Utils to compute the apparent impedance over a layered Earth Model
+def AppImpedance_LayeredEarth(freq, thickness, sig, nlayer, format='Res-Phase', chg=0., tau=0., c=0., mu_r=1., eps_r=1.):
```
lheagy: What do you think of `MT_LayeredEarth` ? (`AppImpedance` isn't necessarily intuitive??) 

also, for a variable name `format` --> `return_type` ? see what you think

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,128 @@
+import numpy as _np
```
thast: It was a try to make things clearer but from a user point of view.
importana
The Tab completion does not look like quite as what I assume we want to look like.

With the above 'trick', here is what I get:
importnsem

It is easy to change back but that is might be something we want to think of.

SimPEG/EM/Analytics/NSEM.py
thast: I like it.

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,128 @@
...
+    return UD, EH, Z, K
+
+# Utils to compute the apparent impedance over a layered Earth Model
+def AppImpedance_LayeredEarth(freq, thickness, sig, nlayer, format='Res-Phase', chg=0., tau=0., c=0., mu_r=1., eps_r=1.):
+    '''
+    This code compute the analytic response of a n-layered Earth to a plane wave (Magnetotellurics).
+    The solution is first developed in Ward and Hohmann 1988.
+    See also http://em.geosci.xyz/content/maxwell3_fdem/natural_sources/MT_N_layered_Earth.html
+
+    freq: frequency, either float or numpy array
```
thast: Thanks! I had no experience with this, good occasion.

SimPEG/EM/Analytics/NSEM.py
rowanc1: You can either specify an `__all__` parameter in the file.

https://stackoverflow.com/questions/2187583/whats-the-python-all-module-level-variable-for

Or probably better is to explicitly import things in the higher level module (`Analytics`).

```pyton
from NSEM import AppImpedance_LayeredEarth
from NSEM import hzAnalyticDipoleF
```

It is more verbose. However, it is explicit, and I think a better option than changing expected variable names.

thast: Thanks @lheagy  and @rowanc1 for the feedbacks. I'm learning a lot like this.
the higher level explicit import in Analytics __init__.py  was already done, so it did not fix the tab completion issue.
I just reverse all the _np import etc. and I set up the __all__ parameter but it does not fix the tab completion issue either (it seems is effect is mainly when we do  from __ import *). 
See below:
__init__ py simpeg 2017-06-08 20-11-19

rowanc1: At some level, I think that you should be accessing this from the `geoana. <tab>`? Does that work as expected (perhaps commenting out the `import *`'s for testing...)?

When you access the module directly, I think this is opening that module and loading everything. It also sounds like the IPyKernel does not respect the `__all__`, although you can configure it to that if you want:

https://github.com/ipython/ipykernel/issues/129

image

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,140 @@
...
+        UD = UD/((np.abs(UD[0, :]+UD[1, :])).max())
+
+    for j in range(0, n+1):
+        EH[:, j] = np.matrix([[1., 1, ], [-1./Z[j], 1./Z[j]]])*UD[:, j]
+
+    return UD, EH, Z, K
+
+# Utils to compute the apparent impedance over a layered Earth Model
+def MT_LayeredEarth(freq, thickness, sig, nlayer, return_type='Res-Phase', chg=0., tau=0., c=0., mu_r=1., eps_r=1.):
+    '''
```
lheagy: This is beautiful, thanks @thast! One small comment: I am not sure if this makes a difference, but could be causing the fail, typically docstrings are inside of `"`, eg. 
```
"""
MT layered earth docs
"""
```

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,141 @@
...
+
+        App_ImpZ[i] = EH[0, 1]/EH[1, 1]
+
+        Res[i] = np.abs(App_ImpZ[i])**2./(mu_0*omega(F[i]))
+        Phase[i] = np.angle(App_ImpZ[i], deg=True)
+
+    if return_type == 'Res-Phase':
+        return Res, Phase
+
+    elif return_type == 'Complex':
```
lheagy: how about labeling this "Impedance" ? (Complex doesn't fully capture what it is). Thoughts @thast?

thast: Sure

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,141 @@
...
+        UD[:, i] = _Tinv(H[i+1], K[i])*_Pinv(Z[i])*_P(Z[i+1])*UD[:, i+1]
+        UD = UD/((np.abs(UD[0, :]+UD[1, :])).max())
+
+    for j in range(0, n+1):
+        EH[:, j] = np.matrix([[1., 1, ], [-1./Z[j], 1./Z[j]]])*UD[:, j]
+
+    return UD, EH, Z, K
+
+# Utils to compute the apparent impedance over a layered Earth Model
+def MT_LayeredEarth(freq, thickness, sig, nlayer, return_type='Res-Phase', chg=0., tau=0., c=0., mu_r=1., eps_r=1.):
```
lheagy: is nlayer necessary? it should be the same as len(thickness), len(sig), right?

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,141 @@
...
+
+# Evaluate Impedance Z of a layer
+_ImpZ = lambda f, mu, k: omega(f)*mu/k
+
+# Complex Cole-Cole Conductivity - EM utils
+_PCC = lambda siginf, m, t, c, f: siginf*(1.-(m/(1.+(1j*omega(f)*t)**c)))
+
+# matrix P relating Up and Down components with E and H fields
+_P = lambda z: np.matrix([[1., 1, ], [-1./z, 1./z]], dtype='complex_')
+_Pinv = lambda z: np.matrix([[1., -z], [1., z]], dtype='complex_')/2.
```
lheagy: what does the underscore do for `dtype='complex_'`??

thast: Shorthand for 'complex128' (64 bit real and imaginary)
https://docs.scipy.org/doc/numpy-1.12.0/user/basics.types.html

SimPEG/EM/Analytics/NSEM.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    This code compute the analytic response of a n-layered Earth to a plane wave (Magnetotellurics).
+    The solution is first developed in Ward and Hohmann 1988.
+    See also http://em.geosci.xyz/content/maxwell3_fdem/natural_sources/MT_N_layered_Earth.html
+
+    :param freq: the frequency at which we take the measurements
+    :type freq: float or numpy.array
+    :param thickness: thickness of the Earth layers in meters
+    :type thickness: float or numpy.array
+    :param sig: electric conductivity of the Earth layers in S/m
```
lheagy: the order here is top to bottom?

SimPEG/EM/Analytics/NSEM.py
lheagy: my suggestion here would be to remove `nlayer` and add a check that sigma and thickness are the correct lengths 
```python
assert len(sigma) == len(thickness) + 1, (
    "len(sigma) should be 1 longer than len(thickness), "
    "the vectors provided  are len(sigma): {}, len(thickness): {}".format(len(sigma), len(thickness))
)
```

SimPEG/EM/Analytics/NSEM.py
thast: Yes, the arrays describe the layers properties from top to bottom

SimPEG/EM/Analytics/NSEM.py
lheagy: ah, cool thanks

SimPEG/EM/Analytics/NSEM.py
lheagy: kk, would you mind updating the docs so that is transparent to the user?

thast: added to documentation

SimPEG/EM/Analytics/NSEM.py
thast: removed

Issue #594, opened 2017-04-27
Passing of `m` (model) in `invProblem` class and the framework.


grosenkj:
I have been looking at the `invProblem` class a bit lately. I find the passing of the model, `m`, into all the methods of various objects in the framework confusing and as I see redundant.

rowanc1:
My comment would be that if we removed the input argument of model to many of the sub functions/classes, they could not be used in a stand alone way. This would be especially true of the regularization and data misfit classes, which you might want to test out as a function (in @lheagy new combo objective function classes). As well as using the survey and problem objects independently in forward modelling situations.

On the `InvProblem` the `model` is primarily stored for faster look ups of fields, as there is some caching going on, as well as storage and being accessible to the directives.

Perhaps though this is a pretty confusing point? Maybe this should be called `current_model` or something? I think that was changed recently...

Let me know if this clears things up.

PR #612, opened 2017-06-11
Em/patch/zero util

lheagy:
Remove use of Utils.Zero() in EM code and instead be explicit about sizes of Zero vectors that are returned and passed around. 

`Utils.Zero()` is too touchy. order of operations matters `a + Utils.Zero()` does not do the same thing as `Utils.Zero() + a` and it behaves differently on different operating systems (including the most recent travis failures on dev: https://travis-ci.org/simpeg/simpeg/builds/238426219). On my machine, I cannot reproduce those errors. 

This also promotes us being clear about sizes of vectors, matrices, which I think is a good thing anyways.

### DIFF
--- a/SimPEG/EM/Base.py
+++ b/SimPEG/EM/Base.py
@@ -140,7 +140,7 @@ class BaseEMProblem(Problem.BaseProblem):
         Derivative of :code:`MfMui` with respect to the model.
         """
         if self.muiMap is None:
-            return Utils.Zero()
+            return Utils.spzeros(self.mesh.nF, len(self.model))
 
         return (
             self.mesh.getFaceInnerProductDeriv(self.mui)(u) * self.muiDeriv
@@ -162,7 +162,7 @@ class BaseEMProblem(Problem.BaseProblem):
         """
 
         if self.muiMap is None:
-            return Utils.Zero()
+            return Utils.spzeros(self.mesh.nF, len(self.model))
 
         if len(self.mui.shape) > 1:
             if self.mui.shape[1] > self.mesh.dim:
@@ -173,7 +173,6 @@ class BaseEMProblem(Problem.BaseProblem):
         dMfMuiI_dI = -self.MfMuiI**2
         dMf_dmui = self.mesh.getEdgeInnerProductDeriv(self.mui)(u)
         return dMfMuiI_dI * (dMf_dmui * self.muiDeriv)
-
 
     @property
     def MeMu(self):
@@ -190,7 +189,7 @@ class BaseEMProblem(Problem.BaseProblem):
         Derivative of :code:`MeMu` with respect to the model.
         """
         if self.muMap is None:
-            return Utils.Zero()
+            return Utils.spzeros(self.mesh.nE, len(self.model))
 
         return (
             self.mesh.getEdgeInnerProductDeriv(self.mu)(u) * self.muDeriv
@@ -212,7 +211,7 @@ class BaseEMProblem(Problem.BaseProblem):
         """
 
         if self.muMap is None:
-            return Utils.Zero()
+            return Utils.spzeros(self.mesh.nE, len(self.model))
 
         if len(self.mu.shape) > 1:
             if self.mu.shape[1] > self.mesh.dim:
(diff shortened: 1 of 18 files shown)


SimPEG/EM/Base.py
```diff
@@ -189,7 +189,7 @@ def MeMuDeriv(self, u):
         Derivative of :code:`MeMu` with respect to the model.
         """
         if self.muMap is None:
-            return Utils.Zero()
+            return Utils.spzeros(self.mesh.nE, len(self.model))
```
lheagy: eventually, these should take vectors (See #594)

SimPEG/EM/FDEM/FieldsFDEM.py
```diff
@@ -351,16 +351,6 @@ def startup(self):
         self._MfMui = self.survey.prob.MfMui
         self._MfMuiDeriv = self.survey.prob.MfMuiDeriv
 
-    def _GLoc(self, fieldType):
```
lheagy: This is already handled by the base class and is therefore unnecessary here

SimPEG/EM/FDEM/SrcFDEM.py
```diff
@@ -435,14 +481,16 @@ def s_eDeriv(self, prob, v, adjoint=False):
...
-                return Zero()
+            elif prob._formulation == 'HJ':
+                if adjoint is True:
+                    return np.zeros_like(prob.model)
+                return np.zeros(prob.mesh.nF)
                 # raise NotImplementedError
-                mu_s = prob.mu - self.mu
-                MMui_s = prob.mesh.getEdgeInnerProduct(mu_s, invMat=True)
-                C = prob.mesh.edgeCurl.T
+                # mu_s = prob.mu - self.mu
```
lheagy: These lines were not hit previously. I would like to leave them as comments for thinking through this problem

SimPEG/EM/NSEM/SrcNSEM.py
```diff
@@ -24,7 +24,8 @@ class BaseNSEMSrc(FDEMBaseSrc):
     '''
 
     freq = None #: Frequency (float)
-
+    _ePrimary = None  #: Primary electric field
+    _bPrimary = None  #: Primary magnetic field
```
lheagy: @grosenkj: I removed these from the base class (there are scenarios, like a prim-sec where we have derivs on the primary that this is dangerous, so I think it should be handled on a case-by-case basis, rather than making an assumption on the base class)

SimPEG/EM/FDEM/SrcFDEM.py
```diff
@@ -17,11 +16,6 @@ class BaseFDEMSrc(BaseEMSrc):
 
     freq = properties.Float("frequency of the source", min=0, required=True)
 
-    _ePrimary = None
-    _bPrimary = None
-    _hPrimary = None
-    _jPrimary = None
-
```
lheagy: stashing these in general is potentially dangerous - later, we should go through and stash on sources where the primary doesn't change.

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -81,20 +83,24 @@ def eval(self, time):
 
 class VTEMWaveform(BaseWaveform):
 
-    eps = 1e-9
-    offTime = 4.2e-3
-    peakTime = 2.73e-3
-    a = 3.
-    hasInitialFields = False
+    offTime = properties.Float(
```
lheagy: @sgkang: I made these properties.

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -346,14 +412,14 @@ class LineCurrent(BaseTDEMSrc):
     :param list rxList: receiver list
     :param bool integrate: Integrate the source term (multiply by Me) [False]
     """
-    waveform = None
```
lheagy: this is handled by the base class

SimPEG/Props.py
```diff
@@ -307,14 +307,14 @@ def get_property(self):
 
         def fget(self):
             if scope.physical_property is None:
-                return Utils.Zero()
+                return None
```
lheagy: we return None's for the Maps if they don't exist, so doing the same for the derivs would seem to make sense - see what you think @rowanc1

lheagy:
image

PR #614, opened 2017-06-13
Em/patch booky

lheagy:
- Include scripts to reproduce the bookpurnong inversions for the paper as examples
- start a "Published" directory of the examples to put examples that are in publications
- make use of properties in waveforms for the TDEM code 
- some travis cleanup


.travis.yml
```diff
@@ -53,9 +53,9 @@ before_install:
 
 install:
   - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then
-      conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib cython ipython h5py vtk;
+      conda install --yes pip python=$TRAVIS_PYTHON_VERSION numpy=1.12.1 scipy matplotlib cython ipython h5py vtk;
```
lheagy: this should be removed now that discretize has been fixed

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -81,14 +98,20 @@ def eval(self, time):
...
+    offTime = properties.Float(
+        "off-time of the source", default=4.2e-3
+    )
+
+    peakTime = properties.Float(
+        "Time at which the VTEM waveform is at its peak", default=2.73e-3
+    )
+
+    a = properties.Float(
+         "parameter controlling how quickly the waveform ramps on", default=3.
```
lheagy: @sgkang: what do you think of this description?

examples/00_published/plot_booky_1Dstitched_resolve_inv.py
```diff
@@ -0,0 +1,363 @@
...
+from scipy.spatial import cKDTree
+
+from SimPEG import (
+    Mesh, Maps, Utils, DataMisfit, Regularization,
+    Optimization, Inversion, InvProblem, Directives
+)
+import SimPEG.EM as EM
+
+
+def download_and_unzip_data(
```
lheagy: @rowanc1: do you think this is something we should generalize as a util?

PR #615, opened 2017-06-13
Em/mt tutorials

sgkang:
Update `SaveOutputEveryIteration` to save smoothness and smalness
Add plotting utils to plot L-curve, misfits ...


lheagy:
the travis fail is still numpy ufunc stuff... 
image

(we are doing `+=` in a few places... if we change those, I *think* we should be good to go)

rowanc1:
Looks like two lines in the `FieldsTDEM`, should be easy to fix.

```
File "/home/travis/build/simpeg/simpeg/SimPEG/EM/TDEM/FieldsTDEM.py", line 252, in _dbdt
    s_m[:, i] += s_m_src


File "/home/travis/build/simpeg/simpeg/SimPEG/EM/TDEM/FieldsTDEM.py", line 352, in _j
    s_e[:, i] += s_e_src
```

PR #618, opened 2017-06-20
Documentation Builds

lheagy:
## Changes in the docs build process
- store pre-built docs on google storage bucket, download and unpack those docs during testing. This way only examples and documentation that changes is re-built.
    - As a result, examples are now only tested on python 2.7. gsutil only works on python 2.7 :( (see issue at GoogleCloudPlatform/gsutil#29). We can write a separate test for examples to ensure they are working in python 3 in a different pr
   - this also removes the link-check on the docs. again, we can work on bringing that back in a different pr
image
- ensure all TDEM tests always use Pardiso for a solver (we will time-out otherwise, and this allows us to test that pymatsolver is still functioning as expected with SimPEG)


.travis.yml
```diff
@@ -59,15 +63,25 @@ install:
...
+      tar -xzf $PWD/docs/_build.tar.gz -C $PWD/docs/_build ;
+      ls $PWD/docs/_build/ ;
     fi
 
 # Run test
 script:
-  - nosetests $TEST_DIR --with-cov --cov SimPEG --cov-config .coveragerc -v -s;
-
+  # - nosetests $TEST_DIR --with-cov --cov SimPEG --cov-config .coveragerc -v -s
+  - if [ "$TEST_DIR" = "$DEPLOY_DIR" -a ${TRAVIS_PYTHON_VERSION:0:1} == "3" ]; then
```
lheagy: on python 3, we only check that the docs build - the examples are not run (they take too long at the moment)

.travis.yml
```diff
@@ -86,17 +100,13 @@ after_success:
...
-  - tar -xvzf credentials.tar.gz
-
   # deploy docs
   - if [ $TRAVIS_PYTHON_VERSION == $DOCS_PY ]; then
-      if [ ! -d ${HOME}/google-cloud-sdk ]; then curl https://sdk.cloud.google.com | bash; fi ;
       if [ "$TRAVIS_BRANCH" = "$MASTER_BRANCH" ]; then
         gcloud auth activate-service-account --key-file credentials/client-secret.json ;
         export GAE_PROJECT=simpegdocs ;
       elif [ "$TRAVIS_BRANCH" = "$DEV_BRANCH" ]; then
+        gsutil cp docs/_build.tar.gz gs://simpeg ;
```
lheagy: whenever docs are updated on dev, we will re-upload to google

tests/em/tdem/test_TDEM_DerivAdjoint_RawWaveform.py
```diff
@@ -1,15 +1,10 @@
...
+from SimPEG import Mesh, Maps, Tests
 from SimPEG import EM
 from scipy.interpolate import interp1d
-
-try:
-    from pymatsolver import PardisoSolver
-    Solver = PardisoSolver
-except ImportError:
-    Solver = SolverLU
+from pymatsolver import Pardiso as Solver
```
lheagy: always using Pardiso as the solver

PR #619, opened 2017-06-21
Ref/saving directive

lheagy:
- save beta, phi_d, phi_m, etc as properties of the directive so they can be accessed interactively

image

- add the option to not write out a text file (eg if you only want to output results interactively in a notebook). The default is true, so it will output the txt file unless otherwise specified


SimPEG/Directives.py
```diff
@@ -288,38 +288,85 @@ class SaveOutputEveryIteration(SaveEveryIteration):
...
-        self.header = "  #     beta     phi_d     phi_m   phi_m_small     phi_m_smoomth_x     phi_m_smoomth_y     phi_m_smoomth_z      f\n"
-        f.write(self.header)
-        f.close()
+        if self.save_txt is True:
+            print(
+                "SimPEG.SaveOutputEveryIteration will save your inversion "
+                "progress as: '###-{0!s}.txt'".format(self.fileName)
+            )
+            f = open(self.fileName+'.txt', 'w')
+            self.header = "  #     beta     phi_d     phi_m   phi_m_small     phi_m_smoomth_x     phi_m_smoomth_y     phi_m_smoomth_z      phi\n"
```
sgkang: can we add phi_m_smooth?

lheagy: it can be computed pretty easily just by adding phi_m_smooth_x, ... do you think it is worth saving separately? we could write it as a property of the directive instead (eg. `save.phi_m_smooth` just adds the components together? )

lheagy: kk, I will start by adding it as a property on the method, and it can easily also be written out beyond that.

lheagy: @sgkang: I think this directive will soon need to be re-written. It is highly specific to tikhonov regularization and I think we want to have it "talk" to the regularization classes (perhaps we add a `name` to each regularization piece and that is what is printed out / saved here). Rather than tacking on more properties, would you mind if we just left this for now and re-think it with the objective function class in mind?

lheagy:
TODO
- [ ] update the load_results() to first check if save_txt is True, otherwise pass

PR #620, opened 2017-06-24
Ref/tdem testing

lheagy:
- clean up in the time domain testing
- efficiency fix in the Problem class. Now we check if the model has truly updated before clearing matrices:
image

image


SimPEG/Problem.py
```diff
@@ -123,10 +123,11 @@ def unpair(self):
     deleteTheseOnModelUpdate = []
 
     @properties.observer('model')
-    def _on_model_update(self, value):
-        for prop in self.deleteTheseOnModelUpdate:
-            if hasattr(self, prop):
-                delattr(self, prop)
+    def _on_model_update(self, change):
+        if self.model is None or not np.allclose(self.model, change['value']):
```
rowanc1: I wonder if an early exit of `self.model is change['value']` would be faster. If the model is large a point wise comparison is going to be slow. If they are pointing to the same block in memory, then we can skip that.

tests/em/tdem/test_TDEM_DerivAdjoint_RawWaveform.py
```diff
@@ -16,298 +17,360 @@
...
 
-        def test_Jvec_j_dhdtx(self):
-            self.JvecTest(prbtype='j', rxcomp='dhdtx')
+        def test_Jvec_adjoint_j_dhdtz(self):
+            self.JvecVsJtvecTest('dhdtz')
 
-        def test_Jvec_j_dhdtz(self):
-            self.JvecTest(prbtype='j', rxcomp='dhdtz')
 
+# class TDEM_DerivTests(unittest.TestCase):
```
rowanc1: Would be nice to have these tests perhaps tested on dev-->master? Just so that we know that it is still working, because they are useful for debugging.

SimPEG/Problem.py
```diff
@@ -123,10 +123,13 @@ def unpair(self):
     deleteTheseOnModelUpdate = []
 
     @properties.observer('model')
-    def _on_model_update(self, value):
-        for prop in self.deleteTheseOnModelUpdate:
-            if hasattr(self, prop):
-                delattr(self, prop)
+    def _on_model_update(self, change):
+        if self.model is change['value']:
```
rowanc1: `change['prev'] is change['value']` the model will always be the value!!

SimPEG/Problem.py
```diff
@@ -123,7 +123,15 @@ def unpair(self):
     deleteTheseOnModelUpdate = []
 
     @properties.observer('model')
-    def _on_model_update(self, value):
+    def _on_model_update(self, change):
+        if change['previous'] is change['value']:
+            return
+        if (
+            not isinstance(change['previous'], properties.utils.Sentinel) and
```
rowanc1: compare to `is properties.undefined`

tests/em/tdem/test_TDEM_DerivAdjoint_RawWaveform.py
lheagy: right now the TDEM tests are still timing out, so this would first take some re-organizing of tests

SimPEG/Problem.py
lheagy: perfect, thanks @rowanc1

tests/em/tdem/test_TDEM_MultiSrc_DerivAdjoint.py
```diff
@@ -1,160 +0,0 @@
-from __future__ import division, print_function
```
lheagy: I have updated test_TDEM_DerivAdjoint to test multiple sources, so this test is now redundant

tests/em/static/test_DC_jvecjtvecadj.py
```diff
@@ -130,7 +130,8 @@ def test_misfit(self):
                 self.survey.dpred(m), lambda mx: self.p.Jvec(self.m0, mx)
             ],
             self.m0,
-            plotIt=False
+            plotIt=False,
+            num=3
```
lheagy: @sgkang: I fixed the number of iterations this runs. If we let it go too far it fails b/c we hit machine precision.

examples/04-grav/plot_laguna_del_maule_inversion.py
```diff
@@ -24,15 +25,16 @@
...
-        'LdM_grav_obs.grv', 'LdM_mesh.mesh',
-        'LdM_topo.topo', 'LdM_input_file.inp'
-    ]
-
-    # Download to Downloads/SimPEGtemp
-    basePath = os.path.expanduser('~/Downloads/simpegtemp')
-    download([url+f for f in cloudfiles], folder=basePath, overwrite=True)
+    # directory where the downloaded files are
+
+    url = "https://storage.googleapis.com/simpeg/Chile_GRAV_4_Miller/Chile_GRAV_4_Miller.tar.gz"
```
lheagy: @fourndo, @craigm, I have zipped up the files into a .tar.gz (the originals are still in the same place, but downloading a .tar.gz is much faster than 4 independent files)

examples/04-grav/plot_laguna_del_maule_inversion.py
```diff
@@ -24,15 +25,16 @@
...
+    downloads = download(url, overwrite=True)
+    basePath = downloads.split(".")[0]
```
lheagy: I have also downloaded to the current directory (a folder called Chile_GRAV_4_Miller will be created). This makes it easier to work with different systems and to keep track of the files (esp on travis)

examples/00_published/plot_booky_1D_time_freq_inv.py
```diff
@@ -392,7 +392,7 @@ def run(plotIt=True, saveFig=False, cleanup=True):
 
     ax0.semilogx(sigma_re, z, 'k', lw=2, label="RESOLVE")
     ax0.semilogx(sigma_sky, z, 'b', lw=2, label="SkyTEM")
-    ax0.set_ylim(-100, 0)
+    ax0.set_ylim(-50, 0)
```
lheagy: this is the range we used in the paper

examples/04-grav/plot_laguna_del_maule_inversion.py
```diff
@@ -24,15 +25,16 @@
...
-    cloudfiles = [
-        'LdM_grav_obs.grv', 'LdM_mesh.mesh',
-        'LdM_topo.topo', 'LdM_input_file.inp'
-    ]
+    # directory where the downloaded files are
 
-    # Download to Downloads/SimPEGtemp
-    basePath = os.path.expanduser('~/Downloads/simpegtemp')
-    download([url+f for f in cloudfiles], folder=basePath, overwrite=True)
+    url = "https://storage.googleapis.com/simpeg/Chile_GRAV_4_Miller/Chile_GRAV_4_Miller.tar.gz"
```
lheagy: @fourndo are you good with this?

tests/em/tdem/test_TDEM_DerivAdjoint_RawWaveform.py
lheagy: I added this back on the test_TDEM_DerivAdjoint

examples/04-grav/plot_laguna_del_maule_inversion.py
craigmillernz: i'm good with that.

SimPEG/Utils/modelutils.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
 from scipy.interpolate import griddata, interp1d
 
-def surface2ind_topo(mesh, topo, gridLoc='CC', method='cubic', fill_value=np.nan):
+def surface2ind_topo(mesh, topo, gridLoc='CC', method='nearest', fill_value=np.nan):
```
lheagy: thanks @fourndo!

lheagy:
Thanks for your input all!

PR #623, opened 2017-06-29
Ref/objfcts

lheagy:
- ensure that all children of a combo regularization are looking at the exact same regmesh, mappings
- do not store both the mesh and the regmesh on the regularization. Only the regmesh is needed. 
- add tests to ensure children of a combo regularization are looking at the same regmesh and mapping objects in memory


tests/base/test_regularization.py
```diff
@@ -302,5 +303,35 @@ def test_update_of_sparse_norms(self):
...
 
+    def test_linked_properties(self):
+        mesh = Mesh.TensorMesh([8, 7, 6])
+        reg = Regularization.Tikhonov(mesh)
+
+        [self.assertTrue(reg.regmesh is fct.regmesh) for fct in reg.objfcts]
+        [self.assertTrue(reg.mapping is fct.mapping) for fct in reg.objfcts]
+
+        D = reg.regmesh.cellDiffx
+        reg.regmesh._cellDiffx = 4*D
```
lheagy: @fourndo: this tests a use-case similar to the one that you showed with updating the cell Diffs

SimPEG/Regularization.py
```diff
@@ -470,58 +459,6 @@ def _nC_residual(self):
         else:
             return self.nP
 
-    @property
-    def mesh(self):
```
rowanc1: This is still accessible from `regmesh.mesh`...

lheagy: yep. But I don't think there is any need to access it from the regularization itself, that is just a potential point on confusion

fourndo:
Good job @lheagy . I merged in my local RotatedObjFun branch in it worked magically.
I vote yes.

PR #626, opened 2017-06-30
delete utils that are already present in discretize

lheagy:
delete util code that has been transferred to discretize so that we are only managing one version!


SimPEG/Utils/coordutils.py
```diff
@@ -1,62 +1 @@
...
-
-    R = rotationMatrixFromNormals(n0, n1)
-
-    assert XYZ.shape[1] == 3, "Grid XYZ should be 3 wide"
-    assert len(x0) == 3, "x0 should have length 3"
-
-    X0 = np.ones([XYZ.shape[0],1])*mkvc(x0)
-
-    return (XYZ - X0).dot(R.T) + X0 # equivalent to (R*(XYZ - X0)).T + X0
+from discretize.utils import rotationMatrixFromNormals, rotatePointsFromNormals
```
rowanc1: Delete all the code!!!
Wonderful. :)

PR #637, opened 2017-09-08
Local problem

fourndo:
New mapping for tiled inversion. 
- [x] Projection matrix for the transformation and derivatives
- [x] Run inversion and check result
- [x] Testing
- [x] Example


SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
...
+
+        return self._P
+
+    @property
+    def Ptransform(self):
+        """
+            Projection for the interpolation of the model (weighted average)
+        """
+        if getattr(self, '_Ptransform', None) is None:
+            sumW = Utils.mkvc(np.sum(self.P, axis=1)+1e-8)
```
lheagy: the 1e-8 is a tolerance to make sure we don't divide by zero? would you mind naming it as a variable for clarity? eg
```
tol = 1e-8
sumW = Utils.mkvc(np.sum(self.P, axis=1)) + tol
```

SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
...
+            self._Ptransform = sp.spdiags(1/sumW,
```
lheagy: could use `Utils.sdiag(1/sumW) * self.P` to simplify this a bit

SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
...
+                                          0, self.meshLocal.nC,
+                                          self.meshLocal.nC) * self.P
+
+        return self._Ptransform
+
+    @property
+    def Psensitivity(self):
```
lheagy: Would it be more clear to name this `Pvol`? `sensitivity` made me think it is a derivative, but it is really that this includes the volume term, and we need that for the sensitivity (right??)

SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
...
+                              shape=(indx.shape[0], self.actvGlobal.sum()))
+            # self.P = Utils.sdiag(self.mesh2.vol[self.actind2])*P.tocsc()
+            self._P = P.tocsr()
+
+            self._shape = indx.shape[0], self.actvGlobal.sum()
+
+        return self._P
+
+    @property
+    def Ptransform(self):
```
lheagy: I am wondering about this name, `transform` isn't overly descriptive (eg we need a bit more clarification as to why this is different than P). Would something along the lines of `Paverage` or `Pave` work?

cc @rowanc1

SimPEG/Maps.py
```diff
@@ -1120,6 +1285,99 @@ def deriv(self, m, v=None):
...
+        assert len(meshes) == 2, "meshes must be a list of two meshes"
+        assert type(actinds) is list, "actinds must be a list of two meshes"
+        assert len(actinds) == 2, "actinds must be a list of two meshes"
+        assert meshes[0].dim == meshes[1].dim, """The two meshes must be the same dimension"""
+
+        self.mesh = meshes[0]
+        self.mesh2 = meshes[1]
+        self.actind = actinds[0]
+        self.actind2 = actinds[1]
+        self._createProjection()
```
lheagy: ideally, we should not create the projection on instantiation, it should only be created when first asked for. This can be achieved through lazy loading
```
@property
def P(self):
    if get_attr(self, '_P', None) is None: 
        # code that creates Projection
        self._P = P
    return self.P
```

SimPEG/Maps.py
```diff
@@ -1120,6 +1285,99 @@ def deriv(self, m, v=None):
...
+    @property
+    def nP(self):
+        """Number of parameters in the model."""
+        # return self.mesh2.nC
+        return self.actind2.sum()
+
+    def _transform(self, m):
+        return self.P*m
+
+    def deriv(self, m):
```
lheagy: In general, this should also be able to take a vector:
```
     def deriv(self, m, v=None):
         if v is not None:
             return self.P * v
         return self.P
```

SimPEG/Directives.py
```diff
@@ -730,60 +733,62 @@ class Update_lin_PreCond(InversionDirective):
...
+            JtJdiag = np.zeros_like(self.invProb.model)
 
-                diagA = np.sum(self.prob.G**2., axis=0) + np.hstack(reg_diag)
+            for prob, dmisfit in zip(self.prob, self.dmisfit.objfcts):
 
-            else:
-                diagA = (np.sum(self.prob.G**2., axis=0) +
-                         self.invProb.beta*(self.reg.W.T*self.reg.W).diagonal())
+                wd = dmisfit.W.diagonal()
+                for ii in range(prob.G.shape[0]):
```
lheagy: This code only works if a problem has a `G`, correct? If that is the case, it is probably worth adding an `assert` statement when the directive is instantiated (in an `__init__`) checking that `hasattr(prob, 'G') is True`

SimPEG/Directives.py
```diff
@@ -49,39 +49,42 @@ def opt(self):
     @property
     def reg(self):
         if getattr(self, '_reg', None) is None:
-            self._reg = self.invProb.reg
+            self.reg = self.invProb.reg  # go through the setter
         return self._reg
 
     @reg.setter
     def reg(self, value):
-        assert any([isinstance(value, regtype) for regtype in self._regPair]),(
```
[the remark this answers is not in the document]
fourndo: Not sure where it went. Back in.

SimPEG/Directives.py
```diff
@@ -49,39 +49,42 @@ def opt(self):
...
     @property
     def dmisfit(self):
-        return self.invProb.dmisfit
+        if getattr(self, '_dmisfit', None) is None:
+            self.dmisfit = self.invProb.dmisfit  # go through the setter
+        return self._dmisfit
 
     @dmisfit.setter
     def dmisfit(self, value):
-        assert any([
```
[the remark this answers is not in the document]
fourndo: Same thing, not sure why the check is gone. Back in.

SimPEG/Directives.py
```diff
@@ -730,60 +733,62 @@ class Update_lin_PreCond(InversionDirective):
...
 
-            # It is a Combo objective, so will have to loop
-            self.ComboObjFun = True
+        for reg in self.reg.objfcts:
+            # Check if he has wire
+            if getattr(reg.mapping, 'P', None) is None:
+                regDiag += (reg.W.T*reg.W).diagonal()
+            else:
+                # He is a snitch!
+                regDiag += reg.mapping.P.T*(reg.W.T*reg.W).diagonal()
```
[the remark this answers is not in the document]
fourndo: Does the same operation assuming that the projection are only 1-0, but your ways is more math explicit.  Made the change in 30fd021b

SimPEG/Directives.py
fourndo: Great. Will also change 'G' for 'getJ', more obvious and will agree with changes made on DC branch.

SimPEG/Directives.py
```diff
@@ -730,60 +733,62 @@ class Update_lin_PreCond(InversionDirective):
...
 
-                diagA = np.sum(self.prob.G**2., axis=0) + np.hstack(reg_diag)
+            for prob, dmisfit in zip(self.prob, self.dmisfit.objfcts):
 
-            else:
-                diagA = (np.sum(self.prob.G**2., axis=0) +
-                         self.invProb.beta*(self.reg.W.T*self.reg.W).diagonal())
+                wd = dmisfit.W.diagonal()
+                for ii in range(prob.G.shape[0]):
+                    JtJdiag += prob.mapping().deriv().T*(wd[ii] * prob.G[ii, :])**2.
```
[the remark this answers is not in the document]
fourndo: Ok good one, I have a question about that. Here the mapping is a projection having the name 'chiMap' because it's the mag problem. But other problems will use other mapping 'sigmaMap', 'muMap' etc. We should just be able to just call 'mapping' from the outside, without having to be specific about a problem. How to do that?

SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
         return self._nP
 
 
+class Tile(IdentityMap):
+    """
+        Mapping for tiled inversion
+    """
+
+    tree = None
+    nCell = 27
```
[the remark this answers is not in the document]
fourndo: Maximum number of neighbors used to interpolate. I though 27 for number of neighbours in a 3x3. Might need to change to a smarter metric later.

SimPEG/Maps.py
```diff
@@ -387,6 +387,171 @@ def nP(self):
         return self._nP
 
 
+class Tile(IdentityMap):
+    """
+        Mapping for tiled inversion
+    """
+
+    tree = None
```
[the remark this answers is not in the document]
fourndo: Changes made on d0990eb8

SimPEG/Maps.py
fourndo: Roger that

SimPEG/Maps.py
fourndo: Ok

SimPEG/Directives.py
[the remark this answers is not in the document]
fourndo: Got it, "prob.mapPair()"

SimPEG/Directives.py
```diff
@@ -233,7 +250,12 @@ def target(self):
         if getattr(self, '_target', None) is None:
             # the factor of 0.5 is because we do phid = 0.5*|| dpred - dobs||^2
             if self.phi_d_star is None:
-                self.phi_d_star = 0.5 * self.survey.nD
+
+                nD = 0
+                for survey in self.survey:
+                    nD += survey.nD
```
lheagy: How we want to apply this is a bit subtle. I am not sure if we want to have a single stopping criteria for all phi_d's. If I am jointly inverting DC (nD = 50) and Richards (nD = 20), then we might want to set the stopping criteria such that we fit both the DC (reaching phi_d* = 25) and Richards (reaching phi_d*= 10) independently rather than reaching a "global" target misfit with total (phi_d* = 35).

jiajiasun: @lheagy You are right! My experiences of doing joint inversion is that, we need to fit each of the observed geophysical data set to about its noise level. Therefore, instead of calculating and monitoring the sum of all Phids, it is important to make sure that each phid actually reaches its target data misfit value. 

This problem become more critical when the number of data from one set of observations is very different from that from the other (say, nD = 200 for DC, nD = 20 for Richards), because the data set with more observations will be likely to dominate the minimization process (because decreasing the data misfit for the data set with more observations will lead to a quick decrease in the total objective function). 

Another way of looking at this problem is, consider the gradient, because we rely on gradient (or some sort of variant of gradient, like conjugate gradient) to minimize the objective function. It is very likely that the gradient will be dominated by the data set with more observations, as well.

You can do a quick test by setting nD for DC equal a very large number and nD for Richards equa a very small number, and letting the inversion only return a desired 'global' target.  A good chance you will see what I described above.

This is an open question for joint inversion of, say, seismic and gravity, as well.

I agree with you that, setting a target for a 'global' misfit is not a good idea.

lheagy: however, for the global / local problems, we will want a single `phi_d_star`. Maybe we use an option
`joint_inversion = True`, which does
```
self.phi_d_star = [0.5 * self.chifact * survey.nD for survey in self.survey]  
```

fourndo: You guys are the best @jiajiasun + @lheagy . I ll have another go at it.

@jiajiasun : Were your problems updating the same physical property (i.e. single vector of density) or one for each? If a single vector of density, do you think the number of data matter - or would the sensitivities (gradient) matter more?

jiajiasun: @fourndo My experiences are mostly with joint inversion of, say, seismic and gravity that deals with different physical properties. But what I described previously also applies to the problem of inverting for, say, a single vector of resistivity, from two different electrical data sets. I guess you guys are more interested in the latter?

Let us consider the total objective function phi(m) = w1 * phi_d1 + w2 * phi_d2 + beta * phi_m, where w1, w2 and beta are scalars, and phi_d1, phi_d2 and phi_m are data misfit and regularization terms. Imagine you set w1 = 1, and w2 = 0.00001, then the inversion would not bother trying to minimize phi_d2 because that term is already very very small due to the very very small value assigned to w2 when evaluating the total objective function. 

Similarly, suppose w1 = w2 = 1.0, but the number of data for the first set of data is 10000, and that for the second set of data is 10, then similar thing would happen; the inversion would focus much more on minimizing the first data misfit term, thereby allowing the first set of data to contribute more to the final inverted model.

Setting a 'global' target data misfit for the above two cases would most likely result in the second set of data contributing much less to the inversion.  Also, even if you achieve your 'global' target, you might end up with one set of data being overfit and the other being under-fit. Therefore, I would recommend monitoring both phid_d1 and phi_d2  and make sure they both achieve their respective target data misfit value.

So, when you have two sets of data with different amount of data, you might want to leverage w1 and w2 to make sure that both sets of data contribute approximately equal to the inversion process, assuming of course that you trust the two sets of data equals and you want them to both contribute to the final model.

We can also try to understand this problem by looking at gradient.  Suppose our total objective function is phi(m) = w1* ||d_1^obs - G1*m||_2^2 + w2* ||d_2^obs - G2*m||_2^2 + w3 * ||W_m*m||_2^2, then the gradient (i.e., our search direction) would be grad = w1*G_1^T*delta_d1 + w2*G_2^T*delta_d2  + beta*W_m^T*W_m*m. It is obvious that the first term is the contribution from the first set of data and so on. If w1 = w2 = 1.0, and suppose number of first set of data is much much larger than the number of second set of data, then it is very likely that G_1^T*delta_d1 will be much larger than G_2*T*delta_d2, because delta_d is data residual and G^T*delta_d is the projection of data residual onto data space spanned by the rows of the G matrix. (This is my guess. I haven't verified it numerically).

In this case, we might consider adjusting w1 and w2 to make sure that both sets of data contribute to the gradient on more or less the same level.

Again, this is still an open question for joint inversion (of multiple sets of data sensitive to different physical properties) because we often have very different number of seismic data than gravity data. I have not seen a  satisfactory answer and even an attempt to investigate this issue in literature. This is a challenging problem!  (What I wrote down above reflects my current thinking, which may change in a few month :)

fourndo: This is great Jiajia, 
I will set it up so everything becomes a list (phi_d*, beta's), just like we did for the misfit and regularization functions. That way we can monitor each individually. We interested in both versions (multiple misfits + multiple phys props  || multiple misfits + one phys props), which we should have the mechanics to tackle both very similarly, maybe just different flavors of directives.

Agreed, still a very open-question I think. Hopefully we can share experiments faster/easier once we have this guy set up properly.

jcapriot: Would it just be a matter of dividing each data misfit term by the number of data? Or at least could that be a good starting point? Essentially it would mean each data misfit term contributes equally to the final objective model at the point when both of them are fit to their noise levels. However, I only have intuitions about how they contribute outside of that point.

fourndo: Hey @jcapriot ,
I like the way to think. I had try that before, and it does work in theory. In practice though, phi_d is also used in the optimization part also, and it becomes less transparent when we start messing up with the measure itself. Other option is to have multipliers on the misfit function (like a beta, but on phid). So that way we could change the weight of each regularizations AND misfit functions independantly. 
something like 
```
phi = beta1 * phid_1 + beta2 * phid_2 + beta3*reg1 + beta4*reg2 + ...
```
Default would be all 1
Thoughts?

jiajiasun: It is an option as well! Actually, some folks from UCSD used this trick in their joint inversion of, say, CSEM and MT, for resistivity. I forgot the names of the authors. Will find out later. 

My personal view on this option is that, if the noise from each set of data can be estimated well, then dividing by the number of data is a good idea. Otherwise, it might not be. But, I have not played with this strategy in my work yet. Need to run some inversions with all these different strategies and find out which one works best.

jcapriot: @fourndo It might almost simplify the ``phi_d`` measure in the optimization, meaning they would only ever have to hit ``1.0`` instead of ``nD``.

@jiajiasun I feel like part of doing joint inversion should be estimating the noise ahead of time. I mean at some point there just gets to be too many unknowns at once, and it's not exactly feasible to black-box it.

lheagy: I would hesitate to divide by the number of data by default - I would suggest we leave it up to the user to define an appropriate multiplier (if we do something under the hood, it has the potential to be quite confusing). 

I could imagine scenarios where one data set has a very small number of data, another has a large number and you may want to effectively try treating all data equally (rather than treating each survey equally, as dividing by the number of data would do). I am sure each approach has its merits, but is should be the user who makes that decision and I think that keeping the definition of the data misfit quite simple should make things more transparent. 

my 2 cents :)

SimPEG/Utils/modelutils.py
```diff
@@ -29,32 +34,42 @@ def surface2ind_topo(mesh, topo, gridLoc='CC', method='nearest', fill_value=np.n
 
         if gridLoc == 'CC':
             XY = ndgrid(mesh.vectorCCx, mesh.vectorCCy)
-            Zcc = mesh.gridCC[:, 2].reshape((np.prod(mesh.vnC[:2]), mesh.nCz), order='F')
-            #gridTopo = Ftopo(XY)
-            gridTopo = griddata(topo[:, :2], topo[:, 2], XY, method=method, fill_value=fill_value)
-            actind = [gridTopo >= Zcc[:, ixy] for ixy in range(np.prod(mesh.vnC[2]))]
+            Zcc = mesh.gridCC[:, 2].reshape((np.prod(mesh.vnC[:2]), mesh.nCz),
+                                            order='F')
+
```
lheagy: This is me being a bit picky, but I find it much easier to read across multiple lines if we break at brackets, eg
```
Zcc = mesh.gridCC[:, 2].reshape(
    (np.prod(mesh.vnC[:2]), mesh.nCz), order='F'
)
gridTopo = griddata(
    topo[:, :2], topo[:, 2], XY, method=method, fill_value=fill_value
)
actind = [
    gridTopo >= Zcc[:, ixy] for ixy in range(np.prod(mesh.vnC[2]))
]
``` 

No worries if you don't want to update this on this pr, just a thought for the future. cc @rowanc1: might be worth putting a few notes on style in the docs. Thoughts on that? Do you have any handy references?

SimPEG/Directives.py
lheagy: Extending this to the joint inversion case will be a bit more involved. @fourndo: would you mind creating an issue and tagging it here? we can address this in a separate pr. For now, I think the current implementation is ok

SimPEG/Directives.py
```diff
@@ -708,82 +735,88 @@ def validate(self, directiveList):
...
                 "Without a Linear preconditioner, convergence may be slow. "
-                "Consider adding `Directives.Update_lin_PreCond` to your "
+                "Consider adding `Directives.UpdateJacobiPrecond` to your "
                 "directives list"
             )
         return True
 
 
-class Update_lin_PreCond(InversionDirective):
+class UpdateJacobiPrecond(InversionDirective):
```
lheagy: I think we could just spell this out `UpdateJacobiPreconditioner`

SimPEG/Maps.py
```diff
@@ -387,6 +387,300 @@ def nP(self):
         return self._nP
 
 
+class Tile(IdentityMap):
+    """
+        Mapping for tiled inversion
+    """
+
+    nCell = 27  # Number of neighbors to use in averaging
```
lheagy: just out of curiosity... where did this come from? (27 just seems oddly specific) :)

fourndo: Short answer : 3**3 = 27
It's a default to cover the case where a cell in the local problem is slightly offseted from the global mesh, so the maximum number of neighbors to worry about is a 3x3x3 block

SimPEG/Directives.py
fourndo: On master it is now ```UpdatePreconditioner``` , having in mind that the function could eventually take different PC styles. Are you good with this?

lheagy:
feel free to ping when the merge conflicts have been addressed and I will do a more thorough review - this is an exciting piece to get plugged in!

SimPEG/InvProblem.py
```diff
@@ -91,28 +94,28 @@ def startup(self, m0):
...
-                if isinstance(objfct, DataMisfit.BaseDataMisfit):
-                    print("""
-    SimPEG.InvProblem is setting bfgsH0 to the inverse of the eval2Deriv.
-    ***Done using same Solver and solverOpts as the {} problem***""".format(
-                            objfct.prob.__class__.__name__
+        if self.bfgs:
+            if isinstance(self.dmisfit, DataMisfit.BaseDataMisfit):
+                print("""SimPEG.InvProblem is setting bfgsH0 to the inverse of the eval2Deriv.
+                      ***Done using same Solver and solverOpts as the problem***"""
+                      )
```
lheagy: sorry - being picky. Would you mind updating bracketing levels here and in the following few lines? So 
```
print("""SimPEG.InvProblem is setting bfgsH0 to the inverse of the eval2Deriv.
          ***Done using same Solver and solverOpts as the problem***"""
)
```

SimPEG/Maps.py
```diff
@@ -1120,6 +1414,99 @@ def deriv(self, m, v=None):
         return self.P
 
 
+# class Mesh2MeshTopo(IdentityMap):
```
lheagy: can this be removed? or was it commented out for another reason??

SimPEG/PF/Magnetics.py
```diff
@@ -28,49 +28,29 @@ class MagneticIntegral(Problem.LinearProblem):
...
+        """
+            Magnetic field data
+        """
+        self.model = m
 
         if self.forwardOnly:
 
             # Compute the linear operation without forming the full dense G
-            fwr_d = self.Intrgl_Fwr_Op(m=m)
+            fwr_d = self.Intrgl_Fwr_Op(m=self.chiMap*m)
```
lheagy: here, you can call 
```
self.chi
```
and that will deal with the mapping (and will work if the user provides `chi` as an input to the forward problem in the case that they are not running an inversion)

SimPEG/PF/Magnetics.py
```diff
@@ -79,18 +59,11 @@ def G(self):
...
-    #     """
-    #     dmudm = self.chiMap.deriv(self.chi)
-    #     return self.G*dmudm
-
     def getJ(self, m, f=None):
         """
             Sensitivity matrix
         """
-        dmudm = self.chiMap.deriv(self.chi)
+        dmudm = self.chiMap.deriv(self.model)
```
lheagy: here, you should be able to do `self.chiDeriv`

SimPEG/PF/Magnetics.py
```diff
@@ -166,14 +132,16 @@ def Intrgl_Fwr_Op(self, m=None, Magnetization="ind"):
 
         if survey.srcField.rxList[0].rxType == 'tmi':
 
-
             # Convert Bdecination from north to cartesian
-            D = (450.-float(survey.srcField.param[2])) % 360.
-            I = survey.srcField.param[1]
+            dec = (450.-float(survey.srcField.param[2])) % 360.
```
lheagy: later, we should pull this into a simple util

SimPEG/PF/Magnetics.py
```diff
@@ -256,6 +224,13 @@ def Intrgl_Fwr_Op(self, m=None, Magnetization="ind"):
 
         return fwr_out
 
+    @property
+    def modelMap(self):
```
lheagy: @fourndo: this is necessary for directives? if so, we can try to address this by providing an `inversion_models` list in the new simulation class so that the inversion is aware of what it is inverting for. (no need to change anything at the moment, just looking forward)

lheagy: ah, I see: also in the Problem. kk, I think we can clean this up a bit in the future

examples/05-mag/plot_inversion_tiled.py
```diff
@@ -0,0 +1,329 @@
+"""
+PF: Tiled magnetic Inversion
+============================
```
lheagy: very cool, thanks for adding this!

examples/05-mag/plot_inversion_tiled.py
```diff
@@ -0,0 +1,329 @@
...
+    # Lets create a simple Gaussian topo and set the active cells
+    [xx, yy] = np.meshgrid(mesh.vectorNx, mesh.vectorNy)
+    zz = -np.exp((xx**2 + yy**2) / 75**2) + mesh.vectorNz[-1]
+
+    # Create an array for topography
+    topo = np.c_[Utils.mkvc(xx), Utils.mkvc(yy), Utils.mkvc(zz)]
+
+    # Go from topo to actv cells
+    actv = Utils.surface2ind_topo(mesh, topo, 'N')
+    actv = np.asarray([inds for inds, elem in enumerate(actv, 1) if elem],
```
lheagy: is this grabbing the active inds as an array? 
`actv.indActive`

lheagy:
Once more request, would you mind updating the pr description so we can use it as release notes? Thanks!

PR #640, opened 2017-09-21
Update Regularization.py

sgkang:
remove Zero class for alpha_xx, alpha_yy, alpha_zz
this makes a problem 
Objectivefunction.py Line 387 when sqrt is used.
`curW = np.sqrt(mult) * fct.W`

### DIFF
--- a/SimPEG/Regularization.py
+++ b/SimPEG/Regularization.py
@@ -1145,7 +1145,7 @@ class Tikhonov(BaseComboRegularization):
     def __init__(
         self, mesh,
         alpha_s=1e-6, alpha_x=1.0, alpha_y=1.0, alpha_z=1.0,
-        alpha_xx=1., alpha_yy=1., alpha_zz=1.,
+        alpha_xx=0., alpha_yy=0., alpha_zz=0.,
         **kwargs
     ):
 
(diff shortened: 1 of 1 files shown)


SimPEG/Regularization.py
```diff
@@ -1145,7 +1145,7 @@ class Tikhonov(BaseComboRegularization):
     def __init__(
         self, mesh,
         alpha_s=1e-6, alpha_x=1.0, alpha_y=1.0, alpha_z=1.0,
-        alpha_xx=Utils.Zero(), alpha_yy=Utils.Zero(), alpha_zz=Utils.Zero(),
+        alpha_xx=1., alpha_yy=1., alpha_zz=1.,
```
lheagy: @sgkang: I think the default for these should be 0? They could be `0.` instead of `Utils.Zero()`, but I don't think we want to be updating this default.

cc @rowanc1

lheagy:
Another thought - we might want to add tests on defaults so we keep track of the assumptions we make on default values in SimPEG cc @rowanc1

SimPEG/Regularization.py
sgkang: Oh, sorry this should be zero.

SimPEG/Regularization.py
```diff
@@ -1145,7 +1145,7 @@ class Tikhonov(BaseComboRegularization):
     def __init__(
         self, mesh,
         alpha_s=1e-6, alpha_x=1.0, alpha_y=1.0, alpha_z=1.0,
-        alpha_xx=Utils.Zero(), alpha_yy=Utils.Zero(), alpha_zz=Utils.Zero(),
+        alpha_xx=0., alpha_yy=0., alpha_zz=0.,
```
lheagy: Thanks @sgkang, this looks good!

PR #641, opened 2017-09-21
Em/dev into nsem/dev

grosenkj:
Pulling new changes into nsem/dev


grosenkj:
Latest and greatest into nsem/dev

PR #645, opened 2017-10-21
Devin VRM

dccowan:
Here, we add functionality for using potential fields to model viscous remanent magnetic responses. A new section of simpeg was made for this because:
* It is a time-dependent problem
* It is formulated using an integral equation
* It is linear, however, explicit storage of the linear operator is prohibitively large


SimPEG/VRM/ProblemVRM.py
```diff
@@ -0,0 +1,609 @@
...
+    refRadius: Distances from source in which cell sensitivities are refined.
+               Must be an array or list with elements equal to the refFact.
+               (default based on minimum cell size)
+    topoMap: A Maps object which operates from mesh cells to cells which are
+             computed in the forward model (default is all mesh cells)
+
+    """
+
+    # SET CLASS ATTRIBUTES
+    _refFact = None
```
lheagy: These can be set using the properties library fairly cleanly. 

```python
refFact = properties.Integer(
    "refinement factor",
    default = 3,
    min = 0
)
```

This will allow you to remove lines 43, 49, 64. To then do the additional checks in 66 - 70, you can use

```python
@properties.validator('refFact')
def check_vs_radius(self, change):
    val = change['value']
    assert val == len(self.refRadius), "Refinement factor no longer matches length of refinement radii array. Please ensure refinement factor is equal or less to number of elements in refinement radii"
    if val > 4:
         print("Refinement factor larger than 4 may result in computations which exceed memory limits")
```

cc @rowanc1

SimPEG/VRM/ProblemVRM.py
```diff
@@ -0,0 +1,609 @@
...
+
+            rxObj = srcObj.rxList[qq]
+            dComp = rxObj.fieldComp
+            locs = rxObj.locs
+            M = np.shape(locs)[0]
+
+            if dComp is 'x':
+                for rr in range(0, M):
+                    u1 = locs[rr, 0] - ax
+                    u1[np.abs(u1) < 1e-10] = np.min(xyzh[:, 0])/1000
```
lheagy: - instead of hard-coding 1e-10, could you use the variable `eps`?
- what is the factor of 1000 for? is it a unit conversion?

lheagy:
another quick thought: in terms of organization, I suggest that we put `VRM` inside of `SimPEG.EM`, so it would be
```
from SimPEG.EM import VRM
```

SimPEG/VRM/ProblemVRM.py
```diff
@@ -0,0 +1,609 @@
...
+    _refRadius = None
+    _topoMap = None
+    _AisSet = False
+    _TisSet = False
+    surveyPair = SurveyVRM
+
+    def __init__(self, mesh, **kwargs):
+
+        # **kwargs
+        self._refFact = kwargs.get('refFact', 3)
```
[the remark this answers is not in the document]
dccowan: refFact refers to refinement factor. The idea is that many cells near the transmitter do not experience uniform magnetization due to the variation in the shape of the primary field. Instead of refining the mesh, we let the physical property of that cell remain constant but we recompute the sensitivity of that cell to the data.

SimPEG/VRM/ProblemVRM.py
```diff
@@ -0,0 +1,609 @@
...
+            print("Refinement factor larger than 4 may result in computations which exceed memory limits")
+
+        super(BaseProblemVRM, self).__init__(mesh, **kwargs)
+
+    @property
+    def refFact(self):
+        return self._refFact
+
+    @refFact.setter
+    def refFact(self, Val):
```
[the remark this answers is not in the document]
dccowan: See comment above.

SimPEG/VRM/ProblemVRM.py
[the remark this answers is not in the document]
dccowan: An explanation for this starts on pg 48 of my thesis

SimPEG/VRM/ProblemVRM.py
dccowan: Values here were determine experimentally and provide stability to the forward model. If this is not done, then there may be locations relative to the cells in which we wind up dividing by zero.

SimPEG/VRM/ProblemVRM.py
dccowan: I like things how I have them right now. But if I can get a detailed explanation as to why this is better, and it does the exact same thing, we could end up changing it.

SimPEG/VRM/ProblemVRM.py
dccowan: This was done

dccowan:
I've looked at the remaining errors and they do not seem to be related to the contribution I am trying to merge. @lheagy  would you mind taking a look?

0.3.6b0
```diff
@@ -0,0 +1,2 @@
+Requirement already satisfied: properties in c:\users\sarah\anaconda3\lib\site-packages
```
lheagy: This file should not have been added, would you mind removing it?

SimPEG/Regularization.py
```diff
@@ -745,16 +746,16 @@ def _update_regmesh_indActive(self, change):
...
-    def _validate_cell_weights(self, change):
-        if change['value'] is not None:
-            # todo: residual size? we need to know the expected end shape
-            if self._nC_residual != '*':
-                assert len(change['value']) == self._nC_residual, (
-                    'cell_weights must be length {} not {}'.format(
-                        self._nC_residual, len(change['value'])
-                    )
-                )
+    # @properties.validator('cell_weights')
```
lheagy: why was this commented out?

SimPEG/VRM/ProblemVRM.py
```diff
@@ -0,0 +1,686 @@
...
+import numpy as np
+import scipy.sparse as sp
+
+############################################
+# BASE VRM PROBLEM CLASS
+############################################
+
+
+class BaseProblemVRM(Problem.BaseProblem):
+    """
```
lheagy: can you please indent this over so that the indentation level of the text matches the quotes and use the sphinx formatting for defining your variables (see for example https://github.com/simpeg/simpeg/blob/master/SimPEG/EM/FDEM/ProblemFDEM.py#L63) , eg
```
    """
    Base class for VRM problem

    **Required Arguments**:
    :param discretize.BaseMesh mesh: a 3D tensor or OcTree mesh

    **Optional Inputs**:
    :param int refFact: maximum refinement factor for sensitivities 
    """
```

lheagy: `refFact` --> `ref_fact`: https://www.python.org/dev/peps/pep-0008/

0.3.6b0
dccowan: Line will be removed

dccowan: correction. File will be removed.

lheagy:
There was another file: python_simlinks - Shortcut.lnk that should also be removed.

SimPEG/Regularization.py
dccowan: No clue. I will uncomment and rerun tests

PR #655, opened 2017-12-04
ref/examples

lheagy:
- update link to paper in the em examples to point to computers and geoscience instead of arxiv
- update pypi credentials so simpeg will auto-deploy
- update dependence on properties and vectormath to be in-sync with discretize (we require `vectormath` and `properties>=0.3.6b0`


lheagy:
@rowanc1, I am planning on merging without bumping version here (only updates in the docs of the examples). Thoughts or concerns with that?

PR #666, opened 2018-01-19
Add SumMap and HomogeneousMap

fourndo:
Two new maps added to the framework.

**HomogeneousMap**: Allows to invert for groups of model cells under the assumption to have homogeneous physical properties. Useful to invert for best-fitting value of geological model. 

**SumMap**: Allow to decompose the physical property model into components, such that the forward model F(m) = F(g(x) + h(y)), 
where g(x) and h(y) are different parameter spaces. For example:
g(x): HomogeneousMap * WireMap
h(y): WireMap
Each model space can be regularized on a different mesh and with different assumptions.

- [ ] Test maps
- [ ] Write example

### DIFF
--- a/SimPEG/Maps.py
+++ b/SimPEG/Maps.py
@@ -254,10 +254,15 @@ class ComboMap(IdentityMap):
                     )
                 )
 
-            if isinstance(m, ComboMap):
-                self.maps += m.maps
-            elif isinstance(m, IdentityMap):
+            if np.any([isinstance(m, SumMap), isinstance(m, IdentityMap)]):
                 self.maps += [m]
+            elif isinstance(m, ComboMap):
+                self.maps += m.maps
+            else:
+                raise ValueError(
+                    'Map[{0!s}] not supported',
+                    m.__class__.__name__
+                )
 
     @property
     def shape(self):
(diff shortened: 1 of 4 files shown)


SimPEG/Maps.py
```diff
@@ -348,6 +353,155 @@ def deriv(self, m, v=None):
...
+
+
+class HomogeneousMap(IdentityMap):
+    """
+        A map to group model cells into an homogeneous unit
+
+        :param list index: list of bool for each homogeneous unit
+
+    """
+    nBlock = 1  # Variable allowing to stack same Map over multiple sets
```
rowanc1: Can we move to lowercase variable names for parameters? `n_blocks`? (or just `blocks`?)

SimPEG/PF/Magnetics.py
```diff
@@ -39,11 +39,11 @@ def fwr_ind(self, m):
 
         else:
 
-            return self.G.dot(m)
+            return self.G.dot(self.chiMap * m)
```
lheagy: here, you can replace `self.chiMap * m` with `self.chi` (does the mapping step for you)

lheagy: same elsewhere

SimPEG/PF/Magnetics.py
```diff
@@ -79,6 +79,11 @@ def G(self):
 
         return self._G
 
+    @property
+    def modelMap(self):
```
lheagy: I think we might want to make this private (eg `_model_map`) as we want to promote users working with the physical properties, but internally, it can be handy for inheriting from the linear problem

lheagy:
@fourndo, what do you think of `SurjectUnits` for the class name instead of `HomogeneousMap`? It would be a bit more consistent with the current names-space (similar to `SurjectFull`, `Surject1D` ...

lheagy:
@fourndo: do you mind if we close this in favour of #724?

fourndo:
You got it

Issue #562, opened 2017-03-11
the problem with "Problem"


lheagy:
I had a chat with @dougoldenburg today and he brought to light some of the confusion introduced by the term `Problem` in the framework, in particular when first introducing the SimPEG framework to new community members, which is perhaps indication of a more systemic issue. 
![image](https://cloud.githubusercontent.com/assets/6361812/23820542/accfe65c-05cf-11e7-8756-d84b39f5f2b4.png)

## Points of confusion
- often a first association with the word "Problem" is the thing you are trying to solve: the application, the reason you are considering inversion in the first place. 
- "Problem" is over-used already. "Forward Problem", "Inverse Problem"
- even in the paper, we use "problem" to mean two different things \cite{SimPEGpaper}

![image](https://cloud.githubusercontent.com/assets/6361812/23820586/8d41ce30-05d0-11e7-9f40-a0da0341be2e.png)

## Potential Alternatives
- `Physics`
- `GoverningEquations`
- `PartialDifferentialEquations`
- `PDE`
- `Equations`

In looking at the current `Problem.py` and considering the construction of a forward simulation, I am (first pass!) partial to `Equations`.
- `BaseProblem` --> `BaseEquations`
- `BaseTimeProblem` --> `BaseTimeEquations`
- perhaps `Problem.py` --> `GoverningEquations.py`

Then in a module, if we consider EM
- `BaseEMequations`
- `BaseFDEMequations`
- `Maxwell3D_e(BaseFDEMequations)` 

```
from SimPEG.EM import FDEM
equations = FDEM.Maxwell3D_e(mesh)
survey = FDEM.Survey(srcList)
equations.pair(survey)  # or fwdSim = simulation(equations, survey) down the road
```

I look forward to hearing thoughts!

lheagy:
Flipping through some presentations as well
![image](https://cloud.githubusercontent.com/assets/6361812/23837758/897ac16e-074a-11e7-8b6a-31242234efdb.png)

fourndo:
I actually like Physics better. Makes it pretty obvious of what is supposed to be in there.
Don't like PDE, too specific.

micmitch:
My preference would be for either Physics, but Equations would also be fine.

lheagy:
Thanks @fourndo, @micmitch. That is fair - physics is a good descriptor. I am wondering if the structure is something along the lines of `Physics` inherits `BaseEquations` (or `BaseTimeEquations`). For example 
```
[...]

PR #672, opened 2018-02-12
Ref/simulation

lheagy:
This is very far from ready to come in, but some input at this point would be valuable. This pr contains a first pass at implementing a `Simulation` class as per #562, and makes more use of the `Data` class as per #202. 

I have attached a notebook that runs the linear problem with exponentially decaying sinusoids using the new Simulation Class ([Simulation.ipynb.zip](https://github.com/simpeg/simpeg/files/1718070/Simulation.ipynb.zip)). 

# Survey
The new survey class is lighter than the previous survey class and only contains details about the sources and receivers. The survey contains 
- `srcList` 
- `counter`
and has the properties / methods
- `nD`, `vnD`
- `nSrc` 
- `getSourceIndex(self, sources)`

Items that have been moved to the data class or the simulation class (or removed)
- `dpred` is now on the simulation
- `make_synthetic_data`   is now on the simulation
- `dobs` is now on the data class
- uncertainties are now on the data class
- `eval` and `evalDeriv` no longer exist 
- and there is no longer a problem to pair to

Some motivations: we may want to have some flexibility down the road with how a survey is constructed (eg. if we are doing a ground survey and all receivers are listening to all sources, then maybe we don't want to provide a receiver list to all the sources, or there might be some utilities around this that make it more transparent for the user). 

# Simulation
The simulation contains everything needed to run a forward simulation. It is instantiated with 
- `mesh`
- `survey`
- `solver`, `solver_opts`, `counter`

It has methods
- `fields(m)`
- `dpred(m, f=None)`
- `Jvec(m, v, f=None)`, `Jtvec(m, v, f=None)`, `Jvec_approx(m, v, f=None)`, `Jtvec_approx(m, v, f=None)`, 
- `residual(m, dobs, f=None)`
- `make_synthetic_data(m, standard_deviation=0.05, f=None)` This now returns a data object

# Fields
Fields objects are instantiated with a simulation (they no longer need a mesh and a survey)
- Fields can access the mesh and survey by `fields.mesh` and `fields.survey` (which just gets the `mesh` and `survey` from the `simulation`)
- `knownFields` and `aliasFields` should be set in the init of a fields object, eg
```python
class Fields3D_e(BaseFDEMFields):

    def __init__(self, **kwargs):
        super(Fields3D_e, self).__init__(**kwargs)
        self.knownFields = {'eSolution': 'E'}
        self.aliasFields = {
            'e': ['eSolution', 'E', '_e'],
            'ePrimary': ['eSolution', 'E', '_ePrimary'],
            'eSecondary': ['eSolution', 'E', '_eSecondary'],
            'b': ['eSolution', 'F', '_b'],
            'bPrimary': ['eSolution', 'F', '_bPrimary'],
            'bSecondary': ['eSolution', 'F', '_bSecondary'],
            'j': ['eSolution', 'CCV', '_j'],  # TODO: move to F
            'h': ['eSolution', 'CCV', '_h'],  # TODO: move to E
        }
```

# Data
Data is a container for observed data. It allows us to assign uncertainties and to index the data by source and receiver. It has properties 
- `dobs`, required. Observed data
- `standard_deviation`, `noise_floor`
- `uncertainty`: this can be set directly or is defined based on the standard deviation and noise floor

The data vector is stored, and unlike before where the data vector was re-arranged so it could easily be indexed by source / receiver, we keep the data as a vector and store the indices associated with each source receiver. Looking forward, there can be further utilities added for slicing the data in sane ways.


SimPEG/EM/FDEM/RxFDEM.py
```diff
@@ -1,37 +1,59 @@
...
-        SimPEG.Survey.BaseRx.__init__(self, locs, rxType=None)
+    # TODO: eventually, this should be a Vector3 and we can allow arbitraty
+    # orientations
+    orientation = properties.StringChoice(
+        "orientation of the receiver 'x', 'y', or 'z'",
+        choices=['x', 'y', 'z'],
+        required=True
+    )
+
+    component = properties.StringChoice(
```
lheagy: @sgkang, @dougoldenburg, @rowanc1, @grosenkj: I would be curious to hear your thoughts on the name for this key-word argument. `component` is perhaps easily confused with spatial component (`orientation`), but I cannot think of another more explicit name. 
```
rx_bz_real = FDEM.Rx.Point_b(orientation='z', component='real') 
rx_bz_imag = FDEM.Rx.Point_b(orientation='z', component='imag') 
```

lheagy:
rather than resolving these conflicts, I would prefer to take these ideas and start fresh from a branch that is up-to-date with master. @grosenkj: does that make a mess with the NSEM stuff?

grosenkj:
@lheagy I am not sure, but I think in both cases the `nsem/ref/simulation` branch is a bit of a mess. So if it is easier for you to start fresh, I support that. That will lead to much faster integration of this work into `master`.

Issue #656, opened 2017-12-11
docs aren't deploying 


lheagy:
<img width="730" alt="simpeg - travis ci 2017-12-11 14-03-45" src="https://user-images.githubusercontent.com/6361812/33856186-2a440606-de7c-11e7-83f4-2cfe734126b1.png">

lheagy:
closed with #674

PR #670, opened 2018-02-08
Improvements: Potential Fields

thast:
A couple of improvements or fix for Potential Fields:

- Fix reading functions for UBC file format (in case of line break)
- add `ModelMap` property
- Change `survey.nD` definition so it is independent of `Problem`
- Use `reg.deriv2` instead of `reg.W` for computing the preconditioner in `Directives.updatePreconditioner`
- general PEP8 improvements (powered by AutoFormat)
- naming convention: `readUBCgravObs` becomes  `readGravityObservations` (for consistency between `Grav` and `Mag` names conventions.)
- Combine all Mag problems (susc, amp, MVI) under ```Magnetics.MagneticIntegral```
- Change coordinate system from legacy x, y, z(down) to convention x, y, z (up)
- Add test+example for MVI-S on TreeMesh
- Add test+example for Amplitude inversion (CSM algo)

PR #674, opened 2018-02-21
Ref/deploy

lheagy:
- clean up deploy sequence so that the most recent docs are deployed to: https://docs.simpeg.xyz
- add `properties.Boolean` to pyclass ignore in the docs (this is temporary and can be removed when aranzgeo/properties#223 is merged) --> this should fix the travis fails in #670, #645


lheagy:
closes #656

PR #678, opened 2018-02-27
Nsem/ref/simulation into ref/simulations

grosenkj:
Fixing up current `NSEM` module to inherit from the "Old" `SimPEG` base classes and `FDEM` classes. A precursor to getting `NSEM` module to relay on the `Simulation` classes being worked out in #672.


lheagy:
Thanks @grosenkj!

PR #682, opened 2018-03-10
Parametric maps

lheagy:
- add a parametric block map (2D and 3D) (eg. as shown in https://github.com/simpeg-research/heagy_2018_AEM/blob/master/Heagy_etal_2018_aem_workshop.pdf) 

image

- rename Parametrized maps to Parametric maps


lheagy:
@sgkang: would you be willing to take a look at this?

lheagy:
@sgkang: would you be good if I merge this?

SimPEG/Maps.py
```diff
@@ -1871,9 +1893,306 @@ def deriv(self, m):
             ]).T)
 
 
-class ParametrizedCasingAndLayer(ParametrizedLayer):
+class ParametricBlock(BaseParametric):
+    """
+        Parametric Block in a Layered Space
```
sgkang: is this in layered earth? 
I thought you are assuming homogenous earth (or background is layered-earth here?)

lheagy: ah, thanks that is an old doc-string that I copied from the Parametric layer map

PR #685, opened 2018-04-04
Bug/nC_residual

lheagy:
A bug was introduced into the way that we set the `nC_residual` in the Regularization. This pr fixes that, cleans up the logic of how we set nC_residual and adds a test to ensure it doesn't break in the future.


thast:
@lheagy : Just to undersand the issue, was the problem when `getattr` returned `None`? (so we would get `None` as `_nC_residual`)

tests/base/test_regularization.py
```diff
@@ -333,5 +333,29 @@ def test_linked_properties(self):
             for fct in reg.objfcts
         ]
 
+    def test_nC_residual(self):
+        cs, ncx, ncz, npad = 1., 10., 10., 20
+        hx = [(cs, ncx), (cs, npad, 1.3)]
+        npad = 12
```
thast: There is two definitions of  parameters `npad`

tests/base/test_regularization.py
```diff
@@ -333,5 +333,29 @@ def test_linked_properties(self):
...
+        temp = np.logspace(np.log10(1.), np.log10(12.), 19)
+        temp_pad = temp[-1] * 1.3 ** np.arange(npad)
+        hz = np.r_[temp_pad[::-1], temp[::-1], temp, temp_pad]
+        mesh = Mesh.CylMesh([hx, 1, hz], '00C')
+        active = mesh.vectorCCz < 0.
+
+        # Step2: Set a SurjectVertical1D mapping
```
thast: Are these comment lines still relevant here?

PR #687, opened 2018-04-10
Static fields fix

micmitch:
Fixed the way E fields were being computed in the fields object for the DC_CC problem. Added tests the check E and J for a electric dipole in a fullspace against the analytic solutions. Still lacking a test for charges.


SimPEG/EM/Static/DC/FieldsDC.py
```diff
@@ -116,9 +116,9 @@ def _e(self, phiSolution, srcList):
         """
             In HJ formulation e is not well-defined!!
             .. math::
-                \vec{e} = -\nabla \phi
+                \vec{e} = \rho \vec{j}
         """
-        return self.prob.MfI*self.prob.MfRhoI * self._j(phiSolution, srcList)
+        return self.prob.MfI*self.prob.MfRho * self._j(phiSolution, srcList)
```
lheagy: Thanks @micmitch! @sgkang: I am not sure if you updated this already on #590, but we can bring this in, along with the additional test and handle the merge after

tests/em/static/test_DC_FieldsDipoleFullspace.py
```diff
@@ -0,0 +1,308 @@
...
+        Bloc = np.r_[-1.0, 0., 0.]
+        M = Utils.ndgrid(x-25., y, np.r_[0.])
+        N = Utils.ndgrid(x+25., y, np.r_[0.])
+
+        rx = DC.Rx.Dipole(M, N)
+        src = DC.Src.Dipole([rx], Aloc, Bloc)
+        survey = DC.Survey([src])
+
+        # Create Dipole Obj for Analytic Solution
+        edipole = fdem.ElectricDipoleWholeSpace(
```
lheagy: might need to add geoana to `requirements_dev.txt`. Nice to see this code in use :)

PR #684, opened 2018-04-03
VRM docstrings

dccowan:
# Documentation and Examples for VRM code

Here, we address remaining work outlined in PR #645 ; i.e. documentation and useful examples.

## Details

* from PR #684 
* commits from @dccowan
* reviewed by @thast and @lheagy 

## Highlights

* [x] forward modeling and inversion example with figures
* [x] add properties using properties package
* [x] generate API
* [x] add square loop receiver class

PR #689, opened 2018-04-11
Bug/ind active priority

lheagy:
- ensure `indActive` set prior to cell weights in the regularization init so the validator looks for cell weights that are the same size as the indactive
- fixes the bug @dccowan was encountering in the example he generated in #684


tests/base/test_regularization.py
```diff
@@ -357,5 +357,26 @@ def test_nC_residual(self):
...
+
+        # z direction
+        npad = 12
+        temp = np.logspace(np.log10(1.), np.log10(12.), 19)
+        temp_pad = temp[-1] * 1.3 ** np.arange(npad)
+        hz = np.r_[temp_pad[::-1], temp[::-1], temp, temp_pad]
+        mesh = Mesh.CylMesh([hx, 1, hz], '00C')
+        active = mesh.vectorCCz < 0.
+
+        active = mesh.vectorCCz < 0.
```
fourndo: Duplicate line

lheagy: thanks!

lheagy:
Thanks @fourndo, I have removed the duplicate and cleaned out some commented code in the tests

lheagy:
@dccowan: it looks like you merged a bunch of stuff from your active branch here. I will start a new pr instead.

PR #691, opened 2018-04-17
ref/em_mass_matrices

lheagy:
- work on simplifying the storage of the inner product matrix derivatives, 
- implement storage for the mu deriv matrices

@sgkang: I would very much appreciate hearing your thoughts on this. I tried to simplify the implementation a bit and also extended it to apply to the mu matrices as well


SimPEG/EM/Base.py
```diff
@@ -202,16 +204,26 @@ def MfMui(self):
...
         Derivative of :code:`MfMui` with respect to the model.
         """
         if self.muiMap is None:
             return Utils.Zero()
 
-        return (
-            self.mesh.getFaceInnerProductDeriv(self.mui)(u) * self.muiDeriv
-        )
+        if getattr(self, '_MfMuiDeriv', None) is None:
+            self._MfMuiDeriv = self.mesh.getFaceInnerProductDeriv(
```
lheagy: instead of creating a separate property, I stash it as a private variable here

SimPEG/EM/Base.py
```diff
@@ -202,16 +204,26 @@ def MfMui(self):
...
+                np.ones(self.mesh.nC)
+            )(np.ones(self.mesh.nF)) * self.muiDeriv
+
+        if v is not None:
```
lheagy: in this case, if v is None, we get the same behaviour as before ant this returns a matrix

SimPEG/EM/Static/DC/ProblemDC.py
```diff
@@ -422,8 +422,6 @@ def getA(self):
         Grad = self.mesh.nodalGrad
         A = Grad.T * MeSigma * Grad
 
-        Vol = self.mesh.vol
```
lheagy: this isn't used in the nodal prob. just cleaning it out

lheagy:
@sgkang: whenever you have a chance, would you mind taking a look at this?

lheagy:
@sgkang: would you mind taking a look at this whenever you have a chance?

SimPEG/EM/Base.py
sgkang: I am not sure this is doing the same thing. My main purpose was to update them for the inversion. 
For instance in second iteration I need to update `_MfMuiDeriv`, but this property is not None, so my 
guess update does not happen with this way.

SimPEG/EM/Base.py
sgkang: nice. it is handy for the later use when we want to generate sensitivity matrix.

SimPEG/EM/Base.py
```diff
@@ -303,41 +331,26 @@ def MeSigma(self):
             self._MeSigma = self.mesh.getEdgeInnerProduct(self.sigma)
         return self._MeSigma
 
-    @property
```
sgkang: similar comment. we need to make sure that we are updating this in the inversion.

SimPEG/EM/Base.py
```diff
@@ -390,39 +394,26 @@ def MfRho(self):
             self._MfRho = self.mesh.getFaceInnerProduct(self.rho)
         return self._MfRho
 
-    @property
-    def MfRhoDerivMat(self):
```
sgkang: similar comment.

SimPEG/EM/Static/DC/ProblemDC.py
sgkang: nice.

SimPEG/EM/Base.py
lheagy: thanks for catching this, I will add MfMuiDeriv to the `_clear_on_mu_update` list above

SimPEG/EM/Base.py
```diff
@@ -71,13 +71,15 @@ def _makeASymmetric(self):
...
+        return [
+            '_MeMu', '_MeMuI', '_MfMui', '_MfMuiI'
+        ]
 
     @property
     def _clear_on_sigma_update(self):
         return [
             '_MeSigma', '_MeSigmaI', '_MfRho', '_MfRhoI',
-            '_MeSigmaDerivMatrix', '_MfRhoDerivMatrix'
+            '_MeSigmaDeriv', '_MfRhoDeriv'
```
lheagy: This ensures that the deriv mats get cleared on any update to sigma

SimPEG/EM/Base.py
lheagy: This does get cleared on updates to sigma or rho - since `_MeSigmaDeriv` is in the list of `_clear_on_sigma_update`, it will be deleted if there is an update to sigma in the inversion

SimPEG/EM/Base.py
lheagy: `_MfRhoDeriv` is cleared on any updates to sigma. See line 82 above

lheagy:
@sgkang: would you mind taking a look at the recent updates here? Please let me know if you have any questions

lheagy:
Thanks @sgkang: I will add some docstrings

PR #692, opened 2018-04-20
L2ObjectiveFunction deriv2

### LINKED ISSUE #686: Objective function
I think that the `L2ObjectiveFunction` does not calculate the second derivate correctly:

image

I think it should use the `W` not the `self.W` in the final line:

`return W.T * W`

lheagy:
fix deriv2 on the L2Objective function if v is None, add a test to ensure behaviour is the same if v is none and if v provided

closes #686


tests/base/test_Objectivefct.py
```diff
@@ -41,6 +41,16 @@ def test_derivs(self):
         objfct = ObjectiveFunction.L2ObjectiveFunction()
         self.assertTrue(objfct.test(eps=1e-9))
 
+    def test_deriv2(self):
```
rowanc1: Thanks @lheagy !

Did this actually fail before?

lheagy: The deriv2 was tested in the test above and it passed because it was passing a vector around (internally, we always pass `v`, so the correct deriv2 was being provided. This test just makes sure that the deriv2 when you provide a vec is the same as if you don't (and this failed before)

PR #695, opened 2018-04-28
simplify getABMN

lheagy:
- move the if else so that we don't have to repeat the rx if / else statement in getABMN

@sgkang: would you be willing to take a look?


thast:
@micmitch : that might concern you. (he is using it the most I believe)

micmitch:
I'm sure that I probably did not write this in the most efficient manner. Improvements in efficiency are always good. This does not change the output of the function does it?

lheagy:
@micmitch, if you have any comments, they are welcome! this doesn't change the output. It is a very lightweight change that just removes duplicate code setting the m and n locations of the receivers

sgkang:
hey @lheagy looks good on my end. Simpler the better.

PR #698, opened 2018-06-07
printversion

### LINKED ISSUE #697: displaying versions
@prisae suggested including a util for printing the version the core packages in your environment for reproducing SimPEG simulations and inversions. 

selection_001

selection_002

**Required Packages**
- numpy
- scipy
- cython
- properties
- vectormath
- discretize
- pymatsolver

**Optional Packages**
- matplotlib
- ipython
- ipywidgets

prisae:
I first go at `versions`. Fixes #697 .

Please have a look, feedback welcome!

selection_001

selection_002


prisae:
There are more options. You can get the html-table as plain text with `SimPEG.versions('html')` (html lower case instead of upper case). You can have that printed nicely by `print(SimPEG.versions('html'))`.

How many columns are used can be set with `SimPEG.versions('html', ncol=5)`, for instance.

Additional packages can be provided via `SimPEG.versions('HTML', add_pckg=[simpegEM1D, geoana])`. They have to be imported in the script beforehand. It only works for packages which use the `__version__` feature though, it is not more sophisticated.

prisae:
Example of the `add_pckg` and `ncol` arguments:

selection_003

SimPEG/Utils/printinfo.py
```diff
@@ -0,0 +1,281 @@
...
+    mklinfo = mkl.get_version_string()
+elif numexpr:
+    mklinfo = numexpr.get_vml_version()
+else:
+    mklinfo = False
+
+__all__ = ['versions', 'versions_html', 'versions_text']
+
+
+def versions(mode='print', add_pckg=None, ncol=4):
```
rowanc1: I wonder if changing the default here depending on if you are in IPython or not?

```python
def versions(mode='html' if IPython else 'print', add_pckg=None, ncol=4):
    ...
```

prisae: I guess you mean `mode='HTML'`, which returns the rendered table (`mode='html'` returns the actual html-code).

You can, but I think is better not to. You would have to make sure you are in an Jupyter Notebook, because otherwise you could also be in an IPython console, and the `mode='HTML'` cannot render.

Also, what speaks for having `mode='print'` is that this looks OK in a notebook too, it would just print it as in the console. So this is the only mode that always works nicely. If there is a nice and easy trick though to ensure if you are in a Jupyter notebook then it could be a good idea, indeed.

prisae: I had another look at this, and I could not find a good way. You could distinguish between an IPython Console and a Jupyter Notebook, but not between a Qt Console and a Jupyter Notebook. But the Qt Console cannot render the `HTML` mode either. So I think sticking with `mode='print'` as default is the safest option in whatever environment someone is using `SimPEG`.

SimPEG/Utils/printinfo.py
```diff
@@ -0,0 +1,281 @@
+"""
+:mod:`printinfo` -- Tools to print date, time, and version information
+======================================================================
+
+Print or return date, time, and package version information in any environment
+(Jupyter notebook, IPython console, Python console, QT console), either as
+html-table (notebook) or as plain text (anywhere).
+
+This script is an adapted version of ``empymod.printversions``, which itself
```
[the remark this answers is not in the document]
prisae: Good point, as I put links as well for `ipynbtools.py` and `watermar.py`. I'll add `https://empymod.github.io`.

PR #699, opened 2018-06-11
Em/feat/3D cyl sources

lheagy:
- rely on geoana for vector potentials in the sources 
- add geoana to the dev requirements

@sgkang: would you be willing to do a quick review on this? I know you have already used it, so if you are happy with these changes, we can put this in with the 3D cyl mesh upgrades

### DIFF
--- a/SimPEG/EM/FDEM/SrcFDEM.py
+++ b/SimPEG/EM/FDEM/SrcFDEM.py
@@ -3,6 +3,8 @@ import numpy as np
 from scipy.constants import mu_0
 import warnings
 
+from geoana.em.static import MagneticDipoleWholeSpace, CircularLoopWholeSpace
+
 from SimPEG.Utils import Zero
 from SimPEG import Survey, Problem, Utils
 
@@ -306,27 +308,13 @@ class MagDipole(BaseFDEMSrc):
         self.freq = freq
         self.loc = loc
 
-
-    @properties.validator('orientation')
-    def _warn_non_axis_aligned_sources(self, change):
-        value = change['value']
-        axaligned = [
-            True for vec in [np.r_[1.,0.,0.], np.r_[0.,1.,0.], np.r_[0.,0.,1.]]
-            if np.all(value == vec)
-        ]
-        if len(axaligned) != 1:
-            warnings.warn(
-                'non-axes aligned orientations {} are not rigorously'
-                ' tested'.format(value)
+    def _srcFct(self, obsLoc, coordinates="cartesian"):
+        if getattr(self, '_dipole', None) is None:
+            self._dipole = MagneticDipoleWholeSpace(
+                mu=self.mu, orientation=self.orientation, location=self.loc,
+                moment=self.moment
             )
-
-
-
-    def _srcFct(self, obsLoc, component):
-        return emutils.MagneticDipoleVectorPotential(
-            self.loc, obsLoc, component, mu=self.mu, moment=self.moment,
-            orientation=self.orientation
-        )
+        return self._dipole.vector_potential(obsLoc, coordinates=coordinates)
 
     def bPrimary(self, prob):
         """
@@ -337,6 +325,7 @@ class MagDipole(BaseFDEMSrc):
         :return: primary magnetic field
         """
         formulation = prob._formulation
+        coordinates = "cartesian"
 
         if formulation == 'EB':
             gridX = prob.mesh.gridEx
@@ -350,19 +339,25 @@ class MagDipole(BaseFDEMSrc):
             gridZ = prob.mesh.gridFz
             C = prob.mesh.edgeCurl.T
 
-        if (
-                prob.mesh._meshType == 'CYL' and
-                getattr(prob.mesh, 'isSymmetric', None) is True
-        ):
-            assert (np.linalg.norm(self.orientation - np.r_[0., 0., 1.]) <
-                    1e-6), ('for cylindrical symmetry, the dipole must be '
-                            'oriented in the Z direction')
-            a = self._srcFct(gridY, 'y')
-        else:
-            ax = self._srcFct(gridX, 'x')
-            ay = self._srcFct(gridY, 'y')
-            az = self._srcFct(gridZ, 'z')
-            a = np.concatenate((ax, ay, az))
+        if prob.mesh._meshType == 'CYL':
+            coordinates = "cylindrical"
+
+            if prob.mesh.isSymmetric is True:
+                if not (
+                    np.linalg.norm(self.orientation - np.r_[0., 0., 1.]) < 1e-6
+                ):
+                    raise AssertionError(
+                        'for cylindrical symmetry, the dipole must be oriented'
+                        ' in the Z direction'
+                    )
+                a = self._srcFct(gridY)[:, 1]
+
+                return C*a
+
+        ax = self._srcFct(gridX, coordinates)[:, 0]
+        ay = self._srcFct(gridY, coordinates)[:, 1]
+        az = self._srcFct(gridZ, coordinates)[:, 2]
+        a = np.concatenate((ax, ay, az))
 
         return C*a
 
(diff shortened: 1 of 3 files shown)


SimPEG/EM/TDEM/ProblemTDEM.py
```diff
@@ -21,6 +21,7 @@ class BaseTDEMProblem(Problem.BaseTimeProblem, BaseEMProblem):
     surveyPair = SurveyTDEM  #: A SimPEG.EM.TDEM.SurveyTDEM Class
     fieldsPair = FieldsTDEM  #: A SimPEG.EM.TDEM.FieldsTDEM Class
     clean_on_model_update = ['_Adcinv']  #: clear DC matrix factors on any model updates
+    dt_threshold = 1e-8
```
sgkang: @lheagy this prevents doing unnecessary factorization. Let me know what do you think

examples/00_published/plot_booky_1D_time_freq_inv.py
```diff
@@ -264,7 +264,7 @@ def run(plotIt=True, saveFig=False, cleanup=True):
     # Bird height from the surface
     b_height_skytem = skytem["src_elevation"].value
     src_height = b_height_skytem[rxind_skytem]
-    srcLoc = np.array([[0., 0., src_height]])
+    srcLoc = np.array([0., 0., src_height])
```
sgkang: This is changed from previous version. Now we need to put 1d array for src location. 
This seems consistent with frequency domain code. Is that what you aimed for?

requirements_dev.txt
```diff
@@ -17,3 +17,4 @@ vectormath
 wheel
 pytest
 git+https://github.com/simpeg/discretize.git@feat/3Dcylmesh
+git+https://github.com/simpeg/geoana@ref/em_and_docs
```
sgkang: I had to make couple of changes here, so take a look @lheagy

examples/00_published/plot_booky_1D_time_freq_inv.py
lheagy: hmm, k I will take a look. Ideally, we should be able to input either.

requirements_dev.txt
lheagy: thanks @sgkang: I would like to merge the geoana pr (simpeg/geoana#9) to master so that we can keep the requirements on the master geoana. For testing at the moment, this is fine though!

lheagy:
The current testing errors and failures are related to the distribution of properties and vectormath on pypi: seequent/properties#247

lheagy:
For versioning here, I suggest this is a `coda minor` (e.g.)
```
SimPEG 0.8.5 --> 0.9.0
```
to follow the discretize change in simpeg/discretize#36 and it contains fairly major additions, particularly in the TDEM code. Thoughts?

examples/00_published/plot_booky_1D_time_freq_inv.py
lheagy: yeah, now we require it to be a vector. If you think this is annoying, I can add the functionality to allow a 2D array to be input and it be cast to a vector. Let me know what you think @sgkang

sgkang: Will be nice, if we can input both forms.

lheagy: kk, I have added a property that will allow us to do that: `LocationVector` in `Props.py`. This is related to: seequent/properties#250

SimPEG/Props.py
```diff
@@ -491,3 +491,26 @@ def summary(self):
             out += ['', str(e)]
 
         return '\n'.join(out)
+
+
+class LocationVector(properties.Array):
```
lheagy: @sgkang: this will allow the user to input a 1D or 2D array and cast it to a 1D vector

tests/em/fdem/forward/test_FDEM_forward.py
```diff
@@ -17,6 +19,42 @@
 SrcList = ['RawVec', 'MagDipole_Bfield', 'MagDipole', 'CircularLoop']
 
 
+class SrcLocTest(unittest.TestCase):
+    def test_src(self):
```
lheagy: and here is a test to make sure that the user can input 1D or 2D arrays (and that incorrect shapes will be caught)

SimPEG/Props.py
sgkang: cool. thanks @lheagy !

SimPEG/EM/FDEM/SrcFDEM.py
```diff
@@ -291,8 +294,8 @@ class MagDipole(BaseFDEMSrc):
     freq = properties.Float(
         "frequency of the source (Hz)", required=True
     )
-    loc = properties.Vector3(
-        "location of the source", default=np.r_[0.,0.,0.]
+    loc = LocationVector(
```
sgkang: so, is this an array or vector3?

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -310,36 +313,24 @@ class MagDipole(BaseTDEMSrc):
     orientation = properties.Vector3(
```
sgkang: not sure it seems previously only 'Z' was working. 
now it's fine for all three directions. if we want to put arbitrary direction, 
can we do np.array([0.1, 0.1, 0.5])?

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -577,3 +583,51 @@ def s_m(self, prob, time):
 
     def s_e(self, prob, time):
         return self.Mejs(prob) * self.waveform.eval(time)
+
+
+# TODO: this should be generalized and plugged into getting the Line current
```
sgkang: right. also, this does not work for octree codes

SimPEG/Props.py
sgkang: nice!

SimPEG/Regularization.py
```diff
@@ -447,7 +447,7 @@ def cellDiffzStencil(self):
         if getattr(self, '_cellDiffzStencil', None) is None:
 
             self._cellDiffzStencil = (
-                self.Pafz.T * self.mesh._cellGradzStencil() * self.Pac
+                self.Pafz.T * self.mesh._cellGradzStencil * self.Pac
```
sgkang: I see. Now `_cellGradzStencil` is property.

SimPEG/Survey.py
```diff
@@ -20,7 +22,7 @@ class BaseRx(object):
 
     def __init__(self, locs, rxType, **kwargs):
         self.uid = str(uuid.uuid4())
-        self.locs = locs
+        self.locs = np.atleast_2d(locs)
```
sgkang: is this okay for DC case (e.g. dipole Rx)?

tests/em/fdem/forward/test_FDEM_forward.py
sgkang: that's great!

tests/em/static/test_DC_FieldsDipoleFullspace.py
```diff
@@ -92,9 +92,9 @@ def setUp(self):
         self.ROIfaceInds = ROIfaceInds
 
 
-    def test_Problem3D_CC_Dirchlet(self, tolerance=0.1):
```
sgkang: thanks @lheagy !

tests/em/tdem/test_TDEM_DerivAdjoint_galvanic.py
```diff
@@ -7,7 +7,7 @@
 
 plotIt = False
 
-testDeriv = True
+testDeriv = False
```
sgkang: why this is changed to False?

lheagy: ah, thanks for catching this - it shouldn't have been (I must have forgotten about it when running some tests). I will update

SimPEG/Survey.py
lheagy: Under the hood, we always assume that the receiver locations are input as a 2D array. This takes care of the case where the user inputs a 1D array, e.g. `np.r_[0., 0., 10.]` and casts it to a 2D array. 

With respect to the DC code - it does mean that the `locs` which were a list are now a 3D numpy array. However, this doesn't change any behavior. Asking for `rx.locs[0]` will still return the 2D array of M-electrode locations, and `rx.locs[1]` will return the N-electrode locations as expected. I tested this on http://docs.simpeg.xyz/content/examples/06-dc/plot_dipoledipole_2Dinversion_twocylinders.html 

image

SimPEG/EM/TDEM/SrcTDEM.py
sgkang: cool!

SimPEG/Survey.py
sgkang: nice! thanks @lheagy

PR #701, opened 2018-06-27
Printversion revisions

lheagy:
## updates
- convert test on printversion from pytest to unittest (see issue #702) 
- updates to docstrings
- add version to a couple of the notebook-styled examples

## todos
- [ ] check into the docs failure on 2.7
- [ ] think about docs: currently, we use the basic, vanilla sphinx formatting (i.e. https://github.com/simpeg/simpeg/blob/master/SimPEG/Utils/codeutils.py#L9) instead of napoleon. @prisae: I haven't done much digging into the napoleon extension - it looks like the end-product is a bit different in terms of style - or should they be identical? Right now, I have enabled the napoleon extension, so we can have both
image see issue #703

### DIFF
--- a/SimPEG/Fields.py
+++ b/SimPEG/Fields.py
@@ -16,12 +16,13 @@ class Fields(object):
 
     """
 
-    #: Known fields,   a dict with locations, e.g. {"e": "E", "phi": "CC"}
+    #: Known fields, a dict with locations, e.g. ``{"e": "E", "phi": "CC"}``
     knownFields = None
-    #: Aliased fields, a dict with [alias, location, function], e.g. {"b":["e","F",lambda(F,e,ind)]}
+    #: Aliased fields, a dict with [alias, location, function], e.g. ``{"b": ["e", "F", lambda(F,e,ind)]}``
     aliasFields = None
     #: dtype is the type of the storage matrix. This can be a dictionary.
     dtype = float
+
 
     def __init__(self, mesh, survey, **kwargs):
         self.survey = survey
--- a/SimPEG/Utils/printinfo.py
+++ b/SimPEG/Utils/printinfo.py
@@ -1,6 +1,6 @@
 """
-:mod:`printinfo` -- Tools to print date, time, and version information
-======================================================================
+`printinfo` -- Tools to print date, time, and version information
+=================================================================
 
 Print or return date, time, and package version information in any environment
 (Jupyter notebook, IPython console, Python console, QT console), either as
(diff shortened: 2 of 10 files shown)


SimPEG/Fields.py
```diff
@@ -16,13 +16,14 @@ class Fields(object):
 
     """
 
-    #: Known fields,   a dict with locations, e.g. {"e": "E", "phi": "CC"}
+    #: Known fields, a dict with locations, e.g. ``{"e": "E", "phi": "CC"}``
     knownFields = None
-    #: Aliased fields, a dict with [alias, location, function], e.g. {"b":["e","F",lambda(F,e,ind)]}
+    #: Aliased fields, a dict with [alias, location, function], e.g. ``{"b": ["e", "F", lambda(F,e,ind)]}``
```
lheagy: these are just for napoleon compatibility (and good for styling anyways)

SimPEG/Utils/printinfo.py
```diff
@@ -223,7 +222,7 @@ def versions_text(add_pckg=None):
 
     # Width for text-version
     n = 54
-    text = '\n' + n*'-' + '\n'
+    text = u'\n' + ''.join(n*['-']) + '\n'
```
lheagy: this is for python 2.7 compatibility

docs/content/api_core/api_Utils.rst
```diff
@@ -5,6 +5,14 @@ Utils
     :members:
     :undoc-members:
 
+Version
```
lheagy: added it to the docs!

examples/06-dc/plot_dc_schlumberger_array.py
```diff
@@ -266,3 +266,10 @@
 ax[1].set_xlabel('Conductivity (S/m)', fontsize=25)
 ax[1].set_ylabel('Depth (m)', fontsize=25)
 ax[1].set_title('(b)', fontsize=25)
+
+###############################################################################
```
lheagy: display the versions in the notebook-styled examples

prisae:
No, napoleon and plain sphinx do use differently. Mainly with regards to the `Parameters`, `Return`, etc, styling of the docstrings. I would recommend napoleon, as SimPEG is also a heavy numpy/scipy user...

examples/06-dc/plot_dc_schlumberger_array.py
prisae: You say "notebook-styled" examples; wouldn't you want to add then `versions('HTML')` instead of `versions()`?

prisae: Never mind. it is a pure Python file I see, so it was your comment that mislead me.

lheagy: We need it to run both as a stand-alone python script and as a notebook, so I figured sticking with the defaults would be simpler

lheagy:
The fail here is just due to the example timing-out. I will resolve that on the other branch when we are doing the merge to master!

PR #706, opened 2018-07-13
Effective medium theory extension

lheagy:
# Effective medium theory map
- extend the effective medium theory mapping to work for preferentially aligned ellipsoids so that  anisotropic conductivities are estimated
- add a calculation of the Hashin-Shtrikman bounds 

# Anisotropic models
- allow the shape of models to include 2D arrays so that anisotropic models can be input into simulations

# New Example
- add an example of the effective medium theory map from my thesis 
image


SimPEG/Props.py
```diff
@@ -60,6 +60,7 @@ class Model(SphinxProp, properties.Array):
 
     class_info = 'a numpy array'
     _required = False
+    _shape = {('*',), ('*', '*')}
```
lheagy: @rowanc1: I added this to allow us to input anisotropic models

lheagy:
👋 hi @sgkang: would you be willing to take a look at this pr?

SimPEG/Props.py
rowanc1: cool!

rowanc1:
Makes a lot of sense to probably pull mappings out into their own package in the future. This is a great place for now. Looking forward to seeing this in SimPEG!!

lheagy:
Thanks @rowanc1: agreed - Maps.py is getting pretty big!

lheagy:
@rowanc1 or @sgkang: would one of you be willing to do a quick review on this?

SimPEG/Maps.py
```diff
@@ -465,10 +465,6 @@ class SelfConsistentEffectiveMedium(IdentityMap, properties.HasProperties):
 
             \chi = \sqrt{\\frac{1}{\\alpha^2} - 1}
```
sgkang: It will be nice clarify what are the input and output parameters of the mapping
So, input is simga0, sigma1, phi, and output is effective sigma

Also, it will be nice to comment what is the inversion parameters in mind for the current set up.

examples/03-maps/plot_effective_medium_theory.py
```diff
@@ -0,0 +1,81 @@
...
+#
+
+fig, ax = plt.subplots(1, 1, figsize=(7, 4), dpi=350)
+
+ax.semilogy(phi, sige)
+ax.grid(which="both", alpha=0.4)
+ax.legend(["{:1.0e} S/m".format(s) for s in sigma1])
+ax.set_xlabel("Volume fraction of proppant $\phi$")
+ax.set_ylabel("Effective conductivity (S/m)")
+
```
sgkang: add `plt.tight_layout`

SimPEG/Maps.py
```diff
@@ -604,40 +715,40 @@ def _sc2phaseEMTRandSpheroidstransform(self, phi1):
...
+        R0 = getR(self.sigma0, sige, self.alpha0, self.orientation0)
+        R1 = getR(self.sigma1, sige, self.alpha1, self.orientation1)
 
         num = -(sigma0 - sige)*R0
         den = (sigma1-sige)*R1 - (sigma0-sige)*R0
 
         return num/den
 
-    def _sc2phaseEMTRandSpheroidstransformDeriv(self, sige, phi1):
+    def _sc2phaseEMTSpheroidstransformDeriv(self, sige, phi1):
```
sgkang: So, I guess inversion parameter is phi1 (volume fraction) ?
I am not sure why phi is not set as a property... any reason why?

examples/03-maps/plot_effective_medium_theory.py
lheagy: thanks seogi, good call

SimPEG/Maps.py
lheagy: here, phi is the inversion model, so it should be an input, rather than a property. I have added docs to hopefully clarify

SimPEG/Maps.py
```diff
@@ -389,10 +389,10 @@ def nP(self):
 class SelfConsistentEffectiveMedium(IdentityMap, properties.HasProperties):
     """
         Two phase self-consistent effective medium theory mapping for
-        ellipsoidal inclusions. The model is the concentration
+        ellipsoidal inclusions. The inversion model is the concentration
```
lheagy: @sgkang: I specified that the inversion model is concentration here

SimPEG/Maps.py
lheagy: Thanks seogi, I have specified it in the docs above

lheagy:
Thanks for your comments @sgkang: I have gone in and improved the docs. Whenever you have a chance, could you take a look and see if that addresses your questions?

lheagy:
👋 Hi @sgkang: I went through and improved the docs to make it more clear what the inversion model it. Would you mind taking a quick look when you have a chance?

SimPEG/Maps.py
sgkang: Thanks @lheagy !

lheagy:
Excellent, thanks @sgkang!

PR #707, opened 2018-07-16
Nsem/ref/simulation into ref/simulation

grosenkj:
Added support for Horizontal Magnetic transfers. Added plotting functions and testing for the code.


lheagy:
I will close this as the `ref/simulation` branch was a bit of a false-start. So for the NSEM updates the work by @JKutt should hopefully bring us back up to speed 😄

PR #724, opened 2018-10-01
Feat/dc fields derivs

lheagy:
## DC
- add fields derivatives with respect to the electric field (in case you want to invert e-field data)
- add e-field receivers to DC problem 

## TDEM 
- add charge and charge density to j-fields
- bug fix in calculation of db/dt in the j-fields object 

## Directives
- add option to save output to a directory, the default is the current working directory 

## TODO:
- [x] tests
- [x] clean up commented-out code


SimPEG/Maps.py
```diff
@@ -347,6 +357,161 @@ def deriv(self, m, v=None):
         return self.P
 
 
+class SumMap(ComboMap):
```
lheagy: @fourndo: can you take a look at this? This should do the same as is in pr #666 (which has to be bad luck if we merge it!, ha)

SimPEG/Maps.py
```diff
@@ -347,6 +357,161 @@ def deriv(self, m, v=None):
...
+            deriv = map_i.deriv(m0, v=deriv)
+            if ii == 0:
+                sumDeriv = deriv
+            else:
+                sumDeriv += deriv
+
+        return sumDeriv
+
+
+class SurjectUnits(IdentityMap):
```
lheagy: and this one too @fourndo? This should replace the HomogeneousMap from #666

SimPEG/Maps.py
```diff
@@ -1852,7 +2017,7 @@ def __init__(self, mesh, pts, ptsv=None, order=3, logSigma=True,
 
 class BaseParametric(IdentityMap):
 
-    slopeFact = 1e2  # will be scaled by the mesh.
+    slopeFact = 1  # will be scaled by the mesh.
```
lheagy: this changes a default - here, the slope factor should not be as severe ... makes the parametric maps converge more slowly

examples/03-maps/plot_sumMap.py
```diff
@@ -0,0 +1,192 @@
...
+from SimPEG import (
+    Mesh, Utils, Maps, Regularization,
+    DataMisfit, Optimization, InvProblem,
+    Directives, Inversion, PF
+)
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def run(plotIt=True):
```
lheagy: Thanks for adding this @fourndo! Rather than using the `def run`, would you mind turning this into a notebook styled example (e.g. http://discretize.simpeg.xyz/en/latest/auto_examples/plot_slicer_demo.html#sphx-glr-auto-examples-plot-slicer-demo-py, the source is: https://github.com/simpeg/discretize/blob/master/docs/examples/plot_slicer_demo.py). 

The notebook-styled examples are a bit more friendly for new users

examples/03-maps/plot_sumMap.py
```diff
@@ -0,0 +1,192 @@
...
+    inv = Inversion.BaseInversion(invProb,
+                                  directiveList=[IRLS, betaest, update_Jacobi])
+
+    # Run the inversion
+    m0 = np.ones(sumMap.shape[1])*1e-4  # Starting model
+    prob.model = m0
+    mrecSum = inv.run(m0)
+    if plotIt:
+
+        mesh.plot3DSlicer(actvMap * model, aspect="equal", zslice=30, pcolorOpts={"cmap":'inferno_r'}, transparent='slider')
```
lheagy: the method is now `plot_3d_slicer`

examples/03-maps/plot_sumMap.py
```diff
@@ -0,0 +1,192 @@
...
+
+        mesh.plot3DSlicer(actvMap * sumMap * mrecSum, aspect="equal", zslice=30, pcolorOpts={"cmap":'inferno_r'}, transparent='slider')
+
+
+
+if __name__ == '__main__':
```
lheagy: if you do the notebook-styled example, this is no longer needed

SimPEG/EM/Static/DC/FieldsDC.py
```diff
@@ -125,15 +126,34 @@ def _j(self, phiSolution, srcList):
         """
         return self._MfRhoI*self._Grad*phiSolution
 
+    def _jDeriv_u(self, src, v, adjoint=False):
```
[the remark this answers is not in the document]
lheagy: yep, I have examples inverting e in my thesis, and there are tests for j, but no examples anywhere yet

tests/em/static/test_DC_jvecjtvecadj.py
```diff
@@ -81,6 +86,69 @@ def test_dataObj(self):
         )
         self.assertTrue(passed)
 
+class DCProblemTestsCC_fields(unittest.TestCase):
```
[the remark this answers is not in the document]
lheagy: correct, it hasn't yet been implemented for the nodal problem

SimPEG/EM/TDEM/SrcTDEM.py
```diff
@@ -611,18 +612,53 @@ def phiInitial(self, prob):
...
-            return - prob.MfRhoI * (Div.T * phi)
+        if self.waveform.hasInitialFields is False:
+            return Zero()
 
-        else:
+        phi = self.phiInitial(prob)
+        Div = Utils.sdiag(prob.mesh.vol) * prob.mesh.faceDiv
+        return - prob.MfRhoI * (Div.T * phi)
+
+    def _getAmmr(self, prob):
```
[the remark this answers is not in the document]
lheagy: yep, only e-b right now, and only for forward simulations. I can add the derivs before we merge this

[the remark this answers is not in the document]
lheagy: Dervis and the adjoints are now added and tested for the h-j formulation

lheagy:
Many thanks @sgkang for the review! I have gone through and added derivatives for the initial fields. Right now, the grounded source where we solve the MMR problem is only the H-J formulation. E-B is not yet implemented. Please let me know if you have any other questions or comments.

PR #729, opened 2018-10-16
Addition regarding inversion to big_picture

prisae:
Feel free to ignore this request, or expand upon it, or modify it. I just thought the big, mystic word of **geophysical inversion** would benefit from some additional explanation. The two main questions in my opinion are:

- What is a geophysical inversion?
Sometimes this is hard to grasp for newbies or for non-geophysicists, when in fact it really is nothing else than iterative modelling with some predefined rules of how to adjust the model and when to stop the iterations. 
- Why do we need to invert the data?
In our case, low-frequency and potential-field methods, there is simply no other way to interpret the data than with this indirect *inversion* method, as we cannot process the data in the same manner one can process, for instance, seismic data or GPR data.

I tried to briefly answer these two question in my text, without going too much into details and becoming too long.


docs/content/basic/big_picture.rst
```diff
@@ -21,6 +21,21 @@ a framework that exposes the details of the implementation to the geophysicist i
 based interrogation. This framework can be easily extended to encompass many geophysical problems and is built with the inverse problem
 as the fundamental goal.
 
+Inversion is the cornerstone of low-frequency and potential-field geophysical
+methods. There are geophysical methods where you can get a subsurface property
+directly from the measured data, an example being the calculation of velocity
+through the normal-moveout (NMO) calculation in seismic data. In low-frequency
```
leouieda: @prisae I really like the idea of adding more context about terms that we might take for granted! I'm not sure I agree with this phrasing, though. In practice, the NMO velocity isn't really the physical property of the rock and the seismic image is not a physical property map (hence the need for tomography, FWI, AVO modeling). And there is such a thing as potential field migration, which gives you an image (like in seismic) but not really a physical property model. There are also other techniques which aren't strictly classified as inversion like Euler deconvolution, interpretation of the radial average power spectrum, MINMAX estimation of magnetization, gradient tensor eigenvector  analysis, etc. Inversion is also a corner-stone for seismic data because you need a velocity model to do migration, and that often comes from tomography.

I would suggest rewording this to include all of geophysics instead of singling out low-frequency and potential fields. Inversion is the apex of geophysical interpretation and is the way we estimate subsurface structure and physical property distributions but geophysical data. It's not the only way, but it is the most sophisticated.

prisae: @leouieda Thanks for your input, I agree with most what you say. I think the focus on low-frequency and potential field data came from the focus of SimPEG. Even though there is a seismic and a GPR module I think most stuff done, at the moment, falls into low-frequency and potential field data.

I agree that NMO velocity isn't a property of the rock, that would be density etc. However, NMO is an interesting example because there you actually move the data in order to get a value, something you hardly ever do in low-freq and potentials (but you give a few examples though).

I think you can edit straight into my edits, right? Would you mind give it a go, and I will have another go at it afterwards to, so we have kind of an iteration here? You could specifically input your knowledge with regards to potential field data, as I have hardly any experience with those.

lheagy: Sorry for dropping the ball here! 

What do you think about starting with something along the lines of
```
There are some geophysical methods where you can make interpretations about the subsurface from (processed) data. For example, geologic contacts can be interpreted from seismic data. Similarly in potential fields data, filtering techniques are often applied to magnetic data and used to interpret geologic maps where rock exposure is limited. In many cases though, the geologic settings and governing physics are sufficiently complex and the data cannot be directly interpreted to construct a geologic model. This is where inversion techniques, which make use of the ability to simulate the governing equations, are a powerful tool for interpretation. 
```
and then we can pick up with your description of how inversion is an iterative process?

I totally agree that NMO is an interesting example, but it is a bit subtle to explain (as is apparent resistivity), so I am struggling to work it in. 

cc @dougoldenburg, if you have a chance to take a look, your input on this too would be appreciated!

prisae: Anything is fine with me, feel free to edit directly. I just thought it would help to expand on 'inversion' a bit, as it is not an easy concept when you are new to it.

lheagy:
We are collaboratively editing this page here: https://docs.google.com/document/d/1y5TKxLhB4AOYPuZW0_oITm-kZhkPNXX7FNeVsIKssig/edit

prisae:
What is the status of this? I cannot recall. Has there be any change and we close this PR? Or is it still an open issue?

prisae:
Over three years old! We should move that in or delete it, and not keep it dead hanging around. I originally just thought that "geophysical inversion" could need a sentence or two more. It might have gotten sidetracked then... What do you think @leouieda , @lheagy ?

prisae:
Over three years old! We should move that in or delete it, and not keep it dead hanging around. I originally just thought that "geophysical inversion" could need a sentence or two more. It might have gotten sidetracked then... What do you think @leouieda , @lheagy ?

jcapriot:
This is probably a good time to revive this PR, as we're updating the guides in the documentation.

PR #730, opened 2018-10-21
fix docs fail

### LINKED ISSUE #726: MASTER is failing!

lheagy:
- comment out napoleon in the docs build until we transition over
- closes #726


thast:
Thanks @lheagy ! You will have to explain that one to me as I couldn't figure it out last week :)

lheagy:
It was subtle, @thast - mostly I just tried poking around with a few things locally until it worked, haha. Eventually, we should transition to using napoleon instead for the docs, but for now, this works

PR #731, opened 2018-10-24
drop support for python 2.7

lheagy:
- remove testing from python 2.7
- add requirement of python >= 3.6 in the setup.py

(follows simpeg/discretize#111)


SimPEG/__init__.py
```diff
@@ -26,4 +25,4 @@
 __version__   = '0.9.2'
 __author__    = 'SimPEG Team'
 __license__   = 'MIT'
-__copyright__ = '2013 - 2018, SimPEG Team, http://simpeg.xyz'
+__copyright__ = f"2013 - {_datetime.datetime.now().year}, {__author__}, http://simpeg.xyz"
```
rowanc1: We could probably `del datetime` so that it is not exposed in the base namespace.

jcapriot:
Should we close this with `simulation`?

lheagy:
closed with #786

PR #732, opened 2018-10-26
Fix Zenodo DOI

prisae:
Just noted that the current Zenodo DOI points to version 0.7.1 from January 2018. Zenodo has a cool feature, a DOI for a project that always links to the newest version, and lists all zenodos ever released for that piece of work. For SimPEG, it is [DOI](https://doi.org/10.5281/zenodo.596373)

Edit: Corrected Zenodo


rowanc1:
Thanks @prisae!

Issue #734, opened 2018-11-16
IO.pseudosections and Utils.plot2Ddata CLIM is not doing anything.


thast:
Thanks @sdevriese for finding it!
Currently the argument `clim` is not doing anything for those two functions, which are used in the binders for visualization.
@sdevriese is trying to use it at her work, using the binders to visualize their own data. I have provided a quick fix, using the DCUtils instead of the IO class but I'd really like to see it fix and merge quickly so we can encourage more users like her.

I will do the fix. It should be very easy (only a few lines of code). @lheagy can we hope for a merge during the weekend so they can resume normal operation and do not have to modify the binder everytime they open it to make it work.

sdevriese:
I really appreciate all the help from @sgkang and @thast this morning! Now on my to-do list is to download and install simpeg (on windows...) so I don't have to use binder and can modify and more easily save notebooks.

thast:
solved in PRs #735 and #738

PR #735, opened 2018-11-16
enhanced plot2DData and IO.pseudosections

thast:
Address issue #734


thast:
image

thast:
As a PR note: colorscale of matplotlib contourf are somehow difficult to handle (@sgkang was also warning me about it before I start). So I went full declarative and control-freak on the levels definition.

PR #736, opened 2018-11-19
Update installing.rst

sdevriese:
Updating link for Anaconda because continuum.io/downloads didn't work


lheagy:
@sdevriese: Thanks for this 🎉. Please add yourself to [AUTHORS.rst](https://github.com/simpeg/simpeg/blob/master/AUTHORS.rst) if you are not there already!

PR #738, opened 2018-11-19
fixed possible issue for levels definition in `SimPEG.Utils.plot2DDdata`

thast:
- fixed possible issue for levels definition in `SimPEG.Utils.plot2DDdata`
- add mask functionality (already proposed as args but not implemented)
- choice of linear or nearest interpolation
- check if there is already a vmin, vmax in the contourOpts dict before adding them based on user-input clim
- improved example for`SimPEG.Utils.plot2DDdata`
- add an example for the IO.pseudosection function

Example output:
image


thast:
@lheagy : Thanks for your patience. I think this is ready for review.

SimPEG/Utils/PlotUtils.py
```diff
@@ -131,30 +148,36 @@ def plot2Ddata(xyz, data, vec=False, nx=100, ny=100,
...
         cont = ax.contourf(
             X, Y, DATA, levels=levels,
             vmin=vmin, vmax=vmax,
             **contourOpts
         )
         ax.streamplot(X, Y, DATAx, DATAy, color="w")
-        if level is not None:
-            CS = ax.contour(X, Y, DATA, level, colors="k", linewidths=2)
+        if level:
+            CS = ax.contour(X, Y, DATA, levels=levels, **levelOpts)
```
lheagy: thanks for this :)

examples/06-dc/read_plot_DC_data_wit_IO_class.py
```diff
@@ -0,0 +1,70 @@
...
+"""
+
+import numpy as np
+import pandas as pd
+import shutil
+import os
+from SimPEG import DC
+
+
+def run(PlotIt=True):
```
lheagy: you don't need to do this inside of the def run. It is okay, and we can merge as-is, but the notebook-styled examples are nicer: e.g. http://docs.simpeg.xyz/content/examples/07-fdem/plot_loop_loop_2Dinversion.html#sphx-glr-content-examples-07-fdem-plot-loop-loop-2dinversion-py

https://github.com/simpeg/simpeg/blob/master/examples/07-fdem/plot_loop_loop_2Dinversion.py. 

If you are game to update this before we get the pr in, that would be appreciated :)

examples/06-dc/read_plot_DC_data_wit_IO_class.py
```diff
@@ -0,0 +1,70 @@
...
+            ncontour=3,
+            ax=ax
+        )
+
+        plt.show()
+
+    # clean up
+    shutil.rmtree(os.path.expanduser('./test_url'))
+
+if __name__ == '__main__':
```
lheagy: if you use the notebook-styled example, this isn't necessary either

examples/14-utils/plot_2D_data.py
```diff
@@ -36,15 +36,30 @@ def run(plotIt=True):
         ax1 = plt.subplot(121)
         ax2 = plt.subplot(122)
         # Plot Real Ex (scalar)
-        cont1, ax1 = Utils.plot2Ddata(xyz, Ex.real, dataloc=True,
-                                      ax=ax1, contourOpts={"cmap": "viridis"})
+        cont1, ax1, cont1l = Utils.plot2Ddata(
+            xyz, Ex.real, dataloc=True,
+            ax=ax1, contourOpts={"cmap": "viridis"},
+            ncontour=5, level=True,
+            levelOpts={'colors': 'k', 'linestyles': 'dashed', 'linewidths': 1}
```
lheagy: is it `linewidths` or `linewidth`?

thast: I confirm `linewidths`: with some of those matplotlib functions, you can give them sequence as argument too. Same for `linestyles`. You can also check at https://matplotlib.org/api/_as_gen/matplotlib.pyplot.contour.html

examples/06-dc/read_plot_DC_data_wit_IO_class.py
thast: Will do

lheagy:
Thanks @thast: this looks good! Once it passes, we can bring it in. I will do a release from master as there are a couple other very small prs that can come in under the same release

PR #740, opened 2018-12-05
Do not install tests in site-packages

cgohlke:


lheagy:
👋 Hi @cgohlke, thanks for your interest in SimPEG. The change you suggest here actually tries to install the tests, which we don't want. Are you running into challenges getting started with SimPEG. If an issue prompted you to suggest this change, please [open an issue](https://github.com/simpeg/simpeg/issues/new) and we can try to help you out.

lheagy:
Sorry for my confusion there @cgohlke. Thanks for this fix!

lheagy:
looks like the travis fail is unrelated and has to do with the docs... I will check in to it

PR #742, opened 2018-12-08
bug-contour-plots

lheagy:
check if there is already a vmin, vmax in the contourOpts dict before adding them based on user-input clim


thast:
@lheagy : you might want to review #738 along with this PR.

PR #745, opened 2018-12-13
fix std in DataMisfit and possible bug in PlotUtils

thast:
Fix for Issue #746

- fixed `DataMisfit.std` that now take into consideration `survey.std`. So you are not forced to set `DataMisfit.W` manually anymore

-  Fixed: `Survey.MakeSyntheticData`, if not given a `std`, was overwriting `survey.std` even if already set.  Now it will use `survey.std` if not given a `std` argument.

Explanation:
`DataMisfit.std` was set to 5% by default, but to change it to `survey.std`, it expected a `None` value, thus it never gets updated...

- bonus: light improvement again of `Plot2DData` (reordering of `if` statement so specifying `contourOpts['vmin','vmax']`  without a `clim` works.

- bonus: Somehow PR #725 did not make it to improve `plot_misfit_curves` so I put it back here. Mainly, more option to plot $\phi_m$ or $\phi_small$  and $\phi_smooth$ independently, plus more options to save the figure.

### DIFF
--- a/SimPEG/DataMisfit.py
+++ b/SimPEG/DataMisfit.py
@@ -58,7 +58,7 @@ class l2_DataMisfit(BaseDataMisfit):
 
     """
 
-    std = 0.05  #: default standard deviation if not provided by survey
+    std = None  #: default standard deviation if not provided by survey
     eps = None  #: default floor
     eps_factor = 1e-5  #: factor to multiply by the norm of the data to create floor
 
@@ -66,11 +66,12 @@ class l2_DataMisfit(BaseDataMisfit):
         BaseDataMisfit.__init__(self, survey, **kwargs)
 
         if self.std is None:
-            if getattr(self.survey, 'std', None) is not None:
+            if getattr(self.survey, 'std', None) is None:
                 print(
                     'SimPEG.DataMisfit.l2_DataMisfit assigning default std '
                     'of 5%'
                 )
+                self.std = 0.05
             else:
                 self.std = self.survey.std
 
--- a/tests/base/test_DataMisfit.py
+++ b/tests/base/test_DataMisfit.py
@@ -44,8 +44,7 @@ class DataMisfitTest(unittest.TestCase):
         self.prob = prob
         self.dobs = dobs
         self.dmis = dmis
-        import SimPEG
-        print(SimPEG.__path__)
+
 
     def test_Wd_depreciation(self):
         with self.assertRaises(Exception):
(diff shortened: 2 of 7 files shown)


micmitch:
Your changes to DataMisfit.std look good to me.

thast:
@lheagy : any idea what happened with PR #725 ? It seems the changes made in `Directives` did not make it in Master even so we merged the PR (and I do not see anything in the history).
I'm curious and I do not understand how that could have happened.

lheagy:
Thanks @thast: don't worry about travis yet... since you updated a computationally-expensive example, we will need to send the updates to the google cloud storage in order for it to pass (so feel free to revert the changes). Would you mind including a small example of the previous (buggy) behaviour and how this fixes it? This would also make a good addition to the tests so that we don't break the expected behaviour in the future.

thast:
@lheagy : Thanks. I will revert back Travis. I do not know when I will be able to provide a test and example though but will do my best to get it quickly.

SimPEG/DataMisfit.py
```diff
@@ -58,19 +58,20 @@ class l2_DataMisfit(BaseDataMisfit):
...
+    std = None  #: default standard deviation if not provided by survey
     eps = None  #: default floor
     eps_factor = 1e-5  #: factor to multiply by the norm of the data to create floor
 
     def __init__(self, survey, **kwargs):
         BaseDataMisfit.__init__(self, survey, **kwargs)
 
         if self.std is None:
-            if getattr(self.survey, 'std', None) is not None:
+            if getattr(self.survey, 'std', None) is None:
```
lheagy: Thanks for catching this @thast. This won't yet catch the case if the user changes the standard deviation on the survey after the data misfit has been created. I think that is okay for now, but lets leave an issue open and we can work to address this with the simulation refactor.

thast:
Hi @lheagy , I chose to go a slightly different way, building the test on already existing tests for DataMisfit.

Also, I've found that the function MakeSyntheticData, if not given a std, would also overwrite survey.std even if already set. So I've fixed it to use it survey.std if not given a std argument.

tests/base/test_DataMisfit.py
```diff
@@ -41,6 +44,8 @@ def setUp(self):
         self.prob = prob
         self.dobs = dobs
         self.dmis = dmis
+        import SimPEG
```
lheagy: is this used in the test at all?

thast: ah good catch. I was experimenting something while writing it and forgot to remove some part of that experimentation....

thast: fixed

thast:
Good job with travis @lheagy!

PR #749, opened 2018-12-30
bug/plot2Ddata

lheagy:
update logic in plot2Ddata so that two vmin, vmax values cannot be provided to contourf (e.g. it was previously possible that an explicit vmin, vmax was provided by the user and that a vmin, vmax was chosen based on the colorbar)

### DIFF
--- a/SimPEG/Utils/PlotUtils.py
+++ b/SimPEG/Utils/PlotUtils.py
@@ -47,12 +47,14 @@ def plot2Ddata(
             if attr is None:
                 attr = contourOpts.pop(key)
             else:
-                if not np.isclose(contourOpts[key], vmin):
+                if not np.isclose(contourOpts[key], attr):
                     raise Exception(
                         "The values provided in the colorbar limit, clim {} "
                         "does not match the value of {} provided in the "
-                        "contourOpts {}. Only one value should be provided or "
-                        "the two values must be equal."
+                        "contourOpts: {}. Only one value should be provided or "
+                        "the two values must be equal.".format(
+                            attr, key, contourOpts[key]
+                        )
                     )
                 contourOpts.pop(key)
 
@@ -88,7 +90,7 @@ def plot2Ddata(
         vmin = DATA[dataselection].min() if vmin is None else vmin
         vmax = DATA[dataselection].max() if vmax is None else vmax
         if scale == "log":
-            if vmin < 0 or vmax < 0:
+            if vmin <= 0 or vmax <= 0:
                 raise Exception(
                     "All values must be strictly positive in order to use the log-scale"
                 )
@@ -144,7 +146,7 @@ def plot2Ddata(
         vmin = DATA[dataselection].min() if vmin is None else vmin
         vmax = DATA[dataselection].max() if vmax is None else vmax
         if scale == "log":
-            if vmin < 0 or vmax < 0:
+            if vmin <= 0 or vmax <= 0:
                 raise Exception(
                     "All values must be strictly positive in order to use the log-scale"
                 )
--- a/SimPEG/Utils/PlotUtils.py
+++ b/SimPEG/Utils/PlotUtils.py
@@ -134,12 +134,23 @@ def plot2Ddata(
             ~np.isnan(DATA),
             np.abs(DATA) != np.inf
             )
+
+        # set vmin, vmax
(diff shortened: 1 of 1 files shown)


lheagy:
Thanks @thast: I agree, we should be consistent. I am inclined to actually enforce that if both are provided that they are identical. This way, we avoid ambiguity of both use-cases happening. Then it doesn't matter which we prioritize (but we can pick `vmin`, `vmax` for consistency). Thoughts?

thast:
@lheagy : Instead of an error, maybe a more forgiving warning that inform the user which args was used and which one was ignored?

lheagy:
@thast: I guess my hesitation with that is that people tend to ignore warnings... However, if we give them a helpful error message, then it should be an easy fix for the user?

prisae:
That's an interesting and good point. Depending on how a user runs Python a Warning might not even be shown or visible (depends on the settings, warnings can be switched off).

thast:
@lheagy : that's a fair point. happy to go that direction.

SimPEG/Utils/PlotUtils.py
```diff
@@ -99,13 +108,8 @@ def plot2Ddata(
...
             DATA = np.ma.masked_array(DATA, mask=MASK)
 
-        if 'vmin' not in contourOpts.keys():
-            contourOpts['vmin'] = vmin
-        if 'vmax' not in contourOpts.keys():
-            contourOpts['vmax'] = vmax
-
         cont = ax.contourf(
-            X, Y, DATA, levels=levels,
+            X, Y, DATA, levels=levels, vmin=vmin, vmax=vmax,
```
thast: @lheagy : Will this work? If vmin, vmax were in ContourOpts, you will get an error because vmin and vmax will be specified twice. That was the reason for the section of code you deleted:
`if 'vmin' not in contourOpts.keys():
            contourOpts['vmin'] = vmin
        if 'vmax' not in contourOpts.keys():
            contourOpts['vmax'] = vmax`

lheagy: when you use `contourOpts.pop("vmin")` it will remove vmin from the `contourOpts` dictionary (see lines 48, 57). Does this make sense?

lheagy: image

thast: Thanks @lheagy , yes it makes sense :)

SimPEG/Utils/PlotUtils.py
```diff
@@ -58,34 +83,18 @@ def plot2Ddata(
...
-        ):
-            raise Exception(
-                """clim must be sctrictly positive in log scale"""
-            )
+
+        # set vmin, vmax if they are not already set
+        vmin = DATA[dataselection].min() if vmin is None else vmin
+        vmax = DATA[dataselection].max() if vmax is None else vmax
+        if scale == "log":
+            if vmin < 0 or vmax < 0:
```
thast: `if vmin <= 0 or vmax <= 0:`

SimPEG/Utils/PlotUtils.py
```diff
@@ -134,28 +138,19 @@ def plot2Ddata(
...
-            raise Exception(
-                """clim must be sctrictly positive in log scale"""
-            )
+
+
+        # set vmin, vmax
+        vmin = DATA[dataselection].min() if vmin is None else vmin
+        vmax = DATA[dataselection].max() if vmax is None else vmax
+        if scale == "log":
+            if vmin < 0 or vmax < 0:
```
thast: `if vmin < 0 or vmax < 0:`

SimPEG/Utils/PlotUtils.py
```diff
@@ -33,10 +33,35 @@ def plot2Ddata(
...
+        vmax = np.max(clim)
+
+    for key, attr in zip(["vmin", "vmax"], [vmin, vmax]):
+        if key in contourOpts.keys():
+            if attr is None:
+                attr = contourOpts.pop(key)
+            else:
+                if not np.isclose(contourOpts[key], vmin):
+                    raise Exception(
+                        "The values provided in the colorbar limit, clim {} "
```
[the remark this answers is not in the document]
lheagy: ah, right :) Thanks

PR #754, opened 2019-01-21
add random seed for VRM

thast:
add random seed for VRM to close #754


thast:
@dom, looks like an issue with the latest PF PR
image

thast:
rerun, slightly different but still failing at the same place
image

PR #755, opened 2019-01-27
numpy matlib import 

lheagy:
import matlib explicitly as in numpy 1.16, it isn't imported in the numpy namespace be default

image


lheagy:
Thanks for the speedy review @thast! I will merge as soon as travis is happy :)

PR #757, opened 2019-02-26
Fix/update pre conditioner

thast:
This formulation takes into account the regularization mapping into the preconditioner and remove the need to check for a specific wiresmap.


fourndo:
@lheagy Looks like we are having some issues with 2.7? Guessing you know about this?

PR #758, opened 2019-02-26
remove testing on python 2.7 from SimPEG

lheagy:
- soft removal of python 2.7: we don't explicitly break compatibility as in #731, but it is no longer tested on Python 2.7
- added a note in the docs saying we will not be working to ensure compatibility with python 2.7 anymore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: python
 
 python:
   - 3.6
-  - 2.7
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


thast:
Do we still need this section?
image

.travis.yml
```diff
@@ -52,12 +51,7 @@ before_install:
   - conda update --yes conda
 
 install:
-  - if [ ${TRAVIS_PYTHON_VERSION:0:1} == "2" ]; then
```
lheagy: @thast: we no longer need this because if was originally checking if we were installing python 2 or python 3. Now, we are only ever installing python 3, so we can follow those download instructions

thast: @lheagy : Sorry if I was not clear. My comment above (with screenshot, as I could not comment a line you did not modidy) concerns line 40 and below. It seems that in the "before_install" section we still have a Python 2.7 if statement.

lheagy: ah, good call - thanks @thast. I just pushed an update.

Issue #761, opened 2019-03-04
Inconsistent naming convention in PF mappings


thast:
There seems to be a naming discrepancy in the `PF` module and that causes me some trouble.
In the `Gravity` module, there are two aliases for `rhoMap`: [`mapping`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Gravity.py#L50) and [`mapPair`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Gravity.py#L191).
In the `Magnetics` module, there is one alias for `chiMap`: [ `modelMap`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Magnetics.py#L255).

@fourndo , what is your opinion on that? Is there any reason to have the two separate names `mapPair` and `modelMap`? If we can it would be nice to be consistent and settle for one or the other (no preference). Even further, can we just limit to just one alias in Gravity, like `mapping`?

thast:
solved in PR #762

PR #762, opened 2019-03-05
Fix mapping naming  convention in PF

thast:
This PR addresses issue #761 and follows PR #758 and PR #757 .

There was a naming discrepancy in the `PF` module and that causes me some trouble.
In the `Gravity` module, there are two aliases for `rhoMap`: [`mapping`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Gravity.py#L50) and [`mapPair`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Gravity.py#L191).
In the `Magnetics` module, there is one alias for `chiMap`: [ `modelMap`](https://github.com/simpeg/simpeg/blob/f4d03c44d7bd0369bf73b5b51131bd7f0c6a718d/SimPEG/PF/Magnetics.py#L255).

This PR settle for `modelMap` as only alias, consistent with the definition of the `LinearProblem` class in `Problem.py`
 
Additional: Fix mapping name in `LinearProblem` (from `mapping` (depreciated) to `modelMap`)


lheagy:
@fourndo: would you mind reviewing this?

fourndo:
@lheagy we need the power

SimPEG/Directives.py
```diff
@@ -974,29 +973,30 @@ def initialize(self):
...
+        m = self.invProb.model
 
         for reg in self.reg.objfcts:
             # Check if he has wire
             regDiag += reg.deriv2(m).diagonal()
         # Assumes that opt.JtJdiag has been updated or static
         diagA = self.opt.JtJdiag + self.invProb.beta*regDiag
-
-        PC = Utils.sdiag((diagA)**-1.)
+        diagA[diagA != 0] = diagA[diagA != 0] ** -1.
```
lheagy: should this be a threshold rather than identical to zero? cc @thast, @fourndo

PR #769, opened 2019-03-25
Replace `func versions` by `class Versions`

prisae:
# Replace `func versions` by `class Versions`

Turns out that there is a way to have html in a notebook and non-html in all the rest without even checking if you are in a notebook or not (thanks @banesullivan; https://github.com/simpeg/discretize/issues/142).

All that is needed is a `Versions`-class which has the right functions linked to `__repr__` and `_repr_html_`:
```
class Versions:

    def __init__(self, add_pckg=None, ncol=4):
        self.add_pckg = add_pckg
        self.ncol = ncol

    def __repr__(self):
        return versions_text(self.add_pckg)

    def _repr_html_(self):
        return versions_html(self.add_pckg, self.ncol)
```

By removing the old functions, respectively integrating the relevant parts into the new class, the whole version printing utility gets a lot simpler (however, it is not backwards compatible).

Before, you had to do
```
from SimPEG import versions
versions()
```
in Python, IPython etc, and in Jupyter
```
from SimPEG import versions
versions('HTML')
```

Now, you simply do
```
from SimPEG import Versions
Versions()
```
EVERYWHERE (note the uppercase `V`), and it will automatically print a html table in a notebook, and plain text everywhere else.


prisae:
(Improvement to #698.)

prisae:
This might be particularly useful for the SimPEG Example Gallery with the sphinx extension, as it now works the same for both IPython and Jupyter...

prisae:
You should use `Versions()` for everything.

Keeping `versions('HTML')`/`versions()` is only for backwards compatibility, so  these scripts don't break. We could add a depreciation warning if you want, or just remove it if you don't want.

prisae:
Btw, if you don't care about backwards compatibility then I could simplify the whole implementation a lot... And only leave `Versions()`.

prisae:
I think you're right @lheagy, and `versions` was probably not widely used so far. So I removed the old `versions`, and simplified everything a lot. Now it is just one class, `Versions`.

I did a quick grep and replaced the existing `versions` in the examples with `Versions`.

lheagy:
Cool, thanks @prisae! Agreed, at this point, the breaking of backwards compatibility isn't too painful, so just removing it and going forward with one `Version` will hopefully simplify our lives down the road. The travis fail is just because the docs build is too big - I can address that this evening.

prisae:
Great, thanks!

prisae:
Restarted the failed test once, but failed again. Not sure what it is about, but as you mentioned, nothing to do with the actual PR.

prisae:
I updated the initial comment to reflect the non-backwards compatibility of the PR.

prisae:
Are these issues resolved @lheagy?

prisae:
This is a small PR, yet also backwards incompatible. Would be great if it gets in before https://github.com/simpeg/simpeg/pull/786

I think this one is ready to go, the failing testes are due to the google sdk, if I see that right.

lheagy:
Thanks for reviving this @prisae! (and sorry for dropping balls): I just merged the latest master and will update the docs so we can get this in. 

P.S. You should add yourself to [AUTHORS.rst](https://github.com/simpeg/simpeg/blob/master/AUTHORS.rst). Please feel free to add that to this pr 😄

PR #772, opened 2019-03-28
Fix for issue 771

Rockpointgeo:
Hopefully I did this correctly.


jcapriot:
Closing this PR, considering it's age, and the fact that as written, this will create a sparse matrix that is actually dense, with  a large size in general. It'll be best to adapt the solver used to either be the sparse version or the dense version depending on the type of `H` Also, If you want to use a `LinearOperator`, you should use the `InexactGaussNewton` method which uses an iterative solver to find the search direction.

PR #781, opened 2019-05-24
Add SG filename pattern

banesullivan:
This adds a filename pattern to search for when the docs build the gallery examples

**TODO: I'm awful with regex expressions so someone more familiar should double check this work.**

Most notably, [this example](https://docs.simpeg.xyz/content/examples/06-dc/read_plot_DC_data_with_IO_class.html) is not running and these changes should fix that.

**NOTE:** This changes still exclude [this example](https://docs.simpeg.xyz/content/examples/11-seis/seis_tomo_joint_with_volume.html) from running so it might need to be renamed?


banesullivan:
Note the DC IO example needs `pandas` so I added it to `requirements_dev.txt`

lheagy:
Thanks @banesullivan! Sorry for dropping the ball on this. I will take a look this week

jcapriot:
So I think this would be better fixed baby just renaming those two example files to the `plot_` format as those are the only two that are done like that, so the regex might not be needed. This something we will take care of quickly with the simulation pr! Again thanks for bringing this up @banesullivan, and sorry it's taken so long to address this.

banesullivan:
Sounds good to me @jcapriot! Closing this in favor of that other PR then...

PR #784, opened 2019-06-04
light refactor in DC receivers

lheagy:
turn IntTrapezoidal into a function that can be reused by both pole and dipole classes rather than duplicating code. 

@micmitch: would you mind taking a quick look at this pr? 

If you take a look at the code 
image

and please leave any comments or questions here
image

Thanks!


lheagy:
Thanks @micmitch!

PR #785, opened 2019-06-06
Tutorials for models and mapping

dccowan:
In this pull request we have tutorials for how to models in different ways and how to use the different kinds of mappings from the model to the mesh. We have this for:
- tensor meshes
- cylindrical meshes
- tree meshes
There is also a small refactoring of the website but I did not spend time parsing this. I merely wanted a well-defined place for the tutorials.


sarahgarre:
**Models and mapping**
clear
**Tensor meshes**
I like the fact that you show a whole bunch of different ways to use this . Pretty much a cookbook of what one might need afterwards to build there own models and mappings. Well done.
_Models with arbitrary shape_
A small thing again is beware of using too many unknown functions without too much explanation here. E.g. Utils.ModelBuilder.PolygonInd() and alike.
_Parameterized block model_
cool! 
This part could use some extra comment text. I know I could look in the API, but a minimum of basic info the first time it appears in the tutorial would be beneficial for the understanding... What do epsilon and p do?
`paramMap = Maps.parmetricBlock(...)`
_Wire maps_
Very nice. This was completely new to me and I think I understood. 

**Cylindrical meshes**
**Tree meshes**
I went through the same for the other mesh types. Looks great. 👍

sarahgarre:
**Models and mapping**

dccowan:
The docs are failing. I believe it is due this intersphinx inventory stuff. I didn't set this up so I'm not exactly sure how to fix it. @lheagy any suggestions??

image

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+    
+
+"""
+
+#########################################################################
+# Import modules
+# --------------
+#
+
+from SimPEG import Mesh
```
lheagy: here, would you mind using `discretize` rather than importing `Mesh`. This will make it a bit more clear for users where they need to look for documentation if they are confused about how to construct a `TensorMesh`, for example 
```suggestion
import discretize
```

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+
+#############################################
+# Defining the mesh
+# -----------------
+#
+# Here, we create the tensor mesh that will be used for all examples.
+#
+
+
+def makeExampleMesh():
```
lheagy: might be worth making `dh`, `npad`, `ncore` inputs to the function so that if the use wants to play around a bit it is easy for them to change the mesh size. 

Also, to be more in-line with [python community practices](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names), lets use lowercase and underscores for names, e.g. `make_example_mesh` 
```suggestion
def make_example_mesh(dh=5, npad=5, ncore=20):
```

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+# In this example we generate a half-space model. Since air cells remain
+# constant during geophysical inversion, the number of model values we define
+# should be equal to the number of cells lying below the surface. Here, we
+# define the model (*mod* ) as well as the mapping (*modMap* ) that goes from
+# the model-space to the entire mesh.
+#
+
+mesh = makeExampleMesh()
+
+m_val = 100.
```
lheagy: `m_halfspace` (`val` is a bit non-descript)

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+
+# Find cells below topography and define mapping
+m_air = 0.
+actv = mesh.gridCC[:, 2] < 0.
```
lheagy: here, can we try to stick with the convention of `ind_active` to denote the active indices? Internally, we are starting to shift to this naming convention

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+modMap = Maps.InjectActiveCells(mesh, actv, m_air)
+
+# Define the model
+mod = m_val*np.ones(actv.sum())
+
+# We can plot a slice of the model at Y=-2.5
+Fig = plt.figure(figsize=(5, 5))
```
lheagy: to simplify things a bit, do you mind using the `subplots` api and lower-case for `fig`, `ax`
```suggestion
fig, ax = plt.subplots(1, 1, figsize=(5, 5))
```

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+Ax = Fig.add_subplot(111)
+mesh.plotSlice(modMap*mod, normal='Y', ax=Ax, ind=int(mesh.nCy/2), grid=True)
+Ax.set_title('Model slice at y = -2.5 m')
+plt.show()
```
lheagy: I don't think the plt.show() is needed for the examples

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+# Define surface topography as an (N, 3) np.array. You could also load a file
+# containing the xyz points
+[xx, yy] = np.meshgrid(mesh.vectorNx, mesh.vectorNy)
+zz = -3*np.exp((xx**2 + yy**2) / 75**2) + 40.
+topo = np.c_[Utils.mkvc(xx), Utils.mkvc(yy), Utils.mkvc(zz)]
+
+# Find cells below topography and define mapping
+m_air = 0.
+actv = Utils.surface2ind_topo(mesh, topo, 'N')
+modMap = Maps.InjectActiveCells(mesh, actv, m_air)
```
lheagy: `mapping` or `model_map` (lower-case + underscores for variable names in python: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+
+# Define the model on subsurface cells
+mod = m_back*np.ones(actv.sum())
+k_dyke = (mesh.gridCC[actv, 0] > 20.) & (mesh.gridCC[actv, 0] < 40.)
```
lheagy: instead of `k` can we go with `ind`

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+mod[k_dyke] = m_dyke
+k_block = ((mesh.gridCC[actv, 0] > -40.) & (mesh.gridCC[actv, 0] < -10.) &
```
lheagy: sorry, this is me being picky but for long statements, would you mind opening the bracket and then entering to a new line (helps with readability and consistency - the tabs are then consistent along the entire program rather than dancing around). 
```          
k_block = (
    (mesh.gridCC[actv, 0] > -40.) & (mesh.gridCC[actv, 0] < -10.) &
    (mesh.gridCC[actv, 1] > -30.) & (mesh.gridCC[actv, 1] < 30.) &
    (mesh.gridCC[actv, 2] > -40.) & (mesh.gridCC[actv, 2] < 0.)
)
```

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+           (mesh.gridCC[actv, 1] > -30.) & (mesh.gridCC[actv, 1] < 30.) &
+           (mesh.gridCC[actv, 2] > -40.) & (mesh.gridCC[actv, 2] < 0.)
+           )
+mod[k_block] = m_block
+
+# Plot
+Fig = plt.figure(figsize=(5, 5))
+Ax = Fig.add_subplot(111)
```
lheagy: same `fig, ax = plt.subplots()` comment as above

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+# Set active cells and define unit values
+m_air = 0.
+actv = Utils.surface2ind_topo(mesh, topo, 'N')
+modMap = Maps.InjectActiveCells(mesh, actv, m_air)
+
+# Define model for cells under the surface topography
+mod = m_back*np.ones(actv.sum())
+
+# Add a sphere
+sphInd = Utils.ModelBuilder.getIndicesSphere(np.r_[-25., 0., -15.],
```
lheagy: Rather than using the `ModelBuilder` code - lets just stick with mappings for now. I think the model builder stuff might eventually get folded into `maps` or moved elsewhere, so mappings are more important to highlight in current tutorials

tutorials/models_mapping/plot_1_tensor_models.py
```diff
@@ -0,0 +1,333 @@
...
+"""
+
+#########################################################################
+# Import modules
+# --------------
+#
+
+from SimPEG import Mesh
+from SimPEG import Utils
+from SimPEG import Maps
```
lheagy: To simplify the transition over to the `simulation` changes, would you mind updating the imports to be 
```python
from SimPEG import Utils as utils
from SimPEG import Maps as maps
```

and use `utils.whatever` and `maps.whatever`

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: sure, ill change this

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: calling each mesh type in the form 'from discretize import TensorMesh'

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: update unit names

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: changed all to ind_active

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: done lower case for standard convention

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: things change to ind_ instead of k_

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: fixed these according to pep8

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: I disagree. I made a conscious decision to setup the tutorials this way with the current status of the SimPEG package. I wanted the users to correctly differentiate the model from the physical property distribution on a mesh. And I wanted to clearly define that a mapping could be defined that would go from the model space to the mesh. Feedback from @sarahgarre  suggests that my tutorial was successful in doing this.

If the way in which functionality like 'ModelBuilder.getIndicesSphere' is called changes in the future, we can always change it in a patch.

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: Whether or not I do this, I will still need to make changes after the simulation class is merged. I'll just wait until then.

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: I've changed things to pep8

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: changed this

tutorials/models_mapping/plot_1_tensor_models.py
dccowan: Unless there is good reason why it should not be there, I would just as rather keep it.

tutorials/models_mapping/plot_1_tensor_models.py
lheagy: Hi Dev, I appreciate that this is the current status. I am trying to advise on how we want to be encouraging users to use SimPEG to avoid frustration as we continue to improve the code. 

In this case, the model builder code in its current form is likely to change, whereas the maps are not. So we want to encourage use of maps. (via the `Maps` module rather that the `Utils.ModelBuilder`). 

If you feel strongly that a `Utils.ModelBuilder` example tutorial should exist, then would it be okay with you if we moved it to a separate tutorial? That way we would have one which uses only maps, and one which uses the model builder code. Separating the two will hopefully clarify that `Maps` and `Utils.ModelBuilder` are not completely interchangeable.

tutorials/models_mapping/plot_1_tensor_models.py
lheagy: okay, I am just trying to help simplify things.

tutorials/models_mapping/plot_1_tensor_models.py
lheagy: This isn't necessary, and adding code that isn't needed just increases complexity. I realize it is only one line, but we don't need it for the plot to show up.

tutorials/models_mapping/plot_1_tensor_models.py
lheagy: It doesn't look like this has been updated. Have you pushed the code?  Similar with the other comments you resolved below.

lheagy:
:wave: @dccowan: it looks like nbsphinx needs to be added to the [dev requirements](https://github.com/simpeg/simpeg/blob/tutorials_models_clean/requirements_dev.txt) here

Issue #751, opened 2019-01-08
run black on SimPEG


lheagy:
similar to simpeg/discretize#132, I think we should be using [black](https://github.com/ambv/black) on SimPEG to format the code. This will help with line-spacing and should simplify the diffs on pull-requests. It also means we do not need to format the code by hand.

prisae:
`black` uses by default a line length of 88.  They say it makes code shorter. Most likely it does, as the line length is longer. However, if you look at code through a terminal or an IPython terminal or vim or emacs etc (which I think still many do), then a line length of 88 is simply horrible.

I am fine with `black`, but I'd suggest to set `--line-length 80` to restore the "normal" behavior.

prisae:
See https://github.com/fatiando/verde/issues/177 for examples and more reasoning why (I think/argue) you should, please please, set `max-line-length` to 80.

prisae:
It is archived, but I still find it hilarious: https://github.com/kenneth-reitz/white
It is basically a wrapper for `$ black --line-length 79`

jcapriot:
This was closed with 0.14.0!

Issue #741, opened 2018-12-07
Module name


prisae:
This is just meant as a discussion. But something that probably would have to be decided before the release of `v1.0`.

What about changing the module name from `SimPEG` to `simpeg`?

- Pro's:
  - Following the convention of most packages, in that the package and module names are identical.
  - Following pep8 guidelines, in which packages and modules should have all-lowercase names.
  - Less confusing. At the moment what a potential user would do (and what happend to me at the beginning several times) is

    1. `pip install simpeg`
    2. Fire-up an IPython shell or a notebook
    3. `import simpeg`

    with the result that it throws an error stating that `simpeg` is not installed. One has then to check the documentation or the examples to find out that the module to import is called `SimPEG`, not `simpeg`.

- Con's:
  Maybe too late for a change, too many scripts and notebooks exist which import `SimPEG`.

Just thought I initiate once the discussion. Maybe it was discussed before, I don't know.

lheagy:
I do agree that this would be a better name.

... but I am afraid of how painful it would be to change at this point. Is there an easy way to depreciate the name? (e.g. allow backwards compatibility for a time?). I don't really think so at the module level??

prisae:
I also think it is not possible on module level, and therefore not very feasibly. I just thought it might be worth to have the discussion before `v1.0`. Because now you could at least still argue "it is not `v1.0` yet, so API changes have to be expected" :wink: 

Also in light that at some point there was discussion to make many of the variable names more pep8ish. So if this would result in a one-off, heavy backwards-incompatible release then it might be worth taking this into account as well.

lheagy:
that is true, it could be lumped under a major refactor

prisae:
I think there is a possibility to deliver a new `simpeg` package and at the same time a `SimPEG` package which will only print information about the name-change. The important bit would be the line 
```
    packages=['simpeg`, `SimPEG'],
```
in `setup.py`. Then, `pip install simpeg` or `pip install SimPEG` (`pip` is case-insensitive, I believe) would install both packages, `simpeg` and `SimPEG`. The directory `SimPEG` would then only contain a `__init__.py`-file. In there, we could put print-statements about all the changes. E.g.: `SimPEG->simpeg`, `EM->electromagnetics`, etc etc.

In this case, if someone does `import simpeg` all is good and as it used to be before with `SimPEG`. However, if someone does `import SimPEG`, it will print all the info we put into `__init__.py`.

[...]

PR #786, opened 2019-06-17
Simulation class refactor

lheagy:
This is a tracking pr for progress on the simulation class. It is very much in-progress at the moment, but I wanted to put it on people's radars so that we can iterate together. So far, the major changes include

# Major changes

## No more `Problem`'s!: `Simulation` class and updates to the `Survey`
Refactoring of the Problem and Survey class into a `Simulation` and a `survey`. Now the survey is much more lightweight and really only tracks sources, receivers and knows how to construct a right hand side. See issue #562 for more discussion 
- The `survey` no longer has a `dpred` method;  this is on the `simulation` now. 
- `survey` no longer has an `eval` method
- Pairing is no longer needed as the `survey` is passed in the instantiation of the `simulation`. 

## Fields
- Fields are now instantiated with a simulation rather than a `survey` and a `mesh`

## Simplify the data class
Simplifying the data class to make it less magical. Now the data class keeps track of `dobs`, `standard_deviation` and `noise_floor`. (previously the survey knew about standard deviation, and this led to confusion about where the data misfit should get weights from - now the `data` object is the source of this. The user can still over-ride this and set their own `W` on the data misfit if desired. See #202 for more discussion
- Data class now has stored properties: `dobs`, `standard_deviation` and `noise_floor`. `uncertainty` can be calculated from these
- still the option to treat the data class as a clever dictionary, but also to simply call `data.dobs` to get the data

## Sources and Receivers
- no more `rxType` on receivers 

## conventions
- module names: now the base python files are named with lower-case letters

## general cleanup
- delete code that duplicated inherited properties / methods
- import `discretize` internally rather than `Mesh`. Note that you can still use `from SimPEG import mesh` to import discretize

# Module specific changes
## StaticUtils
- `writeUBC_DCobs`,  take `data` rather than a `survey` as the second input. 
- `apparent_resistivity` and `plot_pseudoSection` take `data` as a first input rather than a survey
-  `readUBC_DC2Dpre`, `readUBC_DC3Dobs` now just return a data object rather than a dictionary with the survey. 

## Flow
- `RichardsSurvey` --> `Survey`
- `PressureRx` --> `receiver.Pressure`, `SaturationRx` --> `receiver.Saturation`

## VRM
- it is now imported from the electromagnetics module `from SimPEG.electromagnetics import viscous_remanent_magnetization as vrm`

# Todos
- [x] each of the physics modules still needs updating. I have fixed imports in the EM module, but not yet updated the classes to be consistent with the simulation refactor
- [x] update examples
- [x] provide documentation and tips on how to migrate code to the new framework 
- [ ] recommend that versions of SimPEG are pinned on `simpeg-research` repos that have already been published


SimPEG/EM/Static/SIP/SurveySIP.py
```diff
@@ -40,75 +35,75 @@ def dpred(self, m, f=None):
...
-            'v must have the correct number of data.'
-        )
-        indBot, indTop = 0, 0
-        for src in self.survey.srcList:
-            for rx in src.rxList:
-                for t in rx.times:
-                    indTop += rx.nRx
-                    self[src, rx, t] = v[indBot:indTop]
-                    indBot += rx.nRx
+# class Data(BaseData):
```
lheagy: @sgkang: This looks like it was copied verbatim from the base data class, so I am planning to remove it and simply import from the data module. Any concerns with that? Were there any modifications you made?

SimPEG/EM/NSEM/source.py
```diff
@@ -0,0 +1,250 @@
...
+from ...utils import mkvc
+from ..FDEM.source import BaseFDEMSrc
+from ..Utils import omega
+from .Utils.sourceUtils import homo1DModelSource
+
+#################
+###   Sources ###
+#################
+
+# class BaseFDEMSrc(FDEMBaseSrc):
```
lheagy: @grosenkj: It doesn't look to me that this class is adding or changing any functionality from the BaseFDEM source, so I plan to remove it, and instead directly inherit from FDEM. Any concerns with that?

SimPEG/EM/NSEM/survey.py
```diff
@@ -1,86 +1,84 @@
...
-        :retype: SimPEG.EM.NSEM.Data
-        :return: NSEM Data object
-        """
-        data = Data(self)
-        for src in self.srcList:
-            sys.stdout.flush()
-            for rx in src.rxList:
-                data[src, rx] = rx.eval(src, self.mesh, f)
-        return data
+#     def __init__(self, srcList, **kwargs):
```
lheagy: @grosenkj: This functionality is already in the base class, so I plan to remove it. Let me know if you have any concerns!

SimPEG/EM/Static/SIP/SurveySIP.py
sgkang: @lheagy  It is actually bit different. I am taking care of time (`t`) as well here, 
whereas base Data class do not.

lheagy:
Question for folks... I am trying to clean up module naming along the way, e.g. `Simulation.py` --> `simulation.py`. This is fairly straightforward for the base classes, as we mostly import the classes from those modules, so I don't anticipate too many breaking changes there. It is also straightforward to allow the imports from the old api. Currently, I am achieving this through the `old_api.py` doc at the base. 

Where I am very curious to get input is on renaming physics modules. What do folks think about the following 
- `EM` --> `electromagnetics`
    - `FDEM` --> `fdem` (similar for TDEM, NSEM, etc)
    - `DC` --> `dc, `IP` --> `ip`, `SIP` --> `sip`
    - `VRM` --> `vrm`
- `FLOW` --> `flow`
- `SEIS` --> `seismic`
- `PF` --> `potential_fields`
    - `Magnetics` --> `magnetics`
    - `Gravity` --> `gravity`

ngodber:
I like it. I think its more pythonic in that it complies with PEP 8.

 My only other comment is that I personally prefer explicit names over acronyms. So to me electromagnetics.frequency_domain is far prettier than electromagnetics.FDEM, or worse EM.FDEM. 

So I would suggest FDEM -> frequency_domain, time_domain etc. I just think it improves readability all round.

lheagy:
Thanks @ngodber! I was debating about that, `fdem` or `frequency_domain`. Although a bit more verbose, I do agree that is is much more clear (esp if we want to cater to non-specialists or folks newer to the geophysics community, avoiding shorthand is a good idea). And we can always `import electromagentics.frequency_domain as fdem` if it feels too verbose in code. So perhaps the second layer of modules is 
- `FDEM` --> `frequency_domain`
- `TDEM` --> `time_domain`
- `DC` --> `dc_resistivity` or `direct_current_resistivity` ?
- `IP` --> `induced_polarization`
- `SIP` --> `spectral_induced_polarization` 
- `VRM` --> `viscous_remenant_magnetization`

ngodber:
FWIW I think 'resistivity' is probably sufficiently descriptive. I'm not sure if simpeg has modules relating to MMR but even if it does, common parlance is resistivity so I don't think anyone would be confused. 🤷‍♂

lheagy:
Within each method, I am also trying to use a common set of modules. I am starting with EM, so at the moment, if we look at the frequency domain code the structure is now
- `FieldsFDEM` --> `fields`
- `RxFDEM` --> `receiver` (maybe should be plural: `receivers`?)
- `ProblemFDEM` --> `simulation`
- `SrcFDEM` --> `source` (`sources`?)
- `SurveyFDEM` --> `survey`

And similarly for other modules. Does that make sense? Thoughts on singular / plural for Src / Rx ?

ngodber:
Yes that makes a lot of sense to me. Whether to go plural or singular wrt to source or receivers I think comes down to the init statement. As far as I know loc in source and receiver takes a 3 node numpy array, ergo to generate multiple source and receivers you need to loop. If thats correct than the singular source/receiver makes the most sense to me.

prisae:
Linking this https://github.com/simpeg/simpeg/issues/741 , but probably this would be too much of a backwards-incompatible change...

ngodber:
a migration script might be necessary just to capture all the refactors so that older scripts can be quickly converted with a quick run.

lheagy:
Thanks @prisae: this could be a good point to think about making a switch. I will bring it up at the SimPEG meeting this week. It would be a fairly major change, so I want to make sure that we have buy-in from folks before pursuing it. 

@ngodber: We will be sure to provide migration docs, and a script will take some thought. I don't think we will be able to easily write something that covers all cases, but we might be able to get 80-90% of the way, which would still be helpful

SimPEG/electromagnetics/utils/AnalyticUtils.py
```diff
@@ -35,6 +36,12 @@ def MagneticDipoleVectorPotential(srcLoc, obsLoc, component, moment=1.,
     """
     # TODO: break this out!
 
+    warnings.warn(
```
lheagy: :wave: Any concerns (in particular, @sgkang) if we actually deprecate this code now rather than throwing the warning?  The mag dipole and loop code is now all in geoana

.gitignore
```diff
@@ -41,8 +41,17 @@ docs/_build/
 docs/warnings.txt
 .DS_Store
 docs/content/examples/*
-docs/content/tutorials/models_mapping*
+docs/content/tutorials/*
+tutorials/temporary/sensitivity.zarr
```
lheagy: should we just add a blanket `*.zarr` we shouldn't have zarr files in this repo anyways?

lheagy:
Here are a few issues that likely should be addressed in this pr that are still open: 

- [ ] #846: naming around `standard_deviation` and `uncertainty` 
- [ ] #838: serialization and deserialization of the survey class + tests

And a few issues we might want to consider including with this refactor (we can chat through these on slack / at the next meeting):  

**Formatting**
- [ ] #751: run black on SimPEG
- [ ] #741: SimPEG --> simpeg ? 

**Documentation**
- [ ] #858 : adding references to the docs ( and related #857)
- [ ] #859: more explicit documentation around dimensions of `v` for `Jtvec`, `Jvec`
- [ ] #627: units in FDEM docs 

I started a project to track these: https://github.com/simpeg/simpeg/projects/7. Please feel free to add / rearrange!

lheagy:
One question mostly for @jcapriot and @dccowan: for the `examples` in SimPEG, do you think we should remove the `def run():` pattern and reformat to the style of the tutorials as a part of this pr?

rowanc1:
🎉 🚀

PR #788, opened 2019-06-18
MInor edits to MT Rx class and plotUtils

JKutt:
First pull request for MT code

### DIFF
--- a/SimPEG/EM/NSEM/RxNSEM.py
+++ b/SimPEG/EM/NSEM/RxNSEM.py
@@ -66,14 +66,20 @@ class BaseRxNSEM_Point(SimPEG.Survey.BaseRx):
         if self.locs.ndim == 3:
             loc = self.locs[:, :, 0]
         else:
-            loc = self.locs[:, 3:]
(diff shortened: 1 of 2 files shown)


SimPEG/EM/NSEM/RxNSEM.py
```diff
@@ -66,14 +66,14 @@ def _locs_e(self):
         if self.locs.ndim == 3:
             loc = self.locs[:, :, 0]
         else:
```
JKutt: I assume I should make this a condition that if locations are provided for E and B instead making it a default to else?

SimPEG/EM/NSEM/RxNSEM.py
```diff
@@ -66,14 +66,14 @@ def _locs_e(self):
...
-            loc = self.locs
+            loc = self.locs[:, 3:]
         return loc
 
     def _locs_b(self):
         if self.locs.ndim == 3:
             loc = self.locs[:, :, 1]
         else:
```
JKutt: same here

PR #791, opened 2019-06-25
FDEM line source

micmitch:
The addition of a LineSource function for the FDEM survey which integrates over an arbitrary source wirepath to determine correctly define the source term on the edges of an octree mesh. This was originally modelled after the same code for the tensor mesh. New functionally was required find the associated cells and edges in the octree mesh. As a result this code is dependent on the discretize feat/tree_functions branch (see PR https://github.com/simpeg/discretize/pull/177).  Many thanks to @jcapriot for all his help with this. Testing has been done to insure that the discretization of the source term is consistent between octree and tensor meshes with identical core regions.


jcapriot:
I still do not think hanging edges are taken care of properly here. Some care would certainly have to be taken when the wire path crosses different cell levels, and I don't think the way that is implemented here is quite correct (was just a first guess).

In the meantime we could add a check to at least see if all of the returned cells are of the same level, to ensure correctness.

micmitch:
I agree with @jcapriot  that a cell level test would probably be a good idea with the current implementation.

PR #792, opened 2019-07-01
Use scooby for Versions

prisae:
Use `scooby` internally for `Versions`. Now that scooby exists (thanks @banesullivan ) we should use it to simplify our lives. This removes lots of code from SimPEG, and will make maintenance easier down the road.

It is a bit a shame this comes only a few days after https://github.com/simpeg/simpeg/releases/tag/0.11.5 - but the development of scooby came a bit unexpected and fast.

However, there is absolutely NO change to the API / for the user. Everything remains the same.

In the end, the work we did with `Versions` in `SimPEG` (and in `empymod`/`emg3d`) all led to `scooby`, so it was well invested time ;-)


prisae:
Btw, what do you think about having the same in `discretize` too? Now it would just mean a few lines of code...

banesullivan:
Or what if it were only in `discretize` and SimPEG called `discretize`'s function?

prisae:
I changed ```**Parameters**``` to
```
Parameters
--------------
```
I assume SimPEG is moving to numpydoc-style too, right?

prisae:
> Or what if it were only in `discretize` (and SimPEG called `discretize`'s function for backwards compatibility)?

No, I don't think that is good. SimPEG is the main package. So it should have it. It is where everything comes together. discretize could have it too, as it can be used standalone, but not necessary. Because always when you use discretize then you will use something else too; e.g. SimPEG, or emg3d, or similar. And those tools should have it, IMO.

banesullivan:
Thanks make sense so that SimPEG could specify specific packages and hardware that are relevant but that might not be relevant to other packages leveraging `discretize`.

prisae:
We might think of renaming it from `Versions` to `Report`. This would be backwards incompatible (again), so it would have to be before #786 .

The reason is that @banesullivan and I think/hope/expect that `scooby` will become more widely adapted, and in `scooby` it is called `Report` (in the end, it does a bit more than only showing the versions). So it might become common that modules have a `module.Report()`-instance, and then it might be useful if SimPEG follows that.

Just thinking out loud, ideas/opinions welcome.

setup.py
```diff
@@ -43,6 +43,7 @@
         'vectormath>=0.2.0',
         'discretize>=0.4.0',
         'geoana>=0.0.4'
+        'scooby>=0.3.0',
```
lheagy: What do you think of making this a soft dependency as we did for matplotlib in discretize? Since it isn't strictly necessary for running simulations or inversions with SimPEG, I would be inclined to make it a soft rather than a hard dependency

prisae:
Yes, I was thinking about making it a soft dependency just like matplotlib in discretize. I opted not to, as scooby is tiny compared to matplotlib.

I'll implement that! If not today than later in the week.

prisae:
A very quick implementation, not really tested yet...

banesullivan:
My input, FWIW: while yes, we are still house-training `scooby` as it's just a puppy of a project and you might not want users to be forced to install something like that, I went ahead and made `scooby` a hard dependency of PyVista and plan to make it so with other projects I work on.

**Why?**

Lately, PyVista has been gaining a lot of users that don't have a strong Python background and they've been happy to report bugs/generate issues but it was becoming an challenge to work with numerous users to report everything we needed to know about the environment and the computer. My solution: have `scooby` as a hard dependency so that as soon as a user reports a bug, we have them simply run `pyvista.Report()` and **it'll just work** (as having non-Python-aficionados install a new library was too cumbersome at times and they might not follow up on the issue).

So I guess `scooby` started as a reproducibility tool and has morphed into more of a bug-reporting tool for how I use it at least.

prisae:
@banesullivan thanks for the input. @lheagy I can leave the soft dependency in or make it hard again, depending on what you folks want. Maybe you can raise it today at the meeting?

Either way: The `requires`-function could be useful for other soft dependencies too, so it should not live in `printinfo`. I moved it into `codeutils` and renamed it to `requires_module`, as a `requires` exists already in the same file for variables.

Also, `printinfo` (the old `Version`), was quite large, so a separate file was justified. Now, with `scooby`, it is a tiny thing. So I moved it into `codeutils` too, removing the file `printinfo.py` altogether.

prisae:
There is actually a much simpler way in this case than `requires_module`:

```
# scooby is a soft dependency for SimPEG
try:
    from scooby import Report
except ImportError:
    class Report:
        def __init__(self, additional, core, optional, ncol, text_width, sort):
            print('\n  *ERROR*: `SimPEG.Report` requires `scooby`.'
                  '\n           Install it via `pip install scooby`.\n')
```
This is all that is needed to make it a soft dependency. And it prints the useful info.

banesullivan:
@prisae - I like that soft dependency implementation a lot! Maybe we should tag that for `scooby`'s docs for whenever we get around to making those

prisae:
Yes, like this it is very minimal. We could even put it on the `README` of `scooby`, what you think? Or later in the docs...

banesullivan:
Maybe just the docs - maybe make an issue over there so we don't forget about this?

prisae:
Done: https://github.com/banesullivan/scooby/issues/32

prisae:
I believe the fail is ones again related to the Google Cloud SDK thingy...

lheagy:
Thanks for pushing on this! I am at a conference this week and am again behind on things... but I will jump back in in a week or two

prisae:
Two warnings from the Travis-log (run 10 - docs) which might need some action?
```
WARNING: intersphinx inventory 'http://discretize.simpeg.xyz/en/latest/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: http://discretize.simpeg.xyz/en/latest/objects.inv
generating gallery...
WARNING: Duplicate file name(s) found. Having duplicate file names will break some links. List of files: ['../examples/04-grav/plot_inversion_linear.py']
```

prisae:
@jcapriot , could this PR (or the branch `scooby`) be merged into the `simulation` branch?

jcapriot:
I'm going to switch the merge request to the simulation branch, and merge those changes into here to check, but I don't think this will cause any issues.

A few things, cython is not technically a requirement for SimPEG, it is a requirement of discretize. The old `Versions` was the only place cython was ever actually used in SimPEG. I'll be removing it from SimPEG's setup.py shortly.

prisae:
Great, thanks @jcapriot !

PR #802, opened 2019-08-07
Simulation jk

JKutt:
Start of Dask Parallelization over frequencies and forming J for DCIP. Still a bit of work; first plan is to store fields to Zarr for each frequency then have the fields object handle populating.


JKutt:
@lheagy this is the MT and Tipper example script I moved over to simulation that is giving me "nan" for phi_d and f parameters.
[plot_foward_MTandTipper3D.zip](https://github.com/simpeg/simpeg/files/3478302/plot_foward_MTandTipper3D.zip)

JKutt:
Hi @lheagy, the tests still seem to be failing in a few places. Most are single tests in the bunch for each tdem, sies, PF, fdem. The one from tdem states: `test_MagDipole_bPrimaryMu0_e (forward.test_FDEM_sources.TestSimpleSourcePropertiesTensor) ... ERROR` Any guidance on how I could go about fixing these?

JKutt:
@fourndo the following is what I was referring to at the last meeting about the IP `getJ`. I guess what I'm getting at is should the `getJ()` on the IP side check if J is available via zarr from when the DC was run?

https://github.com/simpeg/simpeg/blob/aa868f7760be90d3d0c0e50e8992dd2b48934b02/SimPEG/electromagnetics/static/induced_polarization/simulation.py#L84

SimPEG/electromagnetics/natural_source/simulation.py
```diff
@@ -1,584 +1,584 @@
...
-import dask.array as da
-import multiprocessing
-from scipy.constants import mu_0
-
-try:
-    from pymatsolver import Pardiso as SimpegSolver
-except ImportError:
-    from SimPEG import SolverLU as SimpegSolver
-
-from ...utils import mkvc, setKwargs
```
nishantra4: I have installed "utils" from "https://pypi.org/project/python-utils/" After line 15 getting an error of "ValueError: attempted relative import beyond top-level package". I checked utils package it is not having any mkvc named file.
mkvc3 2019-10-09 10-18-12

lheagy: @nishantra, in this case, `utils` is a module of SimPEG, so the import is relative. You shouldn't need `python-utils` to run SimPEG. Does that make sense?

SimPEG/directives.py
```diff
@@ -106,7 +106,7 @@ def prob(self):
            Assuming that dmisfit is always a ComboObjectiveFunction,
            return a list of problems for each dmisfit [prob1, prob2, ...]
         """
-        return [objfcts.prob for objfcts in self.dmisfit.objfcts]
+        return [objfcts.simulation for objfcts in self.dmisfit.objfcts]
```
lheagy: thanks for all of these! Would you mind also updating line 104 here to be `def simulation(self)` and then for helping with backwards compatibility, we can add a 
```python

@property
def prob(self):
        warnings.warn(
            "The prob property will be depreciated in favor of simulation. "
            "Please update your code accordingly"
        )
        return self.simulation
```

JKutt: Can do! Will get on this shortly.

JKutt: @lheagy Done and Done!

dccowan:
Tutorials have been added for 1D and 2.5D. They can be found in tutorials/temporary. Tutorials consist of:
- 1D forward simulation, 1D inversion and 1D inversion for layer thicknesses
- 2.5D dcip forward simulation and 2.5D dcip inversion

dccowan:
**Issue with 1D Simulation**

For the 1D simulation, we define layer resistivity and thickness 1 to N going from surface to depth. This is not consistent with standard SimPEG convention. The creation of meshes and models for this simulation should be consistent with SimPEG. @sgkang do you have any thoughts?

jcapriot:
@sgkang Your last commit actually caused an issue that is fixed by undoing what you had, can you double check what issue you had?

jcapriot:
I'm going to close and merge this branch into simulation for the time being, to get moving on updates to the potential fields branch.

PR #806, opened 2019-08-29
SphericalMap

thast:
A `mapping` to go from Spherical coordinates (each with its own mapping listed in `maplist`) to Cartesian coordinates.

- Bug fix: The `inverse` of the `IdentityMap` was `not Implemented`. it now returns the input.


jcapriot:
@thast It looks like @fourndo implemented a spherical map in the `simulation` branch. To my eye the implementation looks very similar just with a different name. Can you look over that and double check for me? It's called `SphericalSystem` in that branch.

I do like your changes to the base Identity map though. So we should get those into `simulation`

jcapriot:
Maybe we should create a `BaseMap` that most things inherit instead of an `IdentityMap`?

thast:
@jcapriot : good catch, I did not know. I will leave @fourndo judges of that...

domfournier:
I think we are good with the simulation version. 
@thast You can delete your branch I think

PR #807, opened 2019-09-01
Fix simple regularization

fourndo:
Proposed modification to the model gradient measure for Simple and Sparse regularization.
Normalized length scales are multiplying cell_weights before averaging to cell faces.
- Reduces the dependency of the solution on the change in discretization
- Still allows for IRLS weights to be mesh independent
- Default alpha's remain at 1 (takes care of length scales internally) 

**Before**
image

**After**
LengthScalesApplied

### DIFF
--- a/SimPEG/regularization/sparse.py
+++ b/SimPEG/regularization/sparse.py
@@ -171,7 +171,7 @@ class SparseDeriv(BaseSparse):
     """
 
     def __init__(self, mesh, orientation='x', **kwargs):
-
+        self._length_scales = None
         self.orientation = orientation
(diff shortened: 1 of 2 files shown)


SimPEG/regularization/sparse.py
```diff
@@ -171,7 +171,7 @@ class SparseDeriv(BaseSparse):
     """
 
     def __init__(self, mesh, orientation='x', **kwargs):
-
+        self._length_scales = None
```
lheagy: should we make this a property? That way if we serialize the class, they are included? 
```length_scales = properties.Array("length scales")```

fourndo:
@lheagy Can we bring this in soon. It's a pretty important update.

PR #810, opened 2019-09-11
Small fix for regularization_mesh

thast:
Small fix of the `mkvc` utils call


lheagy:
thanks @thast! I will bring this in and we can include it with the new release coming in with #807

PR #820, opened 2019-10-29
Mag dev df dask nick

fourndo:
@orerocks Feel free to merge.


orerocks:
@fourndo I removed those couple of bottlenecks associated with the types and casting. It now runs as fast as my branch (so a 60% improvement on `MagDev_DF_Dask` branch in my tests). I'm merging. But all my speed testing is on the Cartesian MVI. We still need to confirm performance is ok for SUS, MVI-S, and Gravity.

PR #836, opened 2019-12-24
Fix bug in set_mesh

yezhengkai:
I fixed the topo part in the set_mesh method, hope I did it right.

Following is an example using IO.set_mesh
```python
# read topography
topo1 = read_topo(....)
topo2 = read_topo(....)

# use topo1
mesh, actind = IO.set_mesh(topo=topo1, dx=0.5, dz=0.8, method='nearest')
survey.drapeTopo(mesh, actind, option="top")

# use topo2
mesh, actind = IO.set_mesh(topo=topo2, dx=0.1, dz=0.4, method='cubic')  # can use cubic now
# Force drapTopo using the same survey instance
survey.drapeTopo(mesh, actind, option="top", force=True)
```


lheagy:
Many thanks @yezhengkai for your pr! 

:wave: @sgkang, would you mind taking a look?

SimPEG/EM/Static/DC/IODC.py
```diff
@@ -510,9 +510,26 @@ def set_mesh(self, topo=None,
                 )
             x0 = self.electrode_locations[:, 0].min()
             if topo is None:
-                locs = np.sort(self.electrode_locations, axis=0)
+                # For 2D mesh
```
sgkang: looking great to me @yezhengkai

SimPEG/EM/Static/DC/IODC.py
```diff
@@ -510,9 +510,26 @@ def set_mesh(self, topo=None,
...
+                    # sort by x
+                    row_idx = np.lexsort((self.electrode_locations[:, 0],))
+                # For 3D mesh
+                else:
+                    # sort by x, then by y
+                    row_idx = np.lexsort((self.electrode_locations[:, 1],
+                                          self.electrode_locations[:, 0]))
+                locs = self.electrode_locations[row_idx, :]
             else:
-                locs = np.vstack((topo, self.electrode_locations))
```
sgkang: my thought process here was including both electrode locations and topo locations. 
so we we still need `locs_tmp = np.vstack(np.vstack((topo, self.electrode_locations))`
then line 532 need to be changed as 

`locs = loc_tmp[row_index, :]`

SimPEG/EM/Static/DC/SurveyDC.py
```diff
@@ -131,14 +131,14 @@ def getABMN_locations(self):
...
-    def drapeTopo(self, mesh, actind, option='top', topography=None):
+    def drapeTopo(self, mesh, actind, option='top', topography=None, force=False):
         if self.a_locations is None:
             self.getABMN_locations()
 
         # 2D
         if mesh.dim == 2:
             if self.survey_geometry == "surface":
-                if self.electrodes_info is None:
+                if self.electrodes_info is None or force:
```
sgkang: good catch. can we set `foce=True` as a default?

sgkang:
@lheagy it seems travis is failing due to following:

image

Do you have any idea for this?

lheagy:
Thanks @sgkang. I think this is due to the SimPEG docs not being recently uploaded. I will look in to it tonight

SimPEG/EM/Static/DC/IODC.py
yezhengkai: My consideration is that when the topography and electrode are at the same x (or x, y) coordinates (such as the following settings), surface2ind_topo may have the wrong interpolation.

```
# topo and electrode are at the same x coordinate, but z is different
topo = np.array([[0, 10], [5, 2]])
electrode_locations = np.array([[0, 0], [5, 3]]) 
```

https://github.com/simpeg/simpeg/blob/46c6619c3368dd0000c3b2fe27bdbd042fe8c17a/SimPEG/EM/Static/DC/IODC.py#L592

Should we just keep the topo location and add a warning or just throw an error?

yezhengkai:
For your https://github.com/simpeg/simpeg/pull/836#discussion_r363464696, I have two solutions.

Solution 1 (excluded electrodes):
```python
if dimension == 2:
    mask = np.isin(self.electrode_locations[:, 0], topo[:, 0])
    if np.any(mask):
        warnings.warn(
            "Because the x coordinates of some topo and electrodes are the same,"
            " we excluded electrodes with the same coordinates.",
            RuntimeWarning
        )
    locs_tmp = np.vstack((topo, self.electrode_locations[~mask, :]))
    row_idx = np.lexsort((locs_tmp[:, 0],))
else:
    dtype = [('x', np.float64), ('y', np.float64)]
    mask = np.isin(self.electrode_locations[:, [0, 1]].copy().view(dtype),
                   topo[:, [0, 1]].copy().view(dtype)).flatten()
    if np.any(mask):
        warnings.warn(
            "Because the x and y coordinates of some topo and electrodes are the same,"
            " we excluded electrodes with the same coordinates.",
            RuntimeWarning
        )
    locs_tmp = np.vstack((topo, self.electrode_locations[~mask, :]))
    row_idx = np.lexsort((locs_tmp[:, 1],
                          locs_tmp[:, 0]))
locs = locs_tmp[row_idx, :]
```
Solution 2 (raise an error):
```python
if dimension == 2:
    mask = np.isin(self.electrode_locations[:, 0], topo[:, 0])
    if np.any(mask):
        raise RuntimeError("the x coordinates of some topo and electrodes are the same.")
    locs_tmp = np.vstack((topo, self.electrode_locations[~mask, :]))
    row_idx = np.lexsort((locs_tmp[:, 0],))
else:
    dtype = [('x', np.float64), ('y', np.float64)]
    mask = np.isin(self.electrode_locations[:, [0, 1]].copy().view(dtype),
                   topo[:, [0, 1]].copy().view(dtype)).flatten()
    if np.any(mask):
        raise RuntimeError("the x and y coordinates of some topo and electrodes are the same.")
    locs_tmp = np.vstack((topo, self.electrode_locations[~mask, :]))
    row_idx = np.lexsort((locs_tmp[:, 1],
                          locs_tmp[:, 0]))
locs = locs_tmp[row_idx, :]
```

sgkang:
Hi @yezhengkai, sorry again for being late 2nd solution looks great to me. Can you incorporate your changes?

sgkang:
Hi @yezhengkai, I merged this into `io-dc-topo` branch. 
As long as it passes travis, will merge this in.

PR #841, opened 2020-01-22
Fix bug in set_mesh

sgkang:
This is moved from #836
Comment from @yezhengkai

I fixed the topo part in the set_mesh method, hope I did it right.

Following is an example using IO.set_mesh
```python
# read topography
topo1 = read_topo(....)
topo2 = read_topo(....)

# use topo1
mesh, actind = IO.set_mesh(topo=topo1, dx=0.5, dz=0.8, method='nearest')
survey.drapeTopo(mesh, actind, option="top")

# use topo2
mesh, actind = IO.set_mesh(topo=topo2, dx=0.1, dz=0.4, method='cubic')  # can use cubic now
# Force drapTopo using the same survey instance
survey.drapeTopo(mesh, actind, option="top", force=True)
```


sgkang:
Hi @thast can you quickly review this and approve?

SimPEG/EM/Static/DC/SurveyDC.py
```diff
@@ -131,14 +131,14 @@ def getABMN_locations(self):
...
-    def drapeTopo(self, mesh, actind, option='top', topography=None):
+    def drapeTopo(self, mesh, actind, option='top', topography=None, force=False):
         if self.a_locations is None:
             self.getABMN_locations()
 
         # 2D
         if mesh.dim == 2:
             if self.survey_geometry == "surface":
-                if self.electrodes_info is None:
+                if self.electrodes_info is None or force:
```
thast: What does this force do?

sgkang: This is considering multiple times to set electrode locations, which may not be that much often, but possible

PR #844, opened 2020-02-04
Static receiver midpoint 3D

domfournier:
- Generalize the calculations to 3D arrays.
- Use the `survey.survey_type` rather than a kwarg
- Simplified code 

Works in 3D and 2D
image


SimPEG/electromagnetics/static/utils/StaticUtils.py
```diff
@@ -354,10 +324,8 @@ def plot_pseudoSection(
     if dobs is None:
         dobs = data.dobs
 
-
-    midx, midz = source_receiver_midpoints(
-        data.survey, survey_type=survey_type, dim=dim
-    )
+    midx, midz = source_receiver_midpoints(data.survey)
+    midx = midx[:, 0]  # Only keep the along line
```
jcapriot: I don't like that this assumes every line is run east-west in 3D. Something should be added to check for the orientation of the lines, and choose x or y in 3D.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -355,10 +329,16 @@ def plot_pseudoSection(
...
         dobs = data.dobs
 
-
-    midx, midz = source_receiver_midpoints(
-        data.survey, survey_type=survey_type, dim=dim
-    )
+    midx, midz = source_receiver_midpoints(data.survey)
+    if midx.shape[1] == 2:
+        min_x, min_y = np.min(midx, axis=0)
+        max_x, max_y = np.max(midx, axis=1)
```
fourndo: are you sure about ``axis=1``? Sounds like you a column operation

fourndo:
@jcapriot : All systems green on this side.

image

PR #845, opened 2020-02-10
Tile map

domfournier:
Allow to invert data misfits on tiled meshes.
Add example


domfournier:
Here you @jcapriot . 
@JKutt This should allow us to start using simulation for the gradiometry inversion.

jcapriot:
@domfournier Can you add a simple test for the TileMap? It won't work in the bulk test and will likely needs in own test class written.

jcapriot:
Thanks for getting the example and the test in!

fourndo:
@jcapriot : Well done on the cleanup. Looking fine here.

PR #848, opened 2020-02-17
Fix Ward and Hohmann example

prisae:
There is no specific reference given, so maybe it means another Ward and Hohmann paper. However, if the Ward and Hohmann of the green SEG-EM-book is meant, then I think there are two errors in the figure.

1. ~There is no **Figure 4.56** in Ward and Hohmann, 1988, SEG; I think it tries to reproduce **Figure 4.2** instead.~ With 4.56 was apparently the equation meant, but it is not specified. So 4.56 is the equation, 4.2 the corresponding figure. I adjusted that.
2. The resistivity in Ward and Hohmann is 100 Ohm.m, but in the example 1000 Ohm.m is used (0.001 S/m), so the result is not exactly the same as in Ward and Hohmann.

This PR fixes these two issues.

If I am wrong and another Ward and Hohmann paper is meant then the correct reference should be added.

### DIFF
--- a/SimPEG/EM/Analytics/FDEM.py
+++ b/SimPEG/EM/Analytics/FDEM.py
@@ -7,7 +7,9 @@ from SimPEG import Utils
 
 def hzAnalyticDipoleF(r, freq, sigma, secondary=True, mu=mu_0):
     """
-    4.2 in Ward and Hohmann, 1988.
+    The analytical expression is given in Equation 4.56 in Ward and Hohmann,
+    1988, and the example reproduces their Figure 4.2.
+
 
     .. plot::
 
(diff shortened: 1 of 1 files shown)


prisae:
### Current figure:
w+h-old

### New figure:
w+h

prisae:
Wow, that part goes waaaay back, 19 March 2014. It moved over from

- github.com/lheagy/simpegem/simpegEM/Utils/Ana/FEM.py
- github.com/lheagy/simpegem/simpegEM/Analytics/FDEM.py
- github.com/simpeg/simpegEM/Analytics/FDEM.py

The original commit came from @rowanc1, as far as I can tell.

https://github.com/lheagy/simpegem/commit/dd671e127765dfa547bbda531ff610b7f804129a#diff-7a314400bc887817f94600ec4908ecd2

rowanc1:
But I probably copied it in from someone else. Maybe @sgkang or @lheagy.

prisae:
I find it amazing - believing how quickly 6 years passed?

jcapriot:
I'm going to merge this into master for now, but not update the version number just yet as it is only a documentation update, which should allow us to deal with the conflicts with this with the `simulation` branch.

PR #849, opened 2020-02-18
Update and fix tutorials

dccowan:
Since some changes have been made to the simulation branch, I have fixed and retested:
- Gravity tutorials
- Magnetics tutorials
- 1D DC tutorials
- DCIP2D tutorials


jcapriot:
Hey @dccowan, thanks for catching those. Just to double check, do these get ran during the automated testing? I didn't see anything that jumped out at me saying they do.

dccowan:
> Hey @dccowan, thanks for catching those. Just to double check, do these get ran during the automated testing? I didn't see anything that jumped out at me saying they do.

I haven't added this examples to the testing directly. But if they don't run properly, we will see them fail in the docs test. I believe this is the same thing for any of the pre-existing examples. I figured this would be acceptable.

dccowan:
I ran the docs test locally and it passed. There are some broken links but this should be handled in another pull request. Both the tutorials and the examples gallery are actually up to date with the latest simulation branch. @lheagy can you force merge? Some of the examples in the examples gallery are very large and timeout Travis.

Issue #847, opened 2020-02-13
Renaming of methods' Problems


jcapriot:
In the next release of SimPEG we are renaming many of the base properties to be more in line with pythonic conventions, pep8, etc., and we are also refactoring to get rid of SimPEG's `Problem` and replacing it with `Simulation` (which moves the forward operations from a `Survey` object that has been paired with a `Problem` to just a `Simulation`). As part of this, I'd also like to update the names for each method's problems, to have a consistent convention across SimPEG.

Currently
---------

For example in FDEM and TDEM, the names are 
* `Problem3D_E`
* `Problem3D_B`
* `Problem3D_H`
* `Problem3D_j`

In natural source em:
* `Problem1D_ePrimSec`
* `Problem3D_ePrimSec`

In the static methods IP, and DC, they are
* `Problem2D_CC`
* `Problem2D_N`
* `Problem3D_CC`
* `Problem3D_N`

The simulations in the potential field methods are currently 
* `MagneticIntegral`
* ` Problem3D_DiffSecondary`
* `GravityIntegral`
* `Problem3D_Diff`

In VRM
* `Problem_Linear`
* `Problem_LogUniform`

And the others:
* `RichardsProblem`
* `StraightRayProblem`

As you can see, the convention in naming is certainly not consistent, and we should make a decision moving forward of a good convention that incorporates the `Simulation` name possibly. Some things we might want to keep are, distinguishing between dimensionality of a simulation, possibly being able to use autocomplete within each package to find the simulations (a.k.a. they would all start with `Simulation`), being descriptive with the names, etc. We would like to keep pep8 conventions for class names. Keep in mind that the method names would already be a part of the path to the `Simulation` class.

Mostly try to think about how suggestions would propagate, as some things might get quite long. For example I'll put out three possible formats:

Simulation(Dimension)(ShortDescription), Simulation(Dimension)(LongDescription), (LongDescription)Simulation(Dimension).

For Example
------------
1. `Problem3D_E` &rarr; `Simulation3DE` or `Simulation3DElectricField` or `ElectricFieldSimulation3D`.

2. `Problem3D_B` &rarr; `Simulation3DB` or `Simulation3DMagneticFluxDensity` or `MagneticFluxDensitySimulation3D`

3. `Problem3D_j` &rarr; `Simulation3DJ` or `Simulation3DCurrentDensity` or `CurrentDensity3DSimulation` note here, should `j` be capitalized or not?

[...]

PR #851, opened 2020-02-28
Simulation method names

jcapriot:
Renames the method's simulation classes to have a consistent naming convention.
See issue #847.


jcapriot:
@lheagy Just wanting someone to double check the re-naming is correct.

SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -880,3 +881,43 @@ def getRHSDeriv(self, freq, src, v, adjoint=False):
         s_mDeriv, s_eDeriv = src.evalDeriv(self, adjoint=adjoint)
 
         return RHSDeriv + s_mDeriv(v) + C.T * (MfRho * s_eDeriv(v))
+
+
+############
+# Deprecated
+############
+
+class Problem3D_e(Simulation3DElectricField):
```
rowanc1: This is looking awesome. :)

SimPEG/electromagnetics/static/resistivity/simulation_2d.py
```diff
@@ -578,7 +579,7 @@ class Problem2D_N(BaseDCSimulation_2D):
     fieldsPair_fwd = Fields_N
 
     def __init__(self, mesh, **kwargs):
-        BaseDCSimulation_2D.__init__(self, mesh, **kwargs)
+        BaseDCSimulation2D.__init__(self, mesh, **kwargs)
```
rowanc1: Just a thought, not blocking this: Would `super()` here be better? Just in case some things get switched around in the future, we are only ever changing things in one place.

tests/em/tdem/test_TDEM_DerivAdjoint_RawWaveform.py
```diff
@@ -169,7 +169,7 @@ def test_Jvec_adjoint_e_ey(self):
 
 class DerivAdjoint_B(Base_DerivAdjoint_Test):
 
-    formulation = 'b'
+    formulation = 'MagneticFluxDensity'
```
rowanc1: This is so much better! 🎉

SimPEG/electromagnetics/static/resistivity/simulation_2d.py
jcapriot: I agree, `super()` here would be super. However there are MANY places in the code that need this change, so I'm not as concerned about it just yet.

rowanc1: Absolutely. :)

SimPEG/electromagnetics/frequency_domain/fields.py
```diff
@@ -320,7 +320,7 @@ def _jDeriv(self, src, du_dm_v, v, adjoint=False):
 
 class Fields3D_e(FieldsFDEM):
```
lheagy: Do you think it is worth updating the names of the fields objects here to match that of the simulation? 
```
Fields3DElectricField
```

SimPEG/electromagnetics/natural_source/simulation.py
```diff
@@ -344,7 +345,7 @@ class Problem3D_ePrimSec(BaseNSEMSimulation):
     _sigmaPrimary = None
 
     def __init__(self, mesh, **kwargs):
-        super(Problem3D_ePrimSec, self).__init__(mesh, **kwargs)
+        super(Simulation3DPrimarySecondary, self).__init__(mesh, **kwargs)
```
lheagy: This is looking great 🎉

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -12,8 +13,8 @@
 from ...base import BaseEMSimulation
 
 from ..resistivity.fields import FieldsDC, Fields_CC, Fields_N
-from ..resistivity import Problem3D_CC as BaseProblem3D_CC
-from ..resistivity import Problem3D_N as BaseProblem3D_N
+from ..resistivity import Simulation3DCellCentered as BaseSimulation3DCellCentered
```
lheagy: No 🇨🇦-spelling to mess with everyone `Centred`?? 😉

jcapriot: I did put Alias's for them in the `__init__.py` files already, so someone should be able to access both versions :)

jcapriot:
I'm also going to apply a similar rule for the matching Receivers, and Fields classes.

jcapriot:
With this, I have added a ton of deprecation warnings, and workarounds for older code to work with minimal changes. Hopefully all that is needed is to change the imports to the new names. But we/I cannot guarantee that all old codes will be working. Essentially it now adds functions that mimic the old behavior of `pairing` the simulation and survey. I've also expanded the deprecated `l2_DataMisft` class to work similar to how it previously worked. These are tested using old versions of those tests that have been given very minimal changes (only to import module names).

PR #853, opened 2020-03-09
CSEM TDEM 3D example

prisae:


jcapriot:
switching this to pull in to simulation considering it depends on empymod. I'll make the necessary changes.

jcapriot:
Bringing this example in, Thanks again for writing this @prisae!

PR #854, opened 2020-03-11
Simulation dc2.5 d speedup

jcapriot:
This adds an option to attempt to `miniaturize` a 2D dc survey in the simulation.

This all happens on the backend without the user being wiser (unless `verbose`). It is currently disabled by default, as I believe it needs more testing for it to be a default.

It works as follows, if there are more sources in the survey than there are unique source electrodes. It switches to a miniaturized survey that replaces all sources with `Pole` sources at each unique source electrode. For surveys that have wenner configurations this is very common, and should result in significant speedups for the operations of `dpred`, `Jvec`, and `Jtvec`, by both limiting the amount of `fields` objects stored, and reducing the number of `solve` operations similarly.

Fully building the J matrix, will not have as much of a speedup, as the number of solves will always be at least the number of data. However, the `mini` survey essentially doubles the number of receivers when there are dipole sources, resulting in more overall RHS solutions. But because the number of calls to the solver routine is still at a minimum of the number of source electrodes, we can take advantage of parallelism over RHSs built into the common solvers to still obtain an overall speedup.

There are also a few other small optimizations that should benefit all 2d DC surveys.


jcapriot:
This also adds the ability to auto-generate `TreeMesh`es for 2D dc data.

jcapriot:
All of the tests pass locally, so not quite sure why it is hanging on one of the deprecation tests. I'm still trying to reproduce this. In the meantime I'm going to get this merged in to simulation.

PR #863, opened 2020-04-30
light formatting to be closer to black, update filepaths to be os independent

lheagy:
- light formatting consistent with [black](https://github.com/psf/black) for brackets, indentation and quotes
- use `os.path.sep` to determine which separators should be used 
- use `viridis` instead of `jet` (if you prefer another [perceptually uniform colormap](https://matplotlib.org/3.1.1/gallery/color/colormap_reference.html), feel free to update


lheagy:
I will close this in favour of #869

PR #864, opened 2020-05-01
Sim dask split

jcapriot:
This branch will split out the dask functionality from the Base `SimPEG` operations and moves them all to a specific module `SimPEG.dask`. Meaning that `dask` can now be toggled on (but not off).

This is accomplished by the using purposely doing
`import SimPEG.dask`

This operation will then specifically replace the necessary functions within the classes that have been given attention with `dask`. This should also allow us to more quickly update things specifically for `dask` without having to touch the main code.

I was also able to accomplish this by increasing the amount of operations that could be done on both dask and normal type operations. This should allow us to focus specifically on operations that will benefit from dask the most while also working with and without it.

Also,
I made updates to the `resistivity.Simulation1DLayered` to support empymod <2.0.0 and empymod >= 2.0.0.


requirements_dask.txt
```diff
@@ -0,0 +1,3 @@
+dask
+zarr
+fsspec>=0.3.3
```
lheagy, line 3: do we need `sparse` here too?

SimPEG/simulation.py
```diff
@@ -133,6 +131,11 @@ class BaseSimulation(props.HasModel):
 
     counter = properties.Instance("A SimPEG.utils.Counter object", Counter)
 
+    sensitivity_path = properties.String(
+        'path to store the sensitivty',
+        default="./sensitivity/"
+        )
```
lheagy, line 137:
```suggestion
    )
```

tests/dask/test_IP_jvecjtvecadj_dask.py
```diff
@@ -0,0 +1,333 @@
+from __future__ import print_function
+import unittest
+import discretize
+import numpy as np
+
+import SimPEG.dask
```
lheagy, line 6: this is really cool

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -0,0 +1,99 @@
...
+    m_size = self.model.size
+    count = 0
+    for source in self.survey.source_list:
+        u_source = f[source, self._solutionType]
+        for rx in source.receiver_list:
+            PT = rx.getP(self.mesh, rx.projGLoc(f)).toarray().T
+            df_duT = PT
+            # Find a block of receivers
+            n_block_col = int(np.ceil(df_duT.size *
+                                      8 * 1e-9 / self.max_ram))
```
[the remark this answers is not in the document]
domfournier: that's the conversion from bytes to Gb

SimPEG/dask/potential_fields/magnetics/simulation.py
```diff
@@ -0,0 +1,33 @@
...
+
+    self.model = m
+
+    if W is None:
+        W = np.ones(self.nD)
+    else:
+        W = W.diagonal()
+    if getattr(self, "_gtg_diagonal", None) is not None:
+        return self._gtg_diagonal
+
```
domfournier, line 20: This is not working for Spherical formulation as the JtJdiag changes every iteration. The `chiDeriv` needs to be applied as it was done on current `simulation`

jcapriot: I got ya, would it be better to add these to the `deleteOnModelUpdate` property of the `simulation` class. Contextually of course if chiDeriv is dependent on the model?

SimPEG/electromagnetics/static/resistivity/simulation_1d.py
```diff
@@ -64,11 +76,19 @@ def fields(self, m):
...
+                self.offset, self.fhtfilt,
+                self.hankel_pts_per_dec, factAng=None, ab=33
+            ).real / (2*np.pi)
+        except TypeError:
+            voltage = dlf(
+                PJ, self.lambd,
+                self.offset, self.fhtfilt,
+                self.hankel_pts_per_dec, ang_fact=None, ab=33
+            ).real / (2*np.pi)
+
```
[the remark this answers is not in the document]
jcapriot: yup, the keyword argument changed in empymod 2.0.0. so this will work for older and newer versions.

SimPEG/dask/electromagnetics/static/resistivity/simulation.py
```diff
@@ -0,0 +1,152 @@
...
+        Return the diagonal of JtJ
+    """
+    if (self.gtgdiag is None):
+
+        # Need to check if multiplying weights makes sense
+        if W is None:
+            self.gtgdiag = da.sum(self.getJ(m)**2, axis=0).compute()
+        else:
+            w = da.from_array(W.diagonal())[:, None]
+            self.gtgdiag = da.sum((w*self.getJ(m))**2, axis=0).compute()
```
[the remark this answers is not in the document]
jcapriot: They do the same operation, but only on these lines. Dask's array object functions like numpy's arrays, and thus the ** operation is elementwise. Whereas further down, the object that is created is a sparse matrix, where A**2 is (A@A), and the `power` call is element wise. I think this mostly comes down to scipy's sparse matrix library conforming to numpy's matrix class, and not numpy's array class.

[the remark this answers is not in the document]
jcapriot: generally though from my experience, `**2` element wise operation is faster than `power(2)`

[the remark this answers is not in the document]
jcapriot: Also check out this for some timing examples https://stackoverflow.com/questions/25870923/how-to-square-or-raise-to-a-power-elementwise-a-2d-numpy-array

requirements_dask.txt
jcapriot: nope, sparse has been removed

SimPEG/dask/electromagnetics/static/resistivity/simulation.py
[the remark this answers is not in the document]
domfournier: right on, yep makes sense

SimPEG/dask/potential_fields/magnetics/simulation.py
domfournier: So the trick here is to keep the `_gtg_diagonal` and just apply it to the non-linear part (i.e. `chiDeriv`) at each iteration. I think we probably want to avoid doing things in a for-loop over the data after this too...

jcapriot: Yeah, I gotcha. I’ll get it later tonight then get these pulled in.

jcapriot:
Looks good now, merging this in.

Issue #862, opened 2020-04-24
FDEM: derivative test seems to ignore imag-real-flag


prisae:
I am trying to follow the derivative and adjoint tests in FDEM.

The `adjointTest` (in `simpeg/tests/em/fdem/inverse/adjoint/test_FDEM_adjointEB.py`) uses the `getFDEMProblem`. This function takes a `comp`-parameter, e.g. `exr` for x-directed electric field, real component.

However, inside `getFDEMProblem` the `r`-part is assigned to a `real_or_imag`-parameter, but is then ignored, and the `'imag'`-part is used no matter what. So the adjoint test seem to check, in fact, only the imaginary solutions, no matter if it is called like this
```
adjointTest('e', 'exr')
```
or like this
```
adjointTest('e', 'exi')
```
https://github.com/simpeg/simpeg/blob/2cb11643517881f683cdc33a67f9d28223392609/SimPEG/EM/Utils/testingUtils.py#L32-L36

lheagy:
good catch - thanks!

prisae:
That is fixed (not sure why it didn't got closed automatically).

PR #865, opened 2020-05-02
simulation fdem testing bug 

lheagy:
fix the FDEM testing utils bug described in #862, update the test_muinverse to conform to current API


lheagy:
:wave: Hi @jcapriot, would you mind taking a quick look at this when you have a min?

PR #866, opened 2020-05-02
waveform logic bug 

### LINKED ISSUE #850: Can't set a waveform on initialized BaseTDEMSrc object
I’d like to be able to set the waveform after the TDEMSrc object is initialized, but that doesn’t seem possible at the moment:

As an example, I would expect to be able to do something like this:
```
rx_list = [
    TDEM.RxTDEM.Point_b(
        locs=[np.r_[0.0, 0.0, 0.0]], times=np.r_[1e-3, 2e-3, 5e-3], orientation="z"
    )
]
src_loop = TDEM.Src.CircularLoop(
    rxList=rx_list,
    loc=np.r_[0.0, 0.0, 0.0],
    orientation="z",
    radius=300,
    current=1000.0
)

offTime=1e-3
src_loop.waveform = TDEM.Src.QuarterSineRampOnWaveform(
    ramp_on=np.r_[0.0, 5.0e-4], ramp_off=np.r_[offTime, offTime + 1e-4]
)
```
but this raises an AttributeError due to the call to the non-existent method here https://github.com/simpeg/simpeg/blob/2cb11643517881f683cdc33a67f9d28223392609/SimPEG/EM/TDEM/SrcTDEM.py#L245

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-4e2a9d2765ae> in <module>
     15 offTime=1e-3
     16 src_loop.waveform = TDEM.Src.QuarterSineRampOnWaveform(
---> 17 ramp_on=np.r_[0.0, 5.0e-4], ramp_off=np.r_[offTime, offTime + 1e-4]
     18 )

~/anaconda3/envs/snowshoe_env/lib/python3.7/site-packages/SimPEG/EM/TDEM/SrcTDEM.py in waveform(self, val)
    243             self._waveform = val
    244         else:
--> 245             self._waveform = self.StepOffWaveform(val)
    246 
    247     def __init__(self, rxList, waveform=StepOffWaveform(), **kwargs):

AttributeError: 'CircularLoop' object has no attribute 'StepOffWaveform'
```
Can anyone explain the intended behavior of the waveform setter here?
https://github.com/simpeg/simpeg/blob/2cb11643517881f683cdc33a67f9d28223392609/SimPEG/EM/TDEM/SrcTDEM.py#L240
 The waveform can’t be initialized as None  because of the method call on line 242, and the second branch of the if/else always fails because self.StepOffWaveform is not defined.

lheagy:
- simplify the logic for setting the waveform and allow the waveform to be set after a source is instantiated (will close #850)
- add a test to check if setting a waveform after instantiating the source works


jcapriot:
Would it make more sense to make it a `property.Instance(BaseWaveform)`?

Is there any reason that the waveformPair is necessary? I don't see where it is used anywhere else in the code, outside of validating `waveform`

tests/em/tdem/test_TDEM_crosscheck.py
```diff
@@ -161,17 +162,42 @@ def test_HJ_dhdtz_vtem(self):
...
             rxcomp='MagneticFieldTimeDerivativex', waveform='vtem')
 
-
     def test_MagDipoleSimpleFail(self):
 
         print('\ntesting MagDipole error handling')
 
 
-        with warnings.catch_warnings(record=True):
+        with self.assertWarns(DeprecationWarning):
```
jcapriot, line 170: I cannot figure out what ``error`` this was supposed to test, as it doesn't throw anything on `master`. It does throw some deprecation warnings though... @lheagy any thoughts? Did it not support an arbitrary dimension at some point?

jcapriot: think I found it, had to go back a while. It would originally fail because the `receiver_list` was empty (or because it wasn't given a list of `receiver` object). Validation is handled by properties, and it being a 0 length list is actually okay (i.e. if we only want the fields).

Issue #856, opened 2020-04-15
Accessing Fields for DC 2.5D


thast:
In DC 2.5D, trying to access any other field than "Phi" raises the current error (btw, those other fields appear in the documentation and the 2.5Dcode, they are just not fully implemented yet):
![image](https://user-images.githubusercontent.com/13682747/79374558-40b28d80-7f0c-11ea-988c-c98975be9a44.png)

Side note:
Accessing the 'Phi' field is working. However, I have a question about its shape. See example below. The second dimension is the number of spatial frequencies but I am unsure what is the first dimension: it is neither the size of the whole mesh nor the size of the active cells....
![image](https://user-images.githubusercontent.com/13682747/79374926-ac94f600-7f0c-11ea-9cf4-012879cccb29.png)

Extra information: here is the scripts I ended up using to compute and visualize the field. The em-apps are using a similar approach. So improving on the Fields object will likely improve the em-apps.
```
# Plot vector field of one quadrupole for mtrue
Mx = meshList20[0].gridCC
rx = DC.Rx.Pole_ky(Mx)
src = DC.Src.Dipole([rx],dcsurvey20[0].srcList[0].loc[0],dcsurvey20[0].srcList[0].loc[1])
dcsurvey = DC.Survey_ky([src])
dcprob = DC.Problem2D_N(meshList20[0],sigmaMap=mapping20[0],Solver=PardisoSolver)
dcprob.pair(dcsurvey)

mtrue = 1./res20[0]*1000
phipotential = dcsurvey.dpred(mtrue[actind_air20[0]])
efield = -meshList20[0].cellGrad * phipotential
efieldx = -meshList20[0].cellGradx * phipotential
efieldy = -meshList20[0].cellGrady * phipotential

jfield = dcprob.MfRhoI * efield

from scipy.interpolate import LinearNDInterpolator, NearestNDInterpolator
import matplotlib

tensormesh = Mesh.TensorMesh(meshList20[0].h)
tensormesh.x0 = meshList20[0].x0
indcore, tensormesh = Utils.ExtractCoreMesh(np.c_[[-3,-1],[3,1]],tensormesh)

InterpolateEField_x = LinearNDInterpolator(meshList20[0].gridEy,efieldx)
ExField_tensor = InterpolateEField_x(tensormesh.gridEy)
InterpolateEField_y = LinearNDInterpolator(meshList20[0].gridEx,efieldy)
EyField_tensor = InterpolateEField_y(tensormesh.gridEx)
[...]

PR #868, opened 2020-05-04
This change was necessary for fields to be called correctly for dcip2…

dccowan:
This patch is to fix the error in issue #856 .


jcapriot:
yup, looks like a small bug that was in simulation branch, thanks for catching it.

It should work correctly on master though.

PR #869, opened 2020-05-05
Simulation tutorials 2

jcapriot:
Pulling @dccowan tutorial updates into the simulation branch.


examples/04-grav/Chile_GRAV_4_Miller/LdM_grav_obs.grv
```diff
@@ -0,0 +1,192 @@
+191
```
lheagy, line 1: Would you mind removing these (all of the Chile_GRAV_4_Miller files + the tarball)? The script includes the instructions for downloading them, so we do not need to include them in the repo. We should also update the gitignore to not include these files

tutorials/assets/fdem/fdem_data.obs
```diff
@@ -0,0 +1,243 @@
+1.0000e+02 -2.0000e+02 -2.0000e+02 2.0000e+01 -4.4874e-14 -3.3585e-13
```
lheagy, line 1: instead of including these in the repo, would it work if we archived them on google and just download as needed? I am happy to help with that if that works with folks

jcapriot:
Merging it!

PR #881, opened 2020-05-27
Bug fixes for 3D IP simulation

sgkang:
- fix bug in 3D IP simulation code (computing DC field for each iteration -> compute once)
- clean up 2D code for setting data_type as apparent_chargeability


SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
```diff
@@ -56,10 +65,48 @@ def fields(self, m):
...
+                    for rx in src.receiver_list:
+                        rx._dc_voltage = dc_data[src, rx]
+                        rx.data_type = self.data_type
+                        rx._Ps = {}
+
         self._pred = self.forward(m, f=self._f)
 
         return self._f
 
+    def dpred_dc(self, f=None):
```
sgkang, line 84: @jcapriot: may be there is a better way than this..

jcapriot: part of me was thinking of just simplifying the ip simulation to just accept a dc_simulation as a parameter, which would then allow for arbitrary dimensionality and less repeated code.

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -21,50 +22,59 @@ class BaseIPSimulation(BaseEMSimulation):
 
     eta, etaMap, etaDeriv = props.Invertible("Electrical Chargeability")
 
-    # surveyPair = Survey
+    data_type = properties.StringChoice(
```
jcapriot, line 25: would this be better suited to a be a property of the receiver?

sgkang: agree. receiver also have a property of `data_type`. had to set in the simulation to trigger DC simulation and computing predicted DC data.

SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
sgkang: hmm ... that's a good idea. so take dc simulation as an input parameter of ip simulation, but when dc simulation is none, then generate one.

SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
```diff
@@ -22,40 +24,38 @@ class BaseIPSimulation2D(BaseDCSimulation2D):
 
     eta, etaMap, etaDeriv = props.Invertible("Electrical Chargeability (V/V)")
 
+    data_type = properties.StringChoice(
+        "IP data type",
+        default='volt',
+        choices=['volt', 'apparent_chargeability'],
```
lheagy, line 30: what do you think of using the dictionary for string choice here (to allow the user to input common names and coerce them) 
```
choices={
    "volt": ["voltage"],
    "apparent chargeability": ["apparent_chargeability", "apparent-chargeability", "apparentchargeability"]
}
```

lheagy:
@jcapriot @sgkang: would it work with you both if we consolidate this with #884?

PR #885, opened 2020-06-02
Fix IRLS Tutorial

domfournier:
- Use standard floor error instead of percent
- Increase the number of CG solves 
- Reduce CG tolerance

This should do it.
Figure_3


jcapriot:
mind removing all of the extra added files?

domfournier:
sorry about that, still getting used to git_extentions

PR #889, opened 2020-06-04
Import SolverLU straight from utils.

adamkosik:
This avoids an import error on initialization of SimPEG.


jcapriot:
Again @adamkosik thanks for checking this and also submitting the PR. I updated the Flow simulation to just not set the default solver there, as it is already set in the base `Simulation` As long as these tests pass, we'll pull this in for the next release!

PR #894, opened 2020-06-16
Missing comma fixed

Zhuoliulz:
Add a missing comma to L 175 between "frequency" and "** kwargs".


prisae:
The docs-CI fail (I restarted it, but failed again), but I doubt it is related to this PR (@jcapriot, @lheagy)?

jcapriot:
I think it's the same problem all of these tests are having now (for some unknown reason) everything does pass locally so I'll pull them in and figure out the doc test stall later.

PR #895, opened 2020-06-18
__repr__ for surveys, sources, and receivers

prisae:
I almost forgot about this, and I don't know when I get around to include it for other survey/source/receiver types, so I rather open a PR and have it in for DC at least, so it can also server as an example if someone is interested in implementing it for more classes.


prisae:
### The PR basically changes this:

Selection_001

### to this:

Selection_002

### Example for mixed receivers:

Selection_003

lheagy:
This looks great @prisae! It is much nicer to get a bit of meaningful info from the survey items than just the class info

lheagy:
@jcapriot: what are your thoughts on possibly bringing this in on the next release?

jcapriot:
There's a few little formatting things I would personally change (such as the double "," in the printout of the receiver list)

prisae:
I guess the double comma comes from the fact that the dipoles and poles are two different lists, so the dipole ends and the dipole list ends, which leaves two coma. We could leave out the coma after each source/receiver altogether, there is a line break after all.

Anyway, this is just a suggestion, and it is also only implemented for Poles and Dipoles in `SimPEG/electromagnetics/static/resistivity` - ideally it would be implemented everywhere, for any source and receiver type (and Survey and Simulation type too). But I am missing the experience/knowledge with all the different possibilities in SimPEG.

Feel free to push to this branch if you want, or copy over to somewhere else (or review and tell me what should be changed).

prisae:
(But I think anything informative is better than the status quo which just prints the class info, little formatting improvements or not. So it could also be included and subsequently improved and expanded to other classes over time.)

jcapriot:
ok, I'm happy to bring it in now with these small changes. As a note for the future, I'd like to add some sort of functionality similar to numpy to allow for abbreviations of individual receiver objects when they contain many locations.

ngodber:
Love this change. Great to see this merged.

PR #911, opened 2020-07-20
Allows logic to cull unused kwargs from inputs to a function

jcapriot:
This should allow us to set some default kwargs for certain systems by only accepting them if they are valid to the function, and throwing a warning for invalid kwargs.

### DIFF
--- a/SimPEG/utils/solver_utils.py
+++ b/SimPEG/utils/solver_utils.py
@@ -93,6 +93,9 @@ def SolverWrapD(fun, factorize=True, checkAccuracy=True, accuracyTol=1e-6, name=
             _checkAccuracy(self.A, b, X, self.accuracyTol)
         return X
 
+    def __matmul__(self, other):
+        return self * other
+
     def clean(self):
         if factorize and hasattr(self.solver, "clean"):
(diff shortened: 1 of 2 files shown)


SimPEG/utils/solver_utils.py
```diff
@@ -76,14 +93,22 @@ def __mul__(self, b):
             _checkAccuracy(self.A, b, X, self.accuracyTol)
         return X
 
+    def __matmul__(self, other):
```
lheagy, line 96: :+1:

tests/utils/test_solverwrap.py
```diff
@@ -0,0 +1,64 @@
...
+        self.A = 0.5 * (A + A.T) + n * sp.eye(n)
+
+    def test_Solver(self):
+        x = np.random.rand(self.n)
+        b = self.A @ x
+
+        with self.assertWarns(Warning):
+            Ainv = Solver(self.A, bad_kwarg=312)
+        x2 = Ainv @ b
+        np.testing.assert_almost_equal(x, x2)
```
lheagy, line 23: cool - I didn't know about this!

PR #916, opened 2020-07-22
Update AUTHORS.rst

bluetyson:
Author update as suggested


jcapriot (approved):
Looks good. Also sorry about the delay on getting these in, we started having some issues testing that were very much slowing things down.

PR #917, opened 2020-08-04
change the way of exporting magnetic data

xiaolongw1223:
Here is a minor change: np.savetxt() cannot write the inclination, declination and strength of magnetic data, using the SimPEG utils instead.


domfournier (approved):
Thanks for this @xiaolongw1223

PR #918, opened 2020-08-12
add current argument to LineCurrent

jedman:


jcapriot:
Thanks for this. First off I'd like to apologize for waiting so long to get back on this.

 I'm in support of adding a current option, just want to make sure it's in the right place. @lheagy does this make sense here? or would something like it be better as part of the waveform? I know the `MagDipole` class has a `moment` property, for example, so a `current` makes sense here as well, however it might also be better described as a moment if the transmitter has multiple turns...

jcapriot:
Actually, looking more at the other lines, the `CircularLoop` source has options for `current` and `N` so I think `current` is appropriate here as well.

PR #921, opened 2020-09-04
Mu Inversion Test for FDEM

lheagy:
- improve tolerances on the mu inversion testing
- update the test to use the simulation API


jcapriot:
Can you add these tests to the travis scipt? I don't think they're being run.

Issue #920, opened 2020-08-27
Laguna El Maule Example results with opposites sign


jabascut:
I ran the Laguna El Maule example with Simpeg 0.14.2 and compared the inverted density model with Miller et al. 2016 (Paper presenting the gravity survey of Laguna El Maule). Both results are very similar, but the inverted densities of SimPEG have the opposite sign of the Miller et al. densities. The Miller et al. results, which show a negative density anomaly, have more sense because they are associated with a magmatic camera. Please could confirm if there is an error in the example? 

Regards

Juan A. Bascur

lheagy:
cc @domfournier , @craigmillernz: would you mind looking into this?

rolandhill:
Hi,
There appears to be an issue in the gravity inversion code rather than the example. I ran my own models and found density lows underneath gravity highs. It looks right, just inverted. If you check the SimPEG gravity example on the web site you will find it does the same thing. The gravity high is on the left and the gravity low on the right, however the low density model is on the left and the high one on the right (ie inverted).

https://docs.simpeg.xyz/content/tutorials/03-gravity/plot_inv_1b_gravity_anomaly_irls.html#defining-a-tensor-mesh

Regards,
Roland Hill

rolandhill:
... or is this because of a positive up convention?

jcapriot:
SimPEG does have a consistent right handed coordinate system convention where z is positive up.

craigmillernz:
This is a bug which we have fixed and should be in the next release.  SimPEG expects a "postive up" convention, which is a bit unintuitive for gravity but makes sense for other SimPEG internals.  If you import your data using the `readUBCgravityObservations`  function that reads in a UBC format obs or grv file (where negative residual gravity anomalies are negative sign) then simpeg will now flip the sign to handle it's internal conventions properly.

domfournier:
Will be addressed in PR #922

PR #922, opened 2020-09-10
Update io_utils.py

domfournier:
Flip sign of ubc gravity file for Cartesian convention


domfournier:
That should fix issues with the sign flip seen on the Laguna example.

craigmillernz (changes requested):
How about 
...between Cartesian coordinate system (internal SimPEG format that expects "positive up" gravity signal) and traditional gravity data conventions (positive down).  For example a traditional negative gravity anomaly is described as "positive up" in Cartesian coordinates and hence the sign needs to be flipped for use in SimPEG.

craigmillernz (approved):

craigmillernz:
@domfournier @jcapriot can this be closed? along with #920  ?

PR #923, opened 2020-09-10
Update plot_inv_1b_gravity_anomaly_irls.py

craigmillernz:
Flipped the gravity data sign to be consistent with new Cartesian format.


domfournier (approved):
Nice catch @craigmillernz

jcapriot:
@domfournier, Is this something that should be switched in the downloaded data file itself instead of changing it here?

jcapriot:
Actually... I don't think this needs to bet changed at all, the tutorials look correct to me...
https://docs.simpeg.xyz/content/tutorials/03-gravity/plot_inv_1b_gravity_anomaly_irls.html#sphx-glr-content-tutorials-03-gravity-plot-inv-1b-gravity-anomaly-irls-py
Shows the correct anomalies consistent with the conventions used in SimPEG

PR #928, opened 2020-09-27
WIP: SimPEG(emg3d) integration

prisae:
# Use `emg3d` as backend for `SimPEG`

## Note :: The original examples / curvenotes and instructions do not work any longer, as @seogi and I went down another path after the initial go.

---
Old Stuff

Currently you need to install the branch `dev-solver` form `emg3d` for this to work:
```
pip install git+https://github.com/empymod/emg3d@dev-solver
```

## How to use it

```
from emg3d.solver import Solver
from SimPEG.electromagnetics.frequency_domain.simulation import Simulation3DElectricField

sim = Simulation3DElectricField(
        solver=Solver,
        solver_opts={'max_workes': 4},  # <= You can define the maximum concurrent jobs here
        # all your usual inputs...
)
```

## `emg3d` solver restrictions:

- Currently only works with `TensorMesh`.
- Only works for 3D meshes.
- Only works for `SimPEG.electromagnetics.frequency_domain.simulation.Simulation3DElectricField`.
- **Currently ONLY works for forward modelling (`fields`),  not for inversion (`Jvec`, `Jtvec`).**
- Uses PEC boundary condition.
- The mesh should have a «_good_» number of cells in each direction.

Regarding «_good_» numbers: Each element of `n_cells` (to go with future discretize names) should be dividable as many times by 2 as possible, and ideally stop at a low prime (2, 3, 5, [7]). Good grid sizes (in each direction) up to 1024 are

- 2·2^{2, 3, ..., 9}: 8, 16,  32,  64, 128, 256, 512, 1024,
- 3·2^{2, 3, ..., 8}: 12, 24,  48,  96, 192, 384, 768,
- 5·2^{2, 3, ..., 7}: 20, 40,  80, 160, 320, 640,
- 7·2^{2, 3, ..., 7}: 28, 56, 112, 224, 448, 896,

and preference decreases from top to bottom row (stick to the first two or three rows if possible). Good grid sizes in sequential order: 8, 12, 16, 20, 24, 28, 32, 40, 48, 56, 64, 80, 96, 112, 128, 160, 192, 224, 256, 320, 384, 448, 512, 640, 768, 896, 1024.

## Note
- If you install `tqdm` (`pip install tqdm`) you get nice progress bars for concurrent jobs.
- `max_workers` in `solver_opts` can be used to define the maximum concurrent processes; default is 4.
- For all the other options in `solver_opts` see the [api docs for emg3d.solve](https://emg3d.readthedocs.io/en/stable/api/emg3d.solver.solve.html#emg3d.solver.solve).


prisae:
The fact that I have to check that `self._solutionType == "eSolution"` indicates that it would probably better live in `Simulation3DElectricField`. 

It could be implemented in `Simulation3DElectricField` in the following way:
```
def fields(self, m=None):
    if 'emg3d' in str(self.Solver):
        # Do emg3d stuff
    else:
        return super().fields(m)
````

Opinions, thoughts?

prisae:
Updated these two examples for `emg3d v0.16.0` (branch `dev-solver`) and `SimPEG v0.14.3` (branch `emg3d-solver`). Still works fine:
- https://iooxa.com/@prisae/emg3d-as-solver-for-simpeg/simpegemg3dtaketwo
- https://iooxa.com/@prisae/emg3d-as-solver-for-simpeg/classtest

prisae:
I just realize now that all tests are running... Can they be stopped for Draft-PR's @jcapriot ?

PR #929, opened 2020-09-29
Dev/jpe nsem data fix

jedman:
In modern (>= 1.16) versions of numpy, multifield indexing returns a view rather than a copy (https://numpy.org/devdocs/release/1.16.0-notes.html#multi-field-views-return-a-view-instead-of-a-copy), which breaks some code that converts between structured and unstructured arrays in SimPEG. 

Specifically, when trying to convert a subset of a structured array (the contents of a masked array, say, but not the mask) to an `ndarray`, the following error is raised:

`ValueError: Changing the dtype to a subarray type is only supported if the total itemsize is unchanged`

This PR has a fix, plus some new tests.


jcapriot:
Thanks for adding these changes, I was having issues myself with the rec arrays earlier.

PR #931, opened 2020-10-08
Improve mag3d computation over node locations and more versatile pf IO

dccowan:
This pull request tackled a couple issues:
1. The IO for potential fields can only read and write dobs. It can now read and write survey and dpred.
2. The forward mag problem was not always accurate for survey locations directly above edges and especially nodes. A new strategy and tolerances have been used to improve numerical accuracy.

### DIFF
--- a/SimPEG/potential_fields/magnetics/simulation.py
+++ b/SimPEG/potential_fields/magnetics/simulation.py
@@ -258,7 +258,8 @@ class Simulation3DIntegral(BasePFSimulation):
         # base cell dimensions
         min_hx, min_hy, min_hz = self.mesh.hx.min(), self.mesh.hy.min(), self.mesh.hz.min()
 
-        # comp. pos. differences for tne, bsw nodes
+        # comp. pos. differences for tne, bsw nodes. Adjust if location within
+        # tolerance of a node or edge
         dz2 = self.Zn[:, 1] - receiver_location[2]
         dz2[np.abs(dz2)/min_hz < tol2] = tol2 * min_hz
         dz1 = self.Zn[:, 0] - receiver_location[2]
--- a/SimPEG/potential_fields/magnetics/simulation.py
+++ b/SimPEG/potential_fields/magnetics/simulation.py
@@ -247,8 +247,8 @@ class Simulation3DIntegral(BasePFSimulation):
             Tz = [Tzx Tzy Tzz]
         """
         # TODO: This should probably be converted to C
-        tol1 = 1e-8   # Tolerance 1 for numerical stability over nodes and edges
-        tol2 = 1000. # Tolerance 2 for numerical stability over nodes and edges
+        tol1 = 1e-10    # Tolerance 1 for numerical stability over nodes and edges
+        tol2 = 1e-4     # Tolerance 2 for numerical stability over nodes and edges
 
         rows = {component: np.zeros(3 * self.Xn.shape[0]) for component in components}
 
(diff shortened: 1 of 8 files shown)


dccowan:
**ORIGINAL SIMULATION COMPARED AGAINST UBC-GIF**
image

**NEW SIMULATION COMPARED AGAINST UBC-GIF**
image

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -247,22 +247,32 @@ def evaluate_integral(self, receiver_location, components):
...
         nC = self.Xn.shape[0]
 
+        # base cell dimensions
+        min_hx, min_hy, min_hz = self.mesh.hx.min(), self.mesh.hy.min(), self.mesh.hz.min()
+
         # comp. pos. differences for tne, bsw nodes
-        dz2 = self.Zn[:, 1] - receiver_location[2] + eps
-        dz1 = self.Zn[:, 0] - receiver_location[2] + eps
+        dz2 = self.Zn[:, 1] - receiver_location[2]
+        dz2[np.abs(dz2)/min_hz < tol2] = tol2 * min_hz
```
domfournier, line 263: I am not a huge fan of this, because indexing with logical will likely hurt us when parallelizing. Better if we can stick we adding small constants.

SimPEG/utils/io_utils.py
```diff
@@ -309,16 +315,18 @@ def writeUBCmagneticsObservations(filename, data_object):
...
-    rxLoc = survey.source_field.receiver_list[0].locations
-
-    d = data_object.dobs
-    wd = data_object.standard_deviation
+    data = survey.source_field.receiver_list[0].locations
+    
+    if data_object.dobs is not None:
+    	data = np.c_[data, data_object.dobs]
+    
+    if data_object.standard_deviation is not None:
```
domfournier, line 323: Can `data_object.standard_deviation` be None?

domfournier (commented):
Good job on updating the importer.
Can you check if we can just use the tol value, and avoid doing logical indexing?

SimPEG/utils/io_utils.py
dccowan: Yes. I want you to be able to load predicted data as well.

SimPEG/potential_fields/magnetics/simulation.py
dccowan: It really isn't stable though. The errors above nodes and especially edges are too large. But we can try to figure out a way to accomplish both goals.

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -281,14 +291,14 @@ def evaluate_integral(self, receiver_location, components):
...
+        r2 = np.sqrt(dz2dz2 + R1)
+        r3 = np.sqrt(dz1dz1 + R1)
+        r4 = np.sqrt(dz1dz1 + R2)
+        r5 = np.sqrt(dz2dz2 + R3)
+        r6 = np.sqrt(dz2dz2 + R4)
+        r7 = np.sqrt(dz1dz1 + R4)
+        r8 = np.sqrt(dz1dz1 + R3)
 
         # compactify argument calculations
         arg1_ = dx1 + dy2 + r1
```
domfournier, line 304: Can you try to put the eps on the args instead? Looks like we were not fully protected before...

SimPEG/potential_fields/magnetics/simulation.py
domfournier: Just to leave a trace, @dccowan ran it both ways on his machine on a 500 x 200k size problem. No significant change in runtime, but definitely an improvement in accuracy. Well done.

domfournier (approved):
Good on my side.

jcapriot:
I had brought this up before, but can you check if the same error is present in the GG calculation?

jcapriot (commented):
the changes look fine to help with the stability, just need to fix the merge error at the end of the file.

PR #932, opened 2020-10-16
Load ubc dcip datatypes and plot 3d pseudosections

dccowan:
This pull request accomplishes two things:
- A versatile read function for survey, dpred and obs files for dcip3d
- The ability to plot a pseudo-section directly for 3d dcip data. You can either plot all data are plot data within some distance of a user-defined plane.
- The addition of a primary-secondary approach for 3D simulations (WILL BE IN NEW PULL REQUEST)

**Remaining Items:**

- [x] Split standard formulation and fictitious source formulation into separate simulation classes
- [x] Finalize dcip3d and dcipoctree IOs for DC voltage, IP apparent chargeability and IP secondary voltage and make test
- [ ] Write tests for fictitious source formulation (WILL BE IN NEW PULL REQUEST)
- [x] Ensure all examples and tutorials run without error
- [x] Add a utility which assigns dc data to ip receivers for modeling apparent chargeability directly


jcapriot:
Hey @dccowan, I'm going through this and doing some edits myself, but before I get too far in on the primary-secondary approach. I had a few thoughts/questions.

1: When I had worked this out in my own head I was imagining more of it being taken care of on the source side of the operation. Perhaps even defining a new source class that explicitly used the primary-secondary formulation, which is slightly more in line with, for example, the MagDipole FDEM source. Alternatively, it could just be an option on the Pole and Dipole source. I do see why it was implemented as it is though as you only want 1 call to construct the A matrix for multiple sources.

2: I see some operations in here that set the background model explicitly based on the current resistivity model. Most of the other code that we have that does something like this requires the user to explicitly set the primary model used for the simulation, for example in the MT code base.

3: How accurate is this in the presence of topography? I see a lot of conditions in the code to account for topography (or at least a variable source depth), which is set up to compute the image source as if the source was within the conductive medium.

jcapriot:
Also, the utility functions look really well done!

sgkang:
Hi @jcapriot, interesting points. Below are my two cents.

> 1: When I had worked this out in my own head I was imagining more of it being taken care of on the source side of the operation. Perhaps even defining a new source class that explicitly used the primary-secondary formulation, which is slightly more in line with, for example, the MagDipole FDEM source. Alternatively, it could just be an option on the Pole and Dipole source. I do see why it was implemented as it is though as you only want 1 call to construct the A matrix for multiple sources.
> 
Doing `Ainv*RHS` was much faster than doing for loop to solve multiple right-hand side, and that might be the reason why we did that. 

> 2: I see some operations in here that set the background model explicitly based on the current resistivity model. Most of the other code that we have that does something like this requires the user to explicitly set the primary model used for the simulation, for example in the MT code base.
> 

It will be nice to have this set up.  So, for instance considering a monitoring set up, one could setup more complicated primary resistivity structure. 

> 3: How accurate is this in the presence of topography? I see a lot of conditions in the code to account for topography (or at least a variable source depth), which is set up to compute the image source as if the source was within the conductive medium.

Great point. This was something that I was not sure about, and worthwhile to test. There is half-sphere solution that we can test this.

jcapriot:
I think we should separate the simulation side of the changes then from the io/utils side of the PR so we can discuss this issue on the specific part.

lheagy:
Adding my appreciation for the utils and nice docstrings -- thanks @dccowan! 

As for the questions / comments @jcapriot has raised, here is my 2 cents: 

> 1: When I had worked this out in my own head I was imagining more of it being taken care of on the source side of the operation. Perhaps even defining a new source class that explicitly used the primary-secondary formulation, which is slightly more in line with, for example, the MagDipole FDEM source. Alternatively, it could just be an option on the Pole and Dipole source. I do see why it was implemented as it is though as you only want 1 call to construct the A matrix for multiple sources.

In terms of code organization, I would be supportive of moving more of the primary-secondary work to the source-term. The primary-secondary approach describes how to construct the source-fields. We could still have a light-weight check in the construction of the RHS to see if it needs access to the system matrix or other aspects of the simulation to try to reduce the number of calls to construct the A matrix

> 2: I see some operations in here that set the background model explicitly based on the current resistivity model. Most of the other code that we have that does something like this requires the user to explicitly set the primary model used for the simulation, for example in the MT code base.

+1 for giving the user the option to fix this. We could still provide the option to have the background model be set dynamically (as it is now), but I think that should be an explicit choice by the user. 

> 3: How accurate is this in the presence of topography? I see a lot of conditions in the code to account for topography (or at least a variable source depth), which is set up to compute the image source as if the source was within the conductive medium.

Seconding @sgkang's comments here -- it would be very interesting to see a test of this!

lheagy:
What would be good next steps here? should we sketch out pseudocode for the source-term?

jcapriot:
After talking with Devon, my current thoughts on getting this in to the current code base will be to punt the primary-secondary/fictitious source down the road and then make our current thoughts into a separate PR that would address all of the things we would like to do for those source classes.

Part of that is to first of all split this current implementation into a specific simulation that only does that formulation before pulling this into master.

dougoldenburg:
I'm pretty new at this game but I have a couple of comments/questions:

(a) are there document(s) that encapsulate the details of the mathematics connected with code? The python code is sort of readable at a high level (even for a novice) but some things are a mystery. For example, the choice of a minus sign on line 456. 

(b) In reference to item (a) we do have resources eg pixels and their neighbors, as well as EM modelling papers. How do we tie the software to those references so that a user going through the notebook could follow the base mathematics along with the python code? 

(c) For me, the name fictitious sources, might be not be the best terminology. In DC fictitious sources are often equated to image sources. I would suggest "Corrective Sources" or "Auxilary Sources). We did a fair bit of work on this in the early 2000's when we were grappling with large ground-loop surveys. The best published article is the paper: Geophys. J. Int. (2007) 171, 550–564.
[HOScorrective_source.pdf](https://github.com/simpeg/simpeg/files/5520921/HOScorrective_source.pdf)

(d) a related item that Joe has already brought out is the aspect of topography. It seems that the current formulation is likely to have significant errors. A corrective source approach might we worth investigating if the "analytic" result is replaced by a fine scale Octree numerical solution. It might be worthwhile to consider this.

jcapriot (approved):
@dccowan I've gone through again and edited the branch.
I've noticed that there were some added data and model files that were created by running tutorials, please watch out for accidentally adding these files to the git history.

jcapriot:
As soon as I double check my changes didn't break anything, we'll get this merged.

jcapriot:
Also looks like the output formatting for the gg observations file is off, based on the gif tools documentation here (https://giftoolscookbook.readthedocs.io/en/latest/content/fileFormats/ggfile.html#), I'll switch that too.

PR #936, opened 2020-10-26
fix mapping in LinearSimulation

### LINKED ISSUE #933: LinearSimulation seems to ignore model_map argument
It seems that model_map is not used anywhere in the LinearSimulation class... I tried passing it a maps.Wires but it has no effect (in the examples and tutorial, it is tested with an IdentityMap with a fully active vector, thus it could not be detected.)

I am still new to the simulation class, so let me know if I am mistaken.
If true, would the correct fix will be to write, in `simulation.py`: `self.model = self.model_map * m` instead of `self.model = m`

thast:
Resolve #933 

Mapping was not taken into account in `linearSimulation`.

### DIFF
--- a/SimPEG/simulation.py
+++ b/SimPEG/simulation.py
@@ -538,28 +538,28 @@ class LinearSimulation(BaseSimulation):
         self._G = G
 
     def fields(self, m):
-        self.model = m
-        return self.G.dot(self.model_map * self.model)
+        self.linear_model = m
+        return self.G.dot(self.linear_model)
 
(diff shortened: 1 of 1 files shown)


jcapriot:
@thast 
I don't think this is necessary, as the get operation on the `self.model` should already handle doing the model mapping

jcapriot:
actually I see what the true bug is now...
The model is stored on the `LinearSimulation` class as `linear_model` and thus assignments and gets should be changed from `self.model` to `self.linear_model`

PR #938, opened 2020-10-27
add the Wolfe Curvature condition to the optimization 

thast:
This adds the Wolfe curvature criteria to the optimization checks for line search.

Wolfe conditions are composed of two rules for line search (see https://en.wikipedia.org/wiki/Wolfe_conditions). Only the first one was implemented; this is the second condition.
Further reading: [Nocedal and Wright; section 3.1 ](http://www.apmath.spbu.ru/cnsa/pdf/monograf/Numerical_Optimization2006.pdf)

I do not think that at this stage with current optimization it makes a big difference, so we can discuss making it optional, but it can become handy once we move towards a more advance line-search scheme.


jcapriot:
Just to double check and raise it to everyone's attention, this will essentially double the cost of the line search with these conditions, as it requires calculating the derivative as well for each test point.
My initial inclination would be to leave it as optional and disabled by default. If we are not concerned about the line search time however it should be enabled by default (i.e. for potential fields where doing a single forward modeling is not very expensive in the first place).

thast:
I am in favor of leaving it as optional as well. I'll make the necessary changes.

SimPEG/optimization.py
```diff
@@ -581,11 +596,18 @@ def modifySearchDirection(self, p):
             :return: (xt, passLS) numpy.ndarray, bool
         """
         # Projected Armijo linesearch
-        self._LS_t = 1
+        self._LS_t = 1.
         self.iterLS = 0
         while self.iterLS < self.maxIterLS:
             self._LS_xt = self.projection(self.xc + self._LS_t * p)
-            self._LS_ft = self.evalFunction(self._LS_xt, return_g=False, return_H=False)
+            self._LS_ft, self._LS_ft_descent = self.evalFunction(
```
lheagy, line 603: in order to make this optional, would it be better instead to loop through the `self.stoppersLS` and have this be a method on each stopper?

thast:
@jcapriot @lheagy : I made the Wolfe Curvature optional. I think this one is ready for review (only codecov is unhappy for a 0.01% coverage diff)

jcapriot:
Would you mind adding a test that uses the WolfeCurvature in a line search? Probably just a simple linear inversion with a non-linear mapping of some sort would do.

jcapriot (approved):
Looks good @thast!

PR #940, opened 2020-10-27
plot_utils: add defaults for all contour, streamplot etc. that can be overwritten by inputs

thast:


jcapriot:
These defaults are likely to cause a similar issue to the ones we've been discussing in discretize. see the discussion in https://github.com/simpeg/discretize/pull/227. 

Do not use vmin, vmax as defaults, also this change does not use the created norm from a linear/log scale as a default.

thast:
thanks for the catch @jcapriot (trying to parse out changes from my personal branch, some might date a bit).
See what you think of this next iteration.

thast:
@jcapriot : with the single line, what happens if you pass a 'level' or a 'zorder' argument? Does it overwrite the default or does it throw an error? I am more on the side to give maximum control to the user to overwrite defaults.

jcapriot:
yep, if the option is in the second part, it will overwrite the "default" one.

Issue #941, opened 2020-10-30
Add license and copyright notice to all source files


lheagy:
We should add a copyright notice and license information (names and terms) in every `.py` file. This notice should be at the top of each file 
```
# Copyright (c) YEAR The PROJECT Developers
# Distributed under the terms of the MIT License. 
# SPDX-License-Identifier: MIT
```

Where `PROJECT` 
- [ ] `SimPEG`
- [ ] `discretize`
- [ ] `geoana`
- [ ] `pymatsolver` 

This has been surfaced by @prisae, and again re-iterated by @leouieda in fatiando/maintenance#10 in an unfortunate experience where someone has copied the entire codebase without attributions... 😞

leouieda:
Feel free to use the code in the Fatiando issue to do this in bulk 🙂

PR #942, opened 2020-10-31
Add license and copyright to every .py file

leouieda:
Include the information in case SimPEG code is copied without the license file into another project. This is recommended practice for the GPL at least: https://www.gnu.org/licenses/gpl-faq.en.html#NoticeInSourceFile

Related to #941

Couldn't resist touching every file in the repository as my first SimPEG contribution 🙂


prisae:
One point that might be worth thinking about is if the current year has to be included. I currently have it as well in my project, but I sort of don't like that I then have to update each file in a new year, and I am thinking of changing it. However, here we have not even implemented it yet, so the right point to discuss.

Current version

    # Copyright (c) 2013-2020 The SimPEG Developers. 

As far as I am aware, the MIT license contains the following line:

    # Copyright (c) <year> <copyright holders>

Which would make the following sufficient

    # Copyright (c) 2013 The SimPEG Developers.

leouieda:
Yep, seems like the time to do it would be now. Let me know and I’ll make the changes 🙂

lheagy:
It looks like using just the first year is sufficient (https://choosealicense.com/licenses/mit/), so sticking with 
```
# Copyright (c) 2013 SimPEG Developers.
```
seems to be simplest so that we don't have to update each year! And I don't think we need the "The" at the beginning? 

Thanks so much for this contribution @leouieda!!!

leouieda:
👍🏽  done and rebased to master. Hopefully this fixes the broken tests 🤞🏽

prisae:
It is a difficult topic. I think it is fine, but don't take my word for it. You can read many things online.

Some say that
> Copyright 1995-2020

Is wrong, as a hyphen has no meaning in law. It has to state explicitly each year, in which it was released and had changes to previous releases. E.g.,
> Copyright 1995, 1996, 2000, 2011, 2012, 2013, 2014, 2015, 2020

So it would not have to be changed each year, just each time it is changed and it is a new year.

Anyhow, the purpose of including it in each file is so that the license type is there, and there is attribution to SimPEG, not to battle legal fights. So again I think just having 2013 is fine. The main license file in the repo can be updated... Just my 2c.

leouieda:
The documentation seems to be hanging on the gallery generation. I don't think I've broken anything with the comments so maybe this is coming from somewhere else. Any insights?

jcapriot:
I'd love to get this in soon, we should get this updated for the current master branch.

leouieda:
@jcapriot rebasing this branch is a bit of a nightmare because of the conflicts. It's probably easier to close this and run the formatting script again fresh once you decide that this can be merged (and then avoid merging anything else until this is done).

jcapriot:
@leouieda That sounds like a good plan! I agree it has gotten out of date with newer files and it's probably easier to just not worry about the merge conflicts

leouieda:
Happy to submit this again when you’re ready for it 🙂 Just let me know

PR #944, opened 2020-11-05
Apparent resistivity and phase receivers 

lheagy:
update derivatives and adjoint for apparent resistivity and phase receivers


JKutt (approved):
Right on, looks good! Thanks Lindsey. I'll whip up the phase side of things over the weekend.

PR #945, opened 2020-11-07
light reorganization of nsem receivers and derivatives

lheagy:


JKutt (approved):
Looks great! I appreciate the things like naming the variables fully like "denominator" and "numerator" :)

PR #946, opened 2020-11-07
Simulation mt - Part1

JKutt:
Start of the natural source revamp


JKutt:
You were right Lindsey, there was an error with the importing of NSEM 1D stuff even after the new merge :) Got it figured now.

SimPEG/electromagnetics/stretched_exponential/__init__.py
```diff
@@ -0,0 +1,6 @@
+from __future__ import absolute_import
```
lheagy, line 1: @JKutt: did you want to include the stretched exponential work in this pr? We could work through it, but it does expand the scope quite a bit

SimPEG/electromagnetics/natural_source/receivers.py
```diff
@@ -29,6 +31,8 @@ class BaseRxNSEM_Point(BaseRx):
         {
             "real": ["re", "in-phase", "in phase"],
             "imag": ["imaginary", "im", "out-of-phase", "out of phase"],
+            "apparent_resistivity": ["app_rho"],
```
lheagy, line 34: here, for consistency, I would suggest we consider 
`"apparent resistivity" : ["apparent_resistivity", "app_rho"]`

(e.g. have the default be with the space). What do you think of that?

SimPEG/electromagnetics/natural_source/receivers.py
```diff
@@ -70,14 +74,20 @@ def _locs_e(self):
         if self.locations.ndim == 3:
             loc = self.locations[:, :, 0]
         else:
-            loc = self.locations
+            if self.locations.shape[1] == 6:
```
lheagy, line 77: Just looking back on this, I am not entirely sure why we would end up with `locations` being 6 wide... is this if the e / b receivers are in different locations? If that is the case, I think it might be better to instead have keywords for the e-locations, and b-locations?

SimPEG/electromagnetics/stretched_exponential/__init__.py
JKutt: @lheagy I entirely forgot I brought this into the branch. I would say it is out of scope. Let me remove it for now unless others are interested?

SimPEG/electromagnetics/natural_source/receivers.py
JKutt: @lheagy this was a hack for that case exactly :) Some of the data I work with is roving B and stationary E. I can certainly take the time to add the keywords. I think it would be worth given the rise in airborne AMT techniques.

SimPEG/electromagnetics/stretched_exponential/__init__.py
lheagy: :+1: If you are game to move it elsewhere, I think that will help keep this one scoped more tightly. Thanks @JKutt!

SimPEG/electromagnetics/natural_source/receivers.py
lheagy: :+1: And we also don't have an explicit base station either, which might be interesting to have the possibility for. But for now, we can think through the `locations` assignment. My initial thought would be to have it similar to how we instantiate the dipole receivers. maybe: 

```
locations=None, locations_e=None, locations_b=None
```

and then the user can set either of `locations` (which then assumes e & b are co-located) or `locations_e`, `locations_b` which can be distinct

SimPEG/electromagnetics/natural_source/simulation.py
```diff
@@ -317,7 +390,7 @@ def fields(self, m=None):
 ###################################
 # 3D problems
 ###################################
-class Simulation3DPrimarySecondary(BaseNSEMSimulation):
+class Simulation3DPrimarySecondary(BaseFDEMSimulation):
```
lheagy, line 393: @JKutt: I reverted this just for now to make sure that the tests still pass with a light re-naming, but I think this is a good call- we are not too far off. I will start (or revive) an issue where we can chat through the necessary changes

lheagy:
I started an issue #947 to sketch out a potential path towards having the NSEM implementation rely more heavily on the FDEM class. See what you think when you have a chance!

SimPEG/electromagnetics/natural_source/receivers.py
JKutt: Sounds great. I like it. I've also had plans of evaluating the E receiver as dipole and this plays right into that. But sometime down the road.

grosenkj: `locations` does not assume that e and b are co-located. It allows for the E and B fields to be in separate locations with an additional dimension added to the matrix ([nr_rx, 3, 2] shape (dim=3)). Then the e_loc = locations[:,:,0] and b_loc = locations[:,:,1] and can be assigned independently of each other. This isn't a well-documented feature of the receiver class but I support the idea of having keyword arguments @lheagy. 
I do also think adding better mechanisms for the locations would be great, as @JKutt mention above, especially when using a dipole or not-aligning with grid data.

JKutt:
@lheagy , not sure if talked through this but now that when we get the indices "ind" when setting the fields is a tuple and this lines breaks:

https://github.com/simpeg/simpeg/blob/3b2eef3173b7b789d09e0683adb75f8e6f32bf8f/SimPEG/fields.py#L187

suggestions, or did I mess this up somewhere?

SimPEG/electromagnetics/natural_source/receivers.py
```diff
@@ -29,318 +31,77 @@ class BaseRxNSEM_Point(BaseRx):
...
     )
 
-    def __init__(self, locs, orientation=None, component=None):
+    def __init__(
+        self,
+        locations=None,
+        orientation=None,
+        component=None,
+        locations_e=None,
+        locations_b=None,
```
lheagy, line 49: should we stick with `locations_h` here? Internally, we are working with `h`, so might be more consistent?

SimPEG/electromagnetics/natural_source/receivers.py
```diff
@@ -29,318 +31,77 @@ class BaseRxNSEM_Point(BaseRx):
...
-
-    @f.setter
-    def f(self, value):
-        self._f = value
+        BaseRx.__init__(self, locations)
 
     def _locs_e(self):
-        if self.locations.ndim == 3:
-            loc = self.locations[:, :, 0]
+        if self.locations_e is None:
```
lheagy, line 62: do we still need `_locs_e` or should we just put this logic in the `__init__` and work only with `locations`, `locations_e`, `locations_h` (or `b` depending on what we decide to go with)

SimPEG/electromagnetics/natural_source/receivers.py
```diff
@@ -29,318 +31,77 @@ class BaseRxNSEM_Point(BaseRx):
         {
             "real": ["re", "in-phase", "in phase"],
             "imag": ["imaginary", "im", "out-of-phase", "out of phase"],
+            "apparent resistivity": [
```
lheagy, line 34: Should this be on the base? or should we just over-write `component` for the complex resistivity receiver? I am inclined to do the latter since for example `apparent resistivity` isn't a valid `component` choice for the impedance receiver.

SimPEG/electromagnetics/natural_source/receivers.py
JKutt: I like consistency :), I'm game with changing to h. I can make these changes on my next push.

SimPEG/electromagnetics/natural_source/receivers.py
JKutt: Just to make sure I follow; we strictly use locations_e & locations_h but if initiated with locations, the initialization will parse it into locations_e and locations_b, correct?

SimPEG/electromagnetics/natural_source/receivers.py
JKutt: make sense to me! 👍

SimPEG/electromagnetics/natural_source/receivers.py
lheagy: currently, we have 
```
def __init__(
        self,
        locations=None,
        orientation=None,
        component=None,
        locations_e=None,
        locations_b=None,
    ):
        self.orientation = orientation
        self.component = component
        self.locations_e = locations_e
        self.locations_b = locations_b

        if locations is None:
            locations = np.hstack([locations_e, locations_b])

        BaseRx.__init__(self, locations)

    def _locs_e(self):
        if self.locations_e is None:
            if self.locations.ndim == 3:
                loc = self.locations[:, :, 0]
            else:
                loc = self.locations
            return loc
        else:
            return self.locations_e

    def _locs_h(self):
        if self.locations_b is None:
            if self.locations.ndim == 3:
                loc = self.locations[:, :, 1]
            else:
                loc = self.locations
            return loc
        else:
            return self.locations_b
```

What I think we should do is something like 
```
def __init__(
        self,
        locations=None,
        orientation=None,
        component=None,
        locations_e=None,
        locations_b=None,
    ):
        self.orientation = orientation
        self.component = component

        # check that either locations_e and locations_b provided or locations provided and error if not
        
       if locations_e, locations_b provided: 
            self._locations_e = locations_e
            self._locations_b = locations_b
        
        elif locations is not None: 
            # check shape of locations, 
            if width 2 or a list of length 2: 
                  self._locations_e = locations[:, 0]
                  self._locations_b = locations[:, 1]
            if width 1:
                  self._locations_e = locations
                  self._locations_b = locations
            
        locations = np.hstack([self._locations_e, self._locations_b])

        BaseRx.__init__(self, locations)

    def locations_e(self):
           return self._locations_e

    def locations_b(self): # or h
            return self._locations_b
```

jcapriot: There should also be a check to ensure they're the same length.

SimPEG/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -0,0 +1,433 @@
...
+        :return numpy.ndarray Z: complex impedances at surface for all frequencies (nFreq,)
+        """
+
+        omega = 2 * np.pi * fvec
+        n_layer = len(sigma_1d)
+
+        # Bottom layer quantities
+        alpha = np.sqrt(1j * omega * mu_0 * sigma_1d[-1])
+        ratio = alpha / sigma_1d[-1]
+        Z = -ratio
```
lheagy, line 115: I _think_ this is the opposite convention to what we have currently in the 3D code. (for a halfspace, this will give a +45 degree phase). @grosenkj: do you remember off-hand what the convention is in the 3D code or @JKutt: you might know from your recent simulations / inversions?

grosenkj: Yes, I do remember the 3D convention. I made it consistent with SimPEG so `x` is easting and `y` northing and `z` should is positive up. That is like other codes? 
So `Zxy` has a -135 deg phase and `Zyx` has +45deg for a half-space in the 3D code. This is in contrast with the MT world where usually X is North and y is East and Z positive down (for example in EDI files). For the 1D case, since `Zxy=-Zyx` it is neater to have the phase in the 1-quadrant. 
@lheagy does this address your question?

lheagy: Thanks @grosenkj, I think that makes sense. I am looking at developing an implementation that uses boundary conditions to provide the RHS (rather than using the primary-secondary approach), so will just need to do the book-keeping to make sure everything is consistent

JKutt:
@lheagy the test script `test_Problem1D_AnalyticVsNumeric.py` fails because of an assumption that the given mesh has vectors in the x, y directions at line:
https://github.com/simpeg/simpeg/blob/86b93e9a81e3e1369f678989e7120582743822d2/SimPEG/electromagnetics/natural_source/sources.py#L65

I assume we likely have to change something in the `setup1DSurvey` method where the mesh is created?

lheagy:
Sorry for the delay @JKutt! hmm, it seems there are a couple odd things going on here. The first error I get is 

```
 AttributeError: 'Planewave_xy_1Dprimary' object has no attribute '_sigma1d'

../../../../SimPEG/electromagnetics/natural_source/sources.py:57: AttributeError
```

and then within that there is also the one you pointed to. For a 1D simulation only, I don't think we should need a 3D mesh.

lheagy:
This might be a bit simpler to navigate through if we remove the dependence on `nsem.utils.test_utils.setup1DSurvey` and just set up the survey on our own for these?

JKutt:
That sounds like a great idea! I'll give it a shot.

JKutt:
@jcapriot @lheagy The head banging bug that has eluded me appears to be that variable `i` is not updated in the proceeding loop.... \face palm..

https://github.com/simpeg/simpeg/blob/4284122796d8e5b9617fdc469d822a815392c84e/SimPEG/electromagnetics/frequency_domain/simulation.py#L201

JKutt:
@jcapriot, the following is what I removed to get the fdem & nsem tests to pass. This shouldn't have much repercussion down the road should it?

https://github.com/simpeg/simpeg/blob/a67c8c9a5bad705b2e64cd4280b17d3d5dd1a1dc/SimPEG/electromagnetics/frequency_domain/receivers.py#L106-L109

JKutt:
@lheagy and @jcapriot the second item failing doesn't seem like something i can remove right now like the 1D example. At first it looked like the path was wrong, but it fails entirely with even a proper path. Also this does appear to be affecting all PR's right now. 
https://github.com/simpeg/simpeg/blob/9c29401bd768a4560b6f7555e5c4644758ca610b/tests/docs/test_docs.py#L7-L37

jcapriot:
I think all that is really left at this point is to do a pass through the code to clean up most of the lines that have been commented out.

JKutt:
Awesome stuff @jcapriot. I really like that last commit. All into a single class!

PR #949, opened 2020-11-11
PR for updated fields

jcapriot:
This PR updates the fields, and sources to handle multiple field outputs per source, (i.e. both polarizations of an MT source).
The Rx eval currently runs (but I didn't extensively test that I got the hx_px, hy_py, ex_px, etc all in the correct order, but I think it's right). This eval should be easy to extend to the other receiver classes


JKutt (approved):
Looks great. I'll get to extending this to the other receiver eval's.

PR #951, opened 2020-11-16
reduce duplication in the 1D source definitions

lheagy:
- deprecate `AnalyticPlanewave1D` and `Planewave_xy_1DhomotD` in favour of `Planewave1D` (which is currently a pass through that only tracks frequency)
- remove code that duplicated the base FDEM survey behaviour from the `Survey1D` for the MT code


JKutt (approved):
Looks good to me! We chatted about changing the names anyway one one point 👍

PR #954, opened 2020-11-17
Simulation mt tree

jcapriot:
Generalize background code to also handle OcTree meshes


JKutt (approved):
Super cool Joe 👍 . I'll run the direct comparison with UBC here shortly!

PR #958, opened 2020-11-28
Mag. Simulation: fix if statement for self.chi

thast:
When chi is None, `properties` throw an error on the call of `self.chi`, thus the `if` statement ends in an error when true. (see screenshot below)
this is fixed by using the `getattr` function.
Also, this PR fixes the size of model when the chiMap is a `maps.wires`
image


thast:
An actually, should that not be `getattr(self, 'model', None)` instead? I am having issue with joint inversion because chi and model are of different size and `properties` again is unhappy.

jcapriot (commented):
Shouldn't we just do a `self.model = m` in all cases (similar to the JtJ diag routine)?

thast:
@jcapriot, I stayed close to the original formulation (with the `np.zeros`) but I agree with you. It seems to work the same. @fourndo @domfournier might have an opinion or insights?

jcapriot:
I think this assignment is better, as it would properly update the chiDeriv property as well.

jcapriot:
Also, I was suggesting always assigning `self.model = m` without even doing the check for `self.chi`

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -196,8 +196,8 @@ def Jtvec(self, m, v, f=None):
     @property
     def fieldDeriv(self):
 
-        if self.chi is None:
-            self.model = np.zeros(self.G.shape[1])
+        if getattr(self, 'chi', None) is None:
+            self.model = m
```
jcapriot: I would guess the `self.model = m` line would probably not work right here, as we never pass `m` to fieldDeriv (which might not be a good thing).

thast: I noticed that, it is already been fixed.

thast: For the same reason, I kept the `if getattr(self, 'chi'` ` for that specific case.

PR #939, opened 2020-10-27
PGI

thast:
# PGI

## Status: ready for review
- functionalities: 100 %
- Ease of use: 95%
- Documentation: 90%

## References and Illustrations 
_Thibaut Astic, Douglas W Oldenburg, A framework for petrophysically and geologically guided geophysical inversion using a dynamic Gaussian mixture model prior, Geophysical Journal International, Volume 219, Issue 3, December 2019, Pages 1989–2012, https://doi.org/10.1093/gji/ggz389_

_Thibaut Astic, Lindsey J Heagy, Douglas W Oldenburg, Petrophysically and geologically guided multi-physics inversion using a dynamic Gaussian mixture model, Geophysical Journal International, , ggaa378, https://doi.org/10.1093/gji/ggaa378_

image
image

### Glossary
**PGI**: Petrophysically and Geologically Guided Geophysical Inversion
**GMM**: Gaussian Mixture Model

### regularization
- _adds_ `pgi.py`: 
   - `PGI` and `SimplePGI` regularization (similar to Tikhonov and Simple reg.) 
   - `SimplePGIwithRelationships` (for nonlinear relationships between physical properties

### Directives
- _changes_ `directives.py`: Addition of directives for general joint inversion purposes: 
   - Multiple Targets Misfits monitoring (`MultiTargetMisfits`), 
   - Estimate alphas and data misfit multipliers by eigenvalue (`AlphasSmoothEstimate_ByEig`  `ScalingEstimate_ByEig`, similar to `BetaEstimate_ByEig`) 
   - Schedule for data misfit multipliers as proposed in Astic et al. 2020 (`JointScalingSchedule`, similar to `BetaSchedule`)

- _adds_ `pgi_directives.py`: directives specific to the PGI framework:
   - `PGI_UpdateParameters`: update mref, Ws and GMM
   - `PGI_BetaAlphaSchedule`: update beta and alpha_s, as proposed in Astic & Oldenburg 2019
   - `PGI_AddMrefInSmooth`: add mref to the smoothness when mref is stable, as proposed in Astic & Oldenburg 2019

### Utils:
- _adds_ `pgi_utls.py`: utils for PGI
   - `WeightedGausianMixture`, `GaussianMixtureWithPrior` etc.: adaptation of the `GaussianMixture` class from scikit-learn to the PGI framework
   - make_PGI_regularization, make_SimplePGI_regularization: utils to help in building PGI and Smoothness regularization with multiples physical properties

### Examples:
[...]

PR #959, opened 2020-12-05
Tutorial: Joint PGI Grav + Mag + Octree

thast:
This adds a "Tutorial: Joint PGI Grav + Mag + Octree" to the PGI PR #939.

Note: on my laptop, this tutorial runs fully in under 8 minutes.

image
image
image


RichardScottOZ (commented):
This tutorial shows

it should be, at the start

PR #963, opened 2021-01-13
Merging `simpegEM1D` into `simpeg`

sgkang:
This was written by @jcapriot (transferred from #1013)

This represents my major edits for the em1d branch. It only changes the normal simulation routines to be more flexible and more optimized under the hood to better support the simpeg sources and receivers.

items left to do on the em1d branch include to:
- [ ] Perform further tests with the on-time measurements with arbitrary waveforms. (The magnetic field looks to be shifted, but dbdt looks good).
- [ ] Update the psuedo 3D inversions (the stitched inversions) to this way ( @sgkang we should only need to change a few lines as you and I discussed).
- [ ] better integrate the IO routines


jcapriot:
These are some thoughts I had about bringing in the EM1D code that I will transfer to here:

 - Remove individual frequency_domain_1D, time_domain_1D modules and merge them into the existing frequency_domain/time_domain modules in their own simulation1D.py files.
 - Make use of existing receiver classes (i.e. PointElectricField, PointMagneticField, PointMagneticFluxTimeDerivative).
 - "known systems" seems completely unused, what is it's purpose?
 - make use of waveforms in time_domain/sources.py (these could also be moved into their own waveforms.py file
 - remove `from something import *` style imports outside of `__init__.py` scripts.
 - The names of the `Simulation` classes should be renamed to the established style in SimPEG
 - (kinda big picture) Can we get the Cole-cole and VRM as maps instead of properties on the simulation? (This would require that we are able to accept frequency dependent models

Other things might come to mind, but these are my first thoughts.

dccowan:
I think most of the Feb 17 comments are doable. One of the things that could be a little tricky is organizing waveforms. The waveform classes for 1D and 3D simulations are very different. Do we think it is possible to have a waveform class that works for either simulation, or do we need to keep 1D and 3D waveform classes separate?

dccowan:
There are some challenges with trying to amalgamate the 3D and 1D stuff to live in the same module when it comes to sources and receivers.

**Receivers:** We can use a similar naming convention for the 1D and 3D receiver classes. But unlike the 3D receivers, the 1D codes do not have a comprehensive set of receiver classes. We can only simulate H(w), B(w), h(t), b(t), dh/dt and db/dt but we haven't implemented e-field receivers.

**Sources:** Ultimately the 3D sources can be must more generalized in terms of their orientation compared to the 1D sources.

sgkang:
Most of the errors in the tests seems stemming from out-dated geoana version (for norch?): 

image

jcapriot:
> Most of the errors in the tests seems stemming from out-dated geoana version (for norch?):
> 
> image
> 
> this seems the case. If I pip install geoana in matrix.yml file, the right version of geoana is installed.
> 
> @jcapriot: would be nice if we can update conda-forge one ....

did you try setting a minimum version for `geoana`?

sgkang:
transferred to #1059

PR #965, opened 2021-01-14
Add conda-forge badge

prisae:


thast (approved):
that looks good! Thanks @prisae , great idea!

adjafari:
I had many difficulties to install simpeg on ubuntu 20. Can somebody know ho do?

prisae:
How do you use Python? With conda it should be straight forward. I am on Ubuntu 20, and it is a matter of seconds to install it.

thast:
Hi @adjafari, welcome.
For discussions and help, we have dedicated platforms. Please feel free to reach us on them whenever you are facing challenges:
1) a slack group that you can join following this link: http://slack.simpeg.xyz/
2) a discourse, https://simpeg.discourse.group/

This will make this discussion more visible; thus, more people can help, and people experiencing similar issues will see it.

Thank you for your interest.

PR #967, opened 2021-01-28
Bug fix for gravity simulation 

micmitch:
In the getJtJdiag function I believe that self.nD needs to be self.survey.nD.


micmitch:
Can we get this one line bug fix pulled in before master progresses too far?

jcapriot:
Yup, just needed to make the same change in the magnetic simulation as well.
(it also cause black to reformat that file.

domfournier (approved):
Nice catch.

PR #974, opened 2021-02-27
Clean up and refactor static_utils.py

dccowan:
There are functions within static_utils.py that:
- Are not named according to SimPEG convention
- Are general math and mesh utilities that moved to an appropriate place
- IO functions that should be moved to SimPEG.utils.io_utils
- Are redundant/duplicate and should be removed.
- Are not generalized to arbitrary electrode configurations

**Below are a list of functions within static_utils.py and a suggested course of action:**

- (**done**) *source_receiver_midpoints:* This function doesn't actually compute midpoints. Instead it tries to compute the pseudo-location for the sensitivity for each datum.
    - rename *pseudo_locations*
    - upgrade to properly take into account topography

- (**done**) *apparent_resistivity:* Shouldn't require a data object. Only needs survey geometry and voltage data. This needs to be more clear. And what if the data vector you have is not volts?
    - rename as *apparent_resistivity_from_voltage*
    - change input arguments to survey object and normalized voltage

- (**done**) *plot_layer:* Not very descriptive and assumes you are plotting apparent resistivities. It also requires you to define a 1D mesh which seems unnecessary.
    - rename to *plot_1d_layer_model*
    - made more general and robust

- (**done**) *plot_pseudosection:* Since we also have a 3D pseudosection method, we should distinguish that this plots 2d data. The previous function didn't take into account topography.
    - rename *plot_2d_pseudosection*
    - upgrade to properly take into account topography (done)
    - upgrade to have scatter plot and filled contour plot options (done)

- (**done**) *plot_3d_pseudosection:* This was previously build on matplotlib which is not a 3D plotting engine. Upgrade this utility to be optional if the plotly package is installed.

- *generate_dcip_survey* and *generate_dcip_sources_line:* These do more or less the same thing. However, the latter has better docstrings, comments and take topography into account.

- (**done**) All 3D read and write for UBC formatted data is redundant. Better methods exist in utils.io_utils.io_utils_electromagnetics

- (**done**) 2D and XYZ formatted io methods should be added to utils.io_utils.io_utils_electromagnetics

- *r_unit, gettopoCC, drapeTopotoLoc, closestPointGrid:* These are general math functions and should be moved and likely renamed


jcapriot:
Glad to see this getting pushed on, cleaning up and unifying a lot of these utils will be great! After a cursory look, I've got a few comments,

``apparent_resitivity`` Maybe it should just be ``calculate_apparent_resitivity``? I think that at least apparent resistivity should definitely be in the name.

Can you make deprecated functions that mimic the old versions to ease the transition?

dccowan:
> Glad to see this getting pushed on, cleaning up and unifying a lot of these utils will be great! After a cursory look, I've got a few comments,
> 
> `apparent_resitivity` Maybe it should just be `calculate_apparent_resitivity`? I think that at least apparent resistivity should definitely be in the name.
> 
> Can you make deprecated functions that mimic the old versions to ease the transition?

I'm trying to make deprications when possible.

dccowan:
We should get @thast comments on this as well.

dccowan:
> Glad to see this getting pushed on, cleaning up and unifying a lot of these utils will be great! After a cursory look, I've got a few comments,
> 
> `apparent_resitivity` Maybe it should just be `calculate_apparent_resitivity`? I think that at least apparent resistivity should definitely be in the name.
> 
> Can you make deprecated functions that mimic the old versions to ease the transition?

I called it apparent_resistivity_from_voltage

dccowan:
For some reason, the 3D pseudosection plot being created with sphinx gallery is not assigning the proper color value to each point. Using Spyder IDE and Jupyter notebooks, the plots are correct. The two plots below show the 3D pseudosection plots created in Spyder and created when running *make html*. Both plots are created from the exactly same .py file!!! Everything else about the plots is consistent, but for some reason, the color value assigned to the points is not consistent. I do not have this problem with 2D pseudosection plots.

What I think could be happening is that sphinx_gallery is somehow using an older version of matplotlib when creating the plots. Some aspects of colormaps and their implementation changed in matplotlib in version 3.0.0. Any help on this would be must appreciated.

image

dccowan:
> For some reason, the 3D pseudosection plot being created with sphinx gallery is not assigning the proper color value to each point. Using Spyder IDE and Jupyter notebooks, the plots are correct. The two plots below show the 3D pseudosection plots created in Spyder and created when running _make html_. Both plots are created from the exactly same .py file!!! Everything else about the plots is consistent, but for some reason, the color value assigned to the points is not consistent. I do not have this problem with 2D pseudosection plots.
> 
> What I think could be happening is that sphinx_gallery is somehow using an older version of matplotlib when creating the plots. Some aspects of colormaps and their implementation changed in matplotlib in version 3.0.0. Any help on this would be must appreciated.
> 
> image

I found a way to get plotly images to render on the site. It requires the plotly and python-kaleido packages. And the following alteration is needed in the conf.py file 
image

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
+    Output:
+    mpl_toolkits.mplot3d.axes3d.Axes3D
+        The axis object that holds the plot
 
-    grid_rho = griddata(np.c_[midx, midz], rho.T, (grid_x, grid_z), method="linear")
+    """
 
-    if clim is None:
-        vmin, vmax = rho.min(), rho.max()
+    if np.median(values) > 1.:
```
jcapriot, line 470: just trying to guess at what the property is? I like that this is a general function for plotting 1D models. I would like to see it moved to some other utility file, and then just imported here for backwards compatibility.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
+    else:
+        ax.set_ylim(z.min(), z.max())
+    ax.set_xlabel(x_label)
+    ax.set_ylabel(y_label)
+    
     return ax
 
 
-def plot_3d_pseudosection(
+def plot_2d_pseudosection(
```
jcapriot, line 523: Just asking a general question, but I personally associate pseudo sections as a 2D thing, so is it necessary to specifically say `plot_2d_pseudosection`? I understand the reasoning behind including it for 3D though (`plot_3d_...`), and I think the change in name is probably unnecessary.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -552,132 +560,129 @@ def plot_3d_pseudosection(
...
         )
 
+    else:
+        raise NotImplementedError("plot_type must be 'scatter' or 'tricontourf'")
+        
+    # Use a filled polygon to mask everything above
+    # that has a pseudo-location above the positions
+    # for nearest electrode spacings    
+    
+    if mask_topography:
```
jcapriot, line 642: How robust is this for guessing the topographic surface?

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -552,132 +560,129 @@ def plot_3d_pseudosection(
...
+        )    
+    
+    z_top = np.max(locations[:, -1])
+    z_bot = np.min(locations[:, -1])
+    ax.set_ylim(z_bot - 0.03*(z_top-z_bot), z_top + 0.03*(z_top-z_bot))
+    ax.set_xlabel("Line position (m)")
+    ax.set_ylabel("Pseudo-elevation (m)")
+
     # Define colorbar
     if create_colorbar:
```
jcapriot, line 677: I want to see us use a LogNorm for log scales instead of doing this odd thing of plotting the log values of data and hacking the colorbar formats. I would like to see re-use of the 2D plotting utility in SimPEG/utilities/plot_utils.py instead of yet another recreating of 2D plotting.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -687,29 +692,210 @@ def plot_3d_pseudosection(
...
+        Output:
+        fig: 
+            A plotly figure
+
+        """
+
+        locations = pseudo_locations(survey)
+
+        # Scaling
+        if scale == "log":
```
jcapriot, line 778: I tried looking, but it doesn't look like there's any way to supply a "norm" like argument to poorly for plotting on a log scale like there is in numpy. I would also like to see a better format than plotting as "10^###" as it is very confusing. I imagine there must be some way with plotly to directly control the tick labels, and replace 10^### with an actual interpretable number.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -1859,31 +1343,6 @@ def r_unit(p1, p2):
     return vec, r
 
 
-def getSrc_locs(survey):
```
jcapriot, line 1862: probably just deprecate this function and tell the user to extract from the survey as `locations_a` and `locations_b`

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
-                )
-        return ax.plot(resistivity, z, "k-", **kwargs)
 
 
 ############
 # Deprecated
 ############
 
-
-def plot_pseudoSection(
```
jcapriot, line 2164: Don't touch this one, it matches the call for older versions of SimPEG

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
-        data_type=data_type,
-        space_type=space_type,
-        clim=clim,
-        scale=scale,
-        sameratio=sameratio,
-        pcolor_opts=pcolorOpts,
-        data_locations=data_location,
-        dobs=dobs,
-        dim=dim,
+    return plot_2d_pseudosection(data.survey, data.dobs, 'scatter', ax=ax, scale=scale)
```
jcapriot, line 1583: there's a lot of options that are not passed on to the `plot_pseudosection` function (i.e. pcolor_opts)

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+        "The apparent_resistivity method has been deprecated. Please use "
+        "apparent_resistivity_from_voltage instead. This will be removed in version"
+        " 0.15.0 of SimPEG",
+        DeprecationWarning,
     )
 
+    if dobs is None:
+        dobs = data_object.dobs
+
+    return apparent_resistivity_from_voltage(data_object.survey, dobs, space_type=space_type)
```
jcapriot, line 1597: not passing `eps` or other kwargs on to the called function

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
     warnings.warn(
-        "The gen_DCIPsurvey method has been deprecated. Please use "
-        "generate_dcip_survey instead. This will be removed in version"
+        "The source_receiver_midpoint method has been deprecated. Please use "
+        "pseudo_locations instead. This will be removed in version"
+        " 0.15.0 of SimPEG",
+        DeprecationWarning,
+    )
+
+    return pseudo_locations(survey, kwargs)
```
jcapriot, line 1608: you need to expand kwargs as `**kwargs`

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+
+def plot_layer(rho, mesh, **kwargs):
+    warnings.warn(
+        "The plot_layer method has been deprecated. Please use "
+        "plot_1d_layer_model instead. This will be removed in version"
+        " 0.15.0 of SimPEG",
+        DeprecationWarning,
+    )
+
+    return plot_1d_layer_model(mesh.hx, rho)
```
jcapriot, line 1618: again, not passing any of the kwargs on to the underlying function

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+    )
+
+    return convert_3d_survey_to_2d_lines(survey, lineID)
+
+
+def getSrc_locs(survey):
+    warnings.warn(
+        "The getSrc_locs method has been deprecated. Source "
+        "locations are now computed as a method of the survey "
+        "class. Please use Survey.source_locations(). This method "
```
jcapriot, line 1635: As I mentioned before, point the users to survey.locations_a and locations_b

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+        " will be removed in version 0.15.0 of SimPEG",
+        DeprecationWarning,
+    )
+
+    return survey.source_locations()
```
jcapriot, line 1640: use survey.locations_a and survey.locations_b

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+        write_dcip2d_ubc(
+            fileName,
+            data,
+            'volt',
+            'dobs',
+            format_type=format_type,
+            comment_lines=comment_lines,
+        )
+
+    # if not isinstance(data, Data):
```
jcapriot, line 1703: remove all of these lines

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+            comment_lines=comment_lines,
+        )
+
+
+
+
+
+
+
+    # if not ((dim == 2) | (dim == 3)):
```
jcapriot, line 1959: remove all of these lines

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
+        "The readUBC_DC2Dpre method has been deprecated. Please use "
+        "read_dcip2d_ubc instead. This is imported "
+        "from SimPEG.utils.io_utils. This function will be removed in version"
+        " 0.15.0 of SimPEG",
+        DeprecationWarning,
+    )
+
+    return read_dcip2d_ubc(fileName, 'volt', 'general')
+
+    # # Load file
```
jcapriot, line 2125: remove commented lines

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -2087,145 +1546,777 @@ def gen_3d_survey_from_2d_lines(
...
-    source_list = generate_dcip_sources_line(
-        survey_type, data_type, dim_flag, endl, topo, n, ds, dh
+    return read_dcip3d_ubc(fileName, data_type)
+
+
+
+
+
+
+    # # Load file
```
jcapriot, line 2196: remove commented lines

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -1,35 +1,208 @@
...
+    if data_type not in ["volt", "apparent_resistivity", "apparent_chargeability"]:
+        raise Exception(
+            "data_type must be one of 'volt', 'apparent_resistivity', 'apparent_chargeability'"
+        )
+    
+    # Prevent circular import
+    from ...electromagnetics.static import resistivity as dc
+    from ...data import Data
+    
+    # Load file headers
```
jcapriot, line 50: could you use panda's to load all of these in simpler?

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -1,35 +1,208 @@
...
+    # 2D surface data
+    else:
+        a_cols = [1]
+        b_cols = [2]
+        m_cols = [3]
+        n_cols = [4]
+    
+    # Extract electrode locations
+    if is_surface_data:
+        dummy_elevation = 9999  # Taller than mount Everest
```
jcapriot, line 108: just set it as like, `np.inf` or `np.nan`

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -1,35 +1,208 @@
...
+        a_locations = data_array[:, a_cols]
+        b_locations = data_array[:, b_cols]
+        m_locations = data_array[:, m_cols]
+        n_locations = data_array[:, n_cols]
+    
+    # Set up keeping track of sorting of rows and unique sources
+    k = np.arange(0, n_rows)
+    out_indices = []
+    unique_ab, ab_index = np.unique(np.c_[a_locations, b_locations], axis=0, return_index=True)
+    ab_index = np.sort(ab_index)
```
jcapriot, line 128: I would like to see our generating functions that read in a file, put it in the most memory efficient format for SimPEG. I.E. make a list of all unique sources, and all unique receivers. Create objects for each of them. Then go through and attach the receivers to the sources. (This stops repeated definitions of projection matrices when the receivers are in the same location).

jcapriot: Don't we have a function that is survey_from_abmn, that could be called here?

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -44,51 +217,233 @@ def read_dcip3d_ubc(file_name, data_type):
...
+        n_rows = np.shape(data_array)[0]
+        
+        if np.shape(data_array)[1] > 5:
+            wd = data_array[:, 5]
+        
+        if np.shape(data_array)[1] > 4:
+            d = data_array[:, 4]
+        
+        # Get ABMN electrode locations
+        dummy_elevation = 9999
```
jcapriot, line 276: again, probably set this as `np.inf` or `np.nan` so the simulations will actually fail and not necessarily run while the user ignores the warning.

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -44,51 +217,233 @@ def read_dcip3d_ubc(file_name, data_type):
...
+            # Keep track to sorted output
+            out_indices.append(k[rx_index])
+            
+            # Define Pole or Dipole Receivers
+            if np.all(np.isclose(rx_loc_m, rx_loc_n, atol=1e-3)):
+                rx_list = [dc.receivers.Pole(rx_loc_m, data_type=data_type)]
+            elif np.all(np.any(np.isclose(rx_loc_m, rx_loc_n, atol=1e-3)==False, axis=1)):
+                rx_list = [dc.receivers.Dipole(rx_loc_m, rx_loc_n, data_type=data_type)]
+            else:
+                raise NotImplementedError("An individual source cannot have a mix of Pole and Dipole receivers")
```
jcapriot, line 316: Again, not an actual concern.

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -44,51 +217,233 @@ def read_dcip3d_ubc(file_name, data_type):
...
+        out_indices = []
+        unique_ab, ab_index = np.unique(data_array[:, 0:2], axis=0, return_index=True)
+        ab_index = np.sort(ab_index)
+        
+        a_locations = np.c_[data_array[:, 0], dummy_elevation*np.ones(n_rows)]
+        b_locations = np.c_[data_array[:, 1], dummy_elevation*np.ones(n_rows)]
+        m_locations = np.c_[data_array[:, 2], dummy_elevation*np.ones(n_rows)]
+        n_locations = np.c_[data_array[:, 3], dummy_elevation*np.ones(n_rows)]
+        
+        # Loop over all unique source locations
```
jcapriot, line 288: Likely just use a survey_from_ABMN function

docs/conf.py
```diff
@@ -168,7 +170,7 @@ def linkcode_resolve(domain, info):
     else:
         linespec = ""
 
-    fn = relpath(fn, start=dirname(SimPEG.__file__))
+    # fn = relpath(fn, start=dirname(SimPEG.__file__))
```
jcapriot, line 173: Undo this so documentation has source code links

tests/docs/test_docs.py
```diff
@@ -12,24 +12,28 @@ def path_to_docs(self):
...
-        self.assertTrue(response.returncode == 0)
-        # response = subprocess.call(["make", "html"], shell=True)  # Needed for local test on Windows
-        # self.assertTrue(response == 0)
+        
+        try:
+            response = subprocess.run(["make", "html-noplot"])
+            self.assertTrue(response.returncode == 0)
+        except:
+            response = subprocess.call(["make", "html"], shell=True)  # Needed for local test on Windows
+            self.assertTrue(response == 0)
```
jcapriot, line 21: is this actually needed on windows still? it's been a while. Also this seems odd, because it basically runs the test twice if it fails... You probably need to catch a specific Error.

tests/docs/test_docs.py
```diff
@@ -12,24 +12,28 @@ def path_to_docs(self):
...
         os.chdir(os.path.sep.join(self.path_to_docs))
-
-        response = subprocess.run(["make", "linkcheck-noplot"])
-        print(response.returncode)
-        self.assertTrue(response.returncode == 0)
-        # response = subprocess.call(["make", "linkcheck"], shell=True)  # Needed for local test on Windows
-        # print(response)
-        # self.assertTrue(response == 0)
+        
+        try:
```
jcapriot, line 29: Again, same comment as above. Undo this, or figure out if that change is actually necessary.

tests/em/static/test_DC_Utils.py
```diff
@@ -75,13 +75,13 @@ def test_io_rhoa(self):
             ],
         ):
             print("\n Testing {} ... ".format(survey_type))
-            survey = utils.gen_DCIPsurvey(
```
jcapriot, line 78: leave this as the old way of calling to test both the deprecated and new version at the same time

tests/em/static/test_DC_Utils.py
```diff
@@ -204,21 +204,6 @@ def test_ElecSep(self):
 
         self.assertTrue(passed)
 
-    def test_apparent_resistivity(self):
```
jcapriot, line 207: Why was this removed?

tests/em/static/test_DC_miniaturize.py
```diff
@@ -1,5 +1,5 @@
 from SimPEG.electromagnetics.static import resistivity as dc
-from SimPEG.electromagnetics.static.utils.static_utils import gen_DCIPsurvey
+from SimPEG.electromagnetics.static.utils.static_utils import generate_dcip_sources_line
```
jcapriot, line 2: again, leave these as the old way so you can test the deprecation and the new version (and test that it reproduces the functionality)

jcapriot (changes requested):
I've gone through a lot of the SimPEG code, and there's some things that should be addressed. I did not have a chance to look through all of the tutorials at this point though.

We still need to get the added data files up on the cloud service and removed from here. Also, as we talked about we should probably just not worry about reading in/ saving the "true" models, as it's easier to recreate it in within the tutorials.

SimPEG/electromagnetics/static/resistivity/survey.py
```diff
@@ -106,6 +106,29 @@ def electrode_locations(self):
         loc_n = self.locations_n
         return np.unique(np.vstack((loc_a, loc_b, loc_m, loc_n)), axis=0)
 
+    @property
+    def source_locations(self):
```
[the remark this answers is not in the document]
dccowan: Not quite. When setting the abmn locations, the location of the source is repeated for every receiver and everything is in order. When calling electrode_locations (changed to unique_electrode_locations), we just provide a sorted array of all the locations where there is an electrode. I would like to keep these two functions.

SimPEG/electromagnetics/static/resistivity/survey.py
```diff
@@ -236,4 +259,11 @@ def drapeTopo(self, *args, **kwargs):
 
 @deprecate_class(removal_version="0.15.0")
 class Survey_ky(Survey):
+    warnings.warn(
```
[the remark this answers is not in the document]
dccowan: Do you want to just remove this then? I didn't add this deprecation

SimPEG/electromagnetics/static/utils/__init__.py
```diff
@@ -23,5 +24,12 @@
     genTopography,
     closestPointsGrid,
     gen_3d_survey_from_2d_lines,
-    plot_layer,
```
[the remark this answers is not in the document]
dccowan: The most up to date branch appears to import plot_layer

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -16,6 +17,19 @@
...
+    read_dcip3d_ubc,
+    write_dcip3d_ubc
+)
+
+try:
+    import plotly.graph_objects as grapho
+    has_plotly = True
+except:
+    has_plotly = False
+    pass
```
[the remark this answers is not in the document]
dccowan: it is removed

tests/docs/test_docs.py
dccowan: Yes, It appears to be necessary. And the test failed right away when trying to build the docs so I'm not overly concerned about the timing right now.

tests/em/static/test_DC_Utils.py
dccowan: This may have been rendered redundant by creating another test

tests/docs/test_docs.py
dccowan: If I recall correctly, we used to use Travis which ran on Linux. But now we are using Azure which is a Microsoft product. I think the entire set of tests makes sure that everything works on any OS, but that means the additional testing of things for Windows OS. Make sense?

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -181,40 +198,62 @@ def source_receiver_midpoints(survey, **kwargs):
...
+            ds.append(
+                (rx_midpoints - np.tile(src_midpoint, (n_loc, 1)))
+            )
+    
+    midpoints = np.vstack(midpoints)
+    ds = np.vstack(ds)
+    pseudo_depth = np.zeros_like(midpoints)
+
+    # wenner-like electrode groups (are source and rx midpoints in same place)
+    is_wenner = np.sqrt(np.sum(ds[:,:-1]**2, axis=1)) < 0.1
```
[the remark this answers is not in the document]
dccowan: In a synthetic modeling scenario, the midpoints for the source electrodes and for the receiver electrodes would be identical. But in practice, it may not be the case. I wanted a threshold that was smaller than any reasonable electrodes spacing for a DCIP survey. But I wanted to make the threshold big enough just in case the locations of the Wenner electrodes wasn't perfect.

tests/em/static/test_DC_Utils.py
dccowan: Done

tests/em/static/test_DC_miniaturize.py
dccowan: I'm not sure this one is totally necessary. It's testing things once a survey has been created. Creating the survey itself isn't really part of this test.

SimPEG/electromagnetics/static/resistivity/survey.py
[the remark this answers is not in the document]
jcapriot: The `Survey` class already has this under the `locations_a` and `locations_b` property. Those properties are a list of the source locations for every source receiver combo. There's also the corresponding `locations_m` and `locations_n` as well.

SimPEG/electromagnetics/static/resistivity/survey.py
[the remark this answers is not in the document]
jcapriot: Yes remove it, I don't know how else it would've got in here though.

SimPEG/electromagnetics/static/utils/static_utils.py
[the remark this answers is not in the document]
jcapriot: let's say someone is wanting to plot a pseudo-section over some example (say in a small fish tank) where the electrodes them selves are only centimeters apart. This will incorrectly identify those points as Wenner.

tests/docs/test_docs.py
jcapriot: Please change it to a specific exception then, or replace it with an `if platform.system() == "Windows` block then. See https://docs.python.org/3/library/platform.html#platform.system. I do not like having the catch all except block here.

dccowan: This works for me

jcapriot: Just pointing this out, but your most recent change actually stops the doctests from running on the test suite.

jcapriot: This is now fixed

SimPEG/electromagnetics/static/utils/static_utils.py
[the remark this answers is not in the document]
dccowan: added a keyword argument to see tolerance for identifying wenner data

tests/em/static/test_DC_miniaturize.py
dccowan: There are tests that call gen_DCIPsurvey still. I don't see this as necessary

docs/conf.py
dccowan: Forgot to do this. Good catch. But I would like to fix it so it works when Python and the repo are on different drives in Windows. It's just a pain to remember to do this each time.

SimPEG/electromagnetics/static/utils/__init__.py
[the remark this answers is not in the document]
jcapriot: Again, import ALL of the old deprecate functions here

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +300,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
+        ]
+        m_locs_s = np.c_[
+            np.dot(mn_locs_all[lineID_index, 0:2]-r0[0], uvec), mn_locs_all[lineID_index, 2]
+        ]
+        n_locs_s = np.c_[
+            np.dot(mn_locs_all[lineID_index, 3:5]-r0[0], uvec), mn_locs_all[lineID_index, -1]
+        ]
+        
+        # For each source in the line
+        for ii, ind in enumerate(ab_index):
```
jcapriot, line 394: This could likely call the new `survey_from_ABMN` function (or whatever the actual name is)

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -687,29 +695,345 @@ def plot_3d_pseudosection(
...
+    survey
+        A SimPEG.electromagnetic.static.survey.Survey object
+    sort_index
+        A numpy array which defines any sorting that took place when creating the survey
+
+
+    """
+    
+    if locations_a is None:
+        raise AssertionError("Locations for A electrodes must be provided.")
```
jcapriot, line 950: These should raise `TypeError` and not `AssertionError`

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -687,29 +695,345 @@ def plot_3d_pseudosection(
...
+    if locations_m is None:
+        raise AssertionError("Locations for M electrodes must be provided.")
+
+    assert data_type.lower() in [
```
jcapriot, line 954: do not use `assert`, it is in general bad practice when checking the type as it's possible to turn asserts off. Instead throw a `ValueError`

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -687,29 +695,345 @@ def plot_3d_pseudosection(
...
+    ], "data_type must be one of 'volt', 'apparent_conductivity', 'apparent_resistivity', 'apparent_chargeability'"
+
+    if locations_b is None:
+        locations_b = locations_a
+
+    if locations_n is None:
+        locations_n = locations_m
+    
+    if (locations_a.shape==locations_b.shape==locations_m.shape==locations_n.shape) == False:
+        raise AssertionError("Arrays containing A, B, M and N electrode locations must be same shape.")
```
jcapriot, line 968: Raise `ValueError` here.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -687,29 +695,345 @@ def plot_3d_pseudosection(
...
+                (np.sqrt(np.sum((locations_b - src_loc_b)**2, axis=1)) < 1e-3)
+            )
+        )[0]
+        
+        rx_loc_m = locations_m[rx_index, :]
+        rx_loc_n = locations_n[rx_index, :]
+        
+        # Extract pole and dipole receivers
+        k_ii = k[rx_index]
+        is_pole_rx = np.all(np.isclose(rx_loc_m, rx_loc_n, atol=1e-3), axis=1)
```
jcapriot, line 998: Are there any other conditions we would want to accept as indicative of a `pole` source/receiver? Maybe if B or N is infinity or NaN?

SimPEG/electromagnetics/static/utils/static_utils.py
jcapriot: actually point the user to `survey.source_locations` in the deprecated version of this function

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -1,70 +1,173 @@
...
+        for h in dict_headers:
+            out_dict[h] = data_array[out_indices, file_headers.index(h)]
+        return data_object, out_dict
+    
+    else:
+        return data_object
 
 
-def read_dcipoctree_ubc(file_name, data_type):
+def read_dcip2d_ubc(file_name, data_type, format_type):
```
jcapriot, line 158: can `format_type` be a keyword argument with a default value? probably `general`

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -78,17 +181,205 @@ def read_dcipoctree_ubc(file_name, data_type):
         defined as uncertainties.
 
     """
+    data_type = data_type.lower()
+    format_type = format_type.lower()
 
-    # Unused for now but it will be when we manage IP types better.
-    assert data_type.lower() in [
+    assert data_type in [
```
jcapriot, line 187: Since you're changing this line anyways, again do not use `assert`, instead throw the appropriate error (`ValueError`), also keep the `.lower()` part.

SimPEG/utils/io_utils/io_utils_electromagnetics.py
```diff
@@ -78,17 +181,205 @@ def read_dcipoctree_ubc(file_name, data_type):
...
         "volt",
         "apparent_chargeability",
-    ], "Parameter 'data_type' must be one of {'volt', 'apparent_chargeability'}"
+        "secondary_potential",
+    ], "Parameter 'data_type' must be one of {'volt', 'apparent_chargeability', 'secondary_potential'}"
+    
+
+    assert format_type in [
```
jcapriot, line 194: assert

SimPEG/utils/io_utils/io_utils_electromagnetics.py
jcapriot: nvm on the `.lower()` part, didn't see it a few lines up

docs/conf.py
```diff
@@ -168,7 +170,12 @@ def linkcode_resolve(domain, info):
     else:
         linespec = ""
 
-    fn = relpath(fn, start=dirname(SimPEG.__file__))
+    # Exception for building locally on Windows when Python on C: drive and repo on D: drive
+    # Note that these links will not work on the local build!!!
+    try:
+    	fn = relpath(fn, start=dirname(SimPEG.__file__))
+    except Exception:
```
jcapriot, line 177: Don't catch general `Exception` here, I think you're wanting to specifically catch and ignore the error due to a relative path no being possible on windows when the start and end are on two different drives. (A silly limitation of Windows). But the exception that is raised in those cases is a `ValueError`, so specifically except a `ValueError`

tests/em/static/test_DCIP_io_utils.py
```diff
@@ -0,0 +1,184 @@
...
+               np.isclose(data2D.dobs, data_general.dobs) &
+               np.isclose(data2D.dobs, data_surface.dobs) &
+               np.isclose(data2D.dobs, data_simple.dobs)
+            ))
+
+            self.assertTrue(passed)
+            shutil.rmtree('./dcip2d_tests')
+            print("READ/WRITE METHODS FOR DC2D DATA PASSED!")
+
+        except Exception:
```
jcapriot, line 69: `unittests` can have a `tearDown(self)` method that should be used to erase any generated files. I would recommend this over the try, except method you've used here so that if there is any error in the test it is slightly more clear where it came from. essentially you can put `shut.rmtree('../dcip2d_tests')` in that method.

tests/em/static/test_DCIP_io_utils.py
```diff
@@ -0,0 +1,184 @@
...
+               np.isclose(data3D.dobs, data_general.dobs) &
+               np.isclose(data3D.dobs, data_surface.dobs)
+            ))
+
+            self.assertTrue(passed)
+            shutil.rmtree('./dcip3d_tests')
+            print("READ/WRITE METHODS FOR DCIP3D DATA PASSED!")
+
+        except Exception:
+            print(Exception)
```
jcapriot, line 178: use `tearDown()` instead

tutorials/05-dcr/dcr1d/app_res_1d_data.dobs
```diff
@@ -0,0 +1,25 @@
+-3.0000e+01 0.0000e+00 0.0000e+00 3.0000e+01 0.0000e+00 0.0000e+00 -1.0000e+01 0.0000e+00 0.0000e+00 1.0000e+01 0.0000e+00 0.0000e+00 1.0099e+03
```
jcapriot, line 1: Leftover data file I believe

dccowan: removed

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: I am happy the way it is. But I think once this is released, the community may offer feedback which could lead to some small additions.

SimPEG/utils/io_utils/io_utils_electromagnetics.py
dccowan: Most of the time, people want to read ubc data in surface format. I really wanted to force people to put in the right file format themselves when loading the data.

docs/conf.py
dccowan: sure

tests/em/static/test_DCIP_io_utils.py
dccowan: done

tests/em/static/test_DCIP_io_utils.py
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -181,40 +198,62 @@ def source_receiver_midpoints(survey, **kwargs):
...
+        pseudo_depth[~is_wenner, 0] = (dz/2) * (ds[~is_wenner, 0] / L)
+        if np.shape(ds)[1] > 2:
+            pseudo_depth[~is_wenner, 1] = (dz/2) * (ds[~is_wenner, 1] / L)
+        pseudo_depth[~is_wenner, -1] = np.sqrt(np.sum(ds[~is_wenner,:-1]**2, axis=1))/2
 
-def geometric_factor(survey_object, space_type="half space", **kwargs):
+    return midpoints-pseudo_depth
+
+
+def geometric_factor(survey_object, space_type="half_space", **kwargs):
```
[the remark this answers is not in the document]
dccowan: sounds good

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
 
     return G / (spaceFact * np.pi)
 
-
-def apparent_resistivity(
-    data_object, space_type="half space", dobs=None, eps=1e-10, **kwargs
-):
+def apparent_resistivity_from_voltage(survey, volts, space_type="half_space", eps=1e-10):
```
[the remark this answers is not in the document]
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
-    pcolor_opts={},
-    contour_opts={},
-    cbar_opts={},
-    data_locations=False,
-    dobs=None,
-    dim=2,
-    y_values="n-spacing",
-    **kwargs,
-):
+def convert_survey_3d_to_2d_lines(survey, lineID, data_type='volt', output_indexing=False):
```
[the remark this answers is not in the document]
dccowan: I have found any smart algorithm that tries to automatically figure out the bearing and parse the survey into 2D is unreliable. So to make it always stable for use in practical situations, I like to require a line ID.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
+    mn_locs_all = np.c_[survey.m_locations, survey.n_locations]
+    
+    # For each unique lineID
+    survey_list = []
+    for ID in unique_lineID:
+        
+        source_list = []
+        
+        # Source locations for this line
+        lineID_index = np.where(lineID == ID)[0]
```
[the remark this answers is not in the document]
dccowan: I played around with this a little bit. There was some resorting that I didn't like. Perhaps we can revisit this in the future if we want to optimize it.

SimPEG/electromagnetics/static/utils/static_utils.py
```diff
@@ -258,291 +297,260 @@ def geometric_factor(survey_object, space_type="half space", **kwargs):
...
+            if output_indexing:
+                out_indices.append(kID[rx_index])
+            
+            # Define Pole or Dipole Receivers
+            if np.all(np.isclose(rx_loc_m[:, 0], rx_loc_n[:, 0], atol=1e-3)):
+                rx_list = [dc.receivers.Pole(rx_loc_m)]
+            elif np.all(np.isclose(rx_loc_m[:, 0], rx_loc_n[:, 0], atol=1e-3)==False):
+                rx_list = [dc.receivers.Dipole(rx_loc_m, rx_loc_n)]
+            else:
+                raise NotImplementedError("An individual source cannot have a mix of Pole and Dipole receivers")
```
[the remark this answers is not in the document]
dccowan: I made this improvement. We can have any mix of pole and dipole sources and receivers.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: I moved this to general plotting functions.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: The input arguments of the plot_2d_pseudoection and the original plot_pseudosection are so different that I'm inclined to keep it as is for now.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: It's pretty good. I tested it for peaks and valleys. I wanted to be sure for the latter case. And it's done based on the electrode locations and not topography so I don't believe there is sampling issue there.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: I put it back

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: this is fixed

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: done

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: done

SimPEG/utils/io_utils/io_utils_electromagnetics.py
dccowan: I don't have much familiarity with pandas. Would it that much of an improvement?

SimPEG/utils/io_utils/io_utils_electromagnetics.py
dccowan: I thought about it, but I like this better. Unless we want to make it impossible for the user to apply any operations until proper elevations are applied to the data.

SimPEG/utils/io_utils/io_utils_electromagnetics.py
dccowan: The application of topography to the data is done in the simulations. So if someone goes through the tutorials properly, they will end up applying the topography.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: I think we leave the plot_2d_pseudosection function for now for the above reason. Then after 0.15.0, we can deprecate it and rename it to plot_pseudosection.

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: I tried several ways to implement this but there is an issue in mpl for customizing the ticks on the colorbar when doing a log plot. The original ticks on the colorbar are hard coded somehow when setting the mapping to LogNorm. The new ticks just plot on top of them as you can see below. I literally tried to do cbar.set_ticks([]) to get rid of any ticks. For linear scale, they would disappear. For log-norm they do not. Until this bug is fixed, I am inclined to leave it the way it is because it is stable.

If the data always span several orders of magnitude there isn't much point in playing with this. But if you want to plot app. res. and the values only span 1 order of magnitude, it isn't good.

image

SimPEG/electromagnetics/static/utils/static_utils.py
dccowan: See above comment

Issue #976, opened 2021-02-28
update since 0.14.1 causing error in IP inversion? 


lheagy:
When running the [first notebook in the transform tutorial](https://github.com/simpeg/transform-2020-simpeg/blob/master/1-century-dcip-inversion.ipynb) on the latest version of SimPEG, we run into an error in the IP inversion in cell 56 (thanks @dougoldenburg for catching this!). Does anyone have insights as to what might be the source of error? It was something that changes since the 0.14.1 release 

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-56-7faca2bf9ff5> in <module>
      1 # Run inversion
----> 2 charge_opt = inv_ip.run(m0_ip)

~/git/simpeg/simpeg/SimPEG/utils/counter_utils.py in wrapper(self, *args, **kwargs)
    103         if isinstance(counter, Counter):
    104             counter.countTic(self.__class__.__name__ + "." + f.__name__)
--> 105         out = f(self, *args, **kwargs)
    106         if isinstance(counter, Counter):
    107             counter.countToc(self.__class__.__name__ + "." + f.__name__)

~/git/simpeg/simpeg/SimPEG/inversion.py in run(self, m0)
     62         """
     63         self.invProb.startup(m0)
---> 64         self.directiveList.call("initialize")
     65         print("model has any nan: {:b}".format(np.any(np.isnan(self.invProb.model))))
     66         self.m = self.opt.minimize(self.invProb.evalFunction, self.invProb.model)

~/git/simpeg/simpeg/SimPEG/directives.py in call(self, ruleType)
    186         )
    187         for r in self.dList:
--> 188             getattr(r, ruleType)()
    189 
    190     def validate(self):

~/git/simpeg/simpeg/SimPEG/directives.py in initialize(self)
    239         m = self.invProb.model
    240 
--> 241         dm_eigenvalue = eigenvalue_by_power_iteration(
    242             self.dmisfit, m, n_pw_iter=self.n_pw_iter,
    243         )

~/git/simpeg/simpeg/SimPEG/utils/mat_utils.py in eigenvalue_by_power_iteration(combo_objfct, model, n_pw_iter, fields_list, seed)
    142         for k, obj in enumerate(combo_objfct.objfcts):
    143             if hasattr(obj, "simulation"):
--> 144                 fields_list += [obj.simulation.fields(model)]
    145             else:
    146                 # required to put None to conserve it in the list

[...]

Issue #968, opened 2021-02-03
Looping through data?


domfournier:
https://github.com/simpeg/simpeg/blob/d20eefa6b2238210d1c6c14f998c53bb8ea009a4/SimPEG/potential_fields/gravity/simulation.py#L52

@lheagy 
Why did we switch to a for-loop here?

jcapriot:
I think it was because you had wanted to avoid double storing the sensitivity matrix when it was in memory.

lheagy:
Hmm, yeah I think @jcapriot's comment is right here. I think that the dense matrix multiplication grows memory quite a bit, but there might be a way to do that in-place? That isn't something I have looked into

domfournier:
Oh yea totally, I recall now. The dask operation could do it in place, but now numpy.
Let's leave this guy open for a little while as a reminder.

jcapriot:
Actually... I think the answer here to both questions, is `np.einsum`, and I believe the appropriate line of code will be,
```
np.einsum('i,ij,ij->j', W, G, G)
```

From some quick tests, this is even about 10 times faster than
```
np.sum(W[:, None] * (G * G), axis=0)
```

I still need to double check the memory requirements though.

domfournier:
Wow that was quick. Thanks Eisten (@jcapriot  ;)

Just tried on this:

`import numpy as np
G = np.random.randn(10000,100000)
W = np.ones(G.shape[0])`
This indeed works:
`np.einsum('i,ij,ij->j', W, G, G)`

This blows up:
`np.sum(W[:, None] * G**2., axis=0)`

jcapriot:
Yeah, `einsum` is actually very powerful and I love it, but I’m just still not the best at remembering _how_ to use it.

jcapriot:
We still need to make more use of einsum for this operation in the code base.

PR #977, opened 2021-03-04
Fixing 2D IP bug

jcapriot:
implements testing on apparent changeability data for the IP in 2D and 3D,
Enabled to find runtime bug for 2D IP, however accuracy of the 2D apparent changeability data seems off.

### DIFF
--- a/SimPEG/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/electromagnetics/static/induced_polarization/simulation.py
@@ -46,7 +46,10 @@ class BaseIPSimulation(BaseDCSimulation):
             # loop through receievers to check if they need to set the _dc_voltage
             for src in self.survey.source_list:
                 for rx in src.receiver_list:
-                    if rx.data_type == "apparent_chargeability":
+                    if (
+                        rx.data_type == "apparent_chargeability"
+                        or self.data_type == "apparent_chargeability"
+                    ):
                         scale[src, rx] = self._sign / rx.eval(src, self.mesh, self._f)
             self._scale = scale.dobs
 
--- a/SimPEG/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/electromagnetics/static/induced_polarization/simulation.py
@@ -93,14 +93,6 @@ class BaseIPSimulation(BaseDCSimulation):
 
         return self.gtgdiag
 
-    # def getJ(self, m, f=None):
-    #     """
-    #         Generate Full sensitivity matrix
-    #     """
-    #     if self._Jmatrix is None:
-    #         self._Jmatrix = self._Jtvec(m, v=None, f=f).T
-    #     return self._Jmatrix
-
     # @profile
     def Jvec(self, m, v, f=None):
         return self._scale * super().Jvec(m, v, f)
--- a/SimPEG/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/electromagnetics/static/induced_polarization/simulation.py
@@ -1,14 +1,13 @@
 import numpy as np
 import properties
-from ....utils.code_utils import deprecate_class
+from ....utils.code_utils import deprecate_class, deprecate_property
 
 from .... import props
 from ....data import Data
-from ....utils import mkvc, sdiag
-from ...base import BaseEMSimulation
+from ....utils import sdiag
 
 from ..resistivity.simulation import BaseDCSimulation
-from ..resistivity.fields import FieldsDC, Fields3DCellCentered, Fields3DNodal
+from ..resistivity.fields import Fields3DCellCentered, Fields3DNodal
 from ..resistivity import Simulation3DCellCentered as BaseSimulation3DCellCentered
 from ..resistivity import Simulation3DNodal as BaseSimulation3DNodal
 
@@ -17,8 +16,12 @@ class BaseIPSimulation(BaseDCSimulation):
 
     eta, etaMap, etaDeriv = props.Invertible("Electrical Chargeability")
 
-    data_type = properties.StringChoice(
+    _data_type = properties.StringChoice(
         "IP data type", default="volt", choices=["volt", "apparent_chargeability"],
+    )
+
+    data_type = deprecate_property(
+        _data_type, "data_type", new_name="receiver.data_type", removal_version="0.16.0"
     )
 
     Ainv = None
@@ -48,7 +51,7 @@ class BaseIPSimulation(BaseDCSimulation):
                 for rx in src.receiver_list:
                     if (
                         rx.data_type == "apparent_chargeability"
-                        or self.data_type == "apparent_chargeability"
+                        or self._data_type == "apparent_chargeability"
                     ):
                         scale[src, rx] = self._sign / rx.eval(src, self.mesh, self._f)
             self._scale = scale.dobs
--- a/SimPEG/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/electromagnetics/static/induced_polarization/simulation.py
@@ -7,18 +7,13 @@ from ....data import Data
 from ....utils import mkvc, sdiag
 from ...base import BaseEMSimulation
 
+from ..resistivity.simulation import BaseDCSimulation
 from ..resistivity.fields import FieldsDC, Fields3DCellCentered, Fields3DNodal
 from ..resistivity import Simulation3DCellCentered as BaseSimulation3DCellCentered
 from ..resistivity import Simulation3DNodal as BaseSimulation3DNodal
 
 
-class BaseIPSimulation(BaseEMSimulation):
-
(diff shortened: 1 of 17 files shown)


jcapriot:
In relation to #976

jcapriot:
Ok, so I've come across actually a rather large bug, when a single receiver object is used for multiple sources. It must have multiple "dc_voltage" values. This actually also replicates to a bug in the geometric factor (which is also dependent on the sources. This PR will fix both of these problems soon.

jcapriot:
apparent_resitivity is perfectly fine to store on the receiver, but I think in the future it will be best to store dc_voltages on the simulation

jcapriot:
This PR is now in a good state on my end to merge. @sgkang can you take a once over please?

Some highlights
* Simplification of the IP code for Jvec, JTvec, fields to rely on the parent dc simulation
* Stores geometric factor for apparent resistivity on the receivers as a dictionary indexed by the sources (allows for the same receiver to listen to multiple sources for this data_type)
* Store the _dc_voltage for apparent chargeability on the IP simulation class. This should not be on the receivers because it is not only unique to sources, but also to the mesh and conductivity model. This now allows for a single dc receiver to listen to multiple sources for apparent chargeability data.
* Updates for dcip tutorials for the new changes to discretize, instead of loading in a true_model file it's much easier to just recreate it every time. There's also some streamlining of the plotting commands in those files as well.
* Update the intersphinx mapping for matplotlib

dccowan:
I took a look at the changes. I think this is a lot cleaner and removes some redundancies. Given that it passes all the tests (including reproducing the tutorials) I think it's good to go.

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -7,35 +7,27 @@
...
-    sigma = props.PhysicalProperty("Electrical conductivity (S/m)")
-
-    rho = props.PhysicalProperty("Electrical resistivity (Ohm m)")
-
-    props.Reciprocal(sigma, rho)
+class BaseIPSimulation(BaseDCSimulation):
 
     eta, etaMap, etaDeriv = props.Invertible("Electrical Chargeability")
 
     data_type = properties.StringChoice(
```
sgkang, line 20: @jcapriot: do we need this in the simulation class?

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -101,145 +84,24 @@ def getJtJdiag(self, m, W=None):
...
             else:
-                W = W.diagonal() ** 2
-
-            diag = np.zeros(J.shape[1])
-            for i in range(J.shape[0]):
-                diag += (W[i]) * (J[i] * J[i])
+                W = (self._scale * W.diagonal()) ** 2
 
-            self.gtgdiag = diag
+            self.gtgdiag = np.einsum("i,ij,ij->j", W, J, J)
```
sgkang, line 92: huh, that's cool. is this much more efficient?

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -101,145 +84,24 @@ def getJtJdiag(self, m, W=None):
...
-
-            for rx in src.receiver_list:
-                df_dmFun = getattr(f, "_{0!s}Deriv".format(rx.projField), None)
-                df_dm_v = df_dmFun(src, du_dm_v, v, adjoint=False)
-                Jv.append(rx.evalDeriv(src, self.mesh, f, df_dm_v))
-
-        # Conductivity (d u / d log sigma) - EB form
-        # Resistivity (d u / d log rho) - HJ form
-        return self.sign * np.hstack(Jv)
+        return self._scale * super().Jvec(m, v, f)
```
sgkang, line 98: nice job.

SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
```diff
@@ -40,19 +36,18 @@ def fields(self, m):
...
+        if self._scale is None:
+            scale = Data(self.survey, np.full(self.survey.nD, self._sign))
+            f = self.fields_to_space(self._f)
             # loop through receievers to check if they need to set the _dc_voltage
             for src in self.survey.source_list:
                 for rx in src.receiver_list:
                     if (
                         rx.data_type == "apparent_chargeability"
-                        and rx._dc_voltage is None
+                        or self.data_type == "apparent_chargeability"
```
sgkang, line 47: again do we need this?

could be useful, if one wants to set the data type by passing the `data_type` to a simulation.

SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
```diff
@@ -227,12 +224,12 @@ def MfRhoDerivMat(self):
         Derivative of MfRho with respect to the model
         """
         if getattr(self, "_MfRhoDerivMat", None) is None:
+            drho_dlogrho = sdiag(self.rho) * self.etaDeriv
```
sgkang, line 227: may be a different name `drho_dlogrho_deta_dm` ?

SimPEG/electromagnetics/static/resistivity/receivers.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 from ....utils.code_utils import deprecate_class
+from discretize.utils import Identity
```
sgkang, line 3: I did not know this exists :)

SimPEG/electromagnetics/static/resistivity/receivers.py
```diff
@@ -77,17 +73,45 @@ def eval(self, src, mesh, f):
...
+                    return v / self.geometric_factor[src][:, None]
+                return v / self.geometric_factor[src]
+            except KeyError:
+                raise KeyError(
+                    "Receiver geometric factor has not been set, please execute "
+                    "survey.set_geometric_factor()"
+                )
+        return v
+
+    def evalDeriv(self, src, mesh, f, v=None, adjoint=False):
```
sgkang, line 90: so, this is called many times. 
can we encode `geometric_factor` in `P`? (I guess this was the previous implementation).

tests/em/static/test_DC_2D_analytic.py
```diff
@@ -330,5 +330,54 @@ def test_Simulation2DNodal(self, tolerance=0.05):
         self.assertLess(err, tolerance)
 
 
+class DCSimulationAppResTests(unittest.TestCase):
```
sgkang, line 333: @jcapriot: thanks for adding the test!

tests/em/static/test_IP_2D_fwd.py
```diff
@@ -119,5 +118,122 @@ def test_Simulation2DCellCentered(self):
         self.assertTrue(passed)
 
 
+class ApparentChargeability2DTest(unittest.TestCase):
```
sgkang, line 121: nicely done!

sgkang (approved):
@jcapriot: Great job! and sorry for being late about this review. 
It really simplifies the IP code base, and I loved how you modified. 
Made a few minor comments, and it will be great if you can take a look. 

All the best, 

Seogi.

SimPEG/electromagnetics/static/induced_polarization/simulation.py
jcapriot: We certainly do not, but I left it there for backwards compatibility. I should probably add a deprecation warning saying to define it on the receiver.

SimPEG/electromagnetics/static/induced_polarization/simulation.py
jcapriot: Much more, Dom did a few tests with it to verify in #968

SimPEG/electromagnetics/static/induced_polarization/simulation_2d.py
jcapriot: Again, I think it should be more noted as deprecated behavior.

SimPEG/electromagnetics/static/resistivity/receivers.py
jcapriot: There is an Identity, but I think I actually don't need to import it here...

SimPEG/electromagnetics/static/resistivity/receivers.py
jcapriot: Unfortunately not, which was the reason for switching. The geometric factor is dependent on the src, so either we store a `P` for every [source, receiver] , or a `geometric_factor` for every [source, receiver] pair. This is more memory efficient with only a slight hit to computation time. This also allows for a single receiver object to be attached to many sources.

sgkang: got it! that's a good point that I did not take into account before.

PR #982, opened 2021-04-02
Update plot_inv_1a_gravity_anomaly.py

RichardScottOZ:
Minor edits

### DIFF
--- a/examples/02-gravity/plot_inv_grav_tiled.py
+++ b/examples/02-gravity/plot_inv_grav_tiled.py
@@ -37,12 +37,12 @@ import shutil
 #
 # Define the survey and model parameters
 #
-# Create an a global survey and mesh and simulate some data
+# Create a global survey and mesh and simulate some data
 #
 #
 
 
-# Create and array of observation points
+# Create an array of observation points
 xr = np.linspace(-30.0, 30.0, 20)
 yr = np.linspace(-30.0, 30.0, 20)
 X, Y = np.meshgrid(xr, yr)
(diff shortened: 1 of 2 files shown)


RichardScottOZ:
Questions to add: it says you can create a starting model and/or a reference model....then proceeds to ignore the reference model for a simple background starting model. Should say what it is and why you use one?

thast (approved):
That looks good! Thanks again! Regarding your question about the reference model, I believed you got an answer on Slack (for the record here: if no reference model is provided when defining the regularization, the starting model is used as the reference model).

RichardScottOZ:
Yes, the important part bit being you can access with regularization.ref Dom said - and hence set values where you want to?

thast:
@RichardScottOZ:
In detail, you can set a reference model that you have previously built (let's call it `reference_model`) when defining the regularization. 
- For example `reg = regularization.Tikhonov(mesh=mesh, mref=reference_model)` 
- This is similar to doing what you just said: `reg.mref=reference_model`
- In addition, you can also act on adding bounds constraints. the `ProjectedGNCG` accepts arrays as `lower` and `upper` bounds. For example, you can define `opt=optimization.ProjectedGNCG(maxiter=20, lower=reference_model-1, upper=reference_model+1)`

RichardScottOZ:
Thanks very much forvthe detail.

RichardScottOZ:
So for lower and upper you can vary the array. Say I want a cover sequence of units to stay the same, set the lower=upper at those indices?

thast:
yes that works.

RichardScottOZ:
Thanks.

PR #984, opened 2021-04-08
typo in the init

lheagy:
The 3D `centred` should match `centered`. Fixed a typo in the init that mapped `centred` to the 2D simulation


thast (approved):
good catch! That looks good to me.

PR #989, opened 2021-04-25
Extend BaseScr to that Pole and Dipole DC are simply minor extensions…

ngodber:
Introduce explicit Multipole source and minor fixes to correct runtime warnings around syntax conventions.


ngodber:
@dccowan / @jcapriot  The tests seem to be failing when static source classes are called with a numpy array. I've added code so that when this occurs appropriate type conversion occurs however it seems the properties extension cuts in before this type conversion occurs and throws an error. The only way I see to get around this is to use properties.Union() for location however internally its always a list of 1d arrays, so this is more than a work around then a solution. Is this understanding correct?

PR #991, opened 2021-05-01
Em1d hack 04/30/2021

sgkang:
- Deprecated sources, receivers, and survey of the frequency_domain_1d. 
- Updated `EM1DFMSimulation` to be consistent with the existing receivers, sources, and survey of the 3D FDEM simulation. 
- Made some updates on the side of the sources, receivers, and survey of the 3D FDEM simulation to handle `EM1DFMSimulation`. 
- Updated the tests. 
- Still, some parts of the code assumes that a receiver object contains a single location; so, when multiple locations are put, errors were raised.


sgkang:
# Some notes:

- `EM1DFMSimulation` is optimized. Faster than the previous version. 
- `StitchedEM1DFMSimulation` is in good shape for the serial case, but we are not gaining much from parallelization. It seems related to the overhead of passing source_list.

sgkang:
## All tests passing

 This is a wrap on my end for the first crack of the `em1d` code. So, if both of you: @jcapriot and @dccowan could have a look and provide feed-back that would be great! 

- Now, it uses the existing source, receiver, & survey class of 3d em codes. 
- The code is optimized by sorting for each sounding, and calculating the reflection coefficient once for all data included in a sounding. 
- Added an option for chunk for parallelization (seems working better for the stitched frequency domain)
- This optimization sacrificed a generality in the code. Assumptions that I made are given below. 

## Assumptions
- A receiver object only contains a single location
- Input of a sounding number to a source object (if the user did not input the sounding number, it is assumed that all sources have the same sounding number; might be a bad assumption, and expand this).
- In each sounding, it is assumed: 1) all source locations are the same, 2) all vertical offsets between receiver and source are the same (radial offsets can vary; e.g. Dighem survey)
- For the time-domain, I also assumed that all radial offsets between receiver and source are the same in a sounding.  
- For the stitched problem, I assumed a certain order of generating source_list, see below links for examples:
[https://github.com/simpeg/simpeg/blob/c138d9d5de46ca90e73d98c88e460e62d8785e3f/tests/em/em1d/testGlobalEM1D_FD_jac_layers.py#L49](url)) 
[https://github.com/simpeg/simpeg/blob/c138d9d5de46ca90e73d98c88e460e62d8785e3f/tests/em/em1d/testGlobalEM1D_TD_jac_layers.py#L69](url)

PR #993, opened 2021-05-03
Dc boundary conditions updates

jcapriot:
updates DC boundary condition code to use new implementation in discretize 0.7.0


ngodber:
Awesome work. Can't wait to use it in anger.

PR #1004, opened 2021-05-17
Depth weighting function in "model_utils.py"

xiaolongw1223:
I added the depth weighting function in the "model_utils.py" file as another weighting option for users compared with sensitivity weighting. The constant parameters "v" and "z0" are adjustable to simulate the decay of potential filed data along the depth.
 
We can perform depth weighting function as following:

wr = depth_weighting(mesh, xyz_topo, indActive=ind_active, v=3, z0=5)
reg.cell_weights = wr

I did a few tests shown below:

Gravity inversion:
 
Screen Shot 2021-05-16 at 8 34 20 PM

Magnetic inversion:

Screen Shot 2021-05-16 at 8 35 35 PM


RichardScottOZ:
Thanks, interesting!

xiaolongw1223:
> Thanks, interesting!

You are welcome!

jcapriot:
I'm unsure of the way topography is handled within this function. For one, this is only applicable to `TensorMesh`, also the internal behavior (where there is a lot of shifting of z values relative to topography) is not documented or clear to someone else using this function.

I'll also let @fourndo provide his opinions on this as well.

domfournier:
Right, I got the feeling we could just use cKDTree to find the nearest points from the topo array to all cell centers, using one less dimension than the input mesh (so good for both 2D and 3D), then just apply the delta_(y|z) to the function. Altering to distance weights would be trivial from there...

xiaolongw1223:
> I'm unsure of the way topography is handled within this function. For one, this is only applicable to `TensorMesh`, also the internal behavior (where there is a lot of shifting of z values relative to topography) is not documented or clear to someone else using this function.
> 
> I'll also let @fourndo provide his opinions on this as well.

Thanks for comments! Yes, this function is only applicable to the TensorMesh. I rarely use other mesh, but I think we should make it more applicable.  For the second point, though shifting has not been documented so far, I think shifting is reasonable. Without shifting, the depth would have both positive and negative values. In this case, the W(z) can not monotonically decrease. 

Dom provide another way to go. We can re-do this function.

xiaolongw1223:
> Right, I got the feeling we could just use cKDTree to find the nearest points from the topo array to all cell centers, using one less dimension than the input mesh (so good for both 2D and 3D), then just apply the delta_(y|z) to the function. Altering to distance weights would be trivial from there...

Thanks to comments! Not sure I understand, Does delta_(y|z) mean the distance from topo to cell centers? I would assume this method is applicable to different types of mesh.

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,73 @@ def ismember(a, b):
...
+    """
+    Create depth weighting
+
+    Parameters
+    ----------
+    mesh : Mesh object
+        discretize model space.
+    topo : numpy.ndarray
+        [X,Y,Z] topographic data.
+    indActive : array of bool
```
domfournier, line 275: could be optional

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,73 @@ def ismember(a, b):
...
+        index vector for the active cells on the mesh below the topography
+    v : float
+        exponent constant parameter for depth weighting.
```
domfournier, line 278: probably a good idea to supply a default

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,73 @@ def ismember(a, b):
...
+    z0 : float
```
domfournier, line 279: same here. z0 = 1/2 smallest cell?

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,73 @@ def ismember(a, b):
...
+
+    wz = np.power(
+        -mesh.vectorCCz + factor + z0,
+        -v
+        )
+
+    wz = wz ** 0.5
+
+    # Normalize
+    wz /= np.nanmax(wz)
```
domfournier, line 313: normalization should happen after applying the active indices

domfournier (commented):
Will be a great function to have. See comments below

xiaolongw1223:
> scipy.spatial.cKDTr

Thanks for clarification!

SimPEG/utils/model_utils.py
xiaolongw1223: I agree, will do that!

SimPEG/utils/model_utils.py
xiaolongw1223: Thanks

jcapriot:
In my opinion, I think the clearest implementation would simply be something like (in pseudocode):

```
def depth_weighting(mesh, z0=None, v=2):

    if z0 is None:
       z0 = mesh.nodes[:, -1].max()
    if z0 is scalar:
        dz = mesh.cell_centers[:, -1] - z0
    elif z0 is 2d array of shape (points, dim):
        f = make_interpolate_function(z0)
       dz = mesh.cell_centers[:, -1] - f(mesh.cell_centers[:, :-1])
    
    wz = np.abs(dz)**(-v/2)
    return wz/np.nanmax(wz)
```

Depth weighting doesn't really have anything to do with topography, it is simply the depth from observation points...

domfournier:
Yes and no, usually we like to normalize on the active set only, otherwise the cells containing the topo/obs will skew it.

RichardScottOZ:
What about depth weighting for reasons like basement? e.g. if you have an 800m hill made of sand you might not want to call 200m down from there the same as 200m down from a mostly rest of flat ground.

domfournier:
@RichardScottOZ I hear you, but most use-cases are to counteract the decay of sensitivities. As @jcapriot said, depth/distance from the receiver locations is usually what we would be passing to the function. If you want to weight higher some feature for a good reason, that becomes more a reference/constraints building problem.

RichardScottOZ:
Yep, fair enough.

xiaolongw1223:
> In my opinion, I think the clearest implementation would simply be something like (in pseudocode):
> 
> ```
> def depth_weighting(mesh, z0=None, v=2):
> 
>     if z0 is None:
>        z0 = mesh.nodes[:, -1].max()
>     if z0 is scalar:
>         dz = mesh.cell_centers[:, -1] - z0
>     elif z0 is 2d array of shape (points, dim):
>         f = make_interpolate_function(z0)
>        dz = mesh.cell_centers[:, -1] - f(mesh.cell_centers[:, :-1])
>     
>     wz = np.abs(dz)**(-v/2)
>     return wz/np.nanmax(wz)
> ```
> 
> Depth weighting doesn't really have anything to do with topography, it is simply the depth from observation points...

I think you are right, Joe. Thanks for your comments! z0 with shape of 2d array corresponds to having the topography.

xiaolongw1223:
@domfournier  Hi Dom, could you please take a review on updates. I made some updates and add more documentations. Thanks!

xiaolongw1223:
test results

plt_model_rec

depth_log

jcapriot:
> Yes and no, usually we like to normalize on the active set only, otherwise the cells containing the topo/obs will skew it even though they are not even part of the inversion model.

Yeah, but that is an `ind_active` thing, which I agree should be in there to only normalize by the maximum value of the active cells. Unless you’re suggesting something else in terms of adding some sort of epsilon to the regularize the `1/(z-z0)`  part when they are very close

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,69 @@ def ismember(a, b):
...
+
+        The depth weighting depends on the depth from receiver locations.
+
+
+      """
+
+
+    # Set up default value of z0
+    if z0 is None:
+        z0 = 0.5 * mesh.h_gridded.min()
```
domfournier, line 304: ok I think we are getting confused between @jcapriot and mine. So z0 will be the small constant to basically smooth the function and avoid zero division.
This is different than the topo/obs values (maybe reference_xyz?) that we to compute distances/depth from. It doesn't make much sense here to compute a distance from the smallest cell size...

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,69 @@ def ismember(a, b):
...
+        tree = cKDTree(z0[:, :-1])
+        _, ind = tree.query(mesh.cell_centers[:, :-1])
+        delta_z = np.abs(mesh.cell_centers[:, -1] - z0[ind, -1])
+
+    else:
+        raise Exception(
+            "z0 must be either a scalar or 2d array!"
+            )
+
+    wz = (delta_z) ** (-0.5*v)
```
domfournier, line 322: add back `z0` with the `delta_z`

domfournier (changes requested):
Ok great, getting there. See comment below regarding `z0`

SimPEG/utils/model_utils.py
xiaolongw1223: I see. So now we actually need three variables for this function: v (exponent value), reference_loc (scalar or 2d array to compute delta_z) and z0 (small constant avoiding zero in denominator). Does this make sense?

domfournier: sounds right, and only `mesh` and `reference_locs` are required, the `z0` and `exp` can be kwargs with default as you have it now.

xiaolongw1223: Sounds good. I will update right later. Thanks!

domfournier: sorry for being picky, but can we use plain English too for the variable names. Like `z0` -> `threshold`, `v` -> `exponent`

xiaolongw1223: No worries, sure, that does more explicit.

domfournier:
@xiaolongw1223 Just came to mind. What are you doing with the volumetric part? Are you multiplying outside with mesh.volumes? If not, we should probably multiply in here before normalizing...

xiaolongw1223:
> @xiaolongw1223 Just came to mind. What are you doing with the volumetric part? Technically this is a volume integral and we need to worry about changing cell size... Are you multiplying outside with mesh.volumes? If not, we should probably multiply in here before normalizing...

yup, I multiplied mesh.vol in my script. In the latest updates, I added the volumetric information in the function, which is more convenient.

domfournier (approved):
That looks fantastic. Well done!

xiaolongw1223:
> That looks fantastic. Well done!

Thank you all for help! @domfournier @jcapriot

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,75 @@ def ismember(a, b):
...
+        exponent parameter for depth weighting.
+    threshold : float
+        adjustable constant parameter.
+
+    Returns
+    -------
+    wz : numpy.ndarray
+        normazlied depth weighting captures decay of the potential field data.
+
+
```
jcapriot, line 288: You need to add in a line here distinguishing the rest as Notes

```
Notes
------
```

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,77 @@ def ismember(a, b):
...
+    # Default exponent value
+    if exponent is None:
+        exponent = 2.0
+
+    # Default threshold value
+    if threshold is None:
+        threshold = 0.5 * mesh.h_gridded.min()
+
+    assert isinstance(reference_locs, (float, np.ndarray))
```
jcapriot, line 310: do not use asserts,
also this is way too limiting, best idea would probably be to just do:
`reference_locs = np.asarray(reference_locs)`

Then further down (line 314) just do
`if reference_locs.ndim < 2:`

SimPEG/utils/model_utils.py
```diff
@@ -260,3 +260,77 @@ def ismember(a, b):
...
+    -------
+    wz : numpy.ndarray
+        normazlied depth weighting captures decay of the potential field data.
+
+    Notes
+    -----
+
+    ..math::
+
+        w(z) = (delta_z + threshold) ** (-0.5 * exponent)
```
jcapriot, line 293: volume should also be here

SimPEG/utils/model_utils.py
xiaolongw1223: if '''reference_locs.ndim == 1''', '''reference_locs''' is a 1d array. 
But seems we cannot do '''mesh.cell_centers[:,-1] - reference_locs''' if they have different size.

So I just do:
'''
if reference_locs.ndim == 0:
'''

xiaolongw1223:
test `code`

SimPEG/utils/model_utils.py
jcapriot: That’s why I suggested doing `ndim < 2` also if you try it, there’s a few different cases where it would still work as you’ve written it, if it’s a 0d array, a 1d array with one element, or a 1d array of the same length as the number of cells.

xiaolongw1223: I see, thanks!

xiaolongw1223:
Please find the test results here: https://curvenote.com/@xiaolong_wei/simpeg-depth-weighting/test-depth-weighting

This PR is ready to merge

jcapriot:
So after chatting more with @domfournier, we've decided that this function actually should NOT pre-multiply by volume (or normalize by it either). We will likely be creating an issue and PR soon about updating the current sensitivity weighting code to do this as well, as it is counterintuitive to not multiply by cell volumes within the regularization only for certain weighting schemes.

xiaolongw1223:
> So after chatting more with @domfournier, we've decided that this function actually should NOT pre-multiply by volume (or normalize by it either). We will likely be creating an issue and PR soon about updating the current sensitivity weighting code to do this as well, as it is counterintuitive to not multiply by cell volumes within the regularization only for certain weighting schemes.

Hi Joe, I will update this function accordingly. Thanks for clarification.

PR #1009, opened 2021-05-20
Fix PGI gallery example

thast:
This PR aims at fixing the gallery examples for PGI; they work but do not appear on the documentation website.
I figure it could be because I omitted to name them according to the naming convention.


jcapriot:
ya, that would be why. I was only looking at the tutorials before and didn't realize you had also added gallery examples

PR #1013, opened 2021-06-22
Em1d updates

jcapriot:
This represents my major edits for the em1d branch. It only changes the normal simulation routines to be more flexible and more optimized under the hood to better support the simpeg sources and receivers.

items left to do on the em1d branch include to:
- [ ] Perform further tests with the on-time measurements with arbitrary waveforms. (The magnetic field looks to be shifted, but dbdt looks good).
- [ ] Update the psuedo 3D inversions (the stitched inversions) to this way ( @sgkang we should only need to change a few lines as you and I discussed).
- [ ] better integrate the IO routines


sgkang:
Let's merge this and work with em1d branch.

Issue #771, opened 2019-03-28
Gauss Newton Minimization Error


Rockpointgeo:
I was working through the SimPEG examples trying to understand how the guts of the code work and I find the easiest way to do that is to keep it as simple as possible. So I wanted to try and solve the toy linear problem using GaussNewton (basically something where the matrix is actually being formed and inverted) so I can look at things step by step. Unfortunately, when we I set `opt = Optimization.GaussNewton(maxIter=60)` I get the following error:
```
'_CustomLinearOperator' object has no attribute 'tocsc'
```
The SolverWrapper assumes that the matrix being passed to it is sparse. The error occurs when the code in the wrapper tries to set the sparse matrix to a csc_sparse format. 

There are a few things going on.
1) The wrapper for direct solvers `SolverWrapD` assumes the system matrix is sparse.
2) `SolverWrapD` is currently used for `Solver` which is wraps `scipy.sparse.linalg.spsolve` and `SolverLU` which wraps `scipy.sparse.linalg.splu`. The `splu` code is more restrictive and requires a csc or csr sparse matrix. spsolve accepts an ndarray or sparse matrix.
3) `SolverLU` is used in many places within the code for forward modelling. `Solver` is only used in `Optimization.GaussNewton`
4) The argument `H` that is passed to `Solver` in generated by evalfunction. Unfortunately (for this case) the function creates a linear operator. It is a relatively easy task to generate a sparse representation of `H` from the Linear operator.

My solution is to change the existing code
```
    @Utils.timeIt
    def findSearchDirection(self):
        return Solver(self.H) * (-self.g)
```
to
```
    @Utils.timeIt
    def findSearchDirection(self):
         if isinstance(self.H, sp.linalg.LinearOperator):
            self.H = sp.csr_matrix(self.H.dot(np.identity(self.g.size)))
        elif isinstance(self.H, np.ndarray):
            self.H = sp.csr_matrix(self.H)

        return Solver(self.H) * (-self.g)
```
but maybe the checking isn't necessary and a simpler fix would be
```
    @Utils.timeIt
    def findSearchDirection(self):
        self.H = sp.csr_matrix(self.H.dot(np.identity(self.g.size)))
        return Solver(self.H) * (-self.g)
```

I have made a notebook that shows the error and I wrote code to replicate `inv.run` and was able to fix the problem.

It is here https://colab.research.google.com/drive/1iGJ7XTTLdxrb0JjJgirnsCTFMoI6hIlp

[...]

PR #1014, opened 2021-06-23
Fix for issue 771

WouterDls:
Not sure if I'm supposed to do it like this. 

Using an iterative solver instead of a direct solver solves the bug that was exposed in #771  (the wrapper of the direct solver requires sparse matrices, instead of sparse linear operators).

I'm not completely sure if iterative solvers are optimal here, but it works.

### DIFF
--- a/SimPEG/optimization.py
+++ b/SimPEG/optimization.py
@@ -4,7 +4,7 @@ import numpy as np
 import scipy.sparse as sp
 from six import string_types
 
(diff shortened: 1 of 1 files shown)


jcapriot:
Similar comment to  #772, It'll be best to adapt the solver used to either be the sparse version or the dense version depending on the type of `H` If you want to use a `LinearOperator`, you should use the `InexactGaussNewton` method which uses an iterative solver to find the search direction.

PR #1016, opened 2021-06-28
Remove dead code

prisae:
These four lines are "dead code", as they are overwritten later on.

The alternative of removing them would be to add a `return sx, sy, sz` at the end of the removed lines. However, this if-statement is only ever called if two points are identical (`lx=ly=lz=0`), so should very rarely be the case. Additionally, the speed-gain is probably negligible.


prisae:
This function contains further a repetition, it does 3x the same thing. If there is interest I can push a commit simplifying this as well.

prisae:
I pushed the refactor mentioned above. If you prefer not to have this refactor then simply remove the second commit...

sgkang (approved):
Looks good to me. Thanks for cleaning this up.

PR #1018, opened 2021-07-13
Total vector map

domfournier:
An attempt at reproducing the vector inversion regularized by amplitude where

\phi_m = \| w (a - a_{ref}) \| = \sum w_i^2 (a_i - a_ref)^2
and

a = ((P_x m)^2 + (P_y m)^2 + (P_z m)^2)^0.5


SimPEG/maps.py
```diff
@@ -616,6 +616,66 @@ def deriv(self, m, v=None):
...
+        Shape of the matrix operation (number of indices x nP)
+        """
+        return (self.nP, 3*self.nP)
+
+    def _transform(self, model):
+        """
+        :param model:
+        :return:
+        """
+        return np.linalg.norm(self.wires*model, axis=0)
```
lheagy, line 650: Have you looked at all at using the amplitude squared rather than amplitude? I appreciate it is a different function, but derivatives are much nicer

`np.sum((self.wires*model)**2, axis=0)`

domfournier: So the square is technically happening in the reg.__call__. That's the confusing part. We are sort of mapping the parameters, but with the intention of using it in an L2 function? Or are you suggesting to minimize \| a^2 \|?

SimPEG/maps.py
```diff
@@ -616,6 +616,66 @@ def deriv(self, m, v=None):
...
+            if v is not None:
+                derivs += [A * wire * v]
+            else:
+                derivs += [A * wire.P]
+
+        return derivs
+
+    def deriv2(self, m, v=None):
+        """
+        """
```
lheagy, line 668: I don't think this is quite right? both the numerator and denominator have derivs 

continuous: 
da / dx = x/a
d2a / dx2 = 1/a - x^2 / a^2 = (a - x^2) / a

SimPEG/maps.py
lheagy: here, I am thinking about minimizing | a^2 | (or | a^2 - aref^2| ). It would get rid of the square root and having to deal with a denominator (that could potentially go to zero).

SimPEG/maps.py
domfournier: yea I know it's weird. But the deriv of a^2 is technically 2 (a / a ) * da/dm ?

SimPEG/maps.py
domfournier: yep gotcha, haven't tried that.

SimPEG/maps.py
lheagy: here though we need the second derivative, not the deriv of the squared amplitude. 

For now, if we  stick with amplitude as defined in the transform a = sqrt(x^2 + y^2 + z^2), then we need the second deriv of the amplitude wrt the model. 

so 

da / dm = [diag(x/a), diag(y/z), diag(z/a)]  (* dwires_dm * v) 

and then the second deriv, if we just look at the first part (x) should be 

d2a / dx2 = 1/a - x^2 / a^2 = (a - x^2) / a

lheagy: though for mappings, I don't think we actually need the second deriv. I think we should only ever need the first

PR #1019, opened 2021-07-15
Update receivers.py

lheagy:
typo in the mag receiver docs ("components" rather than "component" is the keyword. 

Out of curiosity, is there a benefit to providing a list here `components = ["bx", "by", "bz"]` as opposed to creating three receivers? 

```
rx_x = mag.receivers.Point(locations=survey_xyz, components="bx")
rx_y = mag.receivers.Point(locations=survey_xyz, components="by")
rx_z = mag.receivers.Point(locations=survey_xyz, components="bz")
```

### DIFF
--- a/SimPEG/potential_fields/magnetics/receivers.py
+++ b/SimPEG/potential_fields/magnetics/receivers.py
@@ -9,7 +9,7 @@ class Point(survey.BaseRx):
     Magnetic point receiver class for integral formulation
 
     :param numpy.ndarray locs: receiver locations index (ie. :code:`np.c_[ind_1, ind_2, ...]`)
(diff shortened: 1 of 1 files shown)


jcapriot:
Thanks for the typo. From what I understood as far as accepting multiple components, it allowed for more operations to be easily combined under the hood.

domfournier:
Correct, creating three receivers (or more) would require multiple calls to the `evaluate_integral`, which is the meat of creating the kernels for mag and grav. It's just faster to do it all at once since they all have the same xyz position. I get it, it's not as pretty, but it's more efficient.

lheagy:
all good :) thanks for clarifying!

PR #1021, opened 2021-07-21
Pgi improvements

thast:
**(This is a draft)**

Various quality of life and numerical improvements for PGI

_done:_
- better numerical implementation of the gradients when the GMM clusters are overlapping
- Implements the Hessian for the case  when the GMM clusters are overlapping
- better default for the `wiresmap` in the `utils` tools used to create the PGI regularization

_To Do List:_
- embedded plotting function for GMM (and subsequently use it in the tutorial)
-


SimPEG/regularization/pgi.py
```diff
@@ -648,6 +725,7 @@ def __init__(
         gmm=None,
```
domfournier, line 725: I am not seeing anything different from the `SimplePGI` here? Any chance we can only keep one of the two. Just a lot of code duplication....

SimPEG/utils/pgi_utils.py
```diff
@@ -95,7 +98,11 @@ def make_SimplePGI_regularization(
     """
 
     if wiresmap is None:
-        wrmp = Wires(("m", mesh.nC))
+        if "indActive" in kwargs.keys():
+            indActive = kwargs.pop("indActive")
+            wrmp = Wires(("m", indActive.sum()))
```
domfournier, line 103: careful with type here, I believe sum() will return a float

SimPEG/utils/pgi_utils.py
```diff
@@ -105,7 +112,8 @@ def make_SimplePGI_regularization(
         mplst = maplist
 
     if cell_weights_list is None:
-        clwhtlst = [Identity() for maps in wrmp.maps]
+        clwhtlst = [np.ones(maps[1].shape[0]) for maps in wrmp.maps]
+        print(clwhtlst[0].shape)
```
domfournier, line 116: guessing for testing - remove

SimPEG/utils/pgi_utils.py
```diff
@@ -115,13 +123,14 @@ def make_SimplePGI_regularization(
         gmm=gmm,
```
domfournier, line 123: It's a bit of a choice here, but we could just ask for kwargs and refer to the inputs of the PGI regularization, instead of duplicating all the input parameters. @jcapriot what's your opinion on this?

SimPEG/utils/pgi_utils.py
```diff
@@ -151,7 +160,7 @@ def make_SimplePGI_regularization(
             alpha_y=alph_y[i],
             alpha_z=alph_z[i],
             cell_weights=clwhtlst[i],
-            **kwargs
+            **kwargs,
```
domfournier, line 163: Could be an issue here if people try to input alpha_s as kwargs. Is there a reason it should always be zero and never set? We should probably assert it in the setter

SimPEG/utils/pgi_utils.py
```diff
@@ -164,6 +173,7 @@ def make_PGI_regularization(
     wiresmap=None,
```
domfournier, line 173: Same thing here, we are duplicating a LOT of code. Can we just keep one and be explicit about the assumption. Even if temporarily the volume part would be an issue in certain cases, we can fix the assumption after.

SimPEG/utils/pgi_utils.py
```diff
@@ -877,10 +894,253 @@ def score_samples_with_sensW(self, X, sensW):
             Log probabilities of each data point in X.
         """
         check_is_fitted(self)
-        X = _check_X(X, None, self.means_.shape[1])
+        X = self._validate_data(X, reset=False)
 
         return logsumexp(self._estimate_weighted_log_prob_with_sensW(X, sensW), axis=1)
 
+    def plot_pdf(
```
domfournier, line 901: Got the feeling this should go under the `plot_utils`?

domfournier (commented):
An impressive piece of work. Few comments below.

SimPEG/utils/pgi_utils.py
thast: hum, it could be. I have not thought about it in that way. the idea was to have it as an object function `gmm.plot_pdf`. Are there specific reasons you would prefer it as `utils.plot_pdf(gmm)`?

domfournier: Oh ok, didn't realize it was under the class. My bad.

SimPEG/utils/pgi_utils.py
thast: good catch!

SimPEG/regularization/pgi.py
thast: In short, the only difference is for the volume weights. All the rest is duplicated. The `pgi` classes mirror what was done with `Simple` and `Tikhonov`:
- `SimplePGI` is equivalent to `Simple`
- `PGI` is equivalent to `Tikhonov`

I am all for merging it in a single class, with a flag to turn volume weights ON and OFF. The main reason was to reflect the original `Simple` and `Tikhonov` classes.

thast: So, if we change `Simple` and `Tikhonov`, we could make the same changes to `pgi`

SimPEG/utils/pgi_utils.py
thast: +1

SimPEG/regularization/pgi.py
domfournier: I get it for sure. It would make my life easier to refactor if that cleanup was happening ahead. But we can tackle it on the next PR if you want to do it there.

thast: or could I use your latest work on `Sparse` with regards to volume weights and mirror that in `pgi`? (just worrying about consistency, but if you already have an example, we can make it happen in this PR).

domfournier: You can take a look at #1038 , but basically the rule now will be that the reg always multiply in the cell volumes. Sensitivity weights we be pre-normalize. So the PGI cell_weights should have volumes - exactly how `PGIsmallness` is implemented. So ideally just keep that one.

SimPEG/regularization/pgi.py
```diff
@@ -610,20 +680,27 @@ def W(self):
...
+                        speye(len(self.wiresmap.maps)),
+                        sdiag(np.sqrt(self.regmesh.vol)),
+                    )
+                    * sdiag(np.sqrt(self.cell_weights))
+                )
             else:
                 return sp.kron(
-                    speye(len(self.wiresmap.maps)), sdiag(np.sqrt(self.regmesh.vol)),
+                    speye(len(self.wiresmap.maps)),
+                    sdiag(np.sqrt(self.regmesh.vol)),
```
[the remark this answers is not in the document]
thast: `sp.kron` is used here with `len(self.wiresmap.maps)` for cases with multiple physical properties. `W` is then a matrix of (n*p) X (n*p) size (n:number of active cells, p: number of physical properties) with repeated volumes weights for each phys. prop (thus, the `sp.kron(speye(len(self.wiresmap.maps)), [...]`.

There might be a better way; we could form the diagonal with a nested loop `[weights for I in range(len(..)]` and then call `sdiag` but would that be a significant gain?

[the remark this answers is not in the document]
domfournier: It would certainly make the code mode readable with less operations. Plus you can avoid the if statement by always kroning your vector with len(self.wiresmap.maps)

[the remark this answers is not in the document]
domfournier: It would certainly make the code mode readable with less operations. Plus you can avoid the if statement by always kroning your vector with len(self.wiresmap.maps)

PR #1026, opened 2021-07-24
Replace np.matlib.repmat by np.tile

prisae:
Importing from `numpy.matlib` is deprecated since NumPy 1.19.0. This PR replaces the `np.matlib.repmat` by `np.tile`, as outlined here https://numpy.org/doc/stable/user/numpy-for-matlab-users.html


jcapriot (approved):
LGTM

PR #1027, opened 2021-07-24
Replace np.matrix by regular ndarrays

prisae:
Done in two distinct ways:
- Replace `np.matrix(x)` by `np.array(x)` where it is obvious that `x` is at least 2D.
- Replace `np.matrix(x)` by `np.atleast_2d(x)` where this is not the case.

This approach will fail if there is any place where the string annotation has been used:
```
np.matrix('1 2; 3 4')
```
I have not seen it though, so fingers crossed. The tests will show.

Also, matrix multiplication and matrix power might be different, which might require further adjustments. Again, the tests will show.

### DIFF
--- a/SimPEG/electromagnetics/analytics/NSEM.py
+++ b/SimPEG/electromagnetics/analytics/NSEM.py
@@ -15,14 +15,14 @@ _PCC = lambda siginf, m, t, c, f: siginf * (
 )
 
 # matrix P relating Up and Down components with E and H fields
-_P = lambda z: np.matrix([[1.0, 1,], [-1.0 / z, 1.0 / z],], dtype="complex_",)
-_Pinv = lambda z: np.matrix([[1.0, -z], [1.0, z]], dtype="complex_") / 2.0
+_P = lambda z: np.array([[1.0, 1,], [-1.0 / z, 1.0 / z],], dtype="complex_",)
+_Pinv = lambda z: np.array([[1.0, -z], [1.0, z]], dtype="complex_") / 2.0
 
 # matrix T for transition of Up and Down components accross a layer
-_T = lambda h, k: np.matrix(
+_T = lambda h, k: np.array(
     [[np.exp(1j * k * h), 0.0], [0.0, np.exp(-1j * k * h)]], dtype="complex_"
 )
-_Tinv = lambda h, k: np.matrix(
+_Tinv = lambda h, k: np.array(
     [[np.exp(-1j * k * h), 0.0], [0.0, np.exp(1j * k * h)]], dtype="complex_"
 )
 
@@ -54,8 +54,8 @@ def _Propagate(f, thickness, sig, chg, taux, c, mu_r, eps_r, n):
     K = k(f, sigcm, mu, eps)
     Z = _ImpZ(f, mu, K)
 
-    EH = np.matrix(np.zeros((2, n + 1), dtype="complex_"), dtype="complex_")
-    UD = np.matrix(np.zeros((2, n + 1), dtype="complex_"), dtype="complex_")
+    EH = np.zeros((2, n + 1), dtype="complex_")
+    UD = np.zeros((2, n + 1), dtype="complex_")
 
     UD[1, -1] = 1.0
 
@@ -65,7 +65,7 @@ def _Propagate(f, thickness, sig, chg, taux, c, mu_r, eps_r, n):
         UD = UD / ((np.abs(UD[0, :] + UD[1, :])).max())
 
     for j in range(0, n + 1):
-        EH[:, j] = np.matrix([[1.0, 1,], [-1.0 / Z[j], 1.0 / Z[j]],]) * UD[:, j]
+        EH[:, j] = np.array([[1.0, 1,], [-1.0 / Z[j], 1.0 / Z[j]],]) * UD[:, j]
 
     return UD, EH, Z, K
 
--- a/SimPEG/electromagnetics/analytics/NSEM.py
+++ b/SimPEG/electromagnetics/analytics/NSEM.py
@@ -15,15 +15,15 @@ _PCC = lambda siginf, m, t, c, f: siginf * (
 )
 
 # matrix P relating Up and Down components with E and H fields
-_P = lambda z: np.array([[1.0, 1,], [-1.0 / z, 1.0 / z],], dtype="complex_",)
-_Pinv = lambda z: np.array([[1.0, -z], [1.0, z]], dtype="complex_") / 2.0
+_P = lambda z: np.array([[1.0, 1,], [-1.0 / z, 1.0 / z]])
+_Pinv = lambda z: np.array([[1.0, -z], [1.0, z]]) / 2.0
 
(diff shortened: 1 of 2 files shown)


prisae:
There are, as expected, some issues with dot products etc...

prisae:
There are only two places where the deprecated `np.matrix` is used, and I think they go back to @thast and to @dccowan  - I therefore add you as requested reviewers, could you check your functions still work as intended with my changes?
- @thast : `SimPEG/electromagnetics/analytics/NSEM.py`
- @dccowan : `SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py`

SimPEG/electromagnetics/analytics/NSEM.py
```diff
@@ -15,14 +15,14 @@
 )
 
 # matrix P relating Up and Down components with E and H fields
-_P = lambda z: np.matrix([[1.0, 1,], [-1.0 / z, 1.0 / z],], dtype="complex_",)
-_Pinv = lambda z: np.matrix([[1.0, -z], [1.0, z]], dtype="complex_") / 2.0
+_P = lambda z: np.array([[1.0, 1,], [-1.0 / z, 1.0 / z],], dtype="complex_",)
```
jcapriot, line 18: These do not need do be declared as a complex type.

SimPEG/electromagnetics/analytics/NSEM.py
```diff
@@ -15,14 +15,14 @@
...
+_Pinv = lambda z: np.array([[1.0, -z], [1.0, z]], dtype="complex_") / 2.0
 
 # matrix T for transition of Up and Down components accross a layer
-_T = lambda h, k: np.matrix(
+_T = lambda h, k: np.array(
     [[np.exp(1j * k * h), 0.0], [0.0, np.exp(-1j * k * h)]], dtype="complex_"
 )
-_Tinv = lambda h, k: np.matrix(
+_Tinv = lambda h, k: np.array(
     [[np.exp(-1j * k * h), 0.0], [0.0, np.exp(1j * k * h)]], dtype="complex_"
```
jcapriot, line 26: same here (regarding explicit complex type)

SimPEG/electromagnetics/analytics/NSEM.py
```diff
@@ -65,7 +65,7 @@ def _Propagate(f, thickness, sig, chg, taux, c, mu_r, eps_r, n):
         UD = UD / ((np.abs(UD[0, :] + UD[1, :])).max())
 
     for j in range(0, n + 1):
-        EH[:, j] = np.matrix([[1.0, 1,], [-1.0 / Z[j], 1.0 / Z[j]],]) * UD[:, j]
+        EH[:, j] = np.array([[1.0, 1,], [-1.0 / Z[j], 1.0 / Z[j]],]) * UD[:, j]
```
jcapriot, line 68: the "*" operator should be changed to an "@" operator, on this line, and on line 64 above. What's going to throw most of the errors here is that the multiplication for a `numpy.matrix` was actual matrix multiplication (which is completely different than the `ndarray` class, clearly confusing and why they are getting rid of the matrix class.)

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
```diff
@@ -695,7 +695,7 @@ def _getGeometryMatrix(self, xyzc, xyzh, pp):
                         G[COUNT, :] = c * np.c_[Gxz, Gyz, Gzz]
                         COUNT = COUNT + 1
 
-        return np.matrix(G)
+        return np.atleast_2d(G)
```
jcapriot, line 698: G is already a 2D ndarray, (see it's initial creation way up on like 183-ish)

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
```diff
@@ -917,8 +917,8 @@ def fields(self, m):
         self.model = m  # Initiates/updates model and initiates mapping
 
         # Project to active mesh cells
-        # m = np.matrix(self.xiMap * m).T
-        m = np.matrix(self.xiMap * m).T
+        # m = np.atleast_2d(self.xiMap * m).T
+        m = np.atleast_2d(self.xiMap * m).T
```
jcapriot, line 921: I believe you should just be able to do...
` m = self.xiMap * m` (no need to convert it to anything at all)

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
```diff
@@ -934,10 +934,10 @@ def Jvec(self, m, v, f=None):
         dxidm = self.xiMap.deriv(m)
 
         # dxidm*v
-        v = np.matrix(dxidm * v).T
+        v = np.atleast_2d(dxidm * v).T
```
jcapriot, line 937: this shouldn't need to be here at all either
`v = dxidm * v`
(if I'm keeping track of dimensions correctly)

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
```diff
@@ -1047,7 +1047,7 @@ def fields(self, m=None):
                     self.tau2,
                 )
 
-                f.append(mkvc((self.A[qq] * np.matrix(eta)).T))
+                f.append(mkvc(np.dot(self.A[qq], np.atleast_2d(eta)).T))
```
jcapriot, line 1050: again, could've used an `@` operator.

jcapriot (changes requested):
Thank you for heading this off @prisae. There's certainly a lot to clean up from these older modules still too. I've added a couple suggestions. However the requested changes for the NSEM module (where `*` should be changed to `@` are required)

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
```diff
@@ -934,10 +934,10 @@ def Jvec(self, m, v, f=None):
         dxidm = self.xiMap.deriv(m)
 
         # dxidm*v
-        v = np.matrix(dxidm * v).T
+        v = np.atleast_2d(dxidm * v).T
 
         # Dot product with A
-        v = self.A * v
+        v = np.dot(self.A, v)
```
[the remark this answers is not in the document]
prisae: Good calls. I am still not used to the @ operator... Changing it now.

SimPEG/electromagnetics/analytics/NSEM.py
prisae: Good, I only looked into the matrix. Will remove these complex declarations.

SimPEG/electromagnetics/viscous_remanent_magnetization/simulation.py
prisae: Done, the tests should show if it works :-)

jcapriot:
Looking through the code, It would appear to me that these changes should be correct, even though they're not tested. In the long term, we will be moving them to `geoana`

PR #1032, opened 2021-08-21
SimPEG(emg3d) integration

prisae:
# SimPEG(emg3d)

PR to implement the required steps on SimPEG-side to use emg3d as a solver for EM modelling and inversion within SimPEG.

## TODO's

- [ ] An electric point receiver object can now have arbitrary rotations, however, these rotations have to be the same for the entire receiver (all locations). This should be generalized so that each location can have its own rotations.


prisae:
As agreed, I will still move the `_jtvec` over to `emg3d`.

prisae:
The relevant changes (only for `data_type='complex'`) are now in the `main` branch of `emg3d`.

prisae:
Sorry I had to force-push undoing the `merge main` -- somehow it didn't work any longer afterwards, postponing the update with main...

prisae:
Superseded by #1076, which is basically a cleaned-up version of this one.

Issue #763, opened 2019-03-05
What is the definition of misfit


dougoldenburg:
I think we need two distinct words: one describes the normed value; the other describes the final item that is put into phi_d in the inversion and includes the normed value multiplied by a scalar. 

![image](https://user-images.githubusercontent.com/6361812/53839437-548b7200-3f4c-11e9-9615-a7c6c3e37dab.png)

![image](https://user-images.githubusercontent.com/6361812/53839452-5e14da00-3f4c-11e9-8530-44d88cfeadfd.png)

thast:
Just to summarize my opinion here:
1) There are many options for normalizing (`1`, or `1/2` or `1/nd` (`nd`: number of data)). So whatever choice we make, we will have to be explicit about it and about the target misfit (`nd`, `nd/2` or `1`.).
2) I like the `1/2`, not (only) for the optimization or because it is already set up, but also (and that is a very biased reason I agree), is that it also comes naturally when thinking of the L2 norm as the representation of a Gaussian uncertainties on the data, and the `1/2` is part of the [definition of a Gaussian](https://en.wikipedia.org/wiki/Multivariate_normal_distribution):
![image](https://user-images.githubusercontent.com/13682747/53845659-c967a780-3f5e-11e9-920e-5dad586729b7.png)
3) If not `1/2`, my second choice would be `1/nd` for the data misfit , as then your target is always 1 no matter the number of data.

lheagy:
This got us (@dougoldenburg, @sgkang and I tripped up again)... I would be quite in favour of removing the factor of 1/2 and pointing to the fact we are building on Parker, Oldenburg, ...

santisoler:
Adding my two cents in the discussion after what @thibaut-kobold mentioned in #1332. I'm completely on board to stick with a convention and be super explicit about the definition and the implementation of objective functions.

On personal preference, I don't quite like the 1/2 for usability, but I get the mathematical aesthetics of it. For ultimate usability purposes, I think normalizing by the number of data points is the best choice: the target for the `phi_d` will be 1 for a `chi` factor of 1, so users don't need to keep the number of points in their buffer memory to interpret the values of `phi_d`.

I don't think we should normalize it by `n_data` right now, and I think removing the 1/2 as @lheagy did in #1326 it's one good move forward.

PR #1040, opened 2021-09-22
Refactor/regularization

domfournier:
- Overhaul of the Base, Simple and Tikhonov regularizations

### DIFF
--- a/SimPEG/regularization/__init__.py
+++ b/SimPEG/regularization/__init__.py
@@ -1,4 +1,4 @@
-from .base import BaseRegularization, BaseComboRegularization
+from .base import BaseRegularization, L2Regularization, BaseComboRegularization, Small, SmoothDeriv, SmoothDeriv2
 from .regularization_mesh import RegularizationMesh
 from .tikhonov import (
     SimpleSmall,
--- a/SimPEG/regularization/base.py
+++ b/SimPEG/regularization/base.py
@@ -20,7 +20,7 @@ class BaseRegularization(BaseObjectiveFunction):
 
     """
     _active_cells = None
-    _fixed_weights = None
+    _cell_weights = None
     _free_weights = None
     _mapping = None
     _reference_model = None
@@ -28,6 +28,7 @@ class BaseRegularization(BaseObjectiveFunction):
     counter = None
     _free_multiplier = 1.
     _model = None
+    _model_units = None
 
     def __init__(self, mesh=None, **kwargs):
         super().__init__()
@@ -37,36 +38,6 @@ class BaseRegularization(BaseObjectiveFunction):
         utils.setKwargs(self, **kwargs)
 
     # Properties
-    @property
-    def reference_model(self) -> np.ndarray:
-        """Reference physical property model"""
-        return self._reference_model
-
-    @reference_model.setter
-    def reference_model(self, values: np.ndarray | float):
-
-        if isinstance(values, float):
-            values = np.ones(self._nC_residual) * values
-
-        validate_array_type("reference_model", values, float)
-        validate_shape("reference_model", values, self._nC_residual)
-        self._reference_model = values
-
-    @property
-    def model(self) -> np.ndarray:
-        """Reference physical property model"""
-        return self._model
-
-    @model.setter
-    def model(self, values: np.ndarray | float):
-
-        if isinstance(values, float):
-            values = np.ones(self._nC_residual) * values
-
-        validate_array_type("model", values, float)
-        validate_shape("model", values, self._nC_residual)
-        self._model = values
-
     @property
     def active_cells(self) -> np.ndarray:
         """Indices of active cells in the mesh"""
@@ -83,15 +54,15 @@ class BaseRegularization(BaseObjectiveFunction):
         self._active_cells = values
 
     @property
-    def fixed_weights(self):
+    def cell_weights(self):
         """Regularization weights applied at cell centers"""
-        return self._fixed_weights
+        return self._cell_weights
 
-    @fixed_weights.setter
-    def fixed_weights(self, values: np.ndarray):
-        validate_array_type("fixed_weights", values, float)
-        validate_shape("fixed_weights", values, self._nC_residual)
-        self._fixed_weights = values
+    @cell_weights.setter
+    def cell_weights(self, values: np.ndarray):
+        validate_array_type("cell_weights", values, float)
+        validate_shape("cell_weights", values, self._nC_residual)
+        self._cell_weights = values
 
     @property
     def free_multiplier(self):
@@ -121,18 +92,19 @@ class BaseRegularization(BaseObjectiveFunction):
             self._free_weights = values
 
     @property
-    def regularization_mesh(self) -> RegularizationMesh:
-        """Regularization mesh"""
-        return self._regularization_mesh
+    def model(self) -> np.ndarray:
+        """Reference physical property model"""
+        return self._model
 
-    @regularization_mesh.setter
-    def regularization_mesh(self, mesh: RegularizationMesh):
-        if not isinstance(mesh, RegularizationMesh):
-            TypeError(
-                f"'regularization_mesh' must be of type {RegularizationMesh}. "
-                f"Value of type {type(mesh)} provided."
-            )
-        self._regularization_mesh = mesh
+    @model.setter
+    def model(self, values: np.ndarray | float):
+
+        if isinstance(values, float):
+            values = np.ones(self._nC_residual) * values
+
+        validate_array_type("model", values, float)
+        validate_shape("model", values, self._nC_residual)
+        self._model = values
 
     @property
     def mapping(self) -> maps.IdentityMap:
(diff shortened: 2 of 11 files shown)


domfournier:
Standby until #1021 is merged in.

SimPEG/regularization/base.py
```diff
@@ -151,10 +290,8 @@ def deriv(self, m):
...
         """
-
-        mD = self.mapping.deriv(self._delta_m(m))
-        r = self.W * (self.mapping * (self._delta_m(m)))
-        return mD.T * (self.W.T * r)
+        r = self.W * self.f_m(m)
+        return self.f_m_deriv(m).T * (self.W.T * r)
 
     @utils.timeIt
     def deriv2(self, m, v=None):
```
jcapriot, line 297: Just a thought, but I wonder if we could implement the regularizations as a `HasModel` class, so we can make use of the automated clearing of certain items if the model is updated, (i.e. here we can avoid creating the derivative mapping matrix at each call to `deriv2`.

jcapriot:
Dropped my adjustments in https://github.com/MiraGeoscience/simpeg/pull/22

jcapriot (commented):
Looks pretty good to me, just need to settle on a time period to remove the deprecated behavior, probably doesn't need to be very long, so maybe say by version `0.19.0`?

domfournier:
Really up to you guys. I would say one or two releases should be enough for people to migrate over?

lheagy:
I will take a look through this coming week. @JKutt has been doing some testing and comparisons between different generations of SimPEG regularization too, so we will take this for a test-drive!

SimPEG/regularization/__init__.py
```diff
@@ -1,21 +1,80 @@
...
+class SimpleSmoothDeriv(SmoothDeriv):
+    """Deprecated class, replaced by SmoothDeriv."""
+
+    def __init__(self, mesh=None, **kwargs):
+        super().__init__(mesh=mesh, **kwargs)
+
+
+@deprecate_class(removal_version="0.x.0", future_warn=True)
+class Simple(WeightedLeastSquares):
+    """Deprecated class, replaced by SmoothDeriv."""
```
lheagy, line 36:
```suggestion
    """Deprecated class, replaced by WeightedLeastSquares."""
```

SimPEG/regularization/base.py
```diff
@@ -28,98 +26,266 @@ class BaseRegularization(BaseObjectiveFunction):
...
+        "indActive",
+        "0.x.0",
+        error=False,
     )
-    regmesh = properties.Instance(
-        "regularization mesh", RegularizationMesh, required=True
+
+    @property
+    def model(self) -> np.ndarray:
+        """Reference physical property model"""
```
lheagy, line 103: question here: if I am understanding correctly, this is different from the `reference_model`? but possibly used in the Sparse? (still looking through everything :) 

If so a couple of suggestions / questions
- it would be worth updating the docstring to clarify the difference from the `reference_model` 
- and second, is this something that we want to expose to users? or should it be kept as a private variable under the hood?

SimPEG/regularization/base.py
```diff
@@ -130,11 +296,19 @@ def __call__(self, m):
 
             r(m) = \\frac{1}{2}
         """
-        r = self.W * (self.mapping * (self._delta_m(m)))
+        r = self.W * self.f_m(m)
```
lheagy, line 299: should we use this as an opportunity to remove the 1/2 here and in the data misfit to address #763? I don't think we should do it in this pr, but it seems like a complementary change to make in either the same release or a new release very quickly after this. To keep conversations scoped, we can continue discussion on #763 and keep this pr focussed on the reg, but reading the doc strings prompted this

SimPEG/regularization/base.py
```diff
@@ -180,12 +352,290 @@ def deriv2(self, m, v=None):
...
+    def _multiplier_pair(self):
+        return f"alpha_{self.orientation}"
+
+    def f_m(self, m):
+        """
+        Model gradient
+        """
+        dfm_dl = self.cell_gradient @ (self.mapping * self._delta_m(m))
+
+        if self.units == "radian":
```
lheagy, line 535: should we do a `.lower()` here? or I might have missed if this is unnecessary and done in the setter

SimPEG/regularization/regularization_mesh.py
```diff
@@ -297,138 +366,129 @@ def aveCC2Fz(self):
...
 
     @property
-    def cellDiffx(self):
+    def cell_gradient_x(self):
         """
-        cell centered difference in the x-direction
+        cell centered gradient in the x-direction
 
         :rtype: scipy.sparse.csr_matrix
         :return: differencing matrix for active cells in the x-direction
```
lheagy, line 393:
```suggestion
        :return: cell centered gradient matrix for active cells in the x-direction
```

SimPEG/regularization/base.py
domfournier: Replicated docstring - will fix.

SimPEG/regularization/base.py
domfournier: Yep, done.

SimPEG/regularization/base.py
domfournier: I am easy either way. Guess once you review the math with your group, you can confirm if it still adds up when those halves are removed, then we just pull them once and for all?

SimPEG/regularization/regularization_mesh.py
domfournier: All docstrings reviewed ... will need another check.

SimPEG/directives/pgi_directives.py
```diff
@@ -63,33 +60,26 @@ class PGI_UpdateParameters(InversionDirective):
...
-            if self.pgi_reg.mrefInSmooth and self.keep_ref_fixed_in_Smooth:
-                self.fixed_membership = np.c_[
-                    np.arange(len(self.pgi_reg.gmmref.cell_volumes)),
-                    self.pgi_reg.compute_quasi_geology_model(),
-                ]
+            # if self.pgi_reg.reference_model_in_smooth and self.keep_ref_fixed_in_Smooth:
+            #     self.fixed_membership = np.c_[
+            #         np.arange(len(self.pgi_reg.gmmref.cell_volumes)),
+            #         self.pgi_reg.compute_quasi_geology_model(),
+            #     ]
```
santisoler, line 82: If these lines are intended to be commented out, might be worth just remove them.

SimPEG/regularization/base.py
```diff
@@ -180,12 +352,290 @@ def deriv2(self, m, v=None):
...
-
-        mD = self.mapping.deriv(self._delta_m(m))
+        f_m_deriv = self.f_m_deriv(m)
         if v is None:
-            return mD.T * self.W.T * self.W * mD
+            return f_m_deriv.T * ((self.W.T * self.W) * f_m_deriv)
+
+        return f_m_deriv.T * (self.W.T * (self.W * (f_m_deriv * v)))
+
+    # TODO move these to general validation functions in code_utils
```
santisoler, line 361: This TODO line could be called to be completed and remove it? If not, would be better to remove it and open an issue for it.

SimPEG/regularization/base.py
```diff
@@ -180,12 +352,290 @@ def deriv2(self, m, v=None):
...
+        if array is not None and (
+            not isinstance(array, np.ndarray) or not array.dtype == dtype
+        ):
+            raise TypeError(
+                f"Values provided for '{attribute}' for {self} must by a"
+                f" {np.ndarray} of type {dtype}. "
+                f"Values of type {type(array)} provided."
+            )
+
+    # TODO move these to general validation functions in code_utils
```
santisoler, line 373: Some with this TODO line.

SimPEG/regularization/base.py
```diff
@@ -28,98 +26,266 @@ class BaseRegularization(BaseObjectiveFunction):
...
-        utils.setKwargs(self, **kwargs)
+    _model = None
+
+    def __init__(
+        self,
+        mesh: RegularizationMesh | BaseMesh,
+        active_cells: np.ndarray | None = None,
+        mapping: maps.IdentityMap | None = None,
+        reference_model: np.ndarray | None = None,
+        units: str | None = None,
```
santisoler, line 37: From what I understood the `units` arguments is new. Might worth adding a description for it on the docstring of the class. I think it currently only does something if `units="radians"`. If we don't add that to the docstring is difficult for users to get and idea of what strings can be passes without digging into the code.

santisoler (changes requested):
Nice work @domfournier! Lots of stuff here! I'm leaving some small comments, mainly about docstrings.

SimPEG/directives/pgi_directives.py
domfournier: Done

domfournier:
https://docs.google.com/document/d/1upczfyFa_phA0cYsrlX3IJ5h9OOCQmq6GVdcHc8D-Z0/edit?usp=sharing

lheagy:
Here are the points we chatted about in the 2022-09-07 meeting 

On naming:
- Propose we rename “Small” to “Smallness”. This is typically what we call the term in literature and when talking about it
- “SmoothDeriv” and “SmoothDervi2” are a bit misleading (SmoothDeriv implies that the derivative is smooth, which would in fact be the second order derivative). Suggestions are: 
    - “SmoothDeriv” → FirstOrderSmoothness or SmoothnessFirstOrder
    - “SmoothDeriv2” → SecondOrderSmoothness or SmoothnessSecondOrder 

- Suggestion for docstrings (and this can be a separate issue / pr if we want), it would be good to include the statement of the integrals that we are solving. 

- Related, but separate PR: Lets get rid of the ½! #763

lheagy (approved):
This looks great. Thanks for the big push @domfournier! It is so nice to have this cleaned up. 

From my perspective, this is good to go. The last suggestion I have is that we rename the datamisfit.L2DataMisfit to be `LeastSquares` for consistency, but I think this should be a separate, small PR as it has a different focus.

domfournier:
@jcapriot merged in. Couple more fixes for stuff encountered during testing.

Issue #985, opened 2021-04-09
QuadTree Regularization requires an IndActive


thast:
When using a 2D quadtree mesh, any regularization, except `Tikhonov`, requires to specify an explicit `indActive`.

Code to reproduce the error:

```
from discretize import TreeMesh
from discretize.utils import refine_tree_xyz
from SimPEG import regularization, maps
import numpy as np


dx, dy = 1. , 1.
nx,ny = 16, 16
hx = [(dx, nx)]
hy = [(dy, ny)]
quadtreemesh = TreeMesh([hx,hy], x0='CC')
pts = np.zeros(2).reshape(1,2)
mesh = refine_tree_xyz(
    quadtreemesh, pts, octree_levels=[3,3], method="radial", finalize=True
)


# just giving the mesh works for tikhonov
reg = regularization.Tikhonov(quadtreemesh)
reg(np.ones(quadtreemesh.nC))

# for any other regularization, you need an explicit indActive, whether or not you give a mapping already
reg = regularization.Simple(
    quadtreemesh,
    indActive=np.ones(quadtreemesh.nC,dtype='bool')
)
reg(np.ones(quadtreemesh.nC))

# Proof
reg = regularization.Simple(
    quadtreemesh,
    #indActive=np.ones(quadtreemesh.nC,dtype='bool')
)
reg(np.ones(quadtreemesh.nC))

# Proof 2
reg = regularization.Simple(
    quadtreemesh,
    mapping = maps.IdentityMap(quadtreemesh.nC),
    #indActive=np.ones(quadtreemesh.nC,dtype='bool')
)
reg(np.ones(quadtreemesh.nC))
```

jcapriot:
Regularizations should now support inversions without specifying and active cells list, in `0.18.0`

PR #1041, opened 2021-09-29
Patch for simple regularization and FDEM base simulation class.

dccowan:
This patch fixes two small bugs:

1) If no active cells model is assigned to the simulation, the projection from cells to active faces (x, y and z) breaks. The patch will simply recognize that *simulation.indActive = None* and temporarily create an active cells model in the functionality of *regularization.Simple*. See issue #985 

2) The base simulation class for frequency domain will store the factorizations of A-inverse at all frequencies even if you are doing a forward simulation. If the forward simulation has many frequencies, you will likely run out of RAM. By introducing a 'forward_only' keyword argument to the simulation, we can delete each factorization of A-inverse once it is used.

### DIFF
--- a/SimPEG/electromagnetics/frequency_domain/simulation.py
+++ b/SimPEG/electromagnetics/frequency_domain/simulation.py
@@ -65,6 +65,10 @@ class BaseFDEMSimulation(BaseEMSimulation):
 
     survey = properties.Instance("a survey object", Survey, required=True)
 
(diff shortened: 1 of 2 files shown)


domfournier:
I will wait for you to fix/pass the tests?
Ideally, brake this into 2 PRs as they are tackling different issues.

dccowan:
The changes made on this branch have been parsed into separate pull requests.

PR #1044, opened 2021-10-05
Add forward only option to 3D FDEM

dccowan:
Currently, the factorizations for A inverse are stored simultaneously for all frequencies. In the case you would like to run a forward simulation on many frequencies, you can run out of RAM. Similar to the option for potential fields, we introduce a 'forward_only' argument.


domfournier (approved):
Still holding fields for all frequencies, but a simple fix for now. 
Fine by me.

Issue #1043, opened 2021-10-05
scikit-learn 1.0 breaking change: _check_X removed.


ngodber:
With 1.0 it appears following function has been removed which is causing: 

 File "\miniconda3\envs\simpeg\lib\site-packages\SimPEG\utils\__init__.py", line 59, in <module>
    from .pgi_utils import (
  File "\miniconda3\envs\simpeg\lib\site-packages\SimPEG\utils\pgi_utils.py", line 22, in <module>
    from sklearn.mixture._base import _check_X, check_random_state, ConvergenceWarning
ImportError: cannot import name '_check_X' from 'sklearn.mixture._base' (C:\Users\njgod\miniconda3\envs\simpeg\lib\site-packages\sklearn\mixture\_base.py)

ngodber:
fixed in merge.

PR #1045, opened 2021-10-05
No active cells bug in regularization mesh

dccowan:
Currently, the projection from cells to active faces breaks in regularization_mesh.py when no active cells model is applied to the simulation. This is a basic fix for the bug.


domfournier (approved):
Well done. Just need to wait for the tests passing.

dccowan:
> Well done. Just need to wait for the tests passing.

I think the break is related to scikit-learn. It's issue #1043

domfournier:
Correct. We discussed this at the meeting. @thast will do a small PR this week to patch it.

PR #1047, opened 2021-10-05
Add truncation factor for sensitivity weights

dccowan:
Currently, sensitivity weights in directives are stabilized through the 'threshold' property. However, it is difficult to adjust this parameter if one would like to truncate the sensitivity weights. Here we introduce a truncation factor as a final step when creating the sensitivity weights.

### DIFF
--- a/SimPEG/directives/directives.py
+++ b/SimPEG/directives/directives.py
@@ -1606,6 +1606,7 @@ class UpdateSensitivityWeights(InversionDirective):
     everyIter = True
     threshold = 1e-12
     switch = True
+    truncation_factor = 1e-10
 
     def initialize(self):
 
@@ -1669,6 +1670,7 @@ class UpdateSensitivityWeights(InversionDirective):
 
             wr = wr ** 0.5
             wr /= wr.max()
+            wr[wr < self.truncation_factor] = self.truncation_factor
         else:
             wr += 1.0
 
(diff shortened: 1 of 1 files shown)


SimPEG/directives/directives.py
```diff
@@ -1669,6 +1670,7 @@ def getWr(self):
 
             wr = wr ** 0.5
             wr /= wr.max()
+            wr[wr < self.truncation_factor] = self.truncation_factor
```
domfournier, line 1673: Isn't the same thing as `threshold`, just moved outside of the sum? I would suggest just using the `threshold` parameter for less refactoring required.

domfournier: they are really doing the same thing, just a smoother corner if you add. 

I am cool with just moving out the threshold out of the sum, but not a fan of adding another term...

dccowan: I see your point, BUT you don't know if the largest value in diag(JtJ) is 1000 or 0.0001. For example, if I set the threshold to be 1e-6 and my largest diag(JtJ) value is 1e-8, my sensitivity weights are essentially just a constant model of 1e-6.

Whether or not we use threshold or truncation factor, it needs to be done AFTER normalizing and not before!!!

domfournier: Agreed.

jcapriot:
With the most recent change, the `threshold` value is always relative to the maximum value of the JtJdiag/mesh.cell_volume function. This should make the behavior much easier to understand and use in practice.

Closing this for now. but if you find the new behavior to also be in need of truncating it like this as well, we can discuss further.

PR #1052, opened 2021-10-12
New test for dc fields.

ngodber:
add test for apparent res dc fields. Testing tanh overflow.


ngodber:
test failures unrelated to test code.

ngodber:
@jcapriot this is ready for review

jcapriot:
@ngodber I know this is really, really late. But I'm going through a lot of the PR's and bringing them up. The issue here wasn't necessarily caused by having small thicknesses (ideally tanh should never overflow when given real values). There was an error when storing the wavenumber domain frequencies as complex numbers instead of real numbers. I've fixed it in #1190

PR #1054, opened 2021-11-03
Exorcise properties pf

dccowan:
This pull request is the start of removing the properties package from SimPEG. We are starting with straight forward properties of classes within potential fields.

The way we set the physical properties, mapping should be in a separate PR because it touches all simulation classes.


SimPEG/potential_fields/base.py
```diff
@@ -70,6 +78,76 @@ def __init__(self, mesh, **kwargs):
...
+        self._ind_active = input_array
+
+    @property
+    def actInd(self):
+        warnings.warn(
+            "The 'actInd' property has been deprecated. "
+            "Please use 'ind_active'. This will be removed in version 0.17.0 of SimPEG.",
+            FutureWarning,
+        )
+        return self._ind_active
```
domfournier, line 138: @jcapriot Do you think we could handle all of these properties -> @property with something like `code_utils.deprecate_property` , but would actually re-route to the setter of the new name? Feels like it would simplify the classes during transition.

SimPEG/potential_fields/base.py
```diff
@@ -14,38 +14,46 @@
...
 class BasePFSimulation(LinearSimulation):
-    actInd = properties.Array(
-        "Array of active cells (ground)", dtype=(bool, int), default=None
-    )
 
-    n_cpu = properties.Integer(
-        "Number of processors used for the forward simulation",
-        default=int(multiprocessing.cpu_count()),
-    )
+    def __init__(self, mesh, ind_active=None, store_sensitivities='ram', **kwargs):
```
domfournier, line 24: I would avoid adding explicit arguments here, and just let the setters handle do the checking/replace when processing the kwargs

PR #1056, opened 2021-11-08
Feature/tiled simulations

domfournier:
Latest changes from Mira R&D


SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -19,13 +21,166 @@
...
+
 Sim.Jvec = dask_Jvec
+
+
+def dask_Jtvec(self, m, v):
+    """
+        Compute adjoint sensitivity matrix (J^T) and vector (v) product.
+    """
+    self.model = m
+    if isinstance(self.Jmatrix, Future):
```
jcapriot, line 148: I haven't looked into it too much, but what _exactly_ is this line causing to happen?

If you want to just cause a wait until `self.Jmatrix` is no longer a `Future` wouldn't you use a 
`while isinstance(self.Jmatrix, Future)` loop condition?

PR #1057, opened 2021-11-16
Fix pgi tutorials accounting volume twice

thast:
The tutorial were still running smoothly but they actually now account volume twice. Final Alphas are still different than before because volume and normalization are done in a different order since 0.15.2.


jcapriot:
I noticed this too going through it myself!

PR #1059, opened 2021-11-25
Bringing in `simpegEM1D` into `SimPEG`

sgkang:
This is a pull request bringing in `simpegEM1D` (https://github.com/simpeg/simpegEM1D) to `SimPEG`
This was a team efforts of @dccowan  and @jcapriot, so thanks for both Joe and Devin!

## Major changes has made for this pull request:

1. Update the em1d codes to be compatible with the latest simulation class
2. Remove the em1d specific source, receiver, and survey classes, and update the em1d codes in a way that can be compatible with the existing em source, receiver, survey classes
3. With the Item 2 users can use the same source, receiver, and survey classes for em1d codes, which use pseudo-analytic solution, and other PDE-based codes

### Summary of technical details about the update (from @jcapriot)

Below are items left for further development, but we are planning to address them once we merge this into `main`


jcapriot:
@sgkang, Can you remove all of the binary files that got mixed into this branch? It looks like a bunch are in "examples/05-fdem/bookpurnong_inversion/"

examples/05-fdem/plot_stitched_1d_bookpurnong.py
```diff
@@ -0,0 +1,276 @@
+
+"""
+Inversion of RESOLVE data acquired at Bookpurnong, Austrailia
+=============================================================
+
+XXX
```
lheagy, line 6: This would be good to fill in (or remote at this point), but even a sentence or 2 that points to EM geosci for more background would be great.

tutorials/07-fdem/plot_inv_1_em1dfm.py
```diff
@@ -0,0 +1,350 @@
...
+#     2) Regularization: constraints placed on the recovered model and a priori information
+#     3) Optimization: the numerical approach used to solve the inverse problem
+#
+#
+
+# Define the data misfit. Here the data misfit is the L2 norm of the weighted
+# residual between the observed data and the data predicted for a given model.
+# The weighting is defined by the reciprocal of the uncertainties.
+dmis = data_misfit.L2DataMisfit(simulation=simulation, data=data_object)
+dmis.W = 1./uncertainties
```
lheagy, line 222: this line isn't necessary as long as the uncertainties are provided to the data object (which they are in this example). In general, I think we want to be encouraging folks to set this through the data misfit object rather than manually, so I would suggest removing this line

examples/05-fdem/plot_stitched_1d_bookpurnong.py
```diff
@@ -0,0 +1,276 @@
...
+dobs = DOBS.flatten()
+std = 0.1
+floor = 20.
+uncertainties = std*abs(dobs)+floor
+data_object = data.Data(survey, dobs=dobs, standard_deviation=uncertainties)
+
+hz = np.r_[thicknesses, thicknesses[-1]]
+mesh_reg = get_2d_mesh(n_sounding, hz)
+# Now we can create the regularization using the 2D mesh
+reg = LaterallyConstrained(mesh_reg, mapping=maps.IdentityMap(nP=mesh_reg.nC))
```
lheagy, line 205: since the mapping is an identity map, it isn't necessary. I would suggest removing it to keep this simpler (in general, there aren't many cases where the mapping is needed in the regularization)

tutorials/08-tdem/plot_fwd_1_em1dtm.py
```diff
@@ -0,0 +1,175 @@
...
+receiver_location = np.array([0., 0., 20.])
+receiver_orientation = "z"                    # "x", "y" or "z"
+times = np.logspace(-5, -2, 31)               # time channels (s)
+
+# Define receiver list. In our case, we have only a single receiver for each source.
+# When simulating the response for multiple component and/or field orientations,
+# the list consists of multiple receiver objects.
+receiver_list = []
+receiver_list.append(
+    tdem.receivers.PointMagneticFluxDensity(
```
lheagy, line 65: For this and all of the following tdem tutorials, I would suggest that we use the `PontMagneticFluxDensityTimeDerivative` receivers. It is much more common to work with db/dt data than it is to work with B data in a tdem experiment

sgkang: True. I did not realize that. @dccowan: what's your thought? If you agree, can you update the tutorial?

tutorials/07-fdem/plot_inv_1_em1dfm.py
sgkang: good point. removed the line.

setup.py
```diff
@@ -35,7 +35,7 @@
     packages=find_packages(exclude=["tests*", "examples*", "tutorials*"]),
     install_requires=[
         "numpy>=1.7",
-        "scipy>=1.0.0",
+        "scipy>=1.0.0,<=1.4.1",
```
jcapriot, line 38: I can't remember if we discussed why the upper bound on scipy was placed here

jcapriot:
Looks like there's still some more binary files left in `tests/em/em1d/em1dtm/`

setup.py
sgkang: my bad. I'll revert this. no need for this.

jcapriot:
@sgkang There's still tons of data files in this PR. Please go to the "Files Changed" tab here and filter out the .py files to see all of them. It looks like more got added in in the "-remove binary" commit. Mostly in the `examples/05-fdem/bookpurnong_inversion` directory and `tests/em/em1d/em1dfm/` directory.

jcapriot:
@sgkang After this round finishes, we'll check the diff coverage and ensure it's well tested.

tutorials/08-tdem/plot_fwd_1_em1dtm.py
lheagy: just pinging again on this @sgkang, @dccowan: I think for the TDEM examples, using the Time Derivative is much more common

SimPEG/electromagnetics/base_1d.py
```diff
@@ -0,0 +1,947 @@
...
+from ..simulation import BaseSimulation
+
+# from .time_domain.sources import MagDipole as t_MagDipole, CircularLoop as t_CircularLoop
+# from .frequency_domain.sources import MagDipole as f_MagDipole, CircularLoop as f_CircularLoop
+
+from .. import utils
+from .. import props
+from empymod.utils import check_hankel
+
+try:
```
jcapriot, line 17: Isn't multiprocessing an included fundamental python package? What environments have you seen that hasn't had multiprocessing? It's likely safe to just use `from multiprocessing import Pool` without any special guards.

jcapriot:
I was chatting with @sgkang about this implementation of the stitched inversion.
As written the `Simulation1DStitched` expects a 1D model for each sounding location identified by the user, keyed by a `i_sounding` argument attached to each source. Meaning sounding `i_snd` corresponds to a 1D vector `sigma[i_snd, :]` whose location is at the sounding location.

I would suggest a slightly different implementation that makes use of our discretize meshes to make this more generalized.
The physical property model would instead live on a 3D mesh, that is defined by a 2D x-y mesh, and a 1D vertical mesh. The physical property is then interpolated from the nodes of the x-y mesh, to the sounding locations. In my opinion this makes it easier to associate locations with their physical properties, as they are defined by the mesh instead of the source locations and an `i_sounding` parameter. It might also be easier for a user to interact with the stitched code as well.

If the user provided a `SimplexMesh` with a node at each sounding location, this would be essentially equivalent to what it is now, however our code could be more generalized to the x-y mesh accepting possibly `TensorMesh` or `TreeMesh` as well.

SimPEG/electromagnetics/base_1d.py
sgkang: sounds good. I'll consider `multiprocessing` as a base python package.

sgkang:
@jcapriot: I like the idea. A logical place to put the mesh seems `discretize`?

tutorials/08-tdem/plot_fwd_1_em1dtm_stitched_sk.py
```diff
@@ -0,0 +1,234 @@
+"""
```
sgkang, line 1: @jcapriot: rewrote the example. have a look and let me know if this makes sense to you.

jcapriot:
> @jcapriot: I like the idea. A logical place to put the mesh seems `discretize`?

I've got a working PR for `SimplexMesh` at https://github.com/simpeg/discretize/pull/263

lheagy:
Does #1087 cover the changes that would have been brought in here? (e.g. should we close this pr? )

PR #1065, opened 2022-01-28
remove disqus from the footer

lheagy:
remove disqus from the footer because all we got was spam! 🍔


lheagy:
💥 Thanks @prisae and @jcapriot!

PR #1070, opened 2022-02-18
Pde base

jcapriot:
This PR creates a common base for the PDE simulations, and dynamically builds mass matrices for a physical property.

The goal is to reduce copy-pasted code between functions, and enable easier testing of all the functions, and also a common structure for all of the calls to these functions/matrices.


jcapriot:
As part of this, I also took the chance to refactor the IP code to combine all of the duplicated code of the 2D and 3D cases, this also simplifies these class significantly.

Essentially all it takes for the IP class is to redefine the `sigmaDeriv` and `rhoDeriv` to be in terms of the `etaDeriv`, and then everything else builds naturally.

SimPEG/base/pde_simulation.py
```diff
@@ -0,0 +1,529 @@
...
+
+            MI_prop = getattr(self, f"Me{arg}I")
+            u = MI_prop @ (MI_prop @ -u)
+            M_prop_deriv = getattr(self, f"Me{arg}Deriv")
+            return M_prop_deriv(u, v, adjoint=adjoint)
+
+        setattr(cls, f"Me{arg}IDeriv", MeIDeriv_prop)
+
+        @property
+        def _clear_on_prop_update(self):
```
lheagy, line 330: this will clear all of them on every model update? We might want to look and see if the mappings are set before choosing what needs to be cleared (e.g. in an EM simulation / inversion, we don't need to be clearing the mu matrices, unless you are actually inverting for magnetic permeability)

SimPEG/base/pde_simulation.py
```diff
@@ -0,0 +1,529 @@
...
+            ]
+            return items
+
+        setattr(cls, f"_clear_on_{arg.lower()}_update", _clear_on_prop_update)
+        return cls
+
+    return decorator
+
+
+class BasePDESimulation(BaseSimulation):
```
lheagy, line 353: I am wondering if there is a way that would make sense to break this up to keep the name-spaces a bit smaller... e.g. the EM simulations don't need Mcc / Mn (unless we need to solve an initial condition for a grounded source in TDEM, but we might want to give that scenario a bit more thought anyways... it would be nice if we could make use of the DC simulation for this rather than what we are doing right now, which is re-implementing it... but this is a tangent :) 

Anyways, my main question here is: do we want to break this up into `Scalar` and `Vector` PDE simulations? or is that just a level of modularity that isn't helpful.

SimPEG/base/pde_simulation.py
jcapriot: This part isn't what is actually called when the property is updated, it just represents a list of items that should be cleared if you update the physical property. This functions as the `_clear_on_mu_update` and `_clear_on_sigma_update` properties on the current `BaseEMSimulation` class.

Down in the `BaseElectricalPDESimulation` you'll see where it performs the deletions, which executes in the same way as our current simulations.

lheagy: :+1: Thanks for clarifying!

SimPEG/base/pde_simulation.py
jcapriot: This begs the question for me, in should  *any* of the Mass Matrices be easily accessible to the user in the namespace of the `Simulation`. In my opinion they should start with an `_`, but as a start I was erring on consistency with the current method.

lheagy: Yeah, I was thinking about that too... In the earlier days of simpeg, I would have leaned towards saying they should be top-level accessible for the educational value, but really now, if you want to learn about building up systems of equations, that should be done at the level of discretize, not simpeg. So I would be supportive of moving towards `_` for the mass matrices

jcapriot: I could probably tackle that pretty quickly with this PR

PR #1074, opened 2022-04-03
Fix type (`solver_opts`, not `solver_spts`)

prisae:
I assume this was a typo @jcapriot


lheagy (approved):

jcapriot:
The entire SP module was never updated for the `Problem` -> `Simulation` transition.

prisae:
I just stumbled over it by looking for something entirely different, thought I'd report.

PR #1075, opened 2022-04-03
Make sklearn a soft dependency

prisae:
A start, potentially some edge casing missing, let's see what the tests do.


thast:
That looks good to me. Can you show me what message is displayed when sklearn is not installed? I just want to make sure it is explicit enough to not deter people from using it.

prisae:
It is probably not ideal currently. We could/should discuss this in the meeting on Wednesday. And we could expand the message in `discretize.utils.code_utils.requires` to state something like «Please install ...».

```
>>> import SimPEG
>>> SimPEG.utils.pgi_utils.GaussianMixtureWithPrior()
Missing dependencies: ['sklearn'].
Not running `__init__`.
```

We might want to adjust it. If you decorate a function, the message is better. But because these are classes, the message is not that useful (`Not running `__init__`).

prisae:
In `emg3d` I have this message, maybe we could do something similar:

```
"emg3d: This feature requires the missing soft dependencies ['sklearn']."
```

thast:
> In `emg3d` I have this message, maybe we could do something similar:
> 
> ```
> "emg3d: This feature requires the missing soft dependencies ['sklearn']."
> ```

I would strongly lean towards a very explicit message that includes instructions on how to install it (a bit like when discretize tells you to rebuild your local repository in place), especially for people less comfortable with Python that just want to access the tools. 
E.G. I am comfortable with Python but not with C++ so I really appreciate when a package gives me instructions in the failing messages on how to install a missing dependencies (like `brew install gcc`). Without it, I would otherwise move on or lose my time googling for the correct instructions (with sometimes multiple try and fail).

prisae:
Maybe. But then `requires` cannot be used, as it is very general. Any specific instruction is probably doomed to fail on some systems. E.g., `brew install gcc` will not be a good instruction on many platforms. Let's discuss it in the meeting!

A line like

> `pip install sklearn` or `conda install -c conda-forge sklearn`.

could be a good idea though.

We should definitely properly document it, I have not done that yet in this PR, but you will know best what is the right place to do so.

thast:
Edit:

> 
`pip install scikit-learn` or `conda install -c conda-forge scikit-learn`

prisae:
I am closing this, as it is heavily outdated (still adjusting, e.g., `setup.py`).

I will create a new PR if we decide to make `sklearn` (and potentially `pandas`) soft dependencies.

PR #1076, opened 2022-04-04
emg3d solver new

prisae:
**Note**: This Draft PR will be replaced by a new one, where most stuff goes into emg3d, and only smaller changes are required to SimPEG. I keep this around until that is done.
- emg3d PR: https://github.com/emsig/emg3d/pull/338
- SimPEG PR: TODO

---
# SimPEG(emg3d)

PR to implement the required steps on SimPEG-side to use emg3d as a solver for EM modelling and inversion within SimPEG.

Superseeding #1032

The old PR was quite old, had a few quirky merges (one from my personal GitHub), things that got completely disbanded again. This is a new try, rebased on the newest brain (and on the branch `soft-sklearn`, see Warning below).

## Warning

There seems to be an issue with `scikit-learn` (`sklearn`) and multiprocessing. For this PR to work you must uninstall `sklearn`, otherwise the computation will silently hang and wait forever (under Linux and OSX).

More info: https://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux

## Related branches
Note, there are a few branches that should be deleted once this is in:
- https://github.com/simpeg/simpeg/tree/emg3d-solver-new : this one.
- https://github.com/simpeg/simpeg/tree/emg3d-solver : the original one of #1032 .
- https://github.com/simpeg/simpeg/tree/emg3d-solver-wip : the cleaned-up version of `emg3d-solver`, all commits cherry-picked and everything from Seogi squashed and everything from Dieter squashed (7 commits total), but NOT rebased to main.
- https://github.com/simpeg/simpeg/tree/soft-sklearn

### DIFF
--- a/SimPEG/data.py
+++ b/SimPEG/data.py
@@ -8,7 +8,7 @@ from . import survey
 from .utils import mkvc
 from .utils.code_utils import deprecate_property
 
-__all__ = ["Data", "SyntheticData"]
+__all__ = ["Data", "ComplexData", "SyntheticData"]
 
 
 class UncertaintyArray(properties.Array):
@@ -21,7 +21,6 @@ class UncertaintyArray(properties.Array):
         elif isinstance(value, float):
             return value
         return super(properties.Array, self).validate(instance, value)
-
 
 class Data(properties.HasProperties):
     """
@@ -307,6 +306,8 @@ class Data(properties.HasProperties):
         error=True,
     )
 
+class ComplexData(Data):
+    dobs = None
 
 class SyntheticData(Data):
     """
--- a/SimPEG/data_misfit.py
+++ b/SimPEG/data_misfit.py
@@ -145,7 +145,7 @@ class L2DataMisfit(BaseDataMisfit):
         "__call__(m, f=None)"
 
         R = self.W * self.residual(m, f=f)
-        return 0.5 * np.vdot(R, R)
+        return 0.5 * np.vdot(R, R).real
 
     @timeIt
     def deriv(self, m, f=None):
--- a/SimPEG/electromagnetics/frequency_domain/__init__.py
+++ b/SimPEG/electromagnetics/frequency_domain/__init__.py
@@ -7,6 +7,7 @@ from .simulation import (
     Simulation3DCurrentDensity,
     Simulation3DMagneticField,
 )
+from .simulation_emg3d import Simulation3DEMG3D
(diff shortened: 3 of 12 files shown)


prisae:
Implemented a fix to avoid hanging of multiprocessing when sklearn is installed:
```
try:
    import sklearn
except ImportError:
    pass
else:
    import multiprocessing
    multiprocessing.set_start_method('forkserver')
```
See https://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux

thibaut-kobold:
Do you have a comparison in term of memory usage and computation time on CPU (or GPU) compared to using the current Pardiso solver?

prisae:
I don't think there is a fair comparison.

First, emg3d only works for EM, not any other method. So it is much less flexible.

Seconds, emg3d is an iterative (multigrid) solver, pardiso is a direct solver. With all its advantages and disadvantages. With emg3d, you never get `A` explicitly, you only get `x`. And you have to get it for every source position independently.

Having said that, the advantage of emg3d is that it scales linearly for both RAM and CPU with the number of cells. (See https://emg3d.emsig.xyz/en/stable/manual/info.html#cpu-ram).

So, I can run models with emg3d with 10 million cells on my laptop in a few minutes. With pardiso, 80,000 is roughly the limit an my machine, and it takes a while. However, and again, it heavily depends on the use case. For small models pardiso is most likely the better choice.

prisae:
I don't think there is a fair comparison.

First, emg3d only works for EM, not any other method. So it is much less flexible.

Seconds, emg3d is an iterative (multigrid) solver, pardiso is a direct solver. With all its advantages and disadvantages. With emg3d, you never get `A` explicitly, you only get `x`. And you have to get it for every source position independently.

Having said that, the advantage of emg3d is that it scales linearly for both RAM and CPU with the number of cells. (See https://emg3d.emsig.xyz/en/stable/manual/info.html#cpu-ram).

So, I can run models with emg3d with 10 million cells on my laptop in a few minutes. With pardiso, 80,000 is roughly the limit, and it takes a while. However, and again, it depends on the user case. For small models pardiso is most likely the better choice.

prisae:
Superseeded by https://github.com/simpeg/simpeg/pull/1515.

PR #1077, opened 2022-04-05
Exorcise properties preliminary

dccowan:
This pull request is a partial exorcism of the properties package and a new website build with latest API. More specifically:

- Build the website with pydata theme
- Exorcise properties from
    - SimPEG.utils
    - All survey, sources and receiver classes
- Build website with partial API (utils, survey, sources and receivers)


tests/docs/test_docs.py
```diff
@@ -7,15 +7,15 @@
...
+        dirname, file_name = os.path.split(os.path.abspath(__file__))
         return dirname.split(os.path.sep)[:-2] + ["docs"]
 
     def test_html(self):
         wd = os.getcwd()
         os.chdir(os.path.sep.join(self.path_to_docs))
 
         if platform.system() != "Windows":
-            response = subprocess.run(["make", "html-noplot"])
+            response = subprocess.run(["make", "html"])
```
jcapriot, line 18: This line, and the one below it, is why the doc test is being timed out on azure, please change it back. All of the examples and tutorials are tested separately from the doc build.

SimPEG/survey.py
```diff
@@ -1,204 +1,522 @@
...
-    _Ps = properties.Dictionary(
-        "dictonary for storing projections",
-    )
 
-    def __init__(self, locations=None, **kwargs):
-        super(BaseRx, self).__init__(**kwargs)
-        if locations is not None:
+    def __init__(self, locations=None, storeProjections=False, **kwargs):
+
+        if locations is None:
```
jcapriot, line 76: best to not have locations as a kwarg here and change it to an argument, since it's actually required. Python will allow the person interacting with the code to still use it as a kwarg.

SimPEG/survey.py
```diff
@@ -208,56 +526,142 @@ def getReceiverIndex(self, receiver):
...
+
+    @source_list.setter
+    def source_list(self, new_list):
+        if not isinstance(new_list, list):
+            new_list = [new_list]
+        
+        if any([isinstance(x, BaseSrc)==False for x in new_list]):
+            raise TypeError("Source list must be a list of SimPEG.survey.BaseSrc")
+
+        assert len(set(new_list)) == len(new_list), "The source_list must be unique. Cannot re-use sources"
```
jcapriot, line 596: Don't use assert here, return a ValueError if they're not the same length

jcapriot:
A few general comments here, Can you add tests specifically for the validator functions you've written that test that they both succeed and fail in expected ways?
Then,, probably go back through and use those functions in more of the codebase (e.g. in simpeg/survey.py).

dccowan:
One of the fdem convergence tests in muinverse is failing (not showing 2nd order). Convergence is fine when running locally. @jcapriot I feel like this could be related to the cell gradient stuff you were working with on discretize. Likely the local and pipelines versions of discretize are different?

image

jcapriot:
I've gone through with a ton of edits on this. 

Code
=========
For the code itself, I mostly just made use of the validation functions in every property that needed them.

I made some changes to some of the validator functions and created some more for common requested tasks.

Docs
=======
For the documentation a few formatting and style errors were common, and I want to address these here so they do not happen in the future. They are:

 - When referring specifically to a NumPy array in the documentation always use `numpy.ndarray` never `np.ndarray`
 - `array_like` is its own keyword for an object type, it is not `numpy.array_like` or `np.array_like`
 - When describing arguments that can only take a set of values, describe those values on the same line of the variable name, i.e. 
    ```
    Paramaters
    -----------
    thing : {"value1", "value2", "value3"}
         Description of thing.
    ```
    The first listed item will always be the default value.

    Do not do this:

    ```
    Paramaters
    -----------
    thing : str
        Description of thing. Must be {"value1", "value2", "value3"}.
    ```

Issue #1078, opened 2022-04-09
from __future__ imports


lheagy:
We have quite a number of `from __future__ import X` statements still around in SimPEG. Since we don't support python 2.7 (#731), these are no longer needed, correct?

mplough-kobold:
There are no Python 2.7 `__future__` statements left anymore (they were cleaned up in https://github.com/simpeg/simpeg/pull/1123), so I think this can be closed as completed.

The only remaining instances of `from __future__ import X` are:

```
SimPEG/regularization/base.py
1:from __future__ import annotations

SimPEG/regularization/pgi.py
1:from __future__ import annotations

SimPEG/regularization/sparse.py
1:from __future__ import annotations

SimPEG/regularization/vector.py
1:from __future__ import annotations

SimPEG/utils/code_utils.py
1:from __future__ import annotations
```

These should stay; according to Python's [`__future__` documentation](https://docs.python.org/3.12/library/__future__.html), annotations are not yet mandatory as of Python 3.12.

jcapriot:
You are correct!

PR #1082, opened 2022-05-10
Quadtree fix for potential fields simulation

nwilliams-kobold:
Prior to v.0.16 it was possible to use 2D quadtree meshes in potential fields simulations by defining `self.Zn` after the `simulation` call:

```
    simulation = gravity.simulation.Simulation3DIntegral(
        survey=survey,
        mesh=mesh,
        rhoMap=model_map,
        store_sensitivities="forward_only",
    )
    simulation.Zn = mesh_elevations
``` 

Fixes in commit [#123be88](https://github.com/simpeg/simpeg/commit/123be88b8ac0654236b8dae6af389f02db639b1c) appear to have broken that function. A fix was included by @dccowan in un-merged branch [mdru_pf_changes](https://github.com/simpeg/simpeg/compare/mdru_pf_changes) to support equivalent source modelling. This just surfaces that fix more explicitly.

I also took a moment to copy some of the comments from the magnetics implementation back to the gravity version.


tests/pf/test_grav_inversion_linear_quadtree.py
```diff
@@ -0,0 +1,136 @@
...
+        )
+
+    def test_grav_inverse(self):
+
+        # Run the inversion
+        mrec = self.inv.run(self.model)
+        residual = np.linalg.norm(mrec - self.model) / np.linalg.norm(self.model)
+        print(residual)
+
+        self.assertLess(residual, 0.7)
```
thibaut-kobold, line 127: Can you comment on the choice of 0.7 here?

tests/pf/test_grav_inversion_linear_quadtree.py
```diff
@@ -0,0 +1,136 @@
+from __future__ import print_function
```
thibaut-kobold, line 1: I do not think it is needed anymore (#1078 )

tests/pf/test_grav_inversion_linear_quadtree.py
nwilliams-kobold: I'll work on improved metrics

nwilliams-kobold: @thibaut-kobold In the new test I added magnetics, and I took your suggestion to revise the test criteria. There are now two tests each for gravity and magnetics:

1. Check the true model and recovered model are close enough, with values set based on the current result with a tolerance
2. Check that the normalized data misfit is less than the target misfit plus 10%.

Does this make the test more useful?

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -116,16 +116,20 @@ def evaluate_integral(self, receiver_location, components):
...
 
         # base cell dimensions
-        min_hx, min_hy, min_hz = (
-            self.mesh.hx.min(),
-            self.mesh.hy.min(),
-            self.mesh.hz.min(),
-        )
+        min_hx, min_hy = self.mesh.hx.min(), self.mesh.hy.min()
+        if self.mesh.hz is None:
+            # Allow for 2D quadtree represnetations by using a dummy cell height.
```
thibaut-kobold, line 125:
```suggestion
            # Allow for 2D quadtree representations by using a dummy cell height.
```

nwilliams-kobold:
@domfournier and @thibaut-kobold I added one more revision that standardizes how the tolerances are used and how the dummy cell size is set. I looked at implementing the tolerances as `kwargs`, but didn't know how and where to propagate that access up through the code to higher levels: ie, I didn't know where you want to set them? As parameters to `simulation`? Perhaps best to leave this PR as the low level function only?

thibaut-kobold:
for passing the tol1  and tol2 as arguments, it seems a multi-layer work to make it work properly (we need to modify all three `potential_fields/base.py` , `potential_fields/gravity/simulation.py` , `potential_fields/magnetic/simulation.py`).

For the scope of this PR, I am happy if it stays as is. Thanks for standardizing the tolerances.

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -254,37 +254,39 @@ def evaluate_integral(self, receiver_location, components):
...
 
         # base cell dimensions
-        min_hx, min_hy, min_hz = (
-            self.mesh.hx.min(),
-            self.mesh.hy.min(),
-            self.mesh.hz.min(),
-        )
+        min_hx, min_hy = self.mesh.h[0].min(), self.mesh.h[1].min()
+        if len(self.mesh.h) < 3:
+            # Allow for 2D quadtree represnetations by using a dummy cell height.
```
thibaut-kobold, line 268:
```suggestion
            # Allow for 2D quadtree representations by using a dummy cell height.
```

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -116,16 +116,20 @@ def evaluate_integral(self, receiver_location, components):
                     g_c = [g_cx g_cy g_cz]
 
         """
-        tol1 = 1e-4
-        tol2 = 1e-10
+        tol1 = 1e-4  # Tolerance 1 for numerical stability over nodes and edges
+        tol2 = 1e-10  # Tolerance 2 for numerical stability over nodes and edges
```
thibaut-kobold, line 120:
```suggestion
       
```
tol2 is not even used for gravity. We should just delete this one (I am ok keeping the other one named `tol1` for consistency with `magnetics`)

domfournier:
It's up to you, but technically you odn't need to change the simulation now, just move the tols to the methods inputs, and docs string them.

tests/pf/test_pf_quadtree_inversion_linear.py
```diff
@@ -0,0 +1,242 @@
...
+        dpred = self.grav_sim.dpred(self.grav_model)
+
+        # Check models match well enough (allowing for random noise)
+        model_residual = np.linalg.norm(mrec - self.grav_model) / np.linalg.norm(
+            self.grav_model
+        )
+        self.assertAlmostEqual(model_residual, 0.57, delta=0.1)
+
+        # Check data converged to less than 10% of target misfit
+        data_rms = (
```
thibaut-kobold, line 205: we could have used the `dmis` object here for simplicity if it was saved as `self.dmis`:
`self.dmis(mrec) == 2. * data_rms` here

 but I am ok if left as is.

tests/pf/test_pf_quadtree_inversion_linear.py
```diff
@@ -0,0 +1,242 @@
...
+        dpred = self.mag_sim.dpred(self.mag_model)
+
+        # Check models match well enough (allowing for random noise)
+        model_residual = np.linalg.norm(mrec - self.mag_model) / np.linalg.norm(
+            self.mag_model
+        )
+        self.assertAlmostEqual(model_residual, 0.03, delta=0.05)
+
+        # Check data converged to less than 10% of target misfit
+        data_rms = (
```
thibaut-kobold, line 226: same suggestion as for grav.

nwilliams-kobold:
Thanks @domfournier I'm not comfortable with adding them to the docstrings either, since I don't understand what they do? Why are they needed? And how does one know how large they need to be?

thibaut-kobold (approved):
Great work, Nick, and thanks for adding tests and consistency.
I left a few final suggestions but none are big enough to withhold approval.

tests/pf/test_pf_quadtree_inversion_linear.py
nwilliams-kobold: Thanks for the tip @thibaut-kobold!

tests/pf/test_pf_quadtree_inversion_linear.py
nwilliams-kobold: Done, as suggested.

domfournier:
See https://github.com/nwilliams-kobold/simpeg/pull/1

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -254,37 +254,39 @@ def evaluate_integral(self, receiver_location, components):
         Tz = [Tzx Tzy Tzz]
         """
         # TODO: This should probably be converted to C
-        tol1 = 1e-10  # Tolerance 1 for numerical stability over nodes and edges
-        tol2 = 1e-4  # Tolerance 2 for numerical stability over nodes and edges
+        tol1 = 1e-4  # Tolerance 1 for numerical stability over nodes and edges
+        tol2 = 1e-10  # Tolerance 2 for numerical stability over nodes and edges
```
nwilliams-kobold, line 258: @dccowan We'd love your input on what these two `tol` parameters really mean, and why there are two with different values? I have this update, and @domfournier has another (https://github.com/nwilliams-kobold/simpeg/pull/1) that simplifies them to just one? Are we on the right track?

domfournier: @nwilliams-kobold I suggest merging https://github.com/nwilliams-kobold/simpeg/pull/1, and let unitest run. We can modify it back here if peeps have issue with it.

domfournier: Cool, all systems green on this end. We will give a chance to @dccowan to chime in, but it looks fine by me.

PR #1085, opened 2022-05-18
Add separate pf quadtree equivalent source simulation

dccowan:
As the name says. Make a separate simulation class for the equivalent source layer. This functionality is much more up front as opposed to modifying things after instantiating a Simulation3DIntegral

### DIFF
--- a/SimPEG/potential_fields/base.py
+++ b/SimPEG/potential_fields/base.py
@@ -191,6 +191,9 @@ class BaseEquivalentSourceLayerSimulation(BasePFSimulation):
     """
 
     def __init__(self, mesh, cell_z_top, cell_z_bottom, **kwargs):
+        
+        if mesh.dim != 2:
+            raise AttributeError("Mesh to equivalent source layer must be 2D.")
 
         super().__init__(mesh, **kwargs)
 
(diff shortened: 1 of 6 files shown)


SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -311,6 +311,22 @@ def evaluate_integral(self, receiver_location, components, tolerance=1e-4):
 
         return np.vstack([rows[component] for component in components])
 
+class SimulationEquivalentSourceLayer(Simulation3DIntegral):
+    """
+    Equivalent source layer simulations
+    """
+
+    def __init__(self, mesh2D, cell_z_top, cell_z_bottom, **kwargs):
+
```
domfournier, line 320: Should probably add some docstrings on those parameters `cell_z_top` and `cell_z_bottom`.

Can likely accept `numpy.ndarray` and `float` for both?

domfournier: Can we actually make those property of the class, and do the typing and asserts in the setter?

domfournier (changes requested):
Love it.

Trying to reduce code duplication, may I suggest that you move the core class with its __init__ to the `potential_fields.base` 

Something like:

```
class BaseEquivalentSourceLayer(LinearSimulation):
    """
    Equivalent source layer simulations
    """

    def __init__(self, mesh2D, cell_z_top, cell_z_bottom, **kwargs):
        ...
```

And replace each of gravity and magnetics EQS to a simple inheritance pattern like this
```
class SimulationEquivalentSourceLayer(BaseEquivalentSourceLayer, Simulation3DIntegral):
    """
    Magnetic equivalent source layer simulations
    """
```
with nothing else. This should do the trick.

See comments below regarding the inputs.

domfournier:
Consolidating the classes might help with codecov

SimPEG/potential_fields/gravity/simulation.py
dccowan: The properties class doesn't play nice when I tried to override with standard @property functionality. For now, I like them being required arguments.
I did make it so that you could enter a numpy array or float.

dccowan: Docstrings were also added

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -312,6 +312,26 @@ def evaluate_integral(self, receiver_location, components, tolerance=1e-4):
...
+        A 2D tensor or tree mesh defining discretization along the x and y directions
+    cell_z_top : numpy.ndarray or float
+        Define the elevations for the top face of all cells in the layer
+    cell_z_bottom : numpy.ndarray or float
+        Define the elevations for the bottom face of all cells in the layer
+    """
+
+    def __init__(self, mesh, cell_z_top, cell_z_bottom, **kwargs):
+
+        BaseEquivalentSourceLayerSimulation.__init__(self, mesh, cell_z_top, cell_z_bottom)
```
jcapriot, line 331: These two individual calls to the parent classes should both be done with a single call to `super().__init__`, just pass everything by keyword and it should all end up in the correct place

jcapriot: actually... with the way this is written, I bet you could just delete the `__init__` functions on this class and it would still work just fine, as the `BaseEquivalentSourceLayerSimulation` class's `__init__` should be at the top of the inheritance.
(Then that classes call to `super` would go up a level)

domfournier: I concur

dccowan: I removed the redundant __init__ functions as suggested.

domfournier (changes requested):
Thanks for this. Last minor edits below

SimPEG/potential_fields/base.py
```diff
@@ -176,6 +176,57 @@ def n_cpu(self, other):
...
+    #     return self._mesh
+
+    # @mesh.setter
+    # def mesh(self, new_mesh):
+
+    #     if new_mesh.dim != 2:
+    #         raise AttributeError("Mesh to equivalent source layer must be 2D.")
+
+    #     self._mesh = new_mesh
+
```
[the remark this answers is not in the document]
dccowan: Done

domfournier (approved):
image

PR #1093, opened 2022-08-09
TDEM line current h-field update 

lheagy:
Bug fixes in the MMR solution for the TDEM h-field problem with a line source. 
- note that `e = - \grad \phi` (hence the sign update)
- the volume term needs to be accounted for in the divergence

Todo:
- [x] add a test that checks consistency with the j formulation


lheagy:
👋 @sgkang and @jcapriot: if either / both of you have a chance to look at this, I would appreciate your review! Just a couple small updates in the TDEM line-current source when we are solving for `h`

jcapriot (approved):
Looks good for the fix. Although those solver lines within the simulation are a little rough, those are a problem for another time.

PR #1094, opened 2022-08-13
fix bracket error

dannis999:


lheagy (approved):
Thanks for catching this @dannis999!

lheagy:
@jcapriot: is it okay with you if I merge this in?

PR #1095, opened 2022-08-23
arbitrary orientation for tdem rx

thibaut-kobold:
Allow for arbitrary orientation of receivers for TDEM surveys; this also allow for TMI data for TDEM such as the HeliSAM survey (see @yangdikun's [paper on the Lalor deposit](https://gif.eos.ubc.ca/sites/default/files/sdevriese/files/eg15070.pdf), equation 4 for the formulation of the data)


jcapriot:
It looks like you’ve removed all type checking for the orientation property, please add this as a property with proper type checking. Ideally the setter would check for `x`, `y`, or `z`, set it to the correct unit vector in those cases, else you would accept any length `dim` numpy array, normalize it so it’s unitary, then stash it on `self`.

PR #1096, opened 2022-08-24
bug fix for inversion with only a single parameter

lheagy:
only squeeze v if it has more than one dimension. otherwise this can turn a vector of a single element into a zero-dimensional array (a scalar), which then doesn't interact well in matrix multiplications 

image


jcapriot:
This does look good, maybe it would be worth adding a test for this sweet https://github.com/simpeg/simpeg/blob/main/tests/base/test_mass_matrices.py?

PR #1097, opened 2022-08-24
Update geoapps 0.8.0

domfournier:


SimPEG/dask/electromagnetics/frequency_domain/simulation.py
```diff
@@ -140,7 +140,7 @@ def eval_store_block(A, freq, df_duT, df_dmT, u_src, src, row_count):
 
             for rx in src.receiver_list:
                 v = np.eye(rx.nD, dtype=float)
-                n_blocs = np.ceil(rx.nD / row_chunks)
+                n_blocs = np.ceil(2 * rx.nD / row_chunks)
```
JKutt, line 143: @domfournier, likely silly question, but the 2 is because of real and imaginary?

JKutt (approved):
That is some darn fine work there @domfournier !

PR #1100, opened 2022-09-02
add exception for B-field computation with EB formulation when the waveform starts at 0.

### LINKED ISSUE #1098: v0.17: LineCurrent with waveform does not work anymore with B-field formulation
I think I am encountering an issue with [this PR ](https://github.com/simpeg/simpeg/pull/1093)that got brought in in v 0.17 of SimPEG.
- in v0.16, while LineCurrent could not work with Simulation3DMagneticFluxDensity and a step-off, it was working if an explicit waveform was providing (like a trapezoid…). This configuration is very useful for any ground EM survey with B field.
- In v0.17, [an notImplementedError is raised](https://github.com/simpeg/simpeg/blob/811209d8663a832e545af4e7f17b0a01d2a14362/SimPEG/electromagnetics/time_domain/sources.py#L1109) when using LineCurrent  with the EB formulation, whether or not a waveform was given! 
- **Thus, SimPEG is less capable than before.**

thibaut-kobold:
resolve #1098

### DIFF
--- a/SimPEG/electromagnetics/time_domain/sources.py
+++ b/SimPEG/electromagnetics/time_domain/sources.py
@@ -1,15 +1,15 @@
-import numpy as np
-from scipy.constants import mu_0
-import properties
 import warnings
-from ...utils.code_utils import deprecate_property
 
-from geoana.em.static import MagneticDipoleWholeSpace, CircularLoopWholeSpace
+import numpy as np
+import properties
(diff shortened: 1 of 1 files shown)


jcapriot:
You will also need to update the logic in a few other places as well to match, `hInitial`, and both `hInitialDeriv` and `bInitialDeriv`

PR #1103, opened 2022-09-15
Amplitude regularization

domfournier:
New implementation for MVI sparse inversion to keep the problem (mostly) linear.
 - Implementation of `VectorAmplitude` regularization

$\phi_s = \|| \mathbf{W} \mathbf{P} \mathbf{a}(m - m_{ref}) \||_2^2$

 - Implementation of `Group` map as extension of `Wire` with derivatives of wires returned as a list.

Solution with amplitude regularizer

Sparse_MVI


domfournier:
@orerocks All your dreams will come true.

micmitch:
I get an error when trying to run unbounded examples with `upper_bound=None` and  `lower_bound=None` in the optimization. I'm not sure if this expected behavior or not. If I explicitly set the bounds to (-np.inf, np.inf) I runs as expected. 

I have several suggestions for changes in the associated example script, but I don't think that I have permission to create a pull request on `MiraGeoscience:amplitude_regularization`. Let me if you would like my input on this and where/how I can share these modifications.

jcapriot:
@domfournier What is the current state of this PR? Are we in a place to move forward with it once the style checks are passed?

domfournier:
It's a bit stale, but I would like to meet with John W before merging or closing. If his approach works better, then there we would no point in bringing this in.
 @lheagy Can we schedule another session soon?

jcapriot:
Hey Dom, I think we're both finding some similar choke points in my cross-reference vector regularization (just opened it at #1214), and this one.
* Had to deal with the UpdateSensitivityWeights directive to handle vector valued models.
   * Mine unconditionally reshapes the output of `reg.mapping * jtj_diag` to be a `(n_cells, -1)` array, so then we can always divide by the mesh volume as `vol[:, None]`
* Needed a way to verify the model weight sizes.
  *  Mine allowed for `(n_cells,)`, (`n_cells*dim,)`,  and `(n_cells, dim)` shapes. Then I conditionally handled them in the `_W` function.

I think we can split a few common things into a `BaseVectorRegularizer` class, and maybe have them all live inside a `SimPEG.regularization.vector.py` module?

Some design thoughts, would it be easier if we didn't have to specify a wiremap for the three components in this regularizer? (The simulations already except models like this without mapping them to the different components, so the model should already be in a `n_cells * dim` array), then it can get reshaped within the calls?

One detail though, why are you passing the model difference (`dm`) to the `mapping.deriv` in `f_m_deriv`? Pretty sure it should just be the `m`, unless you had a specific reason (this error would only have shown up in a non-linear mapping).

domfournier:
@jcapriot Yea I saw your modifications to the UpdateSensWeights. I like yours better. 

Looks like we may be able to share some components indeed. Not sure how you want to proceed?

The wire mapping is kind of nice to keep the operations close to what the linear algebra looks like. Also, for joint problems, we will still need Wires to get the sub-vector from a longer array (e.g. [vector mag, density]. And since Wires doesn't have derivatives, I thought we could just implement it in here.

Good point on `dm` - typo. Won't change anything anyway since the deriv of a wire just returns the projection.

 I will merge main in.

domfournier:
@jcapriot Ready for another review.

PR #1104, opened 2022-09-16
Update SimPEG.Report()

prisae:
- Update the optional package list to include all optional imports.
- Move `requires` ahead of `memProfileWrapper`, to decorate latter with `@requires('memory_profiler')`.


prisae:
We could add the optional imports `omf`, `pyvista`, and `mpl_toolkits` from `discretize`. Opinions?

SimPEG/utils/code_utils.py
```diff
@@ -325,16 +326,32 @@ def __init__(self, add_pckg=None, ncol=3, text_width=80, sort=False):
         core = [
             "SimPEG",
             "discretize",
-            "pymatsolver",
+            "pymatsolver",  # TODO: Will be replaced by pydiso
```
jcapriot, line 329: It won't be replaced by pydiso. pydiso is specifically a wrapper for intel's pardiso libraries. pymatsolver is another level above that provide abstract operations with solver objects.

SimPEG/utils/code_utils.py
```diff
@@ -325,16 +326,32 @@ def __init__(self, add_pckg=None, ncol=3, text_width=80, sort=False):
...
             "vectormath",
-            "properties",
+            "properties",   # TODO: Will be removed eventually
```
jcapriot, line 331: properties (and vectormath) are no longer required by SimPEG.

SimPEG/utils/code_utils.py
prisae: Oh, thanks for the clarification. Should pydiso be another optional or mandatory package then?

jcapriot:
> Oh, thanks for the clarification. Should pydiso be another optional or mandatory package then?

This does beg a question:
Should the reported requirements also include requirements of the required packages?

prisae:
> Should the reported requirements also include requirements of the required packages?

Good question. Not sure, something to be discussed. I am not sure if the list I put together is any good. I think it should contain those packages that are most important for issues/failures/changes/dependencies/whatever-is-needed-in-debugging.

prisae:
Simply the most important ones. Because anything else could be asked for specifically, e.g.,
```
SimPEG.Report(['vtk', 'ipywidgets'])
```
if needed.

PR #1107, opened 2022-09-26
Improve how path to sensitivity files are built

santisoler:
Create the paths to where the sensitivity arrays will be stored by the
simulations using `os.path.join` instead of concatenating strings. This makes it
safer and removes the need to remember adding a trailing slash on paths to
directories.


jcapriot:
Thanks for doing this, I was running into this issue on my own re-do. Eventually I’d love to get these to accept [pathlib.PurePath](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath) objects instead.

santisoler:
Glad to contribute! I was intending to use `pathlib.Path` to define those paths, but I'm afraid that could introduce some breaking changes, so I decided to be more conservative and just use `os.path.join()`.

BTW, is there any reason why you would choose `pathlib.PurePath` instead of `pathlib.Path` for SimPEG?

jcapriot:
Nothing particularly different, just that `PurePath` is the parent class of the `Path` objects.

jcapriot:
@santisoler If you can get main merged into here with the removal of `properties` that would be helpful

santisoler:
Thanks Joe for making the default also be created with `os.path.join`. Nevertheless, it's usually advised not to define default values for parameters through function calls. Check `B008` in [bugbear's warnings](https://github.com/PyCQA/flake8-bugbear#list-of-warnings). This is another reason why would be nice to have predefined `Path` objects instead of using `os.path.join`.

jcapriot:
Yes, I realize this issue. It should probably be set with a None argument then set with the current method.

PR #1108, opened 2022-09-28
Allow `Data` to take `dobs` as a dictionary for multicomponents data

santisoler:
When working with potential field data we might want to perform multicomponent
inversions. In those cases we need to create a single 1D array that host every
observed data from each one of the components. This array should interleave the
values of each component on each receiver in the same order as the components
are laid out in the survey object. Now we would be able to pass the `dobs`
argument as a dictionary whose keys are the components and whose values are
their corresponding observed data values.

Todo:

- [ ] Allow the `relative_error`, `noise_floor` and `standard_deviation` to be
  dictionaries
- [ ] Update the docs (should we still use `properties` here?)

### DIFF
--- a/SimPEG/data.py
+++ b/SimPEG/data.py
@@ -380,7 +380,7 @@ def _check_data_sizes(observed_data):
     Check that all elements inside observed data dictionary have the same shape
     """
     # Check if every array in observed_data has the same size
-    sizes = [d.size for d in observed_data.values()]
+    sizes = np.array([d.size for d in observed_data.values()])
     if not (sizes[0] == sizes).all():
         raise ValueError(
             "All elements in the data dictionary should have the same size"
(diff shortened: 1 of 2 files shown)


jcapriot:
While I appreciate the spirit of this usability enhancement. It is not something that necessarily extends to all of `simpeg` and really only works with the `potential_fields` module, and even then only when there is only a single source and receiver object in the survey.

It is definitely unclear in the documentation what order simpeg expects and outputs data in the 1D array (or how to work with a `Data` object in general, which I usually recommend accessing like an array with indices `data[src1, rx1]` ), but essentially it is ordered as follows:
```
i_data = 0
for source in survey.source_list:
    for receiver in source.receiver_list:
        for location in receiver.locations:
            for component in receiver.components:
                i_data += 1
```
A bit of common confusion in this general schema is that most receivers in SimPEG only support a single `component` at a time (only the potential fields receivers support multiple components), and you must therefore use multiple receivers to represent multiple components (think the real and imaginary components of a NSEM receiver).

For dictionary input I would instead support an instantiation scheme that allows you to pass a dictionary of data, whose keys are the source and receiver for that data, i.e.

```
data_dict = {
(src1, rx1): d_1,
(src2, rx3): d_2,
(src1, rx2): d_3,
...
}
```
which would then read in the data and get it into the correct ordering.

PR #1113, opened 2022-10-20
Pf update

jcapriot:
This PR overhalls the internals of the potential fields integral simulations to make use of kernels defined in `geoana`. There should be no visible change on the end user.

Things it does
* Uses the compiled kernels from geoana.
* Only calculates the kernel functions at unique nodes.
* Adds multiprocessing back as the default way of computing the sensitivity matrices


SimPEG/dask/potential_fields/base.py
```diff
@@ -101,15 +64,14 @@ def dask_linear_operator(self):
...
         with ProgressBar():
             print("Forward calculation: ")
-            pred = (stack @ self.model).compute()
-        return pred
+            kernel = stack.compute()
     else:
         with ProgressBar():
             print("Computing sensitivities to local ram")
-            kernel = array.asarray(stack.compute())
+            kernel = stack.persist()
```
domfournier, line 74: Nice

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -118,200 +123,72 @@ def evaluate_integral(self, receiver_location, components, tolerance=1e-4):
...
+
+        node_evals = {}
+        if "gx" in components:
+            node_evals["gx"] = prism_fz(dy, dz, dx)
+        if "gy" in components:
+            node_evals["gy"] = prism_fz(dz, dx, dy)
+        if "gz" in components:
+            node_evals["gz"] = prism_fz(dx, dy, dz)
+        if "gxy" in components:
+            node_evals["gxy"] = prism_fzx(dy, dz, dx)
```
domfournier, line 139: I am sure you have a smart reason - but why are we getting the `xy` component from the `fzx` call?

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -118,200 +123,72 @@ def evaluate_integral(self, receiver_location, components, tolerance=1e-4):
...
+            node_evals["gyy"] = prism_fzz(dz, dx, dy)
+            if "guv" in components:
+                node_evals["guv"] = (node_evals["gyy"] - node_evals["gxx"]) * 0.5
+                # (NN - EE) / 2
+        inside_adjust = False
         if "gzz" in components:
-            rows["gzz"] = -gxx - gyy
-
-        if "guv" in components:
-            rows["guv"] = -0.5 * (gxx - gyy)
```
domfournier, line 306: Aren't we losing `guv` here? 

image

SimPEG/potential_fields/gravity/simulation.py
jcapriot: you can cycle the inputs to the function to get different components

domfournier: very cool

SimPEG/potential_fields/gravity/simulation.py
jcapriot: It is not lost! If you look, there are two conditions to calculate `gxx`, whether `gxx` or `guv` is in components, and the same for `gyy`, then it is finally computed at during the `gyy` check.

domfournier: Yup, my bad.

jcapriot:
Some timing examples on my laptop (4 processors):
on a 32 x 32 x 32 mesh (32768 cells) with 101 * 101 = 10201 observation locations.

Before and after timings for creations of a sensitivity matrix:
Gravity: 
- single "gz" component:
   ~200 seconds -> ~13 seconds
- All 6 components of GG: "gxx, gxy, gxz, gyy, gyz, gzz"
   ~430 seconds -> ~43 seconds

Magnetics:
- "tmi"
  ~ 500 seconds -> ~ 30 seconds

thibaut-kobold:
> Some timing examples on my laptop (4 processors): on a 32 x 32 x 32 mesh (32768 cells) with 101 * 101 = 10201 observation locations.
> 
> Before and after timings for creations of a sensitivity matrix: Gravity:
> 
> * single "gz" component:
>   ~200 seconds -> ~13 seconds
> * All 6 components of GG: "gxx, gxy, gxz, gyy, gyz, gzz"
>   ~430 seconds -> ~43 seconds
> 
> Magnetics:
> 
> * "tmi"
>   ~ 500 seconds -> ~ 30 seconds

Very cool, congrats!
Quick question: looking at the geoana kernels, I see [there is a pure-python version, and a C (or C++) version of the kernels](https://github.com/simpeg/geoana/blob/3d3e021a859670a675227aa274db6d158fbafe4d/geoana/kernels/potential_field_prism.py#L289) and that it tries to import the C version first and revert to pure python if it fails. I imagine the speed up is achieved by the C version My question: is there a way to display a warning if the pure-python version is imported instead of the accelerated one?

jcapriot:
The speedup between the two different versions is not actually that significant. If you're installing geoana through conda-forge it will come compiled. I.e. the c one is maybe 80% the speed of the numpy one, if i remember correctly. Most of the timing differences here come from the multiprocessing addition, and the non-repeated calls of the functions at nodes.

PR #1118, opened 2022-10-25
TDEM exponential waveform

thibaut-kobold:
This PR adds a flexible exponential growth followed by a linear ramp waveform: `ExponentialWaveform`
- parameters: `start_time`, `ramp_on_tau`, `peak_time`, `off_time`
- `eval`, `eval_deriv`, `time_nodes` are implemented
- the gallery example is updated
image

### DIFF
--- a/SimPEG/electromagnetics/time_domain/sources.py
+++ b/SimPEG/electromagnetics/time_domain/sources.py
@@ -4,19 +4,18 @@ import numpy as np
 from geoana.em.static import CircularLoopWholeSpace, MagneticDipoleWholeSpace
 from scipy.constants import mu_0
 
+from ...utils import Zero, sdiag
 from ...utils.code_utils import (
     deprecate_property,
-    validate_float,
-    validate_string,
-    validate_type,
-    validate_ndarray_with_shape,
-    validate_location_property,
     validate_callable,
     validate_direction,
+    validate_float,
     validate_integer,
+    validate_location_property,
+    validate_ndarray_with_shape,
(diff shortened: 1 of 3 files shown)


SimPEG/electromagnetics/time_domain/sources.py
```diff
@@ -878,6 +877,139 @@ def time_nodes(self):
...
+            max_val=self.off_time,
+            inclusive_min=False,
+        )
+        self._peak_time = value
+
+    @property
+    def ramp_on_tau(self):
+        """Ramp on rate
+
+        Parameter controlling how quickly the waveform ramps on (default is 3)
```
jcapriot, line 964: I don't think the default is 3 anymore, best just to remove reference to it here and leave it as described in the constructor.

SimPEG/electromagnetics/time_domain/sources.py
```diff
@@ -4,19 +4,18 @@
...
     validate_direction,
+    validate_float,
     validate_integer,
+    validate_location_property,
+    validate_ndarray_with_shape,
+    validate_string,
+    validate_type,
 )
-
-from ...utils import sdiag, Zero
```
jcapriot, line 19: Do you have something in your IDE that automatically reorders the imports alphabetically?

thibaut-kobold: I do, yes. I think it works alphabetically but also type of import (whether a package, subplackage, or local import)

PR #1122, opened 2022-10-26
fix unnecessary call to maplist on the reference model

thibaut-kobold:
`maplist` is applied wrongfully to mref [here](https://github.com/simpeg/simpeg/blob/1441840f78e43365947bc1dd620f8cfc822c4ae7/SimPEG/regularization/pgi.py#L132-L138). The underlying formula is `||mapping(model) - reference||^2`. This PR fixes that issue
More in-depth explanation:
- mref lives in the regularized space, so if we regularize amplitude, mref is an amplitude model. [Here](https://github.com/simpeg/simpeg/blob/1441840f78e43365947bc1dd620f8cfc822c4ae7/SimPEG/regularization/pgi.py#L207) we see how it is constructed from the means of the GMM (means would be amplitude in your case). So we do not need a mapping to convert it from cartesian to amplitude.
- It seems to be the only occurence. Other places where it could have happened are clear, like [Here](https://github.com/simpeg/simpeg/blob/1441840f78e43365947bc1dd620f8cfc822c4ae7/SimPEG/regularization/pgi.py#L222) or [here](https://github.com/simpeg/simpeg/blob/1441840f78e43365947bc1dd620f8cfc822c4ae7/SimPEG/regularization/pgi.py#L299)

Additional contributions
+import clean up in PGI example, directive and tutorial
+spelling errors fixes in error message


jcapriot:
How does this work if the reference model and the true model have different shapes? It looks like both of them use the same `wiremap` to split up the models. Which would mean it would expect `m` and `mref` to be the same shape anyways.

I do realize that this small fix on it's own is also a bug that should be updated, but there will be more issues I think that should be addressed.

PR #1126, opened 2022-10-31
fix alpha multipliers in `PGI`

thibaut-kobold:
- small bug where `alpha_pgi` was not taken into account in the definition
- removed unused `alpha_s` argument
- the setter needed to be reworked; use `validate_float` instead (if `None` was passed, which was the default, then we ended up with `None` as multiplier)
- fix example that called alpha_s instead of alpha_pgi

### DIFF
--- a/SimPEG/regularization/pgi.py
+++ b/SimPEG/regularization/pgi.py
@@ -5,8 +5,7 @@ import warnings
 
 import numpy as np
 import scipy.sparse as sp
-from SimPEG.utils.code_utils import (deprecate_property,
-                                     validate_ndarray_with_shape)
+from SimPEG.utils.code_utils import deprecate_property, validate_ndarray_with_shape
 
 from ..maps import IdentityMap, Wires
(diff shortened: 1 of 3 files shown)


SimPEG/regularization/pgi.py
```diff
@@ -764,11 +756,14 @@ def alpha_pgi(self):
 
     @alpha_pgi.setter
     def alpha_pgi(self, value):
-        if isinstance(value, (float, int)) and value < 0:
```
jcapriot, line 767: Just use the `validate_float` function from `SimPEG.utils`

jcapriot: It's up to you to decide if it makes sense to allow passing `None` type to this variable. It seems like it'd be better off just leaving it's default (set during `__init__`) at `1.0` and the not allowing a None type to be set to `alpha_pgi`

thibaut-kobold: Let be explicit and not allow None.

PR #1130, opened 2022-11-09
Fix broken math in regularization docs

santisoler:
Fix math in the API reference page for regularizations so the equations are
properly rendered. Slightly change the style in which the equations are written
so they are easier to read directly from the LaTeX sources.


jcapriot:
Looks like these all check out, thanks for fixing that eye sore on the documentation page!

PR #1132, opened 2022-11-09
Remove unused imports across SimPEG

santisoler:
Remove all the imported packages that are not being used within each file.
Avoid removing all the imports that are carried out inside the `__init__.py`
files. Used `autoflake` to perform the removal automatically.


santisoler:
> A lot of the things in `*.utils` files needs to stay there at the moment, as there is a lot of old code out there that imports things directly from those modules (instead of importing them from `SimPEG.utils` as we would intend). I think this also applies to some of the things in `static_utils` as well.

I didn't know that. By "old code" do you mean within SimPEG or other third-party codes that are relying on those imports to be there? If it's only on SimPEG this could actually be an opportunity to shift them.

Anyway, the files that need to keep those unused imports are `*/utils/__init__.py` or just `*/utils.py`?

santisoler:
I've restored all the unused imports within `utils` that were imported in any `__init__.py`, so now `import SimPEG` should work properly again. I've added a `# noqa: F401` comment on those imports so [`flake8` can ignore them](https://flake8.pycqa.org/en/latest/user/violations.html).

I've also removed unused imports in `tests`, `examples` and `tutorials`. There's only one unused import left in `tests/em/fdem/forward/test_FDEM_primsec.py`:

```python
try:
    from pymatsolver import Pardiso as Solver
except ImportError:
    from SimPEG import Solver as SolverLU
```

The `SolveLU` is not being used, but I'm worried if in any case we would want to test without `pymatsolver`. Should we remove this unused import or just import it as `Solver` (its original name)?

jcapriot:
I'm going to do a pass on this one myself to add back incorrect removed imports.

PR #1134, opened 2022-11-11
Run style checks in Azure Pipelines

santisoler:
Run style checks using flake8 and some of its plugins. Add a new
`.flake8` configuration file to tweak the style checks. Add a new
`check` target in the `Makefile`. Create a new pipeline in Azure to
automatically run style checks on every PullRequest. Add new required
packages to `environment_test.yml` and to a new
`requirements_style.txt`.

### DIFF
--- a/environment_test.yml
+++ b/environment_test.yml
@@ -38,6 +38,7 @@ dependencies:
   - python-kaleido
   # Linters and code style
   - black==21.7b0
+  - click==8.0.4 # pin click needed by the pinned black version
   - flake8
   - flake8-bugbear
   - flake8-builtins
--- a/environment_test.yml
+++ b/environment_test.yml
@@ -36,3 +36,10 @@ dependencies:
   - pyvista
   - pip
   - python-kaleido
+  # Linters and code style
+  - black
+  - flake8
+  - flake8-bugbear
+  - flake8-builtins
+  - flake8-mutable
+  - flake8-rst-docstrings
--- a/environment_test.yml
+++ b/environment_test.yml
@@ -37,7 +37,7 @@ dependencies:
   - pip
   - python-kaleido
   # Linters and code style
-  - black
+  - black==21.7b0
   - flake8
   - flake8-bugbear
   - flake8-builtins
--- a/environment_test.yml
+++ b/environment_test.yml
@@ -43,3 +43,4 @@ dependencies:
   - flake8-builtins
   - flake8-mutable
   - flake8-rst-docstrings
+  - flake8-docstrings
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,35 @@
+# -----------------------------
+# Configuration file for flake8
+# -----------------------------
+
+# Configure flake8
+# ----------------
+[flake8]
+extend-ignore =
+    # Too many leading '#' for block comment
+    E266,
+    # Line too long (82 > 79 characters)
+    E501,
+    # Do not use variables named 'I', 'O', or 'l'
+    E741,
+    # Line break before binary operator (conflicts with black)
+    W503,
+
+exclude =
+    .git,
+    __pycache__,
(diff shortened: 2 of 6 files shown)


santisoler:
I expect the style checks to fail big time now, but sets the basis for solving those issues and also choose which style checks we would like to enforce in SimPEG.

One major thing I haven't added is [`pep8-naming`](https://pypi.org/project/pep8-naming/), a package and `flake8` plugin that checks that variable, functions and classes names follow PEP8. I wouldn't add it right away in SimPEG since it will complain a lot now, but I'm mentioning it to keep record that in the long run would be nice to have it.

jcapriot:
I'm still surprised there's a Makefile left over in this repo, it is barely used for anything, and there's no compiled codes in here anyways that would make it that necessary.

jcapriot:
@santisoler looks like there's an error in the azure-pipelines file with a bad name for the style check stage https://dev.azure.com/simpeg/simpeg/_build/results?buildId=2110&view=results. Probably needs to be renamed "Style_Check". You can always explicitly set it's displayname though.

santisoler:
> I'm still surprised there's a Makefile left over in this repo, It is barely used for anything, and there's no compiled codes in here anyways that would make it that necessary.

I saw @rowanc1's hand in that Makefile. I imagined it's not being heavily used in SimPEG. I usually love to see a Makefile in the repo (even if the repo doesn't involve compilation steps) because it provides an easy way to standardize processes, document them and make them quickly available. For example, instead of adding instructions in the README or in the Contributors Guide on how to run every style check, one could just say "run `make check`". We use this approach in every Fatiando package (see https://github.com/fatiando/verde/blob/main/Makefile) and we use them in our [Contributing Guidelines](https://github.com/fatiando/community/blob/main/CONTRIBUTING.md#code-style). One can also use it in CI to run tests and code styles (like I'm doing here). So if you want to change something about how the code style gets done, you just need to edit the `Makefile` and leave the azure pipeline file intact.

jcapriot:
We probably want to (for now) allow these tests to fail and still continue with unit tests.

environment_test.yml
```diff
@@ -36,3 +36,10 @@ dependencies:
   - pyvista
   - pip
   - python-kaleido
+  # Linters and code style
+  - black
```
jcapriot, line 40: make sure that this version matches the current version in the .pre-commit.yml file. (I think 21.7b0)

environment_test.yml
```diff
@@ -36,3 +36,10 @@ dependencies:
...
+  - flake8
+  - flake8-bugbear
+  - flake8-builtins
+  - flake8-mutable
+  - flake8-rst-docstrings
```
jcapriot, line 45: we might also want to use `flake8-docstrings` for numpy style documentation checking
https://pypi.org/project/flake8-docstrings/

santisoler:
> We probably want to (for now) allow these tests to fail and still continue with unit tests.

Sure. Let's just merge this PR and fix all the style issues in separate PRs.

rowanc1:
Looks like the `make graphs` is still in there as well, which made some figures for various papers about the inheritance structure. Exciting to see these new improvements!!

santisoler:
Thanks @jcapriot for all the changes. I just grouped the target files into a single variable in the Makefile to avoid repetition and to make it easier to change in the future. Feel free to revert it if you don't like it. From my point this PR is ready, so merge it whenever you feel the same.

jcapriot:
Awesome, I think the only thing that looks obvious to turn off is the documentation style checking in tests

santisoler:
> Awesome, I think the only thing that looks obvious to turn off is the documentation style checking in tests

Makes sense. We'll never render docstrings of test functions in Sphinx, so we don't care about their format, they just need to be informative when reading them in plain text.

PR #1136, opened 2022-11-24
Use r-strings on strings and docstrings that contain backslashes

santisoler:
Solve `flake8`'s `W605` and `D301` errors that complains about using invalid
escape sequences in strings and not using r-strings for docstrings that has
backslashes. Minor style changes to math inside docstrings for improved
readability.

This PR is intended to solve a portion of the issues raised by flake8 on #1134


santisoler:
There are some invalid escape sequences `\s` in `SimPEG/utils/drivers/magnetics_driver.py` and `SimPEG/utils/drivers/gravity_driver.py`. They are part of strings that are passed as arguments to `re.split()`. Are we currently testing those lines? If we change those strings will we caught if the behaviour of those functions changes?

jcapriot:
> There are some invalid escape sequences `\s` in `SimPEG/utils/drivers/magnetics_driver.py` and `SimPEG/utils/drivers/gravity_driver.py`. They are part of strings that are passed as arguments to `re.split()`. Are we currently testing those lines? If we change those strings will we caught if the behaviour of those functions changes?

I'm fairly sure they were meant to be the specific sequence of character "\s" in a regexp, which is a catch all for any whitespace. So for the python input the string should probably be r"\s" for it to be literal.

santisoler:
> > There are some invalid escape sequences `\s` in `SimPEG/utils/drivers/magnetics_driver.py` and `SimPEG/utils/drivers/gravity_driver.py`. They are part of strings that are passed as arguments to `re.split()`. Are we currently testing those lines? If we change those strings will we caught if the behaviour of those functions changes?
> 
> I'm fairly sure they were meant to be the specific sequence of character "\s" in a regexp, which is a catch all for any whitespace. So for the python input the string should probably be r"\s" for it to be literal.
> 
> Now it being actually tested is not very likely. I have a feeling these "drivers" are not used very often.

Got it. I'm changing these then.

santisoler:
Got rid of every `W605` and `D301` error when running `make check` 🎉 

I think this should be ready to merge. I'm inclined to solve sets of these errors one PR at a time.

PR #1137, opened 2022-11-30
Rename variables to avoid shadowing builtins

santisoler:
Fix some `flake8` `A00*` warnings that warn the usage of variables and
arguments names that shadow some builtins variables, functions and
modules.

### DIFF
--- a/SimPEG/potential_fields/base.py
+++ b/SimPEG/potential_fields/base.py
@@ -250,7 +250,7 @@ class BaseEquivalentSourceLayerSimulation(BasePFSimulation):
         self._unique_inv = None
 
 
-def progress(iter, prog, final):
+def progress(iteration, prog, final):
     """Progress (% complete) for constructing sensitivity matrix
 
     Parameters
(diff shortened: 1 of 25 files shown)


santisoler:
There are some `A003` warnings left related to the `eval` methods of some classes. I didn't want to rename those in case we could create a backward compatibility issue. You can check them by running:

```bash
make flake | grep -E "A[0-9]+"
```

@jcapriot what's your take on those?

jcapriot:
renaming `eval` straight up might cause a few too many issues. However, what I would like to see done eventually is replace all of the `eval` with `__call__`

santisoler:
You're right, that's why I didn't want to touch them. I think we could add a deprecation warning for the `eval` methods in another PR and ultimately deprecate them in v0.20.

jcapriot:
We can probably tackle it when we finish up all the pep8 renaming

santisoler:
@jcapriot I think this is ready to be merged.

PR #1138, opened 2022-11-30
Remove unused loop variables

santisoler:
Fix `flake8` `A007` warnings. Replace loop variables that weren't being
used inside the loop for a underscore (`_`). Avoid using `enumerate`
whenever the counter is not being used inside the loop.


santisoler:
There are two `B007` left here:

https://github.com/simpeg/simpeg/blob/7ef5223615b339ce8ba88c4e6b4ae83940d41a0c/SimPEG/electromagnetics/natural_source/utils/analytic_1d.py#L79-L92 

and here:

https://github.com/simpeg/simpeg/blob/7ef5223615b339ce8ba88c4e6b4ae83940d41a0c/tests/em/em1d/test_EM1D_FD_fwd.py#L285-L296

Could we safely remove those two loop variables or we actually found a bug where they should be used but aren't?

jcapriot:
> There are two `B007` left here:
> 
> https://github.com/simpeg/simpeg/blob/7ef5223615b339ce8ba88c4e6b4ae83940d41a0c/SimPEG/electromagnetics/natural_source/utils/analytic_1d.py#L79-L92
> 
> and here:
> 
> https://github.com/simpeg/simpeg/blob/7ef5223615b339ce8ba88c4e6b4ae83940d41a0c/tests/em/em1d/test_EM1D_FD_fwd.py#L285-L296
> 
> Could we safely remove those two loop variables or we actually found a bug where they should be used but aren't?

The first, `eps` can be safely removed as it is already accounted for in the definition of `ki` above.

The one in EM1D I updated to make use of the complex conductivity value for testing by splitting and assigning it to the correct parts of the `MagneticDipoleHalfSpace`

PR #1139, opened 2022-11-30
Avoid using getattr with fixed constant strings

santisoler:
Fix `flake8` `B009` warnings. Refactor the way fixed attributes are
obtained: instead of using `getattr` with fixed strings, use the
standard `object.attribute` syntax that is more efficient and readable.
Use `hasattr` for checking if the wanted attribute exists instead of
using `getattr` and returning None if not: improves readability.


santisoler:
Indeed. It's a great tool to catch attributes programatically. But if the attribute's name is already known, having to allocate a string and calling `getattr` is less efficient. In cases where we don't know if the attribute exists, I find the `getattr(obj, attr, None)` less readable than:

```python
if hasattr(obj, attr):
    # Do something with obj.attribute
    ...
```

PR #1140, opened 2022-11-30
Avoid using bare `except`s that catch every possible exception

santisoler:
Fix some `flake8` `B001` warnings. Avoid using bare `except` statements: they
catch every possible exception that silent any unexpected error, might hide
bugs and produce misleading error messages for the user. Use `ImportErrors` on
`try...except` blocks at import level. Expect the same exception type that is
being overwritten when the error message must be changed. Refactor the
`Mesh2Mesh` constructor and the `Regularizationmesh.active_cells` method  with
cleaner code that avoids using bare exceptions.


santisoler:
There are still a few bare `except` statemets left. In some of them I don't know the exception that is intended to be caught. A list of them can be obtained running:

```bash
make flake | grep "B001"
```

SimPEG/utils/code_utils.py
```diff
@@ -821,7 +821,7 @@ def validate_integer(property_name, var, min_val=-np.inf, max_val=np.inf):
     """
     try:
         var = int(var)
-    except:
+    except TypeError:
```
jcapriot, line 824: Should likely be `(ValueError, TypeError)`. If the item is a single entity usually a value error is thrown.

SimPEG/utils/code_utils.py
```diff
@@ -864,7 +864,7 @@ def validate_float(
     """
     try:
         var = float(var)
-    except:
+    except TypeError:
```
jcapriot, line 867: `(ValueError, TypeError)`

SimPEG/regularization/regularization_mesh.py
```diff
@@ -53,20 +53,24 @@ def active_cells(self, values: np.ndarray):
...
-                    tmp[values] = True
-                except:
-                    raise ValueError(
-                        "Values must be an array of integers or an array of bools "
-                        "indicating the active cells"
-                    )
+            # Convert values into an array if needed
+            values = np.asarray(values)
+            # Check if values has the right dimensions and size
+            if values.shape != (self.mesh.nC,):
```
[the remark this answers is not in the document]
santisoler: Sure, that's way better.

SimPEG/maps.py
```diff
@@ -2949,17 +2950,28 @@ class Mesh2Mesh(IdentityMap):
     """
 
     def __init__(self, meshes, indActive=None, **kwargs):
+        # Sanity checks for the meshes parameter
+        if not isinstance(meshes, Iterable):
+            raise ValueError(
```
jcapriot, line 2955: Should be a `TypeError` here (Incorrect type passed to the object)

SimPEG/utils/code_utils.py
```diff
@@ -1070,7 +1070,7 @@ def validate_type(property_name, obj, obj_type, cast=True, strict=False):
     if cast:
         try:
             obj = obj_type(obj)
-        except:
+        except TypeError:
```
jcapriot, line 1073: For this one, there are just so many possible error types that could happen, I imagine that the best way to do this one would be something like:

```
except Exception as err:
    raise TypeError(
                f"{type(obj).__name__} cannot be converted to type {obj_type.__name__} "
                f"required for {property_name}."
            ) from err
```

SimPEG/utils/io_utils/io_utils_pf.py
```diff
@@ -296,7 +296,7 @@ def read_gg3d_ubc(obs_file):
                     d.append(factor * temp[3 : 3 + n_comp])
                     wd.append(temp[3 + n_comp :])
                 ii += 1
-            except:
+            except IOError:
```
jcapriot, line 299: I think there's a couple possible errors that should be caught here, all related to the above lines.
- Should error on file read error
- Should error if the array is unable to be converted to floats,

SimPEG/maps.py
jcapriot: Also, checking if it is an instance of `Iterable` does not always succeed (The documentation states it does not work if the container iterates with a `__getitem__` call.) I would leave it as is, but specifically catch the TypeError that is thrown if the input can't be unpacked.

santisoler: Right. I fell back to the original version but now we are catching the exception. I also removed the extra check for the size of the `meshes`, since we would catch that in the `try`.

SimPEG/utils/code_utils.py
santisoler: Done!

SimPEG/utils/io_utils/io_utils_pf.py
santisoler: You're right. I think the main issue is that it's running a lot of things within a single `try`, shadowing possible errors. I unpacked some portions of this code. The `readline()` is not likely to file, but to return an empty string if it hits the EOF. Now the `try` only contains the parsing of the line to the array and I improved the error messages to be more descriptive. I also fixed the line number shown in the messages: line numbers usually start in 1 and we are already reading the first two lines, so the actual line number at each iteration is `ii + 3`.

SimPEG/maps.py
```diff
@@ -2949,17 +2950,23 @@ class Mesh2Mesh(IdentityMap):
...
     def __init__(self, meshes, indActive=None, **kwargs):
-
+        # Sanity checks for the meshes parameter
         try:
             mesh, mesh2 = meshes
-        except:
-            raise TypeError("meshes must be a list of two meshes")
+        except TypeError:
+            raise TypeError("Couldn't unpack 'meshes' into two meshes.")
+        mesh, mesh2 = meshes
```
jcapriot, line 2958: this line is duplicated from the above try statement (when it succeeds)

SimPEG/utils/drivers/gravity_driver.py
```diff
@@ -5,7 +5,7 @@
 
 try:
     from SimPEG import utils
-except:
+except ImportError:
```
jcapriot, line 8: I'm not even sure how these lines made it into here, but I think it's safe to remove the try and just straight up do:
`from SimPEG import utils`

SimPEG/utils/drivers/magnetics_driver.py
```diff
@@ -5,7 +5,7 @@
 
 try:
     from SimPEG import utils
-except:
+except ImportError:
```
jcapriot, line 8: Same comment from the gravity_driver.py file

examples/02-gravity/plot_inv_grav_tiled.py
```diff
@@ -26,7 +26,7 @@
 try:
     from SimPEG import utils
     from SimPEG.utils import plot2Ddata
-except:
+except ImportError:
```
jcapriot, line 29: Same comment as the gravity_driver.py file.

I suspect that these were do to someone trying to simulatenously support multiple versions of SimPEG

SimPEG/utils/code_utils.py
```diff
@@ -960,7 +960,7 @@ def validate_location_property(property_name, var, dim=None):
     """
     try:
         var = np.atleast_1d(var).astype(float).squeeze()
-    except:
+    except TypeError:
```
jcapriot, line 963:
```suggestion
    except (TypeError, ValueError):
```

PR #1141, opened 2022-12-08
Run flake8 in CI against a selection of rules

santisoler:
Create new `flake-permissive` target in Makefile. The new target runs `flake8`
checking if only a selection of rules are being violated. It's being added with
the intention of start catching `flake8` rules that have already been solved
throughout the entire SimPEG. This new rule should be temporary until every
`flake8` rule has been solved. Make Azure to run two `flake8` checks: the
`flake-permissive` target (that checks only of a selection of flake rules) and
the `flake` target but allowing it to fail (it returns exit code zero even if
some rules are being violated).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,15 @@
 STYLE_CHECK_FILES=SimPEG examples tutorials tests
 
-# Define flake8 warnings that should be catched for now.
+# Define flake8 warnings that shouldn't be catched for now.
 # Every time a new warning is solved througout the entire project, it should be
-# added to this list.
+# removed to this list.
 # This list is only meant to be used in the flake-permissive target and it's
-# a temprary solution until every flake8 warning is solved in SimPEG
-FLAKE8_SELECT=W605,D301
+# a temprary solution until every flake8 warning is solved in SimPEG.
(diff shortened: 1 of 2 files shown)


jcapriot:
This looks good. I think we can now delete the `conditions` line of the Testing stage, and remove the first parts of the `conditions` of the Deploy stage.

```
- stage: Testing
  dependsOn: StyleChecks
  jobs:
...
```

```
- stage: Deploy
  dependsOn: Testing
  condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
  jobs:
...
```

santisoler:
Ok. I decided to go back to have a list of rules that we will ignore (allow to fail for now) than just checking a few selections. I want to catch every future warning that is not being currently triggered in SimPEG. This force us to have a very long list of rules in the Makefile, but since they are meant to be temporary I don't see any problem with that.

jcapriot:
@santisoler if you’re good on this implementation I’m happy with it as well

santisoler:
> @santisoler if you’re good on this implementation I’m happy with it as well

I'm happy with it. I would rather merge this ASAP to avoid confusions like the one we had in #1147 than tweak this further more. In any case we can always modify this in future PRs.

PR #1147, opened 2022-12-19
Enable use of inactive cells in equivalent source models

nwilliams-kobold:
As described in https://simpeg.slack.com/archives/CFJUKBPDZ/p1670976419407039 equivalent source models currently do not allow use of inactive cells. The code requires the `cell_z_top` and `cell_z_bottom` to be the same size as `mesh.nC`. Simply changing this to the internally-set `self.nC` allows use of inactive and active cells with the existing `ind_active` parameter.

This change is backwards compatible since using the full active set of cells is always going to be the same as `mesh.nC`, and `self.nC==mesh.nC` in that case.

### DIFF
--- a/SimPEG/potential_fields/base.py
+++ b/SimPEG/potential_fields/base.py
@@ -1,13 +1,15 @@
 import os
+import warnings
+from multiprocessing.pool import Pool
 
 import discretize
 import numpy as np
-import warnings
-from ..simulation import LinearSimulation
 from scipy.sparse import csr_matrix as csr
+
 from SimPEG.utils import mkvc
-from ..utils import validate_string, validate_active_indices, validate_integer
-from multiprocessing.pool import Pool
+
+from ..simulation import LinearSimulation
+from ..utils import validate_active_indices, validate_integer, validate_string
 
 ###############################################################################
 #                                                                             #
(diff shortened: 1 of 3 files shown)


domfournier:
Only getting 6k+ warnings
image

@jcapriot and @santisoler I know you guys are working on this. Any advice on how to get this PR through?

santisoler:
Thanks @domfournier for pinging me here. Feel free to ignore the flake8 errors for now. We are tackling this issue in https://github.com/simpeg/simpeg/pull/1141. Once that PR is merged we are going to check the entire codebase against the flake8 errors that were already solved throughout the entire project. Sorry for the confusion @nwilliams-kobold

santisoler:
@nwilliams-kobold, #1141 have just been merged, so if you merge the current `main` into your branch you should get rid of the CI failures. There will be two flake8 runs: one that checks the flake warnings we currently want to catch and will fail if the PR violates any of them; and another one that runs flake8 with every warning, but this is allowed to fail so the CI job for that one will never fail.

jcapriot:
@nwilliams-kobold Thanks for updating this, I see you've added a test to check the error message, but can you also add a test that makes sure this works?

nwilliams-kobold:
@jcapriot I have added tests for both gravity and magnetics when using equivalent source forward and inverse models with active cells. The new tests are based on the existing all-active tests but I have set all zero-value cells in the model to inactive.

This is exactly matches one of the two main use cases for having active cells in the eqs code:

1. Avoiding compute for zero-valued cells in a forward model
2. Avoiding compute for cells that have zero height

jcapriot:
Just make sure the tolerance is set to a reasonable level that implies it "worked".

nwilliams-kobold:
I have updated the tolerances. I was surprised it failed as it passed locally well within tolerance. I have also plotted all of the data plots and model maps and confirmed everything makes sense.

eg., here's the active-subset mag inversion result plan view on the left relative to the tru model on the right.
CleanShot 2023-01-18 at 12 28 32@2x

jcapriot:
The tests with randomized inputs are a little tougher to reproduce locally. The seed is set when the test script is imported by pytest, not when the script is ran. I would recommend setting the seed inside the testing function itself to make your t easy to reproduce local behavoir

nwilliams-kobold:
@jcapriot Ok, the seed seemed to stabilise the tests. I'm done unless you have more requests?

PR #1148, opened 2022-12-23
TEM Store factorizations in RAM

dccowan:
In this PR, we add the option to store factorizations in RAM. The ability to read/write factorizations can be handled as part of another project. Just like the FEM problem, this is controlled using the 'forward_only' kwarg

### DIFF
--- a/SimPEG/electromagnetics/time_domain/simulation.py
+++ b/SimPEG/electromagnetics/time_domain/simulation.py
@@ -30,13 +30,11 @@ class BaseTDEMSimulation(BaseTimeSimulation, BaseEMSimulation):
             survey=None,
             dt_threshold=1e-8,
             forward_only=True,
-            write_factorizations=True,
             **kwargs
         ):
         super().__init__(mesh=mesh, survey=survey, **kwargs)
         self.dt_threshold = dt_threshold
         self.forward_only = forward_only
-        self.write_factorizations = write_factorizations
         if self.muMap is not None:
             raise NotImplementedError(
                 "Time domain EM simulations do not support magnetic permeability "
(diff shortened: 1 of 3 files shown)


dccowan:
Is there a particular version of black that must be used? I already ran 'black simulation.py' but it's failing the black test on Azure. Please advise.

jcapriot:
SimPEG currently runs black at v21.7b0

jcapriot:
There’s a lot of logic here to determine the step length indices because you’re storing factored Ainv’s in an array. Store them in a dictionary that is indexed by `dt` and you can simplify this significantly.

dccowan:
> There’s a lot of logic here to determine the step length indices because you’re storing factored Ainv’s in an array. Store them in a dictionary that is indexed by `dt` and you can simplify this significantly.

Oh wow, didn't know you could use a float as a dictionary key. Ya, that's more compact.

SimPEG/electromagnetics/time_domain/simulation.py
```diff
@@ -86,36 +100,79 @@ def fields(self, m):
...
+        f[:, self._fieldType + "Solution", 0] = self.getInitialFields()  # mesh x n_src
+
+        # Store Factorizations
+        if self.forward_only is False:
+
+            # Clean factorizations for preexisting model.
+            if hasattr(self, "Ainv"):
+                {k: v.clean() for k, v in self.Ainv.items()}
+            else:
+                self.time_steps = self.dt_threshold * np.round(
```
jcapriot, line 119: This will fail very quickly if dt_threshold is 0

dccowan: K. So let's update the setter so that dt_threshold must be larger than 0. That work?

PR #1151, opened 2023-01-12
Alternative Store tdem factors

jcapriot:
@dccowan This is more what I had in mind for storing the factorizations.
Less changes to the original code, and I'm actually not concerned about just having a better name off the bat for the option.

### DIFF
--- a/SimPEG/electromagnetics/time_domain/simulation.py
+++ b/SimPEG/electromagnetics/time_domain/simulation.py
@@ -24,9 +24,10 @@ class BaseTDEMSimulation(BaseTimeSimulation, BaseEMSimulation):
     Euler.
     """
 
-    def __init__(self, mesh, survey=None, dt_threshold=1e-8, **kwargs):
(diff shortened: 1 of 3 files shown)


lheagy:
@dccowan: would you have time to review this pr?

SimPEG/electromagnetics/time_domain/simulation.py
```diff
@@ -23,9 +23,13 @@ class BaseTDEMSimulation(BaseTimeSimulation, BaseEMSimulation):
     Euler.
     """
 
-    def __init__(self, mesh, survey=None, dt_threshold=1e-8, **kwargs):
+    def __init__(
+        self, mesh, survey=None, dt_threshold=1e-8, store_factors=True, **kwargs
```
dccowan, line 27: Are we sure we want to set the default for `store_factors` to True? I'm pretty sure that people new to SimPEG won't fully consider the impact of this choice when simulating TDEM data.

PR #1160, opened 2023-01-17
Avoid comparing bool variables to True or False

santisoler:
Solve `flake` `E712` warnings. Avoid comparing bool variables against True or
False, use a more pythonic way like `if variable` or `if not variable`.


SimPEG/electromagnetics/viscous_remanent_magnetization/sources.py
```diff
@@ -564,8 +564,8 @@ def _getRefineFlags(self, xyzc, refinement_factor, refinement_distance):
                 )
 
                 ind = (
-                    (k_pos == False)
-                    & (k_neg == False)
+                    (not k_pos)
+                    & (not k_neg)
```
jcapriot, line 568: These two items are arrays of truth values, need to use `!` to negate them, not `not`.

tests/em/fdem/forward/test_FDEM_sources.py
```diff
@@ -134,7 +134,7 @@ def ana_sol(XYZ):
             look_at_these = np.ones(self.mesh.nEx + self.mesh.nEy, dtype=bool)
 
         look_at_these = np.hstack(
-            [look_at_these, np.array(ignore_these == False, dtype=bool)]
+            [look_at_these, np.array(not ignore_these, dtype=bool)]
```
jcapriot, line 137: Again, needs to use `!` instead of `not`

jcapriot (changes requested):
Just some errors caused by using `not` on a numpy arrays of booleans (should be using the `!` operator instead).

SimPEG/electromagnetics/viscous_remanent_magnetization/sources.py
jcapriot: sorry meant `~`

tests/em/fdem/forward/test_FDEM_sources.py
jcapriot: sorry meant `~`

PR #1162, opened 2023-01-18
Generalize on_model_update call

domfournier:


jcapriot:
In the following places: https://github.com/simpeg/simpeg/blob/33a11c4d1760f3d3a039980f5ceb56a5a046c4ce/SimPEG/props.py#L445 and https://github.com/simpeg/simpeg/blob/33a11c4d1760f3d3a039980f5ceb56a5a046c4ce/SimPEG/props.py#L461

The call to the items in clean_on_model_update should work if it is a dictionary or list as well as an item.

* For a dictionary, iterate through the keys and clean those values
* For a list, iterate through the items and clean them

dccowan:
Some notes:

- This functionality is found in props.py
- 'Clean' on model update and 'Delete' on model update will be two different lists. The former applies to the solver object or dictionary of solver objects.
- Everything in the clean list must also be in the delete list. Since clean only gets rid of the factorization and not the solver object.
- Write unit tests

jcapriot:
I've updated the clean matrix calls. Mainly needs testing and to be made use of inside the relevant simulations.

PR #1163, opened 2023-01-18
Add option for amplitude threshold on UpdateSens directive

domfournier:


dccowan:
I'm thinking we use the following properties:

- **threshold** (float), whose action depends on ...
- **threshold_method** (str). Which can be one of:
    - 'global' - adds constant value prior to normalization
    - 'percentile' - picks threshold based on the percentile
    - 'amplitude' - picks threshold base on fraction of maximum amplitude
- **normalization_method** (None, str). Which can be one of:
    - 'maximum' - normalizes by the maximum so that largest cell weights are 1
    - 'minimum' - normalizes by the lower value, after thresholding, so that smallest cell weights are 1

domfournier:
I would personally vote for shorter keywords (e.g. "method", "normalization"), just for convenience.

dccowan:
I think we need to be a little specific in case we add any other functionality. 'method' at first glance isn't particularly descriptive.

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
...
+            )
+            threshold_value = kwargs.pop("threshold")
+
+        if "normalization" in kwargs.keys():
+            warnings.warn(
+                "'normalization' property is deprecated and will be removed in SimPEG 0.19.0."
+                "Please define normalization using 'normalization_method'."
+            )
+            normalization_method = kwargs.pop("normalization")
+            if normalization_method is True:
```
jcapriot, line 2305: These lines are repeated in the `setter` of `normalization_method` you can just let that function handle changing it from a `bool` to the appropriate `str` (or None) option. Then you also get a message to the user about how boolean input is also deprecated.

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
...
+            )
+            if value:
+                self._normalization_method = "maximum"
+            else:
+                self._normalization_method = None
+
+        else:
+            self._normalization_method = validate_string(
+                "normalization_method", value, string_list=["minimum", "maximum"]
+            ).lower()
```
jcapriot, line 2415: Again, the `.lower()` is unnecessary.

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
...
+
+    deprecate_property(
```
jcapriot, line 2417: Same comment as before, you need to store this on the class as the old name,
`normalization = deprecated_property(...)`

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
...
+        return self._every_iteration
 
-    @everyIter.setter
-    def everyIter(self, value):
-        self._everyIter = validate_type("everyIter", value, bool)
+    @every_iteration.setter
+    def every_iteration(self, value):
+        self._every_iteration = validate_type("every_iteration", value, bool)
+
+    deprecate_property(
```
jcapriot, line 2334: Needs to be stored on the class
`everyIter = deprecate_property(...)`

SimPEG/directives/directives.py
```diff
@@ -2300,24 +2455,47 @@ def update(self):
...
                     (reg.mapping * jtj_diag) / reg.regularization_mesh.vol ** 2.0
                 )
-        if self.normalization:
-            wr /= wr.max()
-        wr += self.threshold
+
         wr **= 0.5
+
+        # Apply thresholding
+        if self.threshold_method == "global":
```
jcapriot, line 2469: Just making a note here that this slightly changes the behavior from before, with the default now being the `global` method meaning the threshold is applied before normalization. Does the `percentile` or `amplitude` threshold option do anything similar to this? (I feel like the `amplitude` threshold comes closest but it still clips and doesn't add.) Not saying this is a blocking change, just wanted to make note and allow others to comment if they wanted to.

SimPEG/directives/directives.py
```diff
@@ -2300,24 +2455,47 @@ def update(self):
...
+        # Add sensitivity weighting to all model objective functions
         for reg in self.reg.objfcts:
             if not isinstance(reg, BaseSimilarityMeasure):
                 sub_regs = getattr(reg, "objfcts", [reg])
                 for sub_reg in sub_regs:
                     sub_reg.set_weights(sensitivity=sub_reg.mapping * wr)
 
     def validate(self, directiveList):
+        """Validate directive against directives list.
+
```
jcapriot, line 2493: Should probably make a comment here (and in the class docstring itself) That this class must come before `BetaEstimate_ByEig` and `UpdatePreconditioner` in the directives list.

tests/base/test_directives.py
```diff
@@ -103,22 +103,78 @@ def test_validation_in_inversion(self):
...
         with self.assertRaises(AssertionError):
             # validation should happen and this will fail
             # (sensitivity_weights needs to be before update_Jacobi)
-            inv = inversion.BaseInversion(self.invProb)
+            inv = inversion.BaseInversion(invProb)
             inv.directiveList = [update_Jacobi, sensitivity_weights]
 
+    def test_sensitivity_weighting(self):
+
+        tests_list = [
```
jcapriot, line 124: I would recommend making these as differently ran tests, just in case the first option fails it will still test the second. You can still re-use most of the code, just give these inputs to the internal test func.

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
 
 
 class UpdateSensitivityWeights(InversionDirective):
+    r"""
+    Sensitivity weighting for linear and non-linear problems.
+
+    Compute and apply root-mean squared sensitivities as cell weights in the regularization.
+    Assumes that the map of the regularization function is either ``Wires`` or ``Identity``.
+    This directive requires that the sensitivities be stored explicitly.
```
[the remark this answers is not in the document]
dccowan: Sounds good

SimPEG/directives/directives.py
```diff
@@ -2205,66 +2205,224 @@ def JtJv(v):
...
+    If **global** thresholding is applied, we add a constant :math:`\tau` to the RMS sensitivities:
+
+    .. math::
+        \mathbf{\tilde{s}} = \mathbf{s} + \tau
+
+    In the case of **percentile** thresholding, we let :math:`s_{\%}` represent a given percentile.
+    Thresholding to set a minimum value is applied as follows:
+
+    .. math::
+        \tilde{s}_j = \begin{bases}
```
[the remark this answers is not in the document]
dccowan: good catch!!!

SimPEG/directives/directives.py
```diff
@@ -2178,66 +2178,249 @@ def JtJv(v):
...
+    ``getJtJdiag`` method.
+
+    Parameters
+    ----------
+    every_iteration : bool
+        When ``True``, update sensitivity weighting at every model update; non-linear problems.
+        When ``False``, create sensitivity weights for starting model only; linear problems.
+    threshold : float
+        Threshold value for smallest weighting value
+    threshold_method : {'global', 'percentile', 'amplitude'}
```
jcapriot, line 2202: 'amplitude' the default should be listed first

SimPEG/directives/directives.py
```diff
@@ -2178,66 +2178,249 @@ def JtJv(v):
...
+        Threshold method for how `threshold_value` is applied:
+
+            - global:
+                `threshold_value` is added to the cell weights prior to normalization; must be greater than 0.
+            - percentile:
+                the smallest root-mean squared sensitivity is set using percentile threshold; must be between 0 and 100
+            - amplitude:
+                the smallest root-mean squared sensitivity is a fractional percent of the largest value; must be between 0 and 1
+
+    normalization_method : {None, 'min_value', 'maximum'}
```
jcapriot, line 2212: 'maximum' the default value, should be listed first.

SimPEG/directives/directives.py
```diff
@@ -2273,24 +2453,55 @@ def update(self):
...
                 for sub_reg in sub_regs:
                     sub_reg.set_weights(sensitivity=sub_reg.mapping * wr)
 
     def validate(self, directiveList):
+        """Validate directive against directives list.
+
+        The ``UpdateSensitivityWeights`` directive impacts the regularization by applying
+        cell weights. As a result, its place in the :class:`DirectivesList` must be
+        before any directives which update the preconditioner for the inverse problem
+        (i.e. :class:`UpdatePreconditioner`), and must be before any directives that
```
jcapriot, line 2495: must be "after" directives that estimate an initial beta.

SimPEG/directives/directives.py
```diff
@@ -2273,24 +2453,55 @@ def update(self):
...
                     (reg.mapping * jtj_diag) / reg.regularization_mesh.vol**2.0
                 )
-        if self.normalization:
-            wr /= wr.max()
-        wr += self.threshold
+
         wr **= 0.5
+
+        # Apply thresholding
+        if self.threshold_method == "global":
```
jcapriot, line 2467: It looks like this option doesn't get tested at the moment.

jcapriot (changes requested):

SimPEG/directives/directives.py
dccowan: Done

SimPEG/directives/directives.py
dccowan: Done

SimPEG/directives/directives.py
dccowan: Are you sure? I would want to define my regularization before computing the balance between the regularization and data misfit.

SimPEG/directives/directives.py
dccowan: Done

SimPEG/directives/directives.py
dccowan: Im not worried

SimPEG/directives/directives.py
domfournier: Agreed - should be computed before the beta estimator.

jcapriot: Of course... for some reason I was thinking of something backwards...

SimPEG/directives/directives.py
jcapriot: Don't think you actually changed this one @dccowan

tests/base/test_directives.py
dccowan: Done

SimPEG/directives/directives.py
dccowan: Oops, you're right

SimPEG/directives/directives.py
```diff
@@ -2514,24 +2700,56 @@ def update(self):
...
+            wr = np.clip(
+                wr, a_min=np.percentile(wr, self.threshold_value), a_max=np.inf
+            )
+        else:
+            wr = np.clip(wr, a_min=self.threshold_value * wr.max(), a_max=np.inf)
+
+        # Apply normalization
+        if self.normalization_method == "maximum":
+            wr /= wr.max()
+        elif self.normalization_method == "minimum":
```
jcapriot, line 2726: @dccowan This method is not tested

PR #1166, opened 2023-01-20
Update current_utils.py

domfournier:
Fix crash on single cell intercept of source edge


jcapriot:
@domfournier Can you merge main in please?

PR #1167, opened 2023-01-20
Rotated Gradients

jcapriot:
Constructs a (optionally) rotated gradient operator on general meshes.

This makes use of discretize's `get_face_inner_product` operator to construct the regularization as an inner product of cell gradients. Essentially the weighting parameters are an anisotropic model used to combine the gradients.

This operator also works on general mesh types to construct the smoothness regularization operators on unstructured meshes as well.

Things to check (in addition to the implementation).
- The name of this regularization operator
- cell weights (are currently mixed into the anisotropic model, which I believe is consistent)
- face weights (not yet implemented, but I think should be `sqrt`-ed and applied on both sides of the current `_W` matrix.)
- Future (how to extend this for non-L1 norms.
     - I think it should be handled similarly to how the current `total` option is handled, but it will require some experimentation about how this applies to rotated systems.


jcapriot:
Ah, forgot there were changes to the DC simulation on here as well, I'm going to parse those out and put them on a separate PR

jcapriot:
image

jcapriot:
imageimage

A comparison of an inversion run with the current smoothness operator on a tree mesh, and this smoothness operator on a triangular mesh.

jcapriot:
Just having a thought here... Is it actually necessary to provide orthogonal vectors for the regularization directions, If someone wanted to regularize in two directions that were not orthogonal, I think this formulation still holds, because of the way the anisotropic weighting parameter is constructed.

SimPEG/regularization/rotated.py
```diff
@@ -0,0 +1,241 @@
...
+                    axis=-1,
+                )
+        self._anis_alpha = anis_alpha
+
+    # overwrite the call, deriv, and deriv2...
+    def __call__(self, m):
+        G = self.cell_gradient
+        M_f = self.W
+        r = G @ (self.mapping * (self._delta_m(m)))
+        return 0.5 * r @ M_f @ r
```
thibaut-kobold, line 174: Here we only have one `M_f` with `M_f = self.W`, don't we miss another to have W^TW ?

thibaut-kobold: (same comment for deriv and deriv2)

jcapriot: Actually no, the way it works here is that `M_f=W.T@W` in the standard way of describing the regularization. Thus since I use discretize to construct the anisotropic inner product matrix. So I just set `W=M_f` since there’s a lot of machinery already hooked into updated the weighting matrix on things like weighting function changes.

thibaut-kobold: Thanks. I do admit I find this extremely confusing and important to document here.

thibaut-kobold: `M_f = self.W`; then should be `M_f = self.WTW` for clarity

SimPEG/regularization/rotated.py
```diff
@@ -0,0 +1,241 @@
...
+    def W(self):
+        """The inner product operator using rotated coordinates
+
+        Returns
+        -------
+        scipy.sparse.csr_matrix
+        """
+        if getattr(self, "_W", None) is None:
+            mesh = self.regularization_mesh.mesh
+            cell_weights = np.ones(len(mesh))
```
thibaut-kobold, line 232: @jcapriot : I don't think this works with `active_cells` because `len(mesh)` here will be the size of the total mesh (active + inactive), but later, you multiply those `cell_weights` by `values` (such as sensitivity weights) which will be the shape of the number of inactive cells.

See the error I am getting:
image

thibaut-kobold: PS: the issue only appears if you are setting weights AND have active_cells.

domfournier:
Hey @thibaut-kobold @JKutt ,

Has promised, here's a VERY old notebook that I had used for plots in my thesis for rotated gradients. I tried updating it, but turned out to be a lot of work. 

image

[Chap5_RotatedGradient.zip](https://github.com/simpeg/simpeg/files/14073880/Chap5_RotatedGradient.zip)

It relies on a `getDiffOpRot` method that still survived here: https://github.com/simpeg/simpeg/blob/ac1a0927a70a844b627e3f83bb0a58b683778650/SimPEG/Regularization.py#L2100

Technically you could just copy that part only and overload the gradient operators on the reg classes to try it.

Enjoy.

SimPEG/regularization/rotated.py
```diff
@@ -0,0 +1,266 @@
...
+        Returns
+        -------
+        scipy.sparse.csr_matrix
+        """
+        if getattr(self, "_W", None) is None:
+            mesh = self.regularization_mesh.mesh
+            n_faces = mesh.n_faces
+            n_cells = self.regularization_mesh.n_cells
+            cell_weights = np.ones(n_cells)
+            face_weights = np.ones(n_faces)
```
thibaut-kobold, line 241: @jcapriot : curious if those face weights will allow for sparsity? For what I gather, it won't, correct?

jcapriot: They'd likely allow for it to be sparse in x-y-z orientations, but I don't think they would work properly for making it sparse in a rotated direction. It might be approximately that on minimization but I'm not sure on that front either because I haven't tried it.

PR #1171, opened 2023-01-26
Use discretize.utils.active_from_xyz to get active topography cells

dccowan:


SimPEG/utils/model_utils.py
```diff
@@ -1,9 +1,10 @@
...
-from scipy.interpolate import griddata, interp1d
-from scipy.interpolate import NearestNDInterpolator, LinearNDInterpolator, interp1d
+from scipy.interpolate import griddata
 from scipy.spatial import cKDTree
 import scipy.sparse as sp
+from discretize.utils import active_from_xyz
+import warnings
 
 
 def surface2ind_topo(mesh, topo, gridLoc="CC", method="nearest", fill_value=np.nan):
```
jcapriot, line 10: Make surface2ind_topo return the same thing it did before please. Even though it is deprecated, it's functionality should be unchanged.

Probably just need to do:

```
active_cells = active_from_xyz(mesh, topo, gridLoc, method)
return np.arange(mesh.n_cells)[active_cells]
```

SimPEG/utils/__init__.py
```diff
@@ -104,6 +104,8 @@
 .. autosummary::
   :toctree: generated/
 
+  active_from_xyz
```
jcapriot, line 107: Not sure what other's opinions are, but mine is that I don't necessarily like importing things defined in `discretize.utils` into `SimPEG.utils` just to shadow them there. I would prefer for the user to import them from `discretize` as needed.

Just trying to say avoid importing *more* things from `discretize` than we already are.

SimPEG/utils/model_utils.py
```diff
@@ -316,6 +139,14 @@ def depth_weighting(mesh, reference_locs, indActive=None, exponent=2.0, threshol
     'exponent' and 'threshold' are two adjustable parameters.
     """
 
+    if "indActive" in kwargs:
+        warnings.warn(
+            "The indActive keyword argument has been deprecated, please use grid_loc. "
```
jcapriot, line 144: grid_loc -> active_cells

SimPEG/utils/mesh_utils.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 
 from discretize.utils import (  # noqa: F401
+    random_model,
```
jcapriot, line 4: Any reason to add the random model import from `discretize` here?

SimPEG/utils/__init__.py
```diff
@@ -104,6 +104,7 @@
 .. autosummary::
   :toctree: generated/
 
+  random_model
```
jcapriot, line 107: I don't think this was imported anywhere in the __init__.py

SimPEG/utils/__init__.py
dccowan: I agree that it's weird to import the same things from two places. If it's a discretize function, it should be imported from discretize. And if it's SimPEG functionality, it should be imported from SimPEG.

SimPEG/utils/mesh_utils.py
dccowan: I can't see why. I'm happy to remove it

dccowan: I would also consider moving it from discretize to SimPEG

dccowan: This can be done in a separate PR

domfournier:
Well done gang. Thanks for this.

PR #1173, opened 2023-01-27
Mira review beta estimator

dccowan:


SimPEG/directives/directives.py
```diff
@@ -221,19 +332,93 @@ def validate(self):
         return True
 
 
-class BetaEstimate_ByEig(InversionDirective):
+class SetInitialBeta(InversionDirective):
```
jcapriot, line 335: I don’t think we need a full fledged class for this option, as you can just set it in the inversion problem.

dccowan: I'm sure the ability to set beta0 when defining the inverse problem has existed for a while now (since it is a property of the inverse problem). But in the past, it was inherently obvious where one could set this parameter directly. I'd like to keep it if you don't mind.

jcapriot: I don't want to add another way of doing something that is already possible.

dccowan: If that's the case, we need to do a better job of highlighting it in tutorials. I'm fairly confident that a majority of new users will assume the starting beta is set exclusively through directives. The problem should go away once improvements are made to the API.

jcapriot: If you add this class, I'm sure it would also be useful to add a tutorial showing use of it, so either way the documentation should be expanded. It should be better documented that there is (and always has been) a way to set beta without using a directive.

Thus I would not want to add a directive that literally just could be accomplished by setting a property\kwarg of `BaseInvProblem`.

dccowan: Sure. That's fine with me.

jcapriot:
@domfournier This looks fine to me, I know we had discussed how the older "beta estimator" was the same as the current with `n_pwr_iter=1`. It my comparison this seemed to be true.

SimPEG/directives/directives.py
```diff
@@ -76,7 +93,13 @@ def verbose(self, value):
 
     @property
     def inversion(self):
-        """This is the inversion of the InversionDirective instance."""
+        """Inversion object associated with the directive
```
domfournier, line 96: Should add punctuation everywhere `.`

domfournier (approved):
I really have not much to say other than missing periods. Looks great.

PR #1189, opened 2023-03-17
Add issue forms for github

jcapriot:
This adds Issue templates using fillable forms. It is modeled after the forms numpy uses for their issues.

These are (mostly complete) starting points, if there's anything else I missed though let me know.

For approximate look at what the forms will look like, check them out here: https://github.com/jcapriot/simpeg/tree/issue_forms/.github/ISSUE_TEMPLATE

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -6,7 +6,7 @@ body:
 - type: markdown
   attributes:
     value: >
-      If you'd like to request a new feature in SimPEG, or suggested changed the
+      If you'd like to request a new feature in SimPEG, or suggest changes in the
       functionality of certain functions, we recommend getting in touch with the
       developers on [slack](https://slack.simpeg.xyz), in addition to opening an
       issue or pull request here.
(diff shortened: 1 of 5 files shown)


.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,21 @@
+name: Feature request
+description: If you have a feature request, or an enhancement suggestion let us know.
+title: "ENH: <Please write a comprehensive title after the 'ENH: ' prefix>"
+
+body:
+- type: markdown
+  attributes:
+    value: >
+      If you'd like to request a new feature in SimPEG, or suggested changed the
```
santisoler, line 9:
```suggestion
      If you'd like to request a new feature in SimPEG, or suggest changes in the
```

santisoler (changes requested):
These are great! I just left a minor correction to one of the texts.

For the bug report, I found very useful to ask them to put the full content of `conda list`, so I can have an idea of what versions of the packages they have installed. Since the list is usually very big, we hide it behind a `<details>` tag to easily fold/unfold it when needed. Something like:

```
<details>
<summary>Output of conda list</summary>
<pre>

PASTE OUTPUT OF CONDA LIST HERE

</pre>
</details>
```
I don't know if there might be a way to put the output of a form inside it. But if not, we can ask for a `markdown` form and include that snippet as the `value`.

jcapriot:
I had put in there to paste the output of SimPEG.Report() which will get the versions of all of SimPEG's dependencies for display in a shorter format (it uses Scooby though which is not an explicit requirement of SimPEG). Despite most of our recommendations, a decent amount of people still `pip install simpeg`, so they might not be in a conda environment either.

PR #1190, opened 2023-03-20
Do not store real lambdas as complex

jcapriot:
Storing reals as complex lead to a tanh overflow... also save a lot of time not computing tanh on complex values.

At some point in the future, the 1D DC resistivity simulator should be brought up to the same level as the 1D em functionality.


jcapriot:
Well, this identified more bugs than I was expecting!

jcapriot:
This will also account for the tanh overflow warnings identified in #1052.

PR #1192, opened 2023-03-22
Improve docstring of depth_weighting

santisoler:
Add `depth_weighting` and `surface2ind_topo` to the toctree so they show up in the API reference.

### DIFF
--- a/SimPEG/utils/model_utils.py
+++ b/SimPEG/utils/model_utils.py
@@ -100,42 +100,64 @@ def surface_layer_index(mesh, topo, index=0):
 def depth_weighting(
     mesh, reference_locs, active_cells=None, exponent=2.0, threshold=None, **kwargs
 ):
-    """A simple depth weighting function
+    r"""
+    Construct diagonal elements of a depth weighting matrix
 
-    This function is a simple form of depth weighting based off of the vertical distance
-    of mesh cell centers from the reference location(s).
-
-    This is commonly used to counteract the natural decay of potential field data at
-    depth.
+    Builds the model weights following the depth weighting strategy, a method
+    to generate weights based on the vertical distance between mesh cell
+    centers and some reference location(s).
+    Use these weights in regularizations to counteract the natural decay of
+    potential field data with depth.
 
     Parameters
     ----------
-    mesh : discretize.base.BaseMesh
-        discretize model space.
-    reference_locs : float or (n, dim) numpy.ndarray
-        the reference values for top of the points
-    active_cells : (mesh.n_cells) numpy.ndarray of bool, optional
-        index vector for the active cells on the mesh.
-        A value of ``None`` implies every cell is active.
+    mesh : :class:`discretize.base.BaseMesh`
+        Discretized model space.
+    reference_locs : float or 2d-array
+        Reference location for the depth weighting.
+        It can be a ``float``, which value is the vertical component for
+        the reference location.
+        Or it can be a 2d array, with multiple reference locations, where each
+        row should contain the coordinates of a single location point in the
+        following order: _x, _y, _z.
+        The vertical coordinate of the reference location for each cell in the
+        mesh will be obtained by the closest point in ``reference_locs`` using
+        only their horizontal coordinates.
+    active_cells : :meth:`discretize.base.BaseMesh.n_cells`, array of bools, optional
+        Index vector for the active cells on the mesh.
+        If ``None``, every cell will be assumed to be active.
+        Default ``None``.
     exponent : float, optional
-        exponent parameter for depth weighting.
-    threshold : float, optional
-        The default value is half of the smallest cell width.
+        Exponent parameter for depth weighting.
+        The exponent should match the natural decay power of the potential
+        field. For example, for gravity acceleration, set it to 2; for magnetic
+        fields, to 3.
+        Default ``2``.
+    threshold : float or None, optional
+        Threshold parameters used in the depth weighting.
+        If ``None``, it will be set to half of the smallest cell width.
+        Default ``None``.
 
     Returns
     -------
-    (n_active) numpy.ndarray
-        Normalized depth weights for the mesh, at every active cell.
+    wz : 1d-array
+        Normalized depth weights for the mesh at every active cell.
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
     Parameters
     ----------
-    mesh : discretize.base.BaseMesh
-        discretize model space.
-    reference_locs : float or (n, dim) numpy.ndarray
-        the reference values for top of the points
-    active_cells : (mesh.n_cells) numpy.ndarray of bool, optional
-        index vector for the active cells on the mesh.
-        A value of ``None`` implies every cell is active.
+    mesh : :class:`discretize.base.BaseMesh`
```
jcapriot, line 114: you don't need the :class: or :meth: rst decorators here for numpy docstrings. It will automatically put those in for you.

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
+    threshold : float or None, optional
+        Threshold parameters used in the depth weighting.
+        If ``None``, it will be set to half of the smallest cell width.
+        Default ``None``.
 
     Returns
     -------
-    (n_active) numpy.ndarray
-        Normalized depth weights for the mesh, at every active cell.
+    wz : 1d-array
```
jcapriot, line 143: The returned class is a `numpy.ndarray`.

so it would be:
`wz : numpy.ndarray`

Also naming the output is just fine, particularly here when you refer to the output in the notes (but in general, when the function has only a single output, it is redundant).

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
+        Reference location for the depth weighting.
+        It can be a ``float``, which value is the vertical component for
+        the reference location.
+        Or it can be a 2d array, with multiple reference locations, where each
+        row should contain the coordinates of a single location point in the
+        following order: _x, _y, _z.
+        The vertical coordinate of the reference location for each cell in the
+        mesh will be obtained by the closest point in ``reference_locs`` using
+        only their horizontal coordinates.
+    active_cells : :meth:`discretize.base.BaseMesh.n_cells`, array of bools, optional
```
jcapriot, line 126: I think `mesh.n_cells` is sufficient as the mesh type is already described above. Also `numpy.ndarray`.

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
+        Default ``None``.
     exponent : float, optional
-        exponent parameter for depth weighting.
-    threshold : float, optional
-        The default value is half of the smallest cell width.
+        Exponent parameter for depth weighting.
+        The exponent should match the natural decay power of the potential
+        field. For example, for gravity acceleration, set it to 2; for magnetic
+        fields, to 3.
+        Default ``2``.
```
jcapriot, line 135: Again, the default is listed in the call signature, no need to repeat it.

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
+    threshold : float or None, optional
+        Threshold parameters used in the depth weighting.
+        If ``None``, it will be set to half of the smallest cell width.
+        Default ``None``.
```
jcapriot, line 139: No need to repeat the default value from the call signature.

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,64 @@ def ismember(a, b):
...
-        A value of ``None`` implies every cell is active.
+    mesh : :class:`discretize.base.BaseMesh`
+        Discretized model space.
+    reference_locs : float or 2d-array
+        Reference location for the depth weighting.
+        It can be a ``float``, which value is the vertical component for
+        the reference location.
+        Or it can be a 2d array, with multiple reference locations, where each
+        row should contain the coordinates of a single location point in the
+        following order: _x, _y, _z.
```
jcapriot, line 122: It also works on 2D problems as well, (x, y)

jcapriot (changes requested):
@santisoler, thanks for the additional detail in the parameter descriptions and the notes section for these. Most of my comments relate to the formatting of the docstring itself.

santisoler:
Thanks for those comments @jcapriot. I've just pushed changes after them. Let me know what do you think.

SimPEG/utils/model_utils.py
```diff
@@ -100,42 +100,63 @@ def ismember(a, b):
...
     Notes
     -----
-    When *reference_locs* is a single value the function is defined as,
+    Each diagonal term of the matrix is defined as:
+
+    .. math::
+
+        w(z) = \frac{1}{(|z - z_0| + \epsilon) ^ {\nu / 2}}
 
-    >>> wz = (np.abs(mesh.cell_centers[:, -1] - reference_locs) + threshold) ** (-0.5 * exponent)
```
jcapriot, line 134: I'm still not sure which of these I like better. The code version is more readable from a text perspective in the docstring itself, but the math portion looks better when rendered to html.

santisoler: I see. The issue with the Python version is that it could be misleading. The `reference_locs` is defined as a float or as a 2d-array with a different shape than the `mesh.cell_centers[:, -1]`. This can confuse the user.

 I think the math version is clearer in that sense: it mathematically defines what the weights should be, while the function implements it, even considering more complex scenarios (`reference_locs` as an array for example).

jcapriot: even more confusing is that there is a hidden functionality to this function as well. If you wanted to, you could create a `reference_locs` array that was 1D with the same length as the number of model cells and that would also be a valid call.

santisoler: I agree with that. Finding the KDTree in the function was a surprise, considering that the docstring doesn't mention it.

The conservative decision would be to simplify this function, so `reference_locs` should be a float or a 1d-array with the reference coordinates for each mesh cell (same size as active cells). But I understand that would break a lot of backward compatibility.

Were you thinking about changing this function? (we can discuss it in another issue so we don't derail this PR).

PR #1196, opened 2023-03-30
Fix check for reference_model in PGI regularization

santisoler:
Fix the checks for the `reference_model` attribute in the `PGIsmallness` regularization class. Replace the old `"mref"` for `"reference_model"`.


jcapriot:
Thanks for catching these items marked for removal in the future, they wouldn’t have been caught if the deprecated name was removed!!!

PR #1199, opened 2023-04-03
Dask MetaSim

jcapriot:
Adds the `dask` version of the `MetaSimulation`.

Essentially it never explicitly sends around fields objects, and it avoids sending around simulations (after they have gone through the setter). Every operation that involves one of the internal simulations is explicitly done on the worker that owns that simulation. The `fields` returned by this class are actually `Future`s that live on the worker that owns the respective simulation.

Most items that live on this class are now `Future`s of the respective item in the `MetaSimulation` class ( e.g. `sim.mappings` are now a list of `Future` maps instead of a `list` of maps). As a side effect, it is actually possible to initialize the class with lists of Futures. This is useful if you want to avoid creating everything on the main thread and then serializing and sending the result out to each worker, instead you can create a function which returns an object and send that function to each worker.

I've tried to add safeguards to the setters to safeguard the use to check the outputs of the Future to ensure they are Futures of the correct types.

- [x] Add unit tests
- [ ] Apply to some larger problems to identify choke points


JKutt (approved):
Works well Joe! I managed to get it working with SSH cluster and it executes on the SLURM. I think I may still have some configuration issues but the implementation itself does what it should, its the SLURM side that seems off in my test. I did a short test with the Kubernetes before I left my last position so I believe its good there, that was some time ago though. I don't think too much changed though.

PR #1200, opened 2023-04-03
Fix flake8 `B015` errors: solve unused comparisons

santisoler:
Fix flake8 `B015` error that highlights unused comparisons. Remove `B015` from the list of ignored warnings in `Makefile`.

### DIFF
--- a/SimPEG/electromagnetics/static/resistivity/simulation.py
+++ b/SimPEG/electromagnetics/static/resistivity/simulation.py
@@ -489,7 +489,7 @@ class Simulation3DNodal(BaseDCSimulation):
         if mesh._meshType == "TREE":
             mesh.nodal_gradient
         elif mesh._meshType == "CYL":
-            bc_type == "Neumann"
+            bc_type = "Neumann"
         self.bc_type = bc_type
         self.setBC()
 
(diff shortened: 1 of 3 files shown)


SimPEG/electromagnetics/static/resistivity/simulation.py
```diff
@@ -489,7 +489,7 @@ def __init__(self, mesh, survey=None, bc_type="Robin", **kwargs):
         if mesh._meshType == "TREE":
             mesh.nodal_gradient
         elif mesh._meshType == "CYL":
-            bc_type == "Neumann"
+            bc_type = "Neumann"
```
santisoler, line 492: I assumed that we are intending to change the `bc_type` in this line. Correct me if I'm wrong.

tests/base/test_maps.py
```diff
@@ -560,7 +560,7 @@ def test_basic(self):
 
         named_model = wires * model
 
-        named_model.sigma == model[: mesh.shape_cells[2]]
+        assert named_model.sigma == model[: mesh.shape_cells[2]]
```
santisoler, line 563: Here I assumed that an assertion was missing, is it right?

jcapriot: I think that's correct, but I think this line (and the line a little below) need to be changed to `np.testing.assert_equal`(...)

jcapriot (changes requested):
Think it just needs to use `np.testing.assert_equal` for the check.

PR #1201, opened 2023-04-03
Multiprocessing MetaSimulation

jcapriot:
Adds a Multiprocessing version of the `MetaSimulation`

This PR is currently in a lesser state than the `dask` one, but it should effectively work the same. It's nice if the user wants to not install dask, or just work on a single machine, in which case this is often faster. There are several limitations though, including not being able to ever directly return the created fields objects (s side effect of us storing simulations on fields and not being able to serialize a simulation that has factors stored on it).

- [x] add unit tests
- [x] add the other two `MetaSims`


JKutt (approved):
I really like how you use and setup the queues. Nicely done, great stuff!

PR #1221, opened 2023-05-05
Refactor `PGI_UpdateParameters` directive

santisoler:
#### Summary

Refactor the `PGI_UpdateParameters` class to improve readability. Add
a constructor to the class that initializes all the attributes used by its
methods, and improve its docstring adding descriptions for each one of the
parameters of the constructor.

#### PR Checklist
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/basic/practices.html#style).
* [ ] Added [tests](https://docs.simpeg.xyz/content/basic/practices.html#testing) to verify changes to the code.
* [ ] Marked as ready for review (ff this is was a draft PR), and converted
      to a Pull Request

Add an `__init__` method with a docstring to the `PGI_UpdateParameters` directive, including descriptions for each parameter of the class. Define a property for the PGI regularization in the inversion. Remove the `initialize` method since it's not needed anymore. Simplify the logic in `endIter` method.

### DIFF
--- a/SimPEG/directives/pgi_directives.py
+++ b/SimPEG/directives/pgi_directives.py
@@ -54,7 +54,8 @@ class PGI_UpdateParameters(InversionDirective):
         the covariances. If False, update the precisions. Ignored if
         ``update_gmm`` is False.
     fixed_membership : None or array_like, optional
-        Keep the membership of specific cells fixed.
+        Array with the indices and membership of cells. Use when the membership
+        of some cells is intended to be fixed.
     verbose : bool, optional
         If True, messages will be printed to stdout during each iteration.
 
--- a/SimPEG/directives/pgi_directives.py
+++ b/SimPEG/directives/pgi_directives.py
@@ -11,7 +11,6 @@ import numpy as np
 from ..directives import InversionDirective, MultiTargetMisfits
 from ..regularization import (
     PGI,
-    PGIsmallness,
     PGIwithRelationships,
     SmoothnessFirstOrder,
     SparseSmoothness,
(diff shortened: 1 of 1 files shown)


SimPEG/directives/pgi_directives.py
```diff
@@ -27,108 +26,142 @@
...
+                precisions_init=self.pgi_regularization.gmm.precisions_,
+                random_state=self.pgi_regularization.gmm.random_state,
+                tol=self.pgi_regularization.gmm.tol,
+                verbose_interval=self.pgi_regularization.gmm.verbose_interval,
+                warm_start=self.pgi_regularization.gmm.warm_start,
+                fixed_membership=self.fixed_membership,
+            )
+            new_gmm = new_gmm.fit(model)
+        else:
+            new_gmm = copy.deepcopy(gmmref)
```
santisoler, line 137: I don't see the point of  creating a deep copy of the GMM after each iteration in case that the GMM is not intended to be updated. It's a waste of memory and time. Is there any reason why we should be doing it that I'm not seeing?

SimPEG/directives/pgi_directives.py
```diff
@@ -27,108 +26,142 @@
...
+
+        # Update the GMM in the PGI regularization
+        self.pgi_regularization.gmm = new_gmm
+        # Compute membership of the current model
+        membership = self.pgi_regularization.gmm.predict(model)
+        # Override the membership if some cells have fixed membership
+        if self.fixed_membership is not None:
+            membership[self.fixed_membership[:, 0]] = self.fixed_membership[:, 1]
```
santisoler, line 145: After updating the GMM, I saw that the membership number could change. I mean that a rock unit associated with a membership `0` in the reference GMM, could be related to the membership `1` in the updated GMM.
Therefore, the membership indices in the `fixed_membership` would be wrong and produce an unwanted result.

I think we should test and fix this.

SimPEG/directives/pgi_directives.py
```diff
@@ -27,108 +26,142 @@
...
+        # Get the new reference model
+        mref = mkvc(self.pgi_regularization.gmm.means_[membership])
+        # Update the reference model in the PGI regularization
+        self.pgi_regularization.reference_model = mref
+        if getattr(self.fixed_membership, "shape", [0, 0])[0] < len(membership):
+            self.pgi_regularization._r_second_deriv = None
```
santisoler, line 151: I'm not entirely sure what's the purpose of these lines... I'll keep looking.

santisoler (commented):
There are a few places for improvement. I'm leaving some comments below so we can discuss about them.

SimPEG/directives/pgi_directives.py
thibaut-kobold: @santisoler : it is to save the derivatives of the polynomial mapping such as shown in this example: https://docs.simpeg.xyz/content/examples/10-pgi/plot_inv_1_PGI_Linear_1D_joint_WithRelationships.html#sphx-glr-content-examples-10-pgi-plot-inv-1-pgi-linear-1d-joint-withrelationships-py

The Polynom class takes for some unknown reason a very long time to compute derivatives. Thus, instead of computing them on-call, these hidden properties saves them.

SimPEG/directives/pgi_directives.py
thibaut-kobold: I think I had funky behaviors back in the days when not creating a new object. But sklearn has deeply changed since so it might not be necessary anymore. Try it :)

SimPEG/directives/pgi_directives.py
thibaut-kobold: > After updating the GMM, I saw that the membership number could change. I mean that a rock unit associated with a membership 0 in the reference GMM, could have a the membership 1 in the updated GMM.

This is supposed to be dealt with by the `order_cluster` function. But it might have some mild assumptions like "the highest proportions is for the background". But in my tests it was very robust. See for example my thesis, Figure 6.5, where a range of proportions are tried, with even some cases where the background weight is very very small.

PR #1228, opened 2023-05-27
remove the BaseSimPEG class

lheagy:
remove the unnecessary inheritance of the `BaseSimPEG` class


lheagy:
After chatting with @jcapriot about this, I will close it. This class may later be used for serializing SimPEG objects

PR #1229, opened 2023-05-31
Generalize __add__  for any ComboObjectiveFunction

santisoler:
#### Summary

Rewrite the `__add__` method of `BaseObjectiveFunction` so by default it performs the same task for any `ComboObjectiveFunction` or any of its child classes. Add a new `unpack_on_add` attribute to `ComboObjectiveFunction` to decide weather to unpack its objective functions after an addition, or if it should be added as a whole. Set `unpack_on_add=True` by default on `ComboObjectiveFunction`, but set it to `False` for regularization classes, like `WeightedLeastSquares` and `PGI`. Add tests for the new changes.

#### PR Checklist
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/basic/practices.html#style).
* [ ] Added [tests](https://docs.simpeg.xyz/content/basic/practices.html#testing) to verify changes to the code.
* [ ] Marked as ready for review (ff this is was a draft PR), and converted
      to a Pull Request

Fixes an unexpected behaviour when subclassing `ComboObjectiveFunction`: by default its `__add__` method would have worked differently than it does for the parent, adding the instance as a whole during the addition. Instead, the `ComboObjectiveFunction` gets unpacked, and every objective function in its `objfcts` attribute is directly included in the addition.

### DIFF
--- a/SimPEG/objective_function.py
+++ b/SimPEG/objective_function.py
@@ -132,7 +132,7 @@ class BaseObjectiveFunction(BaseSimPEG):
             num=num,
             expectedOrder=expectedOrder,
             plotIt=plotIt,
-            **kwargs
+            **kwargs,
         )
 
     def test(self, x=None, num=4, plotIt=False, **kwargs):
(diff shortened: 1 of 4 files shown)


jcapriot:
@santisoler Thanks! This looks really good. Once there's good coverage and the tests are passing I'll merge this in.

santisoler:
Thanks @jcapriot! Spoiler alert: one of the tests will fail. While coding this PR I found an inconsistency when we try to sum a combo directly vs when we want to sum a combo multiplied by a scalar.

For example, if we run this:

```python
import discretize
from SimPEG import regularization

mesh = discretize.TensorMesh([5, 5, 5])

reg_1 = regularization.Smallness(mesh)
reg_2 = regularization.SmoothnessFirstOrder(mesh)
combo_1 = reg_1 + reg_2

reg_3 = regularization.SmoothnessSecondOrder(mesh)
combo_2 = reg_3 + combo_1
print(combo_2.objfcts)
```
We get:
```
[<SimPEG.regularization.base.SmoothnessSecondOrder object at 0x7ff156a06c80>, <SimPEG.regularization.base.Smallness object at 0x7ff19ba23d30>, <SimPEG.regularization.base.SmoothnessFirstOrder object at 0x7ff19ba23ca0>]
```

But, if we multiply that `combo_1` by a scalar before adding it, the multiplication defines a new Combo that only has the `combo_1` as `objfcts`:
```python
combo_2 = reg_3 + 1 * combo_1
print(combo_2.objfcts)
```
we get:
```
[<SimPEG.regularization.base.SmoothnessSecondOrder object at 0x7ff156a06c80>, <SimPEG.objective_function.ComboObjectiveFunction object at 0x7ff156a06920>]
```

And:
```python
print(combo_2.objfcts[1].objfcts)
```
```
[<SimPEG.regularization.base.Smallness object at 0x7fc5a452bd30>, <SimPEG.regularization.base.SmoothnessFirstOrder object at 0x7fc5a452bca0>]
```

In one of the new tests this behaviour is exposed, and therefore the test fails.

I drafted a solution in the `__mul__` method, to modify the multipliers of the objective function (if it's a `ComboObjectiveFunction`), instead of creating a new Combo on top. But that breaks quite a lot of stuff for the regularizations. I'm curious to hear your thoughts on that.

santisoler:
We discussed this with @jcapriot today, and decided that many objective functions rely on this behaviour for the multiplication, so we won't change it and adjust the tests so they check this behaviour.

PR #1231, opened 2023-06-05
Fix IP simulation / inversion with SimPEG.dask

thibaut-kobold:
#### Summary
`fields` needs to be called first to set `_scale` for IP simulation. It is the case without dask by calling `getJ` at the beginning of the `getJtJdiag` function, but not with the dask function `dask_getJtJdiag`.

#### PR Checklist
* [x] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/basic/practices.html#style).
* [x] Added relevant method tags (i.e. `GRAV`, `EM`, etc.) (cannot do it as not a team member)
      to a Pull Request 
* [x] Tagged ``@simpeg/simpeg-developers`` when ready for review. (cannot do it as not a team member)

#### Additional information
You can test it by adding `from SimPEG import dask` to the IP tutorial: https://docs.simpeg.xyz/content/tutorials/06-ip/plot_inv_2_dcip2d.html

### DIFF
--- a/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
@@ -1,8 +1,8 @@
+import dask.array as da
+
 from .....electromagnetics.static.induced_polarization.simulation import (
     BaseIPSimulation as Sim,
 )
-
-import dask.array as da
 
 
 def dask_getJtJdiag(self, m, W=None, f=None):
@@ -13,9 +13,9 @@ def dask_getJtJdiag(self, m, W=None, f=None):
         J = self.getJ(m, f=f)
         # Need to check if multiplying weights makes sense
         if W is None:
-            W = self._scale
+            W = self.scale
         else:
-            W = self._scale * W.diagonal()
+            W = self.scale * W.diagonal()
         w = da.from_array(W)[:, None]
         self._gtgdiag = da.sum((w * J) ** 2, axis=0).compute()
 
--- a/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
@@ -16,7 +16,7 @@ def dask_getJtJdiag(self, m, W=None, f=None):
             W = self._scale
         else:
             W = self._scale * W.diagonal()
-        self._gtgdiag = da.einsum("i,ij,ij->j", W**2.0, J, axis=0).compute()
+        self._gtgdiag = da.einsum("i,ij,ij->j", W**2, J, J).compute()
 
     return self._gtgdiag
 
--- a/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
@@ -13,11 +13,10 @@ def dask_getJtJdiag(self, m, W=None, f=None):
         J = self.getJ(m, f=f)
         # Need to check if multiplying weights makes sense
         if W is None:
-            W = self.scale
+            W = self._scale
         else:
-            W = self.scale * W.diagonal()
-        w = da.from_array(W)[:, None]
-        self._gtgdiag = da.sum((w * J) ** 2, axis=0).compute()
+            W = self._scale * W.diagonal()
+        self._gtgdiag = da.einsum("i,ij,ij->j", W**2.0, J, axis=0).compute()
 
     return self._gtgdiag
 
--- a/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
+++ b/SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
@@ -1,8 +1,6 @@
-from .....electromagnetics.static.induced_polarization.simulation import (
-    BaseIPSimulation as Sim,
-)
-
 import dask.array as da
+
+from .....electromagnetics.static.induced_polarization.simulation import BaseIPSimulation as Sim
 
 
 def dask_getJtJdiag(self, m, W=None, f=None):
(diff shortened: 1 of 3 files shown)


domfournier (commented):
I did have to patch that on our side too. This fix is sort of addressing the issue, but not very transparent. I would vote that we make the `_scale` a @property like `_f`, and just make it compute if None instead.

thibaut-kobold:
I added a test that should case the issue in the future. Because the test called `make_synthetic_data` in the setup, it hid the issue.

thibaut-kobold:
make `scale` a property as per @domfournier 's suggestions

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -1,22 +1,23 @@
...
         # Need to check if multiplying weights makes sense
         if W is None:
-            W = self._scale
+            W = self.scale
         else:
-            W = self._scale * W.diagonal()
+            W = self.scale * W.diagonal()
         w = da.from_array(W)[:, None]
-        self._gtgdiag = da.sum((w * self.getJ(m, f=f)) ** 2, axis=0).compute()
+        self._gtgdiag = da.sum((w * J) ** 2, axis=0).compute()
```
domfournier, line 20:
```suggestion
        self._gtgdiag = da.einsum('i,ij,ij->j', W**2., self.Jmatrix, self.Jmatrix)
```

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -1,22 +1,23 @@
...
 
-import dask.array as da
-
 
 def dask_getJtJdiag(self, m, W=None, f=None):
     """
     Return the diagonal of JtJ
     """
     if getattr(self, "_gtgdiag", None) is None:
+        J = self.getJ(m, f=f)
```
domfournier, line 13:
```suggestion
```

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
domfournier: Not sure if the current setup knows about Jmatrix, but at least einsum should be much faster than `(W * J) ** 2.`

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -40,6 +39,25 @@ def sigmaDeriv(self):
     def rhoDeriv(self):
         return sp.diags(self.rho) @ self.etaDeriv
 
+    @property
```
jcapriot, line 42: do not make this a visible property to the user, it should be hidden, likely just need to make sure fields are called before a getJ

jcapriot:
The only change I would make to solve this problem, is to move the getJ call in the dask function to be where it is in the non-dask version.

thibaut-kobold:
> The only change I would make to solve this problem, is to move the getJ call in the dask function to be where it is in the non-dask version.

The thing I notice too is that the issue is also present when calling `Jvec` and `Jtvec` so it is more than a dask version issue. `@property` solves it all.

thibaut-kobold:
proof
Screenshot 2023-06-05 at 2 03 54 PM

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
thibaut-kobold: keep J as there is no Jmatrix I believe in vanilla SimPEG

jcapriot:
Ah yes, I see what's going on. I like the cached property now :)

SimPEG/dask/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -1,22 +1,22 @@
...
         # Need to check if multiplying weights makes sense
         if W is None:
             W = self._scale
         else:
             W = self._scale * W.diagonal()
-        w = da.from_array(W)[:, None]
-        self._gtgdiag = da.sum((w * self.getJ(m, f=f)) ** 2, axis=0).compute()
+        self._gtgdiag = da.einsum("i,ij,ij->j", W**2.0, J, axis=0).compute()
```
jcapriot, line 19: not quite what needs to happen,

`einsum("i,ij,ij->j", W**2, J, J).compute()`

tests/dask/test_IP_jvecjtvecadj_dask.py
```diff
@@ -1,23 +1,137 @@
...
+)
 from SimPEG.electromagnetics import induced_polarization as ip
-import shutil
+from SimPEG.electromagnetics import resistivity as dc
+from SimPEG.utils.io_utils.io_utils_electromagnetics import read_dcip2d_ubc
 
 np.random.seed(30)
 
 
+class IPProblemTests2DN(unittest.TestCase):
```
jcapriot, line 27: I know having more tests is good (but the testing sweet is already quite large and time consuming), What specifically does this test that is not already tested?

thibaut-kobold: I should put it forward higher (comment on line 88 right now)

- The goal is to make a test without calling` make_synthetic_data` first to simulate a real data case. 

- `make_synthetic_data` was the reason why the issue being fixed here was not detected by the tests.

- This test is built by taking the already existing 2D files used in the [IP2D tutorial](https://docs.simpeg.xyz/content/tutorials/06-ip/plot_inv_2_dcip2d.html), with a much smaller mesh (1600 cells).

- I also did not see any test for IP2D + Dask (all IP tests with Dask are for the 3D case). This added test covers this combination.

thibaut-kobold: added some docstring / comments

jcapriot: I think more so the underlying issue was that the simulation failed when not explicitly calling `fields` before any `Jvec`, `JTvec` and in the dask case `getJTJdiag`.

SimPEG/electromagnetics/static/induced_polarization/simulation.py
```diff
@@ -65,27 +83,10 @@ def __init__(
 
     _Jmatrix = None
     _pred = None
-    _scale = None
 
     def fields(self, m):
         if self.verbose:
             print(">> Compute DC fields")
-        if self._f is None:
```
jcapriot, line 73: I think you need to leave these 3 removed lines in (73-76) as it will never set `self._f` if it was `None` and the user repeatedly calls `fields` (you can theoretically call `fields` without ever calling anything else).

thibaut-kobold: good catch. fixed

tests/dask/test_IP_jvecjtvecadj_dask.py
thibaut-kobold: Indeed, that is the core of the issue

PR #1237, opened 2023-06-15
Anisotropy derivative support

### LINKED ISSUE #146: Full Anisotropy
_From @lheagy on June 19, 2015 18:41_

At the moment, MeSigmaIDeriv, MfRhoDeriv, MfRhoDeriv will not take fully anisotropic media - this actually needs to be taken care of inside of getFaceInnerProductDeriv and getEdgeInnerProductDeriv within simpeg, but I am creating the issue here too so it is documented that the fully anisotropic inversion won't be working yet

_Copied from original issue: simpeg/simpegem#22_

jcapriot:
#### Summary
Adds support for taking derivatives of anisotropic property models

#### PR Checklist
* [x] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/basic/practices.html#style).
* [x] Added [tests](https://docs.simpeg.xyz/content/basic/practices.html#testing) to verify changes to the code.
* [x] Added relevant method tags (i.e. `GRAV`, `EM`, etc.)
      to a Pull Request

#### Reference issue
closes #146

This adds derivative support for anisotropic physical property models to the edge and face mass matrices. As a side effect, this also fixed a bug with derivatives on non-axis aligned meshes that used those mass matrices.

#### Additional information
Previously we were doing a simplification on the derivatives which was only valid in axis-aligned anisotropic cases with axis-aligned meshes (`Tensor`, `Cylindrical`, and `TreeMesh`). This is because the evaluation of `get_edge_inner_product_deriv` in those cases we are able to evaluate the derivative of $M_{e,m} u$ by $diag(u) \frac{d M_{e,m} * ones}{d m}$ because $M_e$ was diagonal. However in the full anisotropic case, and non-axis aligned meshes, this no longer applies, and we must instead actually evaluate the derivative function for each different `u` the proper way as $\frac{d M_{e, m} u}{d m}$.


SimPEG/base/pde_simulation.py
```diff
@@ -1,41 +1,72 @@
...
+            if u.ndim > 1:
+                UM = sp.vstack([sp.diags(u[:, i]) @ M for i in range(u.shape[1])])
+            else:
+                U = sp.diags(u, format="csr")
+                UM = U @ M
+            if adjoint:
+                return UM.T
+            return UM
     else:
+        # assume it was a tuple of M_func, prop_deriv
```
lheagy, line 41: would it be clearer if in the first check we looked to see if M is a tuple? (or list) here and then error if we get something unexpected?

SimPEG/base/pde_simulation.py
```diff
@@ -1,41 +1,72 @@
...
+        M_func, prop_deriv = M
         if u.ndim > 1:
-            UM = sp.vstack([sp.diags(u[:, i]) @ M for i in range(u.shape[1])])
+            Mu = [M_func(u[:, i]) for i in range(u.shape[1])]
+            if v is None:
+                Mu = sp.vstack([M @ prop_deriv for M in Mu])
+                if adjoint:
+                    Mu = Mu.T
+                return Mu
```
lheagy, line 49: Would it be simpler if rather than re-assigning `Mu` we just return? e.g. 
```suggestion
                Mu = sp.vstack([M @ prop_deriv for M in Mu])
                if adjoint:
                    return Mu.T
                return Mu
```

I know it is equivalent, but it is potentially confusing to set the matrix equal to its transpose

SimPEG/base/pde_simulation.py
```diff
@@ -1,41 +1,72 @@
...
-            U = sp.diags(u, format="csr")
-            UM = U @ M
-        if adjoint:
-            return UM.T
-        return UM
+            Mu = M_func(u)
+            if v is None:
+                Mu = Mu @ prop_deriv
+                if adjoint:
+                    Mu = Mu.T
```
lheagy, line 63: same comment as above about returning rather than re-assigning
```suggestion
                    return Mu.T
```

SimPEG/base/pde_simulation.py
```diff
@@ -232,10 +263,21 @@ def MfDeriv_prop(self, u, v=None, adjoint=False):
...
             stash_name = f"_Mf_{arg}_deriv"
             if getattr(self, stash_name, None) is None:
-                M_prop_deriv = self.mesh.get_face_inner_product_deriv(
-                    np.ones(self.mesh.n_cells)
-                )(np.ones(self.mesh.n_faces)) * getattr(self, f"{arg.lower()}Deriv")
-                setattr(self, stash_name, M_prop_deriv)
+                prop = getattr(self, arg.lower())
+                t_type = TensorType(self.mesh, prop)
+
+                M_func = self.mesh.get_face_inner_product_deriv(model=prop)
```
lheagy, line 269: a question / thought on naming... `M_func` doesn't prompt me to think about derivatives. Would `M_deriv_func` work? It is the function of getting the derivative of the face inner product matrix ?

SimPEG/base/pde_simulation.py
```diff
@@ -232,10 +263,21 @@ def MfDeriv_prop(self, u, v=None, adjoint=False):
...
+                prop_deriv = getattr(self, f"{arg.lower()}Deriv")
+                if t_type < 3 and self.mesh._meshType.lower() in (
```
lheagy, line 271: a quick comment to refresh the memory on `t_type` would be helpful here 
```suggestion
                # t_type == 3 for full tensor, t_type < 3 for diagonal anisotropy, isotropic and scalar properties
                if t_type < 3 and self.mesh._meshType.lower() in (
```

SimPEG/base/pde_simulation.py
```diff
@@ -252,10 +294,20 @@ def MeDeriv_prop(self, u, v=None, adjoint=False):
...
             stash_name = f"_Me_{arg}_deriv"
             if getattr(self, stash_name, None) is None:
-                M_prop_deriv = self.mesh.get_edge_inner_product_deriv(
-                    np.ones(self.mesh.n_cells)
-                )(np.ones(self.mesh.n_edges)) * getattr(self, f"{arg.lower()}Deriv")
-                setattr(self, stash_name, M_prop_deriv)
+                prop = getattr(self, arg.lower())
+                t_type = TensorType(self.mesh, prop)
+
+                M_func = self.mesh.get_edge_inner_product_deriv(model=prop)
```
lheagy, line 300: same comment as earlier about naming

lheagy (commented):
This looks great Joe! I am looking forward to playing with it. My comments are pretty minor, mostly trying to think about clarity. One thing that wasn't obvious to me is when we would encounter a scenario where `v.dim > 2`? I don't think we should encounter this for the forward? but maybe these is a possibility with the adjoint? (or maybe I am just missing the use case!)

jcapriot:
> This looks great Joe! I am looking forward to playing with it. My comments are pretty minor, mostly trying to think about clarity. One thing that wasn't obvious to me is when we would encounter a scenario where `v.dim > 2`? I don't think we should encounter this for the forward? but maybe these is a possibility with the adjoint? (or maybe I am just missing the use case!)

`v.dim` > 2 happens when we are doing an adjoint multiplication, with multiple field polarizations, and many `vs` (i.e. constructing the sensitivity matrix for an MT problem). So `v` in that case would be like (`v` n_m by n_d by n_pol)

lheagy:
> > This looks great Joe! I am looking forward to playing with it. My comments are pretty minor, mostly trying to think about clarity. One thing that wasn't obvious to me is when we would encounter a scenario where `v.dim > 2`? I don't think we should encounter this for the forward? but maybe these is a possibility with the adjoint? (or maybe I am just missing the use case!)
> 
> `v.dim` > 2 happens when we are doing an adjoint multiplication, with multiple field polarizations, and many `vs` (i.e. constructing the sensitivity matrix for an MT problem). So `v` in that case would be like (`v` n_m by n_d by n_pol)

That makes sense. Thanks for clarifying!

SimPEG/base/pde_simulation.py
jcapriot: Sure thing, will be helpful for future iterations to remind ourselves.

SimPEG/base/pde_simulation.py
jcapriot: I've added a more explicit check and an appropriate error, instead of just an assumption. This should hopefully only ever be an issue if a developer does something unexpected with new code.

lheagy:
💥

PR #1257, opened 2023-07-05
unpack the data misfits for plotting tikhonov curves

lheagy:
#### Summary
Small pr to fix a bug in the plotting of tikhonov curves. Since we turn the data misfit into a combo objective function, we need to unpack it to get the number of data  

#### PR Checklist
      to a Pull Request

### DIFF
--- a/SimPEG/directives/directives.py
+++ b/SimPEG/directives/directives.py
@@ -1765,7 +1765,9 @@ class SaveOutputEveryIteration(SaveEveryIteration):
         plot_small=False,
         plot_smooth=False,
     ):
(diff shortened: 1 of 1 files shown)


jcapriot:
Unfortunate that this error wasn't caught by any tests. Is there anyway we could add a test to make sure it doesn't break in the future?

lheagy:
It is a bit tricky to add a test on this one because it is only plotting code. Do you have any examples handy from discretize of tests we could add? 

(I also think this should be re-written as we revisit directives, but that is for a later date!)

jcapriot:
For starters, I guess we should make sure it at least runs without any errors?

PR #1261, opened 2023-07-12
Set storage type of pf sensitivity matrix

jcapriot:
#### Summary
Allows users to control the type of the stored sensitivity matrix.

#### PR Checklist
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/practices.html#style).
      to a Pull Request

In the non-dask version the sensitivity was being stored as a `np.float64` implicitly. This changes it to be in line with the dask version (`np.float32`), but also now allows the user to change the default if they so desire.

### DIFF
--- a/SimPEG/potential_fields/gravity/simulation.py
+++ b/SimPEG/potential_fields/gravity/simulation.py
@@ -46,7 +46,7 @@ class Simulation3DIntegral(BasePFSimulation):
             # Compute the linear operation without forming the full dense G
             fields = mkvc(self.linear_operator())
         else:
-            fields = self.G @ (self.rho).astype(np.float32)
+            fields = self.G @ (self.rho).astype(self.sensitivity_dtype)
 
         return np.asarray(fields)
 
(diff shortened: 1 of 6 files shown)


SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -46,7 +46,7 @@ def fields(self, m):
             # Compute the linear operation without forming the full dense G
             fields = mkvc(self.linear_operator())
         else:
-            fields = self.G @ (self.rho).astype(np.float32)
+            fields = self.G @ (self.rho).astype(self.sensitivity_dtype)
```
santisoler, line 49: @jcapriot, after checking [`astype` docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html), I think the method by default copies the array even if it has the same type as the given one. I think we should enforce not to copy the array with `copy=False`.
```suggestion
            fields = self.G @ (self.rho).astype(self.sensitivity_dtype, copy=False)
```

I've just tested its behaviour with this little snippet:
```python
import numpy as np
a = np.array([1, 2, 3], dtype=np.float64)
b = a.astype(np.float64)
c = a.astype(np.float64, copy=False)
```
```python
a is b
```
```
False
```
```python
a is c
```
```
True
```

jcapriot: Ah yes, thanks

PR #1276, opened 2023-08-09
added getJ for fdem and nsem simulations

JKutt:
#### Summary
A PR to bring in methods inorder to calculate sensitivity and sensitivity weights for fdem and nsem simulations with the `MetaSimulation` class.

Added methods for the FDEM simulation class are:
- `getJ`
- `getJtJdiag`

#### PR Checklist
      to a Pull Request

#### Reference issue
NA

Allows the calculation of the sensitivity matrix for fdem and nsem

#### Additional information
NA


SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -239,6 +239,85 @@ def Jtvec(self, m, v, f=None):
                 Jtv += np.real(df_dmT_sum)
 
         return mkvc(Jtv)
+    
+    def getJ(self, m, f=None):
```
domfournier, line 243: Would there be an issue putting this in the getter of Jmatrix instead?

SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -239,6 +239,85 @@ def Jtvec(self, m, v, f=None):
...
+            J = self.getJ(m, f=f)
+
+            if W is None:
+                W = np.ones(J.shape[0])
+            else:
+                W = W.diagonal() ** 2
+
+            diag = np.zeros(J.shape[1])
+            for i in range(J.shape[0]):
+                diag += (W[i]) * (J[i] * J[i])
```
domfournier, line 270: Would be faster to use `np.einsum`

SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -239,6 +239,85 @@ def Jtvec(self, m, v, f=None):
...
+        if getattr(self, "_Jmatrix", None) is None:
+            if f is None:
+                f = self.fields(m)
+            self._Jmatrix = self.compute_J()
+        return self._Jmatrix
+    
+
+    def getJtJdiag(self, m, W=None, f=None):
+        """
+            Return the diagonal of JtJ
```
domfournier, line 257:
```suggestion
            Return the diagonal of JtJ
            
            Parameters
            ------------
            m: np.ndarray
                Model values
            ...
```

domfournier (commented):
Glad to see this coming along. See comments

SimPEG/electromagnetics/frequency_domain/simulation.py
JKutt: It shouldn't be and issue so can and will do!

SimPEG/electromagnetics/frequency_domain/simulation.py
JKutt: good call! I forgot about `einsum` .

SimPEG/electromagnetics/frequency_domain/simulation.py
jcapriot: Is there a point to having a Jmatrix property? J is dependent on the passes model, so it makes sense that it is still a function.

jcapriot:
My question would be do we need another way to call getting the J function? I.e. the `compute_J` function. We should probably just leave it as a single function using what we’ve already established.

JKutt:
@jcapriot , https://github.com/JKutt/simpeg/blob/c5d54c5b1c31eea9e9e3e9b281e40d9f8b030135/SimPEG/meta/simulation.py#L312C17-L313C53

these lines don't seem to execute and the model does not get passed down to the sub simulations. Is there a place I should be setting this `self._repeat_sim`?

SimPEG/electromagnetics/frequency_domain/simulation.py
domfournier: @jcapriot So just store an hidden attribute `_Jmatrix`?

jcapriot:
> @jcapriot , https://github.com/JKutt/simpeg/blob/c5d54c5b1c31eea9e9e3e9b281e40d9f8b030135/SimPEG/meta/simulation.py#L312C17-L313C53
> 
> these lines don't seem to execute and the model does not get passed down to the sub simulations. Is there a place I should be setting this `self._repeat_sim`?

No, you shouldn't mess with that property, it is only used when there is actually a `RepeatedSimulation` (i.e. a single simulation is used repeatedly for multiple models). If you look further up in that function you linked, when it sets `self.model = m`, it should go through the `MetaSimulation`'s `model` setter, which then will echo the model to all sub simulations.

SimPEG/electromagnetics/frequency_domain/simulation.py
jcapriot: But then where would you pass the model?

what your proposing would always require you to do:
```
sim.model = m
sim.Jmatrix @ x
```

domfournier: Well Jmatrix always requires fields, and model gets set in the `fields` call. A bit hidden, but that's the common path for all inversions.

jcapriot: I think the common path for all the inversions in SimPEG is to evaluate and stash fields for a given model, then pass those to dpred, Jvec and JTvec.

The general form of those functions is to use a fields object if given, or else compute fields given the model then use them and discard the fields.

SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -240,6 +240,86 @@ def Jtvec(self, m, v, f=None):
...
+                        du_dmT = -dA_dmT
+                        
+                        if not isinstance(dRHS_dmT, Zero):
+                            du_dmT += dRHS_dmT
+                        if not isinstance(df_dmT[0], Zero):
+                            du_dmT += np.hstack(df_dmT)
+
+                        block = np.array(du_dmT, dtype=complex).real.T
+                        Jmatrix[block_count:(block_count + rx.nD)] = block
+                        block_count += rx.nD
```
jcapriot, line 291: I don't think you can use the "block count" way of doing this to reliably index the matrix for the frequency domain problem.

The `get_sources_by_frequency`  function does not necessarily return the sources in data order, it returns them grouped by common frequency. Normally this isn't an issue for magnetotellurics, however, it's certainly possible to have multiple sources with the same frequency in airborne surveys, and we cannot expect a fixed ordering of them. I think we need to make use of a `data` object from the survey here to get it into the correct rows of `J`.

`data` objects have an `index_dictionary` property that you can use for this. The general way to use it would be:

```
data = Data(self.survey)

for freq in self.survey.frequencies:
    ...
    for src in self.survey.get_sources_by_frequency(freq):
        ...
        for rx in src.receiver_list:
            ...
            row_inds = data.index_dictionary[src][rx]
            J[row_inds] = block
```

jcapriot (changes requested):
I think you'll need to change indexing the J matrix

SimPEG/electromagnetics/frequency_domain/simulation.py
JKutt: Thanks @jcapriot , good catch, I'll make use of the `data` object for this.

JKutt:
@simpeg/simpeg-developers

JKutt:
@jcapriot , @santisoler , examples are working and tests are passing! Though have a look at the `pde_simulation.py` modifications. specifically  the function `__inner_mat_mul_op`. For the adjoint, I had to add a case for when the second dimension is larger than 2. When building the full J matrix I am passing `v` as an identity matrix. Perhaps there is a more elegent way to do this? Let me know what you think.

SimPEG/base/pde_simulation.py
```diff
@@ -23,6 +23,10 @@ def __inner_mat_mul_op(M, u, v=None, adjoint=False):
                 u = u[:, None, :]
             if adjoint:
                 if u.ndim > 1 and u.shape[-1] > 1:
+                    if v.shape[-1] > 2:
+                        return M.T * (
+                            u[:, None, :] * v.reshape((u.shape[0], -1, u.shape[1]))
+                        ).sum(axis=2)
```
jcapriot, line 29: Can you tell me what input shapes you were sending to this function to make this necessary? I thought it already supported multiple polarizations already.

jcapriot (changes requested):

SimPEG/electromagnetics/frequency_domain/simulation.py
```diff
@@ -240,6 +240,84 @@ def Jtvec(self, m, v, f=None):
...
+    def getJ(self, m, f=None):
+        """
+        Method to form full J given a model m
+
+        :param numpy.ndarray m: inversion model (nP,)
+        :param SimPEG.electromagnetics.frequency_domain.fields.FieldsFDEM u: fields object
+        :rtype: numpy.ndarray
+        :return: J (ndata, nP)
+        """
+
```
lheagy, line 252:
```suggestion
self.model = m
```

Should set the model in case it has changed so that the sensitivity is recomputed if necessary (see #1358)

jcapriot:
This is now good to go on my end, had to fix a few implementation errors

jcapriot (approved):

PR #1279, opened 2023-08-23
add plausible analytics to simpeg docs

lheagy:
#### Summary
- add plausible analytics to SimPEG docs and remove google analytics
- this will require the docs to be re-deployed (ideally as soon as possible so that we aren't losing out on analytics) 

#### PR Checklist
      to a Pull Request

#### Reference issue
N/A

We were previously using google analytics which imposes more tracking than what we would like. Also, now that google is requiring v4 (which we haven't set up), we are no longer receiving analytics.


santisoler (approved):
Looks good to me!

PR #1283, opened 2023-09-10
Switch from current `flat-layout` to `src-layout`

omid-b:
#### Summary

There are multiple benefits to the `src-layout` packaging over the lagacy `flat-layout`, specially when it comes to distribution of a python package to the community (e.g., [see this](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/#:~:text=The%20src%20layout%20requires%20installation,installation%20is%20used%20for%20testing)). As SimPEG is an open source project, it is important to make its distribution and installation as easy as possible.

This PR is simply the first step toward setting up a `pyproject.toml` rather than depending on `setup.py` for metadata and setup/build configurations (I will submit another PR in this regard, once this gets approved).

#### PR Checklist
* [X] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html).
      to a Pull Request

#### Reference issue

N/A

- Without modifying the source codes, `SimPEG/` was move to `src/SimPEG/`
- Updated `.gitignore`, `Makefile`, and `setup.py` to take the new `src-layout` into account.

#### Additional information

This is the first step for switching SimPEG over a `pyproject.toml` configuration to better handle the dependencies and making it more future proof.


omid-b:
@simpeg/simpeg-developers

lheagy:
Thanks for your pr @omid-b! Updating to the `pyproject.toml` doesn't strictly require that we change the directory layout (e.g. see discretize: https://github.com/simpeg/discretize) and we would prefer to stick with this current layout. So I will close this pr.

PR #1285, opened 2023-09-19
Gravity simulation with Choclo as engine

santisoler:
#### Summary

Add the option to use [Choclo](https://www.fatiando.org/choclo) as the engine for running the gravity simulations, considerably reducing the computation time of the forward simulations and the creation of the sensitivity matrix. Add Choclo as one of the optional dependencies. Update and extend tests for the gravity simulation using pytest. By default, the simulation will use `geoana` as the engine. Users can alternatively choose `choclo` instead if it's installed.

#### PR Checklist
      to a Pull Request


thibaut-kobold:
@santisoler : does this mean SimPEG PF will support GPU with this change?

santisoler:
> @santisoler : does this mean SimPEG PF will support GPU with this change?

Not with the changes in this PR. I'm using Numba, but running the simulation in CPU.

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -27,27 +242,132 @@ class Simulation3DIntegral(BasePFSimulation):
...
+                "The choclo package couldn't be found."
+                "Running a gravity simulation with 'engine=\"choclo\"' needs "
+                "choclo to be installed."
+                "\nTry installing choclo with:"
+                "\n    pip install choclo"
+                "\nor:"
+                "\n    conda install choclo"
+            )
+        # Warn if n_processes has been passed
+        if self.engine == "choclo" and "n_processes" in kwargs:
```
jedman, line 330: where is `n_processes` being ignored? I might be missing something but I see it still being passed to `BasePFSimulation` and set as normal

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -27,27 +242,132 @@ class Simulation3DIntegral(BasePFSimulation):
...
 
     rho, rhoMap, rhoDeriv = props.Invertible("Density")
 
-    def __init__(self, mesh, rho=None, rhoMap=None, **kwargs):
+    def __init__(
+        self,
+        mesh,
+        rho=None,
+        rhoMap=None,
+        engine="geoana",
```
jedman, line 287: Ultimately, I think the engine/parallelism should be inherited from `BasePFSimulation`. The interface (should) work the same for both gravity and magnetics eventually; all we'd need for now is to add a check that the mag simulation is using `engine == "geoana"`, and raise if not

I know it's outside the scope of this PR, but I'm strongly in favor of deprecating the multiprocessing implementation as soon as possible-- it's slow, and very memory inefficient, and with the addition of `choclo`, it makes the interface quite clunky.

SimPEG/potential_fields/gravity/simulation.py
santisoler: Thanks @jedman for the review!

You're right, `n_processes` is still being passed to `BasePFSimulation` and set as an attribute. But when using the `choclo` engine, it's not being used at all. `n_processes` is used within the `BasePFSimulation.linear_operator` method, which is not called when using `choclo` as engine.

That being said, I agree it's not 100% clear that `n_processes` is being ignored. Do you think we should remove it from the `kwargs` or set it to `None` after raising the warning?

SimPEG/potential_fields/gravity/simulation.py
santisoler: I agree. Ultimately, engine and parallelism should be handled by the parent class, since they will be used for both mag and gravity simulations. I added them to the gravity simulation for this PR just to narrow its scope to only the gravity simulation. I think we can move these bits to the `BasePFSimulation` when implementing the mag simulation with choclo.

> I know it's outside the scope of this PR, but I'm strongly in favor of deprecating the multiprocessing implementation as soon as possible-- it's slow, and very memory inefficient, and with the addition of choclo, it makes the interface quite clunky.

I think these are good reasons. For now we are going to leave the two implementations and decide to ditch the multiprocessing one or not later on. There are still some features that the choclo version is not implementing, like working with Dask to distribute the forward among cluster nodes, or storing the sensitivity to disk using compressed zarr arrays, which have higher io speed and require less disk space.

SimPEG/potential_fields/gravity/simulation.py
jedman: ah, that's the code path I was missing, thanks. Yes, I think it would be clearer to set `n_processes=None` after raising the warning

santisoler: Done!

jedman (approved):
fantastic work @santisoler , this addition dramatically improves the performance of the gravity simulation. looking forward to propagating it to the magnetic simulation too!

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -205,6 +363,164 @@ def evaluate_integral(self, receiver_location, components):
...
+        Returns
+        -------
+        (nD,) array
+            Always return a ``np.float64`` array.
+        """
+        # Gather active nodes and the indices of the nodes for each active cell
+        active_nodes, active_cell_nodes = self._get_active_nodes()
+        # Allocate fields array
+        fields = np.zeros(self.survey.nD, dtype=self.sensitivity_dtype)
+        # Start filling the sensitivity matrix
```
santisoler, line 385:
```suggestion
        # Start computing the fields
```

santisoler:
@domfournier would you like to take a look at this PR? I would love to have another pair of eyes looking at it before we decide to merge it.

domfournier:
> @domfournier would you like to take a look at this PR? I would love to have another pair of eyes looking at it before we decide to merge it.

I ll take a look in the morning

SimPEG/potential_fields/gravity/_numba_functions.py
```diff
@@ -0,0 +1,177 @@
...
+            sensitivity_matrix[i, k] = constant_factor * (
+                -kernels[cell_nodes[k, 0]]
+                + kernels[cell_nodes[k, 1]]
+                + kernels[cell_nodes[k, 2]]
+                - kernels[cell_nodes[k, 3]]
+                + kernels[cell_nodes[k, 4]]
+                - kernels[cell_nodes[k, 5]]
+                - kernels[cell_nodes[k, 6]]
+                + kernels[cell_nodes[k, 7]]
+            )
```
domfournier, line 170: Is it possible to reduce code duplication here and move this to a sub-function?

SimPEG/potential_fields/gravity/_numba_functions.py
```diff
@@ -0,0 +1,177 @@
...
+    index the elements that corresponds to each component.
+    """
+    n_receivers = receivers.shape[0]
+    n_nodes = nodes.shape[0]
+    n_cells = cell_nodes.shape[0]
+    # Evaluate kernel function on each node, for each receiver location
+    for i in prange(n_receivers):
+        # Allocate vector for kernels evaluated on mesh nodes
+        kernels = np.empty(n_nodes)
+        for j in range(n_nodes):
```
domfournier, line 75: I am guessing this is where the parallel C process happens? Otherwise it would be really slow to loop - this could technically be done with linear algebra right?

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -27,27 +76,134 @@ class Simulation3DIntegral(BasePFSimulation):
...
+            warnings.warn(
+                "The 'n_processes' will be ignored when selecting 'choclo' as the "
+                "engine in the gravity simulation.",
+                UserWarning,
+                stacklevel=1,
+            )
+            self.n_processes = None
+        # Sanity checks for sensitivity_path when using choclo and storing in disk
+        if self.engine == "choclo" and self.store_sensitivities == "disk":
+            if os.path.isdir(self.sensitivity_path):
```
domfournier, line 175: Feels like the same check should happen regardless if the engine is choclo or geoana. A bit of change of behaviour, but generally safer. I like it.

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -27,27 +76,134 @@ class Simulation3DIntegral(BasePFSimulation):
...
+        (nD,) array
+            Gravity fields generated by the given model on every receiver
+            location.
+        """
+        self.model = m
         if self.store_sensitivities == "forward_only":
             # Compute the linear operation without forming the full dense G
-            fields = mkvc(self.linear_operator())
+            if self.engine == "choclo":
+                fields = self._forward(self.rho)
```
domfournier, line 202: Not sure if we want to have different attribute names based on the engine. Feels like the `linear_operator` method should handle the logic...

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -95,8 +251,10 @@ def G(self):
         Gravity forward operator
         """
         if getattr(self, "_G", None) is None:
-            self._G = self.linear_operator()
-
+            if self.engine == "choclo":
+                self._G = self._sensitivity_matrix()
```
domfournier, line 255: **suggestion**: Same here. The issue with new attribute names is that it makes grav differ a lot from the base class (and sub-classes). I would vote for `engine` to be on the base class, such that we can eventually roll out different engines on the subclasses.

domfournier (approved):
I like it! See suggestions below.

SimPEG/potential_fields/gravity/_numba_functions.py
santisoler: Yes, it's possible. I just moved the evaluation of the kernels and the sums over the nodes for each cell to two different private functions.

SimPEG/potential_fields/gravity/_numba_functions.py
santisoler: All these functions are decorated with `@numba.jit`, so Numba is compiling these functions using LLVM and then using these compiled versions when they are called.

If by linear algebra you mean vectorized functions, it's possible, although those operations involve some level of abstractions that usually result in less efficient code (allocation of large arrays, unnecessary allocation of memory for intermediate results, etc). That's why I went with plain for loops.

Does it make sense?

SimPEG/potential_fields/gravity/simulation.py
santisoler: As I replied before to @jedman, I didn't want to touch the parent class too much here. I was focusing on implementing the forward for gravity only. If I would have added the `engine` attribute to the parent class in this PR, I would also had to modify the behaviour of all child classes. For example, the mag simulation should throw an error if the user is trying to pass `engine="choclo"` (at least for now). And I didn't want to do that at this stage.

I think once we are happy with the design of the class and how we expect users to interact with it, we can move stuff like this one to the parent class. Specially when implementing the mag simulation using Choclo, where the two child classes can reuse some methods.

Does it make sense?

SimPEG/potential_fields/gravity/_numba_functions.py
domfournier: yea it does. So the loops are turned into C loops, which is basically what numpy does (I think?) anyway.
Thanks

SimPEG/potential_fields/gravity/simulation.py
santisoler: This is related to my reply to the next comment. Since I didn't want to touch the parent class in this PR, I thought having specific private methods for the gravity simulation was the sensitive choice (at least for now).

Also, the design of `linear_operator` + `evaluate_integral` doesn't work well with the new implementation of the simulation: the two for loops (over receivers and over nodes) happen within a single numba-decorated function.

So, considering these two reasons, I think it's better to have these two new private methods in the simulation class: `_forward` and `_sensitivity_matrix`.

santisoler:
Hi @domfournier! Thanks for those suggestions. I've applied some of them, specially the ones related to reducing code duplication. I've also replied the other ones explaining better my design decisions.

I think we can move forward with this PR if you are happy with it, and continue the discussion about design on a future PR. Specially the one I'm going to open implementing the mag simulation.

SimPEG/potential_fields/gravity/_numba_functions.py
```diff
@@ -0,0 +1,256 @@
...
+    This function should be used with a `numba.jit` decorator, for example:
+
+    ..code::
+
+        from numba import jit
+
+        jit_sensitivity = jit(nopython=True, parallel=True)(_sensitivity_gravity)
+        jit_sensitivity(
+            receivers, nodes, densities, fields, cell_nodes, kernel_func, const_factor
+        )
```
santisoler, line 124: I would just remove these lines. They had errors and they are hard to detect. Besides, they are just reproducing the function signature, so they aren't too informative.
```suggestion
```

SimPEG/potential_fields/gravity/_numba_functions.py
```diff
@@ -0,0 +1,256 @@
...
+    This function should be used with a `numba.jit` decorator, for example:
+
+    ..code::
+
+        from numba import jit
+
+        jit_forward_gravity = jit(nopython=True, parallel=True)(_forward_gravity)
+        jit_forward_gravity(
+            receivers, nodes, densities, fields, cell_nodes, kernel_func, const_factor
+        )
```
santisoler, line 39: Same here
```suggestion
```

SimPEG/potential_fields/gravity/simulation.py
santisoler: I'll keep that in mind! We should explore adding this to the parent class then.

domfournier (approved):
Thanks for the changes. Looking good on this side.

domfournier:
> Hi @domfournier! Thanks for those suggestions. I've applied some of them, specially the ones related to reducing code duplication. I've also replied the other ones explaining better my design decisions.
> 
> I think we can move forward with this PR if you are happy with it, and continue the discussion about design on a future PR. Specially the one I'm going to open implementing the mag simulation.

Sounds good yea, we can generalize on the second round for mag. Looking forward to testing the speedup!

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -205,6 +363,164 @@ def evaluate_integral(self, receiver_location, components):
...
+        fields = np.zeros(self.survey.nD, dtype=self.sensitivity_dtype)
+        # Compute fields
+        index_offset = 0
+        for components, receivers in self._get_components_and_receivers():
+            n_components = len(components)
+            n_elements = n_components * receivers.shape[0]
+            for i, component in enumerate(components):
+                kernel_func = CHOCLO_KERNELS[component]
+                conversion_factor = _get_conversion_factor(component)
+                vector_slice = slice(
```
jcapriot, line 393: I believe that you are properly taking care of the ordering for multiple components but it could be worth adding a check to make sure that it is consistent (here and the sensitivity as well).

SimPEG/potential_fields/gravity/simulation.py
```diff
@@ -205,6 +363,164 @@ def evaluate_integral(self, receiver_location, components):
...
+        """
+        if isinstance(self.mesh, discretize.TreeMesh):
+            cell_nodes = self.mesh.cell_nodes
+        elif isinstance(self.mesh, discretize.TensorMesh):
+            cell_nodes = self._get_tensormesh_cell_nodes()
+        else:
+            raise TypeError(f"Invalid mesh of type {self.mesh.__class__.__name__}.")
+        return cell_nodes
+
+    def _get_tensormesh_cell_nodes(self):
```
jcapriot, line 464: Let's get a new version of discretize out and tagged so we don't need to include this bit.

jcapriot (changes requested):
I've been playing around a bit with `numba` myself the last couple days. I like that it's able to jit a lot of the functions in here so we start avoiding the costly operations of initializing new (rather large) numpy arrays and also avoid the stacking operation. Avoiding the costly multiprocessing step here is great too. I've been looking into updating `geoana` as well so those functions could be used inside `numba`'s functions. At that point the speed is roughly the same between the two (geoana vs choclo).

Eventually we should put in the simplifications when you have gxx, gyy, and gzz, but as those are not currently in the code (for accuracy reasons) no reason to add them in now.

SimPEG/potential_fields/gravity/simulation.py
santisoler: Nice catch. This was one of trickiest parts of this PR: making sure that we are filling the right rows of the sensitivity matrix and the right elements of the fields array.

The tests pass several components of the acceleration and the tensor, so we are actually testing this bits. In case these slices were wrongly defined, those tests should be failing.

Is there any other test/check we could be running?

SimPEG/potential_fields/gravity/simulation.py
santisoler: Sure!

santisoler:
Thanks @jcapriot for the review! Yes, numba has neat things indeed!

> Eventually we should put in the simplifications when you have gxx, gyy, and gzz, but as those are not currently in the code (for accuracy reasons) no reason to add them in now.

Yes. We might want to add some optimizations to this simulation when having multiple components of acceleration or tensor. These all also relevant for the mag simulation that's coming shortly.

SimPEG/potential_fields/gravity/simulation.py
jcapriot: I see where those tests are now, just didn't pop out to me on the first pass

santisoler:
@jcapriot I've updated the simulation to use the new `TensorMesh.cell_nodes` method in `discretize` `v0.10.0`. I've also increased the minimum version of `discretize` in `setup.py` and `environment_test.yml`.

jcapriot (changes requested):
Throughout the documentation, instead of writing `array` write either `numpy.ndarray` (or `array_like` where appropriate). Also, it could be worth it to run a comparison with the most recent `geoana` with the jittable prism kernels now available there.

santisoler:
@jcapriot do you think we should change anything else on this PR, or is it ready to be merged?

jcapriot:
> @jcapriot do you think we should change anything else on this PR, or is it ready to be merged?

The only thing I would change is the actual argument named after `choclo`, (`choclo_parallel`) I would rename it something related to `numba` (`numba_parallel`?) as `geoana` prism kernels should now support being jitted. We should add another PR to update the internals to also support the `geoana` ones as well if it's not too different.

I feel like the default behavior (in the future) should be: If `numba` is installed, use `numba` parallelized `geoana` (since `geoana` is already a requirement), else use the vectorized calls, We can add `choclo` as a further optional thing if that behavior is actually needed.

I also think we should probably advertise it as being `numba` accelerated as well in that case.

lheagy:
I am really excited to get this in, congrats @santisoler! This accomplishes a major goal of having a concrete connection between the Fatiando and SimPEG communities and the benchmarks you have shown are impressive!

We now have a couple of implementations, and that can be useful for a number of reasons, including testing. I think the discussion about the default behaviour is an important one. `choclo` is a lightweight dependency and from my perspective, the explicit connection between the SimPEG and Fatiando communities is something that should be valued as we consider this. We can chat more at some of the upcoming meetings. 

Again, nice work Santi!

jcapriot:
Oh yeah, I fully support keeping `choclo` as an option, I just don’t want people to get a bad impression of `geoana` either… (I.e geoana=slow, choclo=fast). It’s more like they are both good and fast!

santisoler:
Thanks @jcapriot and @lheagy!

Considering the work you've been doing in allowing geoana kernels to be called within Numba jitted functions, it makes sense to rename the argument to `numba_parallel`. I've just pushed that change.

> Oh yeah, I fully support keeping `choclo` as an option, I just don’t want people to get a bad impression of `geoana` either… (I.e geoana=slow, choclo=fast). It’s more like they are both good and fast!

I'm sorry for making that impression. The main difference in performance doesn't come from `geoana`, but from the Python implementation of the simulation in SimPEG. The choclo and geoana kernels have equivalent runtimes.

santisoler:
I just merged this! 🚀 

Thanks a lot for the reviews, help and comments. I'm looking forward to open the PR for the mag simulation now!

I'll keep this branch up for a while, so I don't break any environment in case someone was using it. But I'll eventually delete it, so please install the latest version on `main` or wait for the next release (`v0.21.0`).

lheagy:
💥 So excited to see this come in!

PR #1286, opened 2023-09-20
Allow to use random seed in make_synthetic_data

santisoler:
#### Summary

Add a new `random_seed` argument to the `make_synthetic_data` method.
Switch to the Numpy's random number generator object and pass the
`random_seed` argument to `np.random.default_rng`, allowing users to
specify a random seed for the synthetic noise. Set random seeds in some 
tests that were failing due to the new random state that the usage of 
`make_synthetic_data` with `random_seed` created in the test suite.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/SimPEG/simulation.py
+++ b/SimPEG/simulation.py
@@ -328,7 +328,7 @@ class BaseSimulation(props.HasModel):
         f : array or None
             Fields for the given model (if pre-calculated).
         add_noise : bool
-            Wether to add gaussian noise to the synthetic data or not.
+            Whether to add gaussian noise to the synthetic data or not.
         random_seed : int or None
             Random seed to pass to `numpy.random.default_rng`.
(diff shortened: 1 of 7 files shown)


santisoler:
@jcapriot I managed to get the tests pass here. But at some point I would love to see all tests run with a hard coded seed. Do you think it worth tackling that in this PR or should we merge this one and tackle the rest in a separate one?

lheagy (approved):
Thanks @santisoler! It will be valuable to have better practices in place for dealing with randomness. This looks good to me 🚀 

I know there are a lot of other places across SimPEG to do this, one of the higher priority places to look to next would be the directives: I am specifically thinking of the beta-estimator. We can start an issue if you like, I am just commenting here as it is top-of-mind as I look at these updates

PR #1288, opened 2023-09-20
Triaxial magnetic gradient forward modelling

thibaut-kobold:
#### Summary

This PR adds magnetic forward modelling for the gradient of the TMI, measured when several magnetometers are mounted on the same aircraft. A good example is [the Heli-GT system from SHA geophysics ](https://www.shageophysics.com/#js__scroll-to-section_heli-gt) but many airborne systems have several magnetometers mounted.

#### PR Checklist
      to a Pull Request

#### Additional information

Screenshot 2023-09-19 at 7 38 12 PM

### DIFF
--- a/SimPEG/potential_fields/magnetics/simulation.py
+++ b/SimPEG/potential_fields/magnetics/simulation.py
@@ -243,7 +243,7 @@ class Simulation3DIntegral(BasePFSimulation):
 
         components: list[str]
             List of magnetic components chosen from:
-            'bx', 'by', 'bz', 'bxx', 'bxy', 'bxz', 'byy', 'byz', 'bzz'
+            'tmi', 'bx', 'by', 'bz', 'bxx', 'bxy', 'bxz', 'byy', 'byz', 'bzz', 'tmi_x', 'tmi_y', 'tmi_z'
 
         OUTPUT:
         Tx = [Txx Txy Txz]
--- a/SimPEG/potential_fields/magnetics/simulation.py
+++ b/SimPEG/potential_fields/magnetics/simulation.py
@@ -1,27 +1,24 @@
 import numpy as np
 import scipy.sparse as sp
-from scipy.constants import mu_0
-
-from SimPEG import utils
-from ..base import BasePFSimulation, BaseEquivalentSourceLayerSimulation
-from ...base import BaseMagneticPDESimulation
-from .survey import Survey
-from .analytics import CongruousMagBC
-
-from SimPEG import Solver
-from SimPEG import props
-
-from SimPEG.utils import mkvc, mat_utils, sdiag
-from SimPEG.utils.code_utils import validate_string, deprecate_property, validate_type
 from geoana.kernels import (
-    prism_fzz,
-    prism_fzx,
-    prism_fzy,
-    prism_fzzz,
     prism_fxxy,
     prism_fxxz,
     prism_fxyz,
+    prism_fzx,
+    prism_fzy,
+    prism_fzz,
+    prism_fzzz,
 )
+from scipy.constants import mu_0
+
+from SimPEG import Solver, props, utils
+from SimPEG.utils import mat_utils, mkvc, sdiag
+from SimPEG.utils.code_utils import deprecate_property, validate_string, validate_type
+
+from ...base import BaseMagneticPDESimulation
+from ..base import BaseEquivalentSourceLayerSimulation, BasePFSimulation
+from .analytics import CongruousMagBC
+from .survey import Survey
 
 
 class Simulation3DIntegral(BasePFSimulation):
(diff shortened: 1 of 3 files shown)


thibaut-kobold:
@domfournier @jcapriot : what would be the best way to test it? Would we need to update `geoana` prior to this PR or is there a path to keep this PR self-sufficient ?

thibaut-kobold:
@simpeg/simpeg-developers

lheagy (commented):
@santisoler: would you have the bandwidth to review this pr?

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -283,33 +280,33 @@ def evaluate_integral(self, receiver_location, components):
             #     # inside an active cell.
             #     node_evals["gzz"] = -node_evals["gxx"] - node_evals["gyy"]
 
-        if "bxx" in components:
+        if "bxx" in components or "tmi_x" in components:
             node_evals["gxxx"] = prism_fzzz(dy, dz, dx)
             node_evals["gxxy"] = prism_fxxy(dx, dy, dz)
             node_evals["gxxz"] = prism_fxxz(dx, dy, dz)
-        if "bxy" in components:
+        if any(s in components for s in ["bxy", "tmi_x", "tmi_y"]):
```
santisoler, line 287: I know this is a mouthful, but it's more efficient than having to build a list and a tuple on each iteration and then checking if any of its values is True.

What do you think?

```suggestion
        if "bxy" in components or "tmi_x" in components or "tmi_y" in components:
```

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -355,6 +352,57 @@ def evaluate_integral(self, receiver_location, components):
                     + tmi[1] * node_evals["gyz"]
                     + tmi[2] * node_evals["gzz"]
                 )
+            elif component == "tmi_x":
+                tmi = self.tmi_projection
+                vals_x = (
```
santisoler, line 357: For some reason, these lines are highlighted as not being covered by tests, but they are actual ran in `test_forward_Mag_Linear.py`. I don't think this is an issue of this PR, since these lines aren't being marked as covered even in `main`.

I'm mentioning it because it would nice to find out why this is happening and how to solve it.

tests/pf/test_forward_Mag_Linear.py
```diff
@@ -101,6 +105,98 @@ def get_block_inds(grid, block):
     np.testing.assert_allclose(d_t, d @ tmi)
 
 
+def test_ana_mag_tmi_grad_forward():
```
santisoler, line 108: Thanks for adding this test function. I think it's nice to have the Simulation compared with the forwards in geoana.

Nevertheless, I think it would be nice to add a finite difference test: forward model the TMI of a set of prisms on points separated by a small distance, so we can then approximate the derivatives of the TMI by finite differences of these values. This way we can have a test that don't rely on the third-order derivatives of the kernels, but on the second-order ones.

What do you think?

santisoler (changes requested):
Thanks @thibaut-kobold for this PR, and sorry for the delayed review.

I think this is good and we shouldn't have any issue in merging it soon. I left a few comments below. Let me know what do you think.

SimPEG/potential_fields/magnetics/simulation.py
domfournier: even cleaner would be to do a set().intersection

```suggestion
        if any(set(["bxy", "tmi_x", "tmi_y"]).intersection(components)):
```

thibaut-kobold: @santisoler : you are right, your solution is faster.
Cool syntax @domfournier , noted. But I will go with Santi's suggestion.
image

SimPEG/potential_fields/magnetics/simulation.py
thibaut-kobold: I have no idea why. Maybe @jcapriot would know.

tests/pf/test_forward_Mag_Linear.py
thibaut-kobold: a bit cuncumber some if we try to keep the problem small, but I cooked up something. see what you think.

jcapriot: The ones in geoana are tested as derivatives of the kernels, so I think it's sufficient to compare against geoana.

santisoler: Sorry for the misunderstanding. I wasn't saying to increase the number of receivers in the grid, but to create "shifted" grids that could have the original 5 x 5 shape. For example, define a pair of grids shifted along `x`:

```python 
    nx, ny = 5, 5
    xr = np.linspace(-20, 20, nx)
    yr = np.linspace(-20, 20, ny)
    X, Y = np.meshgrid(xr, yr)

    delta_x = 0.01
    receivers_loc = np.c_[X.reshape(-1), Y.reshape(-1), Z.reshape(-1)]
    receivers_loc_pos = np.c_[X.reshape(-1) + delta_x / 2, Y.reshape(-1), Z.reshape(-1)]
    receivers_loc_neg = np.c_[X.reshape(-1) - delta_x / 2, Y.reshape(-1), Z.reshape(-1)]
```

Then compute:
* `tmi_x` on `receivers_loc`
* `tmi` on `receivers_loc_pos` and `receivers_loc_neg`

And approximate the `tmi_x` by finite differences:
```python
tmi_x_finite_diff = (tmi_pos - tmi_neg) / delta_x
```

santisoler: > The ones in `geoana` are tested as derivatives of the kernels, so I think it's sufficient to compare against `geoana`.

I'm not saying to retest `geoana` kernels here. In the tests that @thibaut-kobold wrote, we are implementing a simplified version of the forward of the derivatives of tmi, particularly in these lines:
```python
    d = (
        prism_1.magnetic_field_gradient(locXyz)
        + prism_2.magnetic_field_gradient(locXyz)
        + prism_3.magnetic_field_gradient(locXyz)
    ) * mu_0
    tmi_x = (d[:, 0, 0] * b0[0] + d[:, 0, 1] * b0[1] + d[:, 0, 2] * b0[2]) / H0[0]
    tmi_y = (d[:, 1, 0] * b0[0] + d[:, 1, 1] * b0[1] + d[:, 1, 2] * b0[2]) / H0[0]
    tmi_z = (d[:, 2, 0] * b0[0] + d[:, 2, 1] * b0[1] + d[:, 2, 2] * b0[2]) / H0[0]
    np.testing.assert_allclose(d_x, tmi_x, rtol=1e-10, atol=1e-12)
    np.testing.assert_allclose(d_y, tmi_y, rtol=1e-10, atol=1e-12)
    np.testing.assert_allclose(d_z, tmi_z, rtol=1e-10, atol=1e-12)
```

In the event that both the implementation in the simulation and this simplified implementation have the same error, tests will pass and won't be able to catch the bug. By adding the finite difference test we can solve that.

It's not critical, it's just something it would be nice to have.

tests/pf/test_forward_Mag_Linear.py
```diff
@@ -101,6 +105,116 @@ def get_block_inds(grid, block):
...
+        + prism_3.magnetic_field_gradient(locXyz)
+    ) * mu_0
+    tmi_x = (d[:, 0, 0] * b0[0] + d[:, 0, 1] * b0[1] + d[:, 0, 2] * b0[2]) / H0[0]
+    tmi_y = (d[:, 1, 0] * b0[0] + d[:, 1, 1] * b0[1] + d[:, 1, 2] * b0[2]) / H0[0]
+    tmi_z = (d[:, 2, 0] * b0[0] + d[:, 2, 1] * b0[1] + d[:, 2, 2] * b0[2]) / H0[0]
+    np.testing.assert_allclose(d_x, tmi_x, rtol=1e-10, atol=1e-12)
+    np.testing.assert_allclose(d_y, tmi_y, rtol=1e-10, atol=1e-12)
+    np.testing.assert_allclose(d_z, tmi_z, rtol=1e-10, atol=1e-12)
+
+    # finite difference test y-grad
```
thibaut-kobold, line 202: @santisoler : I did add a finite difference test here.
I had to increase the number of receiver to obtain an ok response.
This is similar to what you propose (but instead of one global receiver grid, you have several). I think having a single  grid make the test syntax less cumbersome and easier to debug later on.

The test is still very fast to run with the increased number of receivers (<3 s on my machine).

See if that solution works for you.

tests/pf/test_forward_Mag_Linear.py
thibaut-kobold: Thanks @santisoler, I did add a finite difference test yesterday as you suggested, but with one global receiver grid rather than 5 separate grids as you propose.
I think having a single grid make the test syntax less cumbersome and easier to debug later on.
The test is still very fast to run with the increased number of receivers (<3 s on my machine).
See if that solution works for you.

tests/pf/test_forward_Mag_Linear.py
santisoler: Sounds good! I think we can merge this now 🚀

santisoler (approved):

tests/pf/test_forward_Mag_Linear.py
thibaut-kobold: ~5s on Azure
image (92)

SimPEG/potential_fields/magnetics/simulation.py
santisoler: I'm marking this conversation as resolved since this coverage issue is not caused by this PR. It should be solved by a different one.

PR #1291, opened 2023-10-10
pgi doc

thibaut-kobold:
Correction for the `approx_eval` explanation

@dccowan


jcapriot:
Thanks for the updates Thibaut

PR #1294, opened 2023-10-15
Update sources.py

xionche:
#### PR Checklist
      to a Pull Request

### DIFF
--- a/SimPEG/electromagnetics/frequency_domain/sources.py
+++ b/SimPEG/electromagnetics/frequency_domain/sources.py
@@ -743,8 +743,6 @@ class CircularLoop(MagDipole):
         Source location.
     moment : float
         Magnetic dipole moment amplitude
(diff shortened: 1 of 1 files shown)


xionche:
> Hey @xionche, Is there a specific reason you think we should remove the documentation of this keyword argument? The circular loop certainly supports an orientation.
Hello.  I saw that there is already a CircularLoop ‘Orientation’ keyword below in this code.  But I wanna ask another question, Can we put 2 sources at the same mesh in SimPEG?  For example, I wanna make a source with 2 circular loops.

dccowan:
@xionche Are you asking about a survey comprised of multiple sources? Or the case where two sources are exciting the Earth simultaneously?

The first case is easily handled by putting all source objects in a list, and assigning that list to the survey object. In the second case, you could simply simulate the response for each source separately, then add the responses together.

At this points, should we close this pull request?

xionche:
I’ve resolved the issue.Thanks a lot.

PR #1299, opened 2023-11-08
Fix issue with lengthscales in coterminal angle calculations

domfournier:
#### Summary
The initial implementation of coterminal angles used an average of cell sizes to determine the cell centered length. Those distances are slightly different (by 88%) than the cell gradient distances used by the `regularization_mesh.cell_gradient_` operators on octree changes. 

#### PR Checklist
* [X] If this is a work in progress PR, set as a Draft PR
* [X] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html).
* [X] Added [tests](https://docs.simpeg.xyz/content/getting_started/practices.html#testing) to verify changes to the code.
* [N/A] Added necessary documentation to any new functions/classes following the
* [X] Marked as ready for review (if this is was a draft PR), and converted 
      to a Pull Request
* [X] Tagged ``@simpeg/simpeg-developers`` when ready for review.

Fixes large gradients computed on the -pi/pi transition, resulting in artifacts in the model.

Current results from `examples\03-magnetics\plot_inv_mag_MVI_Sparse_TreeMesh.py.`:

Title should be L2 Spherical model...

Figure_3

After fix:

Figure_3_new

### DIFF
--- a/SimPEG/regularization/regularization_mesh.py
+++ b/SimPEG/regularization/regularization_mesh.py
@@ -551,7 +551,7 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the x-direction.
         """
         if getattr(self, "_cell_distances_x", None) is None:
-            self._cell_distances_x = np.max(self.cell_gradient_x, axis=1).data ** (-1.0)
+            self._cell_distances_x = self.cell_gradient_x.max(axis=1).data ** (-1.0)
 
         return self._cell_distances_x
 
@@ -565,7 +565,7 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the y-direction.
         """
         if getattr(self, "_cell_distances_y", None) is None:
-            self._cell_distances_y = np.max(self.cell_gradient_y, axis=1).data ** (-1.0)
+            self._cell_distances_y = self.cell_gradient_y.max(axis=1).data ** (-1.0)
 
         return self._cell_distances_y
 
@@ -579,7 +579,7 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the z-direction.
         """
         if getattr(self, "_cell_distances_z", None) is None:
-            self._cell_distances_z = np.max(self.cell_gradient_z, axis=1).data ** (-1.0)
+            self._cell_distances_z = self.cell_gradient_z.max(axis=1).data ** (-1.0)
 
         return self._cell_distances_z
 
--- a/SimPEG/regularization/regularization_mesh.py
+++ b/SimPEG/regularization/regularization_mesh.py
@@ -551,8 +551,8 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the x-direction.
         """
         if getattr(self, "_cell_distances_x", None) is None:
-            Ave = self.aveCC2Fx
-            self._cell_distances_x = Ave * (self.Pac.T * self.mesh.h_gridded[:, 0])
+            self._cell_distances_x = np.max(self.cell_gradient_x, axis=1).data ** (-1.0)
+
         return self._cell_distances_x
 
     @property
@@ -565,8 +565,8 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the y-direction.
         """
         if getattr(self, "_cell_distances_y", None) is None:
-            Ave = self.aveCC2Fy
-            self._cell_distances_y = Ave * (self.Pac.T * self.mesh.h_gridded[:, 1])
+            self._cell_distances_y = np.max(self.cell_gradient_y, axis=1).data ** (-1.0)
+
         return self._cell_distances_y
 
     @property
@@ -579,8 +579,8 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the z-direction.
         """
         if getattr(self, "_cell_distances_z", None) is None:
-            Ave = self.aveCC2Fz
-            self._cell_distances_z = Ave * (self.Pac.T * self.mesh.h_gridded[:, 2])
+            self._cell_distances_z = np.max(self.cell_gradient_z, axis=1).data ** (-1.0)
+
         return self._cell_distances_z
 
 
--- a/SimPEG/regularization/regularization_mesh.py
+++ b/SimPEG/regularization/regularization_mesh.py
@@ -551,7 +551,7 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the x-direction.
         """
         if getattr(self, "_cell_distances_x", None) is None:
-            self._cell_distances_x = self.cell_gradient_x.max(axis=1).data ** (-1.0)
+            self._cell_distances_x = self.cell_gradient_x.max(axis=1).toarray().flatten() ** (-1.0)
 
         return self._cell_distances_x
 
--- a/SimPEG/regularization/regularization_mesh.py
+++ b/SimPEG/regularization/regularization_mesh.py
@@ -551,7 +551,9 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the x-direction.
         """
         if getattr(self, "_cell_distances_x", None) is None:
-            self._cell_distances_x = self.cell_gradient_x.max(axis=1).toarray().flatten() ** (-1.0)
+            self._cell_distances_x = self.cell_gradient_x.max(
+                axis=1
+            ).toarray().flatten() ** (-1.0)
 
         return self._cell_distances_x
 
@@ -565,7 +567,9 @@ class RegularizationMesh(props.BaseSimPEG):
             Cell center distance array along the y-direction.
         """
         if getattr(self, "_cell_distances_y", None) is None:
-            self._cell_distances_y = self.cell_gradient_y.max(axis=1).data ** (-1.0)
+            self._cell_distances_y = self.cell_gradient_y.max(
+                axis=1
+            ).toarray().flatten() ** (-1.0)
 
         return self._cell_distances_y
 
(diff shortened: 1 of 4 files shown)


SimPEG/regularization/regularization_mesh.py
```diff
@@ -579,8 +579,8 @@ def cell_distances_z(self) -> np.ndarray:
             Cell center distance array along the z-direction.
         """
         if getattr(self, "_cell_distances_z", None) is None:
-            Ave = self.aveCC2Fz
-            self._cell_distances_z = Ave * (self.Pac.T * self.mesh.h_gridded[:, 2])
+            self._cell_distances_z = np.max(self.cell_gradient_z, axis=1).data ** (-1.0)
```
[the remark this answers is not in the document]
domfournier: Done

santisoler (approved):
Thanks @domfournier for this, and sorry leaving you waiting for this review.

I think this is ready to be merged. I just pushed a small change to improve the test function: now it runs the same test in the three orientations, not only on `"y"`. I also renamed the function to be more descriptive.

Let me know what do you think. If you like it, ping me so I can merge this.

BTW, I just opened #1333 that it's related to this. If you can take a look at it, that would be great.

domfournier:
> Thanks @domfournier for this, and sorry leaving you waiting for this review.
> 
> I think this is ready to be merged. I just pushed a small change to improve the test function: now it runs the same test in the three orientations, not only on `"y"`. I also renamed the function to be more descriptive.
> 
> Let me know what do you think. If you like it, ping me so I can merge this.
> 
> BTW, I just opened #1333 that it's related to this. If you can take a look at it, that would be great.

Looks good man. I pushed a small fix to address your question on #1333 . Result looks good. Good to go on this end.

SimPEG/utils/mat_utils.py
```diff
@@ -411,7 +411,7 @@ def coterminal(theta):
     """
 
     sub = theta[np.abs(theta) >= np.pi]
-    sub = -np.sign(sub) * (2 * np.pi - np.abs(sub))
+    sub = np.sign(sub) * (2 * np.pi - np.abs(sub))
```
jcapriot, line 414: I don't think this will give you correct values now, simple testing:

```
>>> coterminal(np.pi + np.pi/4)
2.356194490192345  #3pi/4
```
it should give you -3pi/4.

I think the cleanest solution to this is to write the whole function as:

```
def coterminal(theta):
    return (theta + np.pi) % (2 * np.pi) - np.pi
```
Which will give you back a number on the domain of [-\pi, \pi)

santisoler: I agree that using the mod is the cleanest way to solve this.

I would like to see that bugfix in another PR though. I think one of the reasons why this wasn't caught before was due to lack of tests, so besides the fix we should add more tests for it. And also update the docstrings for that function.

domfournier: And why would it be negative you think? That's what it was returning before.
I guess my initial implementation was to return the negative angle of the complement, thinking that if the gradient between two angles is large, we would want to force the model update to reduce both. But now I am not so sure if that logic works.

domfournier:
I get a line search failure with this @jcapriot 

```
def coterminal(theta):
    return (theta + np.pi) % (2 * np.pi) - np.pi
```

I don't that's quite it, since the modulo in python takes the sign of the dividor. For example
```
coterminal(-np.pi - np.pi/4) -> 3/4 np.pi
```
But would expect -pi/4

But this works

```
coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
```
new

jcapriot:
> I get a line search failure with this @jcapriot
> 
> ```
> def coterminal(theta):
>     return (theta + np.pi) % (2 * np.pi) - np.pi
> ```
> 
> I don't think that's quite it, since the modulo in python takes the sign of the dividor. For example
> 
> ```
> coterminal(-np.pi - np.pi/4) -> 3/4 np.pi
> ```
> 
> But would expect -pi/4
> 
> But this works
> 
> ```
> coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
> ```
> 
> new

$3\pi/4$ is the correct coterminal angle for $-\pi - \pi/4=-5\pi/4$

jcapriot:
Why did you put a negative in front there?:
```

coterminal = -((theta + np.pi) % (2 * np.pi) - np.pi)

```

instead of 

```

coterminal = (theta + np.pi) % (2 * np.pi) - np.pi

```

jcapriot:
What layer are you cutting this at? Just so I can compare locally later this evening. I think I understand what you’re saying regarding what  needs to happen during minimization. Just need to test a few things.

jcapriot:
Also probably the easiest way to “undo” the cell difference would be to just use `dA = cell_gradient.sign() @ angles` instead of `dA= cell_distances * cell_gradient @ angles`

domfournier:
I sliced it at the same index, but Z axis. Also adjusted the limits to [-75, 75] for close up.

domfournier:
> Also probably the easiest way to “undo” the cell difference would be to just use `dA = cell_gradient.sign() @ angles` instead of `dA= cell_distances * cell_gradient @ angles`

Sounds good. Done. 
I reverted the changes to coterminal as requested by @santisoler , to be addressed instead on PR #1334

jcapriot:
This are results I'm getting:
Currently on main:
Figure_current
Correct coterminal:
Figure_mod
This branch at the moment:

jcapriot:
So these are the current results I'm getting on main:
image
image
These results mostly agree with what's on the current documentation page:
https://docs.simpeg.xyz/content/examples/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh.html#sphx-glr-content-examples-03-magnetics-plot-inv-mag-mvi-sparse-treemesh-py

These are the results I'm getting with this PR:
image
image

santisoler:
> Forget about the negative sign. I think we need to preserve the sign of the incoming angle.
> 
> We can argue all day about what the true coterminal should be mathematically, in the end we need the example in `03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh` to return a smooth model, and not get weird stuff on the wrap around [-pi, pi]. That was the intent of this function.
> 
> Not this image

I totally agree that the ultimate goal is to penalize non-smoothness in the neighboring angles in the model. That's also why I suggested of fixing the bug in `coterminal` in a separate PR: it's related to this one, but in the end they are different things.

I think the problem here is that we need to compute angles differences for each neighboring pair, and having a single array of coterminal angles moved to $[-\pi, \pi)$ (or any other interval of length of $2\pi$) won't solve it.

For example, if we have `theta_1 = np.pi - 0.001` and `theta_2 = -np.pi + 0.001`, their difference is just `0.002`, but if we compute the differences between their coterminal angles we'll get `2 * np.pi - 0.002`, leading to an unwanted penalization for the smoothness of these two.

I think the solution is to have a function that can compute the difference between two angles that's smaller or equal than `np.pi`.

jcapriot:
> > Forget about the negative sign. I think we need to preserve the sign of the incoming angle.
> > We can argue all day about what the true coterminal should be mathematically, in the end we need the example in `03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh` to return a smooth model, and not get weird stuff on the wrap around [-pi, pi]. That was the intent of this function.
> 
> I totally agree that the ultimate goal is to penalize non-smoothness in the neighboring angles in the model. That's also why I suggested of fixing the bug in `coterminal` in a separate PR: it's related to this one, but in the end they are different things.
> 
> I think the problem here is that we need to compute angles differences for each neighboring pair, and having a single array of coterminal angles moved to [−π,π) (or any other interval of length of 2π) won't solve it.
> 
> For example, if we have `theta_1 = np.pi - 0.001` and `theta_2 = -np.pi + 0.001`, their difference is just `0.002`, but if we compute the differences between their coterminal angles we'll get `2 * np.pi - 0.002`, leading to an unwanted penalization for the smoothness of these two.
> 
> I think the solution is to have a function that can compute the difference between two angles that's smaller or equal than `np.pi`.

I think it's important to point out that we are already computing the coterminal angle after taking differences.

jcapriot:
In the meantime, this looks fine to me now, only issue is that the second order smoothness isn't being tested here (but it wasn't before either).

domfournier:
Ok so can we move this one forward, and battle it out further on #1334 ?

domfournier:
> So these are the current results I'm getting on main: image image These results mostly agree with what's on the current documentation page: https://docs.simpeg.xyz/content/examples/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh.html#sphx-glr-content-examples-03-magnetics-plot-inv-mag-mvi-sparse-treemesh-py
> 
> These are the results I'm getting with this PR: image image

domfournier:
> So these are the current results I'm getting on main: image image These results mostly agree with what's on the current documentation page: https://docs.simpeg.xyz/content/examples/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh.html#sphx-glr-content-examples-03-magnetics-plot-inv-mag-mvi-sparse-treemesh-py

The top plot is the smooth Cartesian result, no problem there. This what the smooth Spherical looks like on main

image

SimPEG/regularization/regularization_mesh.py
```diff
@@ -551,8 +551,8 @@ def cell_distances_x(self) -> np.ndarray:
             Cell center distance array along the x-direction.
         """
         if getattr(self, "_cell_distances_x", None) is None:
-            Ave = self.aveCC2Fx
-            self._cell_distances_x = Ave * (self.Pac.T * self.mesh.h_gridded[:, 0])
+            self._cell_distances_x = self.cell_gradient_x.max(axis=1).data ** (-1.0)
```
jcapriot, line 554: Sorry just one last change before this is ready (as I'm thinking about it).

It would likely be safer to instead use `.to_array()` instead of `.data`, That way you can be sure of the ordering of the output.

```suggestion
            self._cell_distances_x = self.cell_gradient_x.max(axis=1).toarray().flatten() ** (-1.0)
```

There are certain cases where the `.data` attribute wouldn't work out to being the correct ordering/ have the correct length, but `.toarray()` should work in all cases.

jcapriot (approved):

PR #1300, opened 2023-11-15
Em1d stitched update

sgkang:
#### Summary
Finished implementing stitched 1D electromagnetic (EM) inversion (or often referred to as laterally-constrained or spatially-constrained inversion) suitable for inverting large-scale airborne EM (AEM) data or surface-based loop EM data. 

#### PR Checklist
      to a Pull Request

- Inversion of time-domain and frequency-domain EM data for imaging subsurface resistivity
- Inversion of time-domain and frequency-domain EM data for imaging subsurface resistivity and source and receiver height [there was a bug in `base_1d.py`; also inverting height raises an issue for storing coefficients @jcapriot ]
- Parallelization using `multiprocessing` [thought about using `Meta` class, but I felt the current implementation is not relevant; but open for any other opinions]
- Visualization of inverted resistivity model on a vertical section
- Creation of a 3D voxel model from the inverted resistivity model

#### Additional information

- There was a bug in `base_1d.py`; also inverting height raises an issue for storing coefficients @jcapriot 
- For parallelization, I thought about using `Meta` class, but I felt the current implementation is not relevant; but open for any other opinions
- An example for inverting synthetic frequency domain EM data to recover both subsurface resistivity and source height is available: https://curvenote.com/@sgkang09/researches/inv-1d-stitched-inv-alt-fdem
- Results of the example are provided below:

**Estimated source height over a profile:**
image

Estimated and True resistivity models over the same profile
image
image

### DIFF
--- a/SimPEG/electromagnetics/base_1d_stitched.py
+++ b/SimPEG/electromagnetics/base_1d_stitched.py
@@ -318,7 +318,8 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
     @property
     def H(self):
         if self.hMap is None:
-            return np.ones(self.n_sounding)
+            h = self.source_locations_for_sounding[:,2] - self.topo[:,2]
+            return h
         else:
             return self.h
 
@@ -360,7 +361,6 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
             self.Tau2[i_sounding, :],
             self.H[i_sounding],
             output_type,
-            self.invert_height,
             False,
             self._coefficients[i_sounding],
         )
@@ -383,7 +383,6 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
             self.Tau2[i_sounding, :],
             self.H[i_sounding],
             'forward',
-            self.invert_height,
             True,
             [],
         )
@@ -412,30 +411,34 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
         return self._sounding_number
 
     @property
-    def sounding_number_chunks(self):
-        self._sounding_number_chunks = list(self.chunks(self.sounding_number, self.n_sounding_for_chunk))
-        return self._sounding_number_chunks
-
-    @property
     def n_chunk(self):
         self._n_chunk = len(self.sounding_number_chunks)
         return self._n_chunk
+    @property
+    def source_locations_for_sounding(self):
+        if getattr(self, '_source_locations_for_sounding', None) is None:
+            self._source_locations_for_sounding = np.vstack([self.survey._source_location_by_sounding_dict[ii][0] for ii in range(self.n_sounding)])
+        return self._source_locations_for_sounding
 
-    def chunks(self, lst, n):
-        """Yield successive n-sized chunks from lst."""
-        for i in range(0, len(lst), n):
-            yield lst[i:i + n]
+    # def chunks(self, lst, n):
+    #     """Yield successive n-sized chunks from lst."""
+    #     for i in range(0, len(lst), n):
+    #         yield lst[i:i + n]
 
-    def input_args_by_chunk(self, i_chunk, output_type):
-        args_by_chunks = []
-        for i_sounding in self.sounding_number_chunks[i_chunk]:
-            args_by_chunks.append(self.input_args(i_sounding, output_type))
-        return args_by_chunks
+    # @property
+    # def sounding_number_chunks(self):
+    #     self._sounding_number_chunks = list(self.chunks(self.sounding_number, self.n_sounding_for_chunk))
+    #     return self._sounding_number_chunks
+
+    # def input_args_by_chunk(self, i_chunk, output_type):
+    #     args_by_chunks = []
+    #     for i_sounding in self.sounding_number_chunks[i_chunk]:
+    #         args_by_chunks.append(self.input_args(i_sounding, output_type))
+    #     return args_by_chunks
 
     def set_null_topography(self):
-        self.topo = np.vstack(
-            [np.c_[src.location[0], src.location[1], 0.] for i, src in enumerate(self.survey.source_list)]
-        )
+        self.topo = self.source_locations_for_sounding.copy()
+        self.topo[:,2] = 0.
 
 
     def set_ij_n_layer(self, n_layer=None):
@@ -538,15 +541,3 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
         if self.fix_Jmatrix is False:            
             toDelete += ['_Sigma', '_J', '_Jmatrix_sigma', '_Jmatrix_height', '_gtg_diag']
         return toDelete
-
-    # def _run_simulation_by_chunk(self, args_chunk):
-    #     """
-    #     This method simulates the EM response or computes the sensitivities for
-    #     a single sounding. The method allows for parallelization of
-    #     the stitched 1D problem.
-    #     """
-    #     n = len(args_chunk)
-    #     results = [
-    #                 self.run_simulation(args_chunk[i_sounding]) for i_sounding in range(n)
-    #     ]
-    #     return results
--- a/SimPEG/electromagnetics/base_1d_stitched.py
+++ b/SimPEG/electromagnetics/base_1d_stitched.py
@@ -139,7 +139,6 @@ class BaseStitchedEM1DSimulation(BaseSimulation):
         Returns
         -------
         numpy.ndarray of float
-        test
         """
         return self._topo
 
--- a/SimPEG/electromagnetics/base_1d_stitched.py
+++ b/SimPEG/electromagnetics/base_1d_stitched.py
@@ -1,0 +1,552 @@
+from scipy.constants import mu_0
+import numpy as np
+from ..simulation import BaseSimulation
+from .. import props
+from .. import utils
+from ..utils.code_utils import (
+    validate_integer,
+    validate_location_property,
+    validate_ndarray_with_shape,
+    validate_type,
+)
+###############################################################################
+#                                                                             #
+#                             BaseStitchedEM1DSimulation                      #
+#                                                                             #
+###############################################################################
+
+__all__ = ["BaseStitchedEM1DSimulation"]
+
+class BaseStitchedEM1DSimulation(BaseSimulation):
+    """
+    Base class for the stitched 1D simulation. This simulation models the EM
+    response for a set of 1D EM soundings.
+    """
+
+    _formulation = "1D"
+    _coefficients = []
+    _coefficients_set = False
+
+    # _Jmatrix_sigma = None
+    # _Jmatrix_height = None
+    # _J = None
+
+    # Properties for electrical conductivity/resistivity
+    sigma, sigmaMap, sigmaDeriv = props.Invertible(
+        "Electrical conductivity at infinite frequency (S/m)"
+    )
+
+    eta = props.PhysicalProperty("Intrinsic chargeability (V/V), 0 <= eta < 1")
+    tau = props.PhysicalProperty("Time constant for Cole-Cole model (s)")
+    c = props.PhysicalProperty("Frequency Dependency for Cole-Cole model, 0 < c < 1")
+
+    # Properties for magnetic susceptibility
+    mu, muMap, muDeriv = props.Invertible(
+        "Magnetic permeability at infinite frequency (SI)"
+    )
+    chi = props.PhysicalProperty(
+        "DC magnetic susceptibility for viscous remanent magnetization contribution (SI)"
+    )
+    tau1 = props.PhysicalProperty(
+        "Lower bound for log-uniform distribution of time-relaxation constants for viscous remanent magnetization (s)"
+    )
+    tau2 = props.PhysicalProperty(
+        "Upper bound for log-uniform distribution of time-relaxation constants for viscous remanent magnetization (s)"
+    )
+
+    # Additional properties
+    h, hMap, hDeriv = props.Invertible("Receiver Height (m), h > 0")
+
+    thicknesses, thicknessesMap, thicknessesDeriv = props.Invertible(
+        "layer thicknesses (m)"
+    )
+
+    def __init__(
+        self,
+        sigma=None,
+        sigmaMap=None,
+        thicknesses=None,
+        thicknessesMap=None,
+        mu=mu_0,
+        muMap=None,
+        h=None,
+        hMap=None,
+        eta=None,
+        tau=None,
+        c=None,
+        dchi=None,
+        tau1=None,
+        tau2=None,
+        fix_Jmatrix=False,
+        topo=None,
+        parallel=False,
+        n_cpu=None,
+        **kwargs,
+    ):
+        super().__init__(mesh=None, **kwargs)
+        self.sigma = sigma
+        self.sigmaMap = sigmaMap
+        self.mu = mu
+        self.muMap = muMap
+        self.h = h
+        self.hMap = hMap
+        if thicknesses is None:
+            thicknesses = np.array([])
+        self.thicknesses = thicknesses
+        self.thicknessesMap = thicknessesMap
+        self.eta = eta
+        self.tau = tau
+        self.c = c
+        self.dchi = dchi
+        self.tau1 = tau1
+        self.tau2 = tau2
+        self.fix_Jmatrix = fix_Jmatrix
+        self.topo = topo
+        if self.topo is None:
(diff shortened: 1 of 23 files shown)


SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+        eta=None,
+        tau=None,
+        c=None,
+        dchi=None,
+        tau1=None,
+        tau2=None,
+        fix_Jmatrix=False,
+        topo=None,
+        parallel=False,
+        n_cpu=None,
```
domfournier, line 84: Starting to be a pretty long list of arguments. We should revisit this later and try to reduce complexity

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+            self.set_null_topography()
+
+        self.parallel = parallel
+        self.n_cpu = n_cpu
+
+        if self.parallel:
+            if self.verbose:
+                print(">> Use multiprocessing for parallelization")
+                if self.n_cpu is None:
+                    self.n_cpu = multiprocessing.cpu_count()
```
domfournier, line 116: You could technically set it directly as default in your kwargs

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+        """
+        return self._topo
+
+    @topo.setter
+    def topo(self, value):
+        self._topo = validate_ndarray_with_shape("topo", value, shape=("*", 3))
+
+    @property
+    def parallel(self):
+        """Parallel
```
domfournier, line 152:
```suggestion
        """
        Run the computation as a parallel process.
```

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+    def n_sounding(self):
+        return len(self.survey.source_location_by_sounding_dict)
+
+    @property
+    def data_index(self):
+        return self.survey.data_index
+
+    # ------------- For physical properties ------------- #
+    @property
+    def Sigma(self):
```
domfournier, line 210: Technically not PEP8, attributes should be lower case...

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+        if getattr(self, "_Sigma", None) is None:
+            # Ordering: first z then x
+            self._Sigma = self.sigma.reshape((self.n_sounding, self.n_layer))
+        return self._Sigma
+
+    @property
+    def Thicknesses(self):
```
domfournier, line 217: Same here

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+            if self.tau is None:
+                self._Tau = 1e-3 * np.ones(
+                    (self.n_sounding, self.n_layer), dtype=float, order="C"
+                )
+            else:
+                self._Tau = self.tau.reshape((self.n_sounding, self.n_layer))
+        return self._Tau
+
+    @property
+    def C(self):
```
domfournier, line 253: Add docstrings

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+    #         self.Chi[i_sounding, :],
+    #         self.dChi[i_sounding, :],
+    #         self.Tau1[i_sounding, :],
+    #         self.Tau2[i_sounding, :],
+    #         self.H[i_sounding],
+    #         'forward',
+    #         True,
+    #         [],
+    #     )
+    #     return output
```
domfournier, line 389:
```suggestion
```

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+    # @property
+    # def sounding_number_chunks(self):
+    #     self._sounding_number_chunks = list(self.chunks(self.sounding_number, self.n_sounding_for_chunk))
+    #     return self._sounding_number_chunks
+
+    # def input_args_by_chunk(self, i_chunk, output_type):
+    #     args_by_chunks = []
+    #     for i_sounding in self.sounding_number_chunks[i_chunk]:
+    #         args_by_chunks.append(self.input_args(i_sounding, output_type))
+    #     return args_by_chunks
```
domfournier, line 445:
```suggestion
```

SimPEG/electromagnetics/frequency_domain/simulation_1d.py
```diff
@@ -149,28 +149,31 @@ def getJ(self, m, f=None):
...
+                #             for rx in src.receiver_list
+                #         )
+                #     else:
+                #         nD = sum(rx.locations.shape[0] for rx in src.receiver_list)
+                #     ip1 = i + nD
+                #     # v = np.exp(-lambs[i:ip1] * h)
+                #     C0s_dh[i:ip1] *= - lambs[i:ip1]
+                #     C1s_dh[i:ip1] *= - lambs[i:ip1]
+                #     i = ip1
+                # J will be n_d * n_src (each source has it's own h)...
```
domfournier, line 172:
```suggestion
```

SimPEG/electromagnetics/time_domain/simulation_1d.py
```diff
@@ -261,17 +264,23 @@ def getJ(self, m, f=None):
...
+                #     # if rx.use_source_receiver_offset:
+                #     #     dz = rx.locations[:, 2]
+                #     # else:
+                #     #     dz = rx.locations[:, 2] - src.location[2]
+                #     nD = sum(rx.locations.shape[0] for rx in src.receiver_list)
+                #     ip1 = i + nD
+                #     C0s_dh[i:ip1] *= 2 * -lambs[i:ip1]
+                #     C1s_dh[i:ip1] *= 2 * -lambs[i:ip1]
+                #     i = ip1
+                #     # J will be n_d * n_src (each source has it's own h)...
```
domfournier, line 281:
```suggestion
```

SimPEG/electromagnetics/frequency_domain/survey.py
```diff
@@ -16,13 +16,22 @@ def __init__(self, source_list, **kwargs):
...
         _frequency_dict = {}
+        _source_location_dict = {}
+        _source_location_by_sounding_dict = {}
         for src in self.source_list:
             if src.frequency not in _frequency_dict:
                 _frequency_dict[src.frequency] = []
             _frequency_dict[src.frequency] += [src]
+            if src.i_sounding not in _source_location_dict:
+                _source_location_dict[src.i_sounding] = []
+                _source_location_by_sounding_dict[src.i_sounding] = []
```
domfournier, line 27: Would suggest creating your dictionary with the sources directly as keys and location as values, so we can avoid looping again and directly form arrays on `dict.values()`

SimPEG/electromagnetics/base_1d_stitched.py
```diff
@@ -0,0 +1,544 @@
...
+    def n_chunk(self):
+        self._n_chunk = len(self.sounding_number_chunks)
+        return self._n_chunk
+
+    @property
+    def source_locations_for_sounding(self):
+        if getattr(self, "_source_locations_for_sounding", None) is None:
+            self._source_locations_for_sounding = np.vstack(
+                [
+                    self.survey._source_location_by_sounding_dict[ii][0]
```
domfournier, line 425: See comment below about changing the dictionary, so you can do directly `np.vstack(self.surveys._source_location_by_sounding.value)`

SimPEG/electromagnetics/frequency_domain/survey.py
```diff
@@ -16,13 +16,22 @@ def __init__(self, source_list, **kwargs):
         super(Survey, self).__init__(source_list, **kwargs)
 
         _frequency_dict = {}
+        _source_location_dict = {}
+        _source_location_by_sounding_dict = {}
         for src in self.source_list:
```
domfournier, line 21: Not ideal to have a long loop in the init. Probably better to do the operation on-demand only under the property.

SimPEG/electromagnetics/frequency_domain/survey.py
```diff
@@ -97,3 +106,34 @@ def get_sources_by_frequency(self, frequency):
             frequency in self._frequency_dict
         ), "The requested frequency is not in this survey."
         return self._frequency_dict[frequency]
+
+    @property
+    def source_location_by_sounding_dict(self):
```
domfournier, line 111: Type the return property, so no need to put it in the name.
```suggestion
    def source_location_by_sounding(self) -> dict:
```

SimPEG/electromagnetics/frequency_domain/survey.py
```diff
@@ -97,3 +106,34 @@ def get_sources_by_frequency(self, frequency):
...
+        :rtype: dictionary
+        :return: sources at the sepcified source location
+        """
+        assert (
+            i_sounding in self._source_location_dict
+        ), "The requested sounding is not in this survey."
+        return self._source_location_dict[i_sounding]
+
+    @property
+    def vnD_by_sounding_dict(self):
```
domfournier, line 130: Same here, type it and remove from name.
```suggestion
    def vnD_by_sounding(self) -> dict:
```

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
```diff
@@ -0,0 +1,208 @@
...
+from scipy import sparse as sp
+from ... import utils
+from ..base_1d_stitched import BaseStitchedEM1DSimulation
+from .simulation_1d import Simulation1DLayered
+from .survey import Survey
+from ... import maps
+from multiprocessing import Pool
+
+
+def run_simulation_time_domain(args):
```
domfournier, line 11: Why not expose the args? We could probably use a dataclass at some point to reduce the complexity.

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
```diff
@@ -0,0 +1,208 @@
...
+    sim = Simulation1DLayered(
+        survey=local_survey,
+        thicknesses=thicknesses,
+        sigmaMap=sigma_map,
+        hMap=h_map,
+        eta=eta,
+        tau=tau,
+        c=c,
+        topo=topo,
+        hankel_filter="key_101_2009",
```
domfournier, line 71: Should probably be passed as a keyword with default.

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
```diff
@@ -0,0 +1,208 @@
...
+
+                pool = Pool(self.n_cpu)
+
+                # Deprecate this for now, but revisit later
+                # It is an idea of chunking for parallelization
+                # if self.n_sounding_for_chunk is None:
+                self._J = pool.map(
+                    run_simulation,
+                    [
+                        self.input_args(i, output_type="sensitivity")
```
domfournier, line 170: Looks like we are re-instantiating classes on every call? Can we not store the simulation?

domfournier (commented):
Thanks for the work Seog. Left some comments, but we can always tackle some of those later for efficiency. See what you think.

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
sgkang: That's a good point. Yeah, I am instantiating classes on every call. I felt that this is more efficient than storing it and passes to child, but I could be wrong.

SimPEG/electromagnetics/base_1d_stitched.py
sgkang: I am open to suggestions.

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
sgkang: no particular reasons. my dirty implementation for multiprocessing.

SimPEG/electromagnetics/time_domain/simulation_1d_stitched.py
sgkang: again open to any suggestions.

sgkang:
@domfournier thanks for the reviews! I updated the codes correspondingly except for your suggestions about h--> height. This need changes in original 1d code. Using `kwargs` seems a good idea, but I was not sure where to start, so any suggestions are welcomed!

domfournier (approved):
Thanks @sgkang . I left another batch of comments, but the majority is stuff we can tackle later for efficiency.

jcapriot:
Hey @sgkang,

I haven't had a lot of time to delve deeply into this, but my initial impression is there are a lot of code duplications inside that could likely be simplified by internally using a `MetaSimulation`. Which would then enable us to take advantage of the multiple parallelization approaches used therein.

Also, can you potentially parse out the bug fixes vs the new additions?

SimPEG/electromagnetics/frequency_domain/survey.py
```diff
@@ -16,13 +16,22 @@ def __init__(self, source_list, **kwargs):
...
+        _source_location_by_sounding = {}
         for src in self.source_list:
             if src.frequency not in _frequency_dict:
                 _frequency_dict[src.frequency] = []
             _frequency_dict[src.frequency] += [src]
+            if src.i_sounding not in _source_location:
+                _source_location[src.i_sounding] = []
+                _source_location_by_sounding[src.i_sounding] = []
+            _source_location[src.i_sounding] += [src]
+            _source_location_by_sounding[src.i_sounding] += [src.location]
```
thibaut-kobold, line 29: @sgkang : this seems to break the NSEM survey that inherits from FDEM survey (because NSEM has no source location)

`'PlanewaveXYPrimary' object has no attribute '_location'`

domfournier:
@sgkang I realize that we are sort of dogpiling on your PR. Do you need/want help with this? I don't mind making a PR onto yours for some of those suggestions.

SimPEG/electromagnetics/frequency_domain/survey.py
thibaut-kobold: PS: it can be solved by explicitly setting `location=None` for NSEM.

lheagy:
Hi all, we have been leaving this pr hanging for quite a while..... Part of the challenge is that the comments here and on #1314 are not easily actionable. @jcapriot, @santisoler, @sgkang, and @domfournier: do you think it would be feasible to capture some todos here or start to sketch out a path forward? We can dedicate some of the upcoming SimPEG meetings to talking about this, but it would be great if we could get a brain-dump from everyone written down so that we can turn it into some actionable steps that can be picked up again.

santisoler:
Thanks for commenting on this @lheagy.

I can leave a few points I think might be relevant for this PR:

- We need to update this against the latest `main` branch. There are quite a bunch of conflicts with `main`, and some code and docs style should be updated (check against Black and flake8, and use numpydoc style for the docstrings).
- @sgkang mentioned a bug in `base_1d.py`. I suggest opening a different PR to solve it, and/or an issue to report it. Bugfixes should be solved as quickly as possible, and not blocked by all the design decisions and work needed when adding a new feature.
- One major concern I have in this PR is the addition of attributes to parent classes that are only needed for particular cases. For example, the new `i_sounding` attribute in the `BaseTDEMSrc` and `BaseFDEMSrc` classes. Outside a stitched inversion, these new attributes don't have any use.
- Moreover, the `i_sounding` attribute in particular is an index for a given source. I don't quite like that objects in a collection keep track of their indices within the collection. A float inside a list doesn't know that it's part of a collection, it's the responsibility of the collection to keep track of its elements and their order. I think we could keep track of source indices from outside the sources, making them agnostic that they belong to a collection.

I think more points will raise, but I think we can start a conversation from these ones.

jcapriot:
I would love to see this implemented with the `MetaSimulation` instead, which should remove the need for the `i_sounding` attribute. I'd be curious to see if there's much slowdown between them, but using the `MetaSimulation` would allow us to easily support multiple parallelism schemes without any extra work on this class.

jcapriot:
Honestly, I would like to see this parsed into at least three separate PR's.
* One for the 1D simulation bug fix
* One for just the regularization
* One implementing the 1D stitched simulation.

Issue #1301, opened 2023-11-25
Possibly wrong derivative in `VectorAmplitude`


thibaut-kobold:
https://github.com/simpeg/simpeg/blob/211f627874c46d3a03450b840970b9a10a7fbe4b/SimPEG/regularization/vector.py#L429

@domfournier : For the first derivative `deriv`, we are calling twice `f_m_deriv`. I think the second one is potentially wrong. It should be `f_m` (and potentially need then to remove `d_m`)

See for reference the derivative of the base class: https://github.com/simpeg/simpeg/blob/211f627874c46d3a03450b840970b9a10a7fbe4b/SimPEG/regularization/base.py#L502C17-L502C17

lheagy:
@johnweis0480 has run into some questions here to. Would you have an example to share?

johnweis0480:
I was looking at this a while back and saw that the derivative test for the smoothness would fail.

![image](https://github.com/simpeg/simpeg/assets/49649694/89759feb-9e0f-468c-bdc5-98e0628a3c15)

What I initially implemented and suggested last spring was to take the amplitude of the cell gradients of the components of magnetization instead of taking the cell gradient of the amplitude. I believe that the failure is because the derivative is actually matching my initial implementation of this regularization for the smoothness.

Currently, the f_m of the vector smoothness class is doing this:

![image](https://github.com/simpeg/simpeg/assets/49649694/bc317b56-95c7-4aae-a051-8e3bfab7da35)

Instead, I believe is should be doing this
![image](https://github.com/simpeg/simpeg/assets/49649694/6c5e00f6-6b0b-4718-bc8b-23a3356e9d57)

to match the derivative and my initial implementation. By changing to the second, the derivative test will pass.

![image](https://github.com/simpeg/simpeg/assets/49649694/e081680b-ee13-4bbd-98ad-156d7fdf82f3)

I do not believe that the format of f_m is well defined as it provides no information on the direction of magnetization. I did compare using cartesian smoothness (top row) to my implementation (bottom row) for MVI with positive results.

jcapriot:
@johnweis0480 What you're suggesting is taking the amplitude of the gradient of the vector components, which is not the same thing as the gradient of the vector amplitudes.

lheagy:
[...]

PR #1303, opened 2023-12-01
Update test_regularization.py

lheagy:
Following on the discussion in #1301, it doesn't look like we are testing the derivatives for the vector amplitude. This pr adds a derivative test for the vector amplitude regularization. When running locally, I obtain the following: 

```
Testing VectorAmplitude Deriv
==================== check_derivative ====================
iter    h         |ft-f0|   |ft-f0-h*J0*dx|  Order
---------------------------------------------------------
 0   1.00e-01    9.980e-04     9.697e-03      nan
 1   1.00e-02    1.858e-03     9.881e-04      0.992
 2   1.00e-03    2.060e-04     1.190e-04      0.919
 3   1.00e-04    2.080e-05     1.210e-05      0.993
*********************************************************
<<<<<<<<<<<<<<<<<<<<<<<<< FAIL! >>>>>>>>>>>>>>>>>>>>>>>>>
*********************************************************
Boooooooo  :(
```

We can chat further on #1301 on how best to handle this.

### DIFF
--- a/tests/base/test_regularization.py
+++ b/tests/base/test_regularization.py
@@ -574,6 +574,8 @@ class RegularizationTests(unittest.TestCase):
             reg.objfcts[0].f_m(model.flatten(order="F")), np.linalg.norm(model, axis=1)
         )
 
+        reg.test(model.flatten(order="F"))
+
 
 def test_WeightedLeastSquares():
     mesh = discretize.TensorMesh([3, 4, 5])


domfournier:
Just brought this back from out of the freezer.

Interestingly, the Amplitude reg was tested and passing under the generic `ReguarizationTests.test_regulatization`, which only uses a single component vector. For multi-components, the `AmplitudeSmallness` passes second-order by itself. But as you pointed out @lheagy on #1301 , the `AmplitudeSmoothnessFirstOrder` is only first-order. Sort of makes sense that applying the derivatives of amplitude evenly on all components isn't really penalizing like we want.

Few options here I see:
1- We just add a warning that the AmplitudeSmoothness is first-order (potential issues)
2- We change the implementation on `AmplitudeSmoothmessFirstOrder` to use @johnweis0480 approach. We might have to re-think the base Amplitude class, but not a big deal.

lheagy:
Thanks @domfournier! My suggestion would be that we go with (2). I think it is misleading to have an objective function that doesn't match what the derivative is doing. 

As @johnweis0480, pointed out, the derivatives match an objective function that is written as follows

image

where we take the difference and then take derivatives. 

This is different than what our current objective function states where we compute the amplitude and then differences 

image

In John's research, he tested out both approaches. If we first compute amplitudes and then take differences, there are a lot of local minima (e.g. Cell 1: [1, 0, 0], cell 2: [0, 1, 0] both have amplitude 1. If we compute amplitude and then take differences the objective function value is zero, whereas if we were to take differences and then compute amplitude there is still progress to be made). He found that this approach ends up line searching a lot because it is littered with local minima.  

Another argument for updating the objective function and leaving the derivatives alone and updating the objective function definition is that the regularization is working well. The influence of the regularization is entirely based on the derivatives as they set the direction of the gradient updates (we don't do anything with the actual evaluation of the regularization other than look to see if it decreases). So making this change would not change the behaviour of the reg, it would only change what is printed in the table that we output

domfournier:
I like it. @johnweis0480 do you have a branch ready for PR? Sounds like an easy quick win.

johnweis0480:
@domfournier Yep sounds good just put up a new pull request with the changes

lheagy:
Closing this in favor of #1329

PR #1310, opened 2023-12-20
Re-implement distance weighting and add a strategy comparison

thibaut-kobold:
#### Squash and Merge summary

Add a new `distance_weighting` function with a new Numba-based implementation that makes it more memory efficient that using Scipy's `spatial_distance.cdist`. Add more tests and a tutorial comparing depth, distance, and sensitivity weighting.

---

#### Summary
distance weighting was sitting in an old location, with old naming convention, and quite inefficient to compute.
This PR brings distance weighting at the same level of maturity as depth weighting, with a 10x speed up.
This PR also adds a new tutorial comparing depth, distance, and sensitivity weighting.

#### PR Checklist
      to a Pull Request

#### Additional information

**Visualizing weights**
image

**efficiency comparison with former `get_dist_wgt` on a small example**

Screenshot 2023-12-19 at 6 26 22 PM

**comparison with former `get_dist_wgt` on a small example**
image
image


jcapriot:
Where was distance weighting previously? I don’t see any removals here in my quick glance.

The other thing is that this creates multiple large dense matrices of (n_rx, n_cells), which will likely put a lot of pressure on memory.

jcapriot:
Also, I don’t think you need to put in the deprecated keyword argument check here since it’s a new function. No need to create something that is automatically deprecated.

thibaut-kobold:
Hi @jcapriot : this is where distance weighting is located at the moment: https://docs.simpeg.xyz/content/api/generated/SimPEG.potential_fields.get_dist_wgt.html#SimPEG.potential_fields.get_dist_wgt

I have not removed or depreciated it yet. It is not used anywhere (and is not even tested). What do you think? Can we simply remove it, or depreciate it but still functional, or depreciate it and throw an error pointing to the new function?

thibaut-kobold:
> The other thing is that this creates multiple large dense matrices of (n_rx, n_cells), which will likely put a lot of pressure on memory.

There is one big matrix `distance` indeed. For potential field, this is the same size as the sensitivity matrix that most keep in memory. I don't think the code is keeping it once we return the final (n_cells) array.

jcapriot:
Thanks for pointing to that. I would say we should just mark that function as deprecated and point to the new one.

Just so that everything is clear, the old function computes the radial distance to 8 points surrounding a cell center, and uses the average weight, as opposed to this implementation which only computes to the cell center. This likely accounts for much of the speedup you're seeing (8 calculations in 3D vs 1 calculation here). Let me track through the history to see who originally wrote the function so we can understand the reasoning

jcapriot:
Looks like this is the oldest bit I can find before the history is lost: https://github.com/simpeg/simpeg/blame/487d178b706e0d5549a7f9a2494d5b737f18474e/SimPEG/potential_fields/magnetics/simulation.py#L1068

In there it says that @domfournier created it in 2015.

domfournier:
I probably got this from a paper or the old Fortran code. I don't have an opinion to go whichever way, other than intuitively the nodal approach would be closer to the sensitivity calc.
I can do a review too now.

SimPEG/utils/model_utils.py
```diff
@@ -195,3 +198,95 @@ def depth_weighting(
...
+        cell_centers = mesh.cell_centers[active_cells]
+        cell_volumes = mesh.cell_volumes[active_cells]
+        n, d = cell_centers.shape
+        t, d1 = reference_locs.shape
+
+        if not d == d1:
+            raise Exception("vectors must have same number of columns")
+
+        # vectorized distance calculations
+        distance = (
```
jcapriot, line 276: This one calculation requires several times the sensitivity matrix size:

It creates one when you do:
`np.dot((cell_centers**2, np.ones((d, t))  #(n_cell, 3) * (3, n_rx) = (n_cell, n_rx)`

It creates another when you do:
`np.dot(np.ones([n, d]), (reference_locs**2.0).T) #(n_cell, 3) * (3, n_rx) = (n_cell, n_rx)`

It creates a third when you do:
`np.dot(cell_centers, reference_locs.T)  #(n_cell, 3) * (3, n_rx) = (n_cell, n_rx)`


it creates another when you add a few together another when you do subtract (at which point some of the intermediaries can be garbage collected).

thibaut-kobold: what would be your preferred path forward?

SimPEG/utils/model_utils.py
```diff
@@ -195,3 +198,95 @@ def depth_weighting(
...
+            np.dot((cell_centers**2.0), np.ones([d, t]))
+            + np.dot(np.ones([n, d]), (reference_locs**2.0).T)
+            - 2.0 * np.dot(cell_centers, reference_locs.T)
+        ) ** 0.5
+
+    dist_weights = (
+        (
+            (
+                (cell_volumes.reshape(-1, 1) / ((distance + threshold) ** exponent))
```
jcapriot, line 285: this line then again will increase the memory usage by creating yet another intermediate sensitivity matrix sized array

SimPEG/utils/model_utils.py
jcapriot: The only way to get around it is to painfully do a python loop over the receiver locations... I think this is just one of those times where you can't avoid it and must take the time hit (alternatively this would make a great candidate for `numba` acceleration @santisoler).

thibaut-kobold: `for` loops across datapoint to keep `n_cells` array?

domfournier: I would agree that a parallel loop would make sense here, since we are just summing all the rows at the end.

jcapriot: yeah, usually `n_rx` << `n_cells` so if we have to do a loop, make it do as much work as possible

thibaut-kobold:
I am not familiar with `numba` but I took a very quick shot at it, proposing the option to either use `numba` or `numpy` (similar in idea with the choice of `geoana` or `choclo` for PF)

SimPEG/utils/model_utils.py
thibaut-kobold: took a stab at it (not familiar with numba coding). See what you think

SimPEG/utils/model_utils.py
```diff
@@ -1,9 +1,11 @@
 import warnings
+from typing import Literal
 
 import discretize
 import numpy as np
 import scipy.sparse as sp
 from discretize.utils import active_from_xyz
+from numba import njit
```
santisoler, line 8: Numba is not a required dependency, so we should try to import it:

```python
try:
    from numba import njit
except ImportError:
    # Define dummy njit decorator
    def njit(*args, **kwargs):
        return lambda f: f
```

SimPEG/utils/model_utils.py
```diff
@@ -258,17 +263,45 @@ def distance_weighting(
...
+        ):
+            distance_weights = np.zeros(len(cell_centers))
+            for _, rl in enumerate(reference_locs):
+                dst_wgt = (
+                    np.sqrt(((cell_centers - rl) ** 2).sum(axis=1)) + threshold
+                ) ** exponent
+                dst_wgt = (cell_volumes / dst_wgt) ** 2
+                distance_weights += dst_wgt
+
+            return distance_weights
```
santisoler, line 292: I would move this function definition outside of the public function, for example move it to 
```python
@njit(parallel=True)
def _distance_weighting_numba(...):
    ...
```

By defining numba decorated functions within another function body, the interpreter will compile it when it needs to run it, but as soon as we leave the scope of `distance_weighting`, it will throw the compiled bits to the garbage collector. Therefore, every time a user runs this function, it will have to recompile the Numba function.

On the other hand, if it lives outside of it, it will compile it only the first time it needs to use it. Then, any further call to `distance_weighting` will reuse those compiled bits.

SimPEG/utils/model_utils.py
```diff
@@ -258,17 +263,45 @@ def distance_weighting(
...
+        @njit(parallel=True)
+        def distance_weighting_numba(
+            cell_centers: np.ndarray,
+            cell_volumes: np.ndarray,
+            reference_locs: np.ndarray,
+            threshold: float,
+            exponent: float = 2.0,
+        ):
+            distance_weights = np.zeros(len(cell_centers))
+            for _, rl in enumerate(reference_locs):
```
santisoler, line 285: No need to use `enumerate` if you're not assigning the index. Also, this for loop is not being run in parallel. To do so you need to use `numba.prange`: https://numba.pydata.org/numba-doc/latest/user/parallel.html

Something like:
```python
n_reference_locs = reference_locs.shape[0]  # check this, not sure if it's right
for i in prange(n_reference_locs):
    rl = reference_locs[i]
    ...
```

santisoler (commented):
Just leaving two cents on improving the numba implementation. See comments below.

SimPEG/utils/model_utils.py
```diff
@@ -1,10 +1,30 @@
...
+        return lambda f: f
+
+    # Define dummy prange function
+    prange = range
+
+    warnings.warn(
+        "numba is not installed. Some computations might be slower.",
+        type=ImportWarning,
+        stacklevel=2,
+    )
```
santisoler, line 27: I would move this warning to the function body, otherwise every user will receive it after importing SimPEG. Moreover, I would be more explicit on what computations might be slower and maybe point people to instructions on how to install Numba.
 
To move it to the function's body you could include an `import numba` in the `try` statement and during the exception assign it to `None`:

```python
try:
    import numba
    from numba import njit, prange
except ImportError:
    numba = None
    # Define dummy njit decorator
    def njit(*args, **kwargs):
        return lambda f: f

    # Define dummy prange function
    prange = range
```

And then in the function:

```python

def depth_weighting(
    mesh, reference_locs, active_cells=None, exponent=2.0, threshold=None, **kwargs
):
    if numba is None:
        warnings.warn(...)
```

santisoler (commented):
That looks better! I left another minor suggestion below

SimPEG/utils/model_utils.py
thibaut-kobold: Done!
Note: i removed the `ImportWarning` because it is ignored by default (thus it would not have shown except in debug mode)

santisoler: Oh, yes. We shouldn't raise an `ImportWarning` in this case, just a `UserWarning` should be fine.

dccowan:
It's great that you made a tutorial. Please note that new tutorials are being developed within github.com/simpeg/user-tutorials ; see https://simpeg.xyz/user-tutorials/ .

Another option would be add the discussion of weighting to the current tutorials for magnetic and gravity inversion.

santisoler (changes requested):
Hi @thibaut-kobold! Sorry for taking a while to review your PR. I went through
it today and it looks great. I appreciate you took the effort to implement the
distance weights with Numba.

I left some minor comments below (check the inline suggestions and comments),
and I have some few thoughts regarding this PR.

## Values for the `engine` parameter

Without any context, the strings `"loop"` and `"cdist"` are not that meaningful
when choosing which method to compute the distance weights. What if we change
these values to `"numba"` and `"scipy"` instead? I'm open to hear other
ideas for them!


## Definition of the distance weights

While reading the code it took me a while to figure out how the distance
weights are defined. Would be nice to include a mathematical expression for
each element of the distance weights array so users can explicitly read its
definition, and contributors can have a better understanding of what the
implementation should calculate.

By reading the code, I tracked down the following definition of the distance
weights. If I define $\mathbf{w}$ as the *unnormalized* weights array, the
$j$-th element of this array (related to the $j$-th active cell) $w_j$ can be
computed as:

$$
w_j =
\frac{1}{V_j}
\sqrt{
  \sum\limits_{i=1}^N
  \left[
    \frac{V_j}{(|\mathbf{p}_i - \mathbf{q}_j| + \epsilon)^\nu}
  \right]^2
}
$$

where:

- $N$ is the total number of reference locations
- $\mathbf{p}_i$ is the $i$-th reference location
- $\mathbf{q}_j$ is the $j$-th cell center
- $V_j$ is the volume of the $j$-th cell
- $\epsilon$ is the threshold
- $\nu$ is the exponent

In this definition, the weights $w_j$ are independent of the cell volume since
they cancel out:

$$
w_j =
\frac{1}{V_j}
\sqrt{
  V_j^2
  \sum\limits_{i=1}^N
    (|\mathbf{p}_i - \mathbf{q}_j| + \epsilon)^{-2\nu}
} 
$$

$$
w_j =
\sqrt{
  \sum\limits_{i=1}^N
    (|\mathbf{p}_i - \mathbf{q}_j| + \epsilon)^{-2\nu}
}
$$

Is this the intended definition of the distance weights? If that's so, then the
numba function doesn't need to take the cell volumes. Feel free to correct me
if I'm wrong.


## Numba implementation

The current Numba implementation is working great, it's parallelizing over the
reference locations and computing the distance weights with less memory
requirements than trying to build the distance matrix with `cdist`.

Nonetheless, there are a few things we could improve to make it run faster and
be more memory efficient.

Since each element of the *unnormalized* distance weights array is independent
from each other, it sounds better to me to parallelize the loop over the active
cells in the mesh, so each thread will be writing to a single element of the
`distance_weights` array. This way we avoid racing conditions in which two
threads try to write the same memory address, and therefore will have to wait
for the first one to finish.

The following line is creating an array with the same shape as the
`cell_centers` just to then reduce it by an aggregation. I'm not 100% sure on
this, but I suspect that Numba will try to allocate that array in memory before
performing the aggregation, a process that would take more time and memory.

```python
dst_wgt = (
    np.sqrt(((cell_centers - rl) ** 2).sum(axis=1)) + threshold
) ** exponent

```

Instead, we can perform the same task with a second for loop that iterates over
the active cells. And as I mentioned before, I would make this loop the first
one and parallelize over it.


With these ideas in mind, I wrote this other implementation that according to
some quick benchmarks I ran, it performs better than the current one:

```python
@njit(parallel=True)
def _distance_weighting_numba(
    cell_centers: np.ndarray,
    cell_volumes: np.ndarray,
    reference_locs: np.ndarray,
    threshold: float,
    exponent: float = 2.0,
) -> np.ndarray:
    """
    ...
    """
    n_active_cells = cell_centers.shape[0]
    n_reference_locs = len(reference_locs)

    distance_weights = np.zeros(len(cell_centers))
    for j in prange(n_active_cells):
        cell_center = cell_centers[j]
        for i in range(n_reference_locs):
            reference_loc = reference_locs[i]
            distance = np.sqrt(((cell_center - reference_loc) ** 2).sum())
            distance_weights[j] += (distance + threshold) ** (-2 * exponent)

    distance_weights **= np.sqrt(distance_weights)
    distance_weights /= np.nanmax(distance_weights)
    return distance_weights

```

Worth noting that this function doesn't use the `cell_volumes`. If the
definition is wrong, this implementation should be modified accordingly.

simpeg/potential_fields/base.py
```diff
@@ -486,6 +487,12 @@ def get_dist_wgt(mesh, receiver_locations, actv, R, R0):
     wr : (n_cell) numpy.ndarray
         Distance weighting model; 0 for all inactive cells
     """
+    warnings.warn(
+        "The get_dist_wgt function has been deprecated, please import "
+        "SimPEG.utils.distance_weighting. This will be removed in SimPEG 0.22.0",
+        FutureWarning,
+        stacklevel=2,
```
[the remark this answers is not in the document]
jcapriot: It should be at 2, then the warning is issued on the line that calls `get_dist_wgt`.

jcapriot:
> Since each element of the unnormalized distance weights array is independent from each other, it sounds better to me to parallelize the loop over the active cells in the mesh, so each thread will be writing to a single element of the distance_weights array. This way we avoid racing conditions in which two threads try to write the same memory address, and therefore will have to wait for the first one to finish.

Just fyi, from what I understand the llvm compiler should recognize that there is a reduction operation here when in parallel. They don’t write to the same location during parallel execution, but parallelizing over the active cells is likelier faster as well, because the operation over the data is more aligned in memory.

simpeg/potential_fields/base.py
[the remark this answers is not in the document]
jcapriot: 2 is the correct level to get what you aim at. You want it to issue the warning at the location the function is called.

1 would make it warn specifically on this line within the `get_dst_wgt` function.

thibaut-kobold:
i am only coming back to this PR now: thank you for the in-depth comment, and proposing an actual working implementation of an alternative; this is great stuff.
Anyway, I swapped my function with yours; things looks good, but tests are failing for things that seems unrelated to that PR

simpeg/utils/model_utils.py
```diff
@@ -150,3 +169,174 @@ def depth_weighting(
...
+    return distance_weights
+
+
+def distance_weighting(
+    mesh: discretize.base.BaseMesh,
+    reference_locs: np.ndarray,
+    active_cells: Optional[np.ndarray] = None,
+    exponent: Optional[float] = 2.0,
+    threshold: Optional[float] = None,
+    engine: Literal["numba", "scipy"] = "numba",
```
jcapriot, line 232: Usually not good to have a string be the default parameter,

I'd recommend:
* Default to None:
    * If `numba` is available, use that
    * Else use `scipy`

tests/base/test_model_utils.py
```diff
@@ -77,6 +74,98 @@ def test_depth_weighting_2D(self):
...
+class TestDistancehWeighting:
+    def test_distance_weighting_3D(self):
+        # Mesh
+        dh = 5.0
+        hx = [(dh, 5, -1.3), (dh, 40), (dh, 5, 1.3)]
+        hy = [(dh, 5, -1.3), (dh, 40), (dh, 5, 1.3)]
+        hz = [(dh, 15)]
+        mesh = TensorMesh([hx, hy, hz], "CCN")
+
+        actv = np.random.randint(0, 2, mesh.n_cells) == 1
```
jcapriot, line 86: Use numpy's `default_rng(seed)` with a set seed for repeatable tests

tests/base/test_model_utils.py
```diff
@@ -77,6 +74,98 @@ def test_depth_weighting_2D(self):
...
+        hx = [(dh, 5, -1.3), (dh, 40), (dh, 5, 1.3)]
+        hz = [(dh, 15)]
+        mesh = TensorMesh([hx, hz], "CN")
+
+        actv = np.random.randint(0, 2, mesh.n_cells) == 1
+
+        reference_locs = (
+            np.random.rand(1000, 2) * (mesh.nodes.max(axis=0) - mesh.nodes.min(axis=0))
+            + mesh.origin
+        )
```
jcapriot, line 120: Again, use numpy's `default_rng(seed)` with a set seed for repeatable tests

tests/base/test_model_utils.py
```diff
@@ -77,6 +74,98 @@ def test_depth_weighting_2D(self):
...
+        dh = 5.0
+        hx = [(dh, 5, -1.3), (dh, 40), (dh, 5, 1.3)]
+        mesh = TensorMesh([hx], "C")
+
+        actv = np.random.randint(0, 2, mesh.n_cells) == 1
+
+        reference_locs = (
+            np.random.rand(1000, 1) * (mesh.nodes.max(axis=0) - mesh.nodes.min(axis=0))
+            + mesh.origin
+        )
```
jcapriot, line 143: Use numpy's `default_rng(seed)` with a set seed for repeatable tests

jcapriot (changes requested):
One comment regarding the default value of engine,
And make your tests repeatable using a `numpy.random.Generator` object.

simpeg/utils/model_utils.py
santisoler: I see your point. I think I'd prefer having code that run slow than code that might fill all available memory if the 2d array that scipy creates is too large.

I'd like to leave the default behavior as it is for now at least.

santisoler (approved):

simpeg/utils/model_utils.py
jcapriot: If that’s so, the default should be a for loop over data points, with vectorized operations along the model cells. Not as fast as numba, but still memory efficient.

jcapriot: But, future issues.

santisoler: You are right. Let's work on that later on, so we don't block this PR now.

Issue #1298, opened 2023-11-06
DOC: Typo in codes comments


anna1963:
### Issue with current documentation:

Here is a screenshot from: https://docs.simpeg.xyz/content/tutorials/05-dcr/plot_fwd_2_dcr2d.html
It says “10% Gaussian noises” in the comment but using np.random.rand (uniform distribution) to generate the noises instead of using np.random.normal, so it's supposed to be "5% uniform distribution". Either correct codes or correct comments. Thanks!
<img width="393" alt="Screen Shot 2023-11-05 at 5 22 32 PM" src="https://github.com/simpeg/simpeg/assets/68674174/50c05570-d7f9-4bcf-8bb0-8ac3eba668c8">

PR #1312, opened 2023-12-27
ISSUE-1298: Use normal distributed noise in example.

domfournier:
Change reference to random.rand to random.randn for a centred normal distribution.

Address issue #1298

### DIFF
--- a/tutorials/03-gravity/plot_1a_gravity_anomaly.py
+++ b/tutorials/03-gravity/plot_1a_gravity_anomaly.py
@@ -232,6 +232,6 @@ if save_output:
 
     np.random.seed(737)
     maximum_anomaly = np.max(np.abs(dpred))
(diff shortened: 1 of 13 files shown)


jcapriot (approved):
Looks fine to me

santisoler (approved):
Looks good! Feel free to merge 🚀 

Just one comment for future PRs on Numpy's random number generation: I think we should move from using the `np.random....` functions for the new random number generator class: https://numpy.org/doc/stable/reference/random/generator.html

This class introduces a way more transparent way on handling random seeds and also a more verbose interface: numbers with normal distributions can be generated with `rng.normal()`, while using a uniform distribution can be done with `rng.uniform()`.

I've started applying these changes in https://github.com/simpeg/simpeg/pull/1286

Again, we should merge this PR as it is!

domfournier:
Ok, rreat. 
You will have to merge @santisoler, I don't have the rights.

PR #1313, opened 2023-12-29
ISSUE-1301: Explicit definition of derivatives on AmplitudeSmoothnessFirstOrder

domfournier:
Addresses issue #1301

### DIFF
--- a/SimPEG/regularization/vector.py
+++ b/SimPEG/regularization/vector.py
@@ -923,6 +923,29 @@ class AmplitudeSmoothnessFirstOrder(SparseSmoothness, BaseAmplitude):
         return self.cell_gradient @ a
 
     def f_m_deriv(self, m):
+        r"""
+        Derivative of the regularization kernel function.
+
+        For first-order smoothness regularization in the x-direction, the derivative of the
+        regularization kernel function with respect to the model is given by:
+
+        .. math::
+            \frac{\partial \mathbf{f_m}}{\partial \mathbf{m}} = diag(\frac{(\mathbf{G_x} \mathbf{G_x} \mathbf{a})}{\mathbg{a}})
+
+        where :math:`\mathbf{a}` is the amplitude of the vector model.
+
+
+
+        Parameters
+        ----------
+        m : numpy.ndarray
+            The model.
+
+        Returns
+        -------
+        numpy.ndarray
+            The derivative of the first-order smoothness regularization.
+        """
         amplitude = self.amplitude(m) + 1e16
         f_m_deriv = sdiag(
             (
@@ -957,7 +980,7 @@ class AmplitudeSmoothnessFirstOrder(SparseSmoothness, BaseAmplitude):
 
         d_m = self.mapping * self._delta_m(m)
 
-        f_m_deriv = self.f_m_deriv(m)
+        f_m_deriv = self.f_m_deriv(d_m)
 
         return self.mapping.deriv(m).T * (
             f_m_deriv * d_m.reshape((-1, self.n_comp), order="F")
--- a/SimPEG/regularization/vector.py
+++ b/SimPEG/regularization/vector.py
@@ -7,6 +7,7 @@ from .base import Smallness
 from discretize.base import BaseMesh
 from .base import RegularizationMesh, BaseRegularization
 from .sparse import Sparse, SparseSmallness, SparseSmoothness
+from ..utils import sdiag
 
 if TYPE_CHECKING:
     from scipy.sparse import csr_matrix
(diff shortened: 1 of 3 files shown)


SimPEG/regularization/vector.py
```diff
@@ -924,33 +922,56 @@ def f_m(self, m):
     def f_m_deriv(self, m) -> csr_matrix:
         r"""Derivative of the regularization kernel function.
 
-        For first-order smoothness regularization in the x-direction, the derivative of the
-        regularization kernel function with respect to the model is given by:
+        See :func:`base.SmoothnessFirstOrder.f_m_deriv` for more information.
+        """
+        return self.cell_gradient
```
thibaut-kobold, line 927: if `f(m) = cell_gradient @ amplitude(m)`, why do we do not see the amplitude derivative in `f_m_deriv` and only cell_gradient as if `f_m` was linear? What do I miss?

domfournier: because we measure amplitude**2, ends up being the sum of gradients of components square. So we really don't gain much in L2, only does something different in the Lp

thibaut-kobold (approved):
Thanks @domfournier for championing this. The rest looks sound.

SimPEG/regularization/vector.py
```diff
@@ -421,12 +421,10 @@ def deriv(self, m) -> np.ndarray:
         (n_param, ) numpy.ndarray
             Gradient of the regularization function evaluated for the model provided.
         """
-        d_m = self._delta_m(m)
+        d_m = self.mapping * self._delta_m(m)
```
lheagy, line 424: This is a comment for a later pr, but it would be nice to have the option to apply the mapping to the reference model or not. In the amplitude regularization, this would be nice because we would then not have to provide a reference model in cartesian coordinates, we could simply provide a reference amplitude. The derivatives for this would be okay for the smallness term (but would get ugly in smoothness terms. But we could just not allow a reference amplitude in the smoothness). We can create an issue on this to follow up later

lheagy:
Hi @domfournier, would you mind adding a test for the derivatives on this? I tried running this branch with the same test as in #1303, and that still seems to fail for me

domfournier:
> Hi @domfournier, would you mind adding a test for the derivatives on this? I tried running this branch with the same test as in #1303, and that still seems to fail for me. Was the goal of this pr to fix the derivs? or just simplify the code?

While this issue is just to re-write the operations so it reflects the math better. The test you've added on #1303 will make this fail since it looks like this is first order only. 

I am fine just switching over to john's formulation, and we can abandon this implementation altogether.

lheagy:
:+1: Thanks for clarifying! Feel free to keep this pr open, as the refactor might still be helpful for making the flow a bit simpler, but we can assess with John's input

jcapriot:
If we’re going to go that route, I’d recommend adapting the JTV regularization to handle arbitrary norms (right now it is only L1). As that is already in SimPEG and tested.

jcapriot:
So these derivatives are still not completely correct:
I adjusted my previous example to include mappings

```
def f(x, map, r):
    dm = x - r
    dm = dm.reshape((n_cells, -1), order='F')
    m = map(x) * dm
    y = m * m
    z = np.sum(y, axis=-1)
    a = np.sqrt(z)
    t = W @ G @ a
    u = t * t
    v = sum(u)
    return v

def fp(x, map, r):
    dm = x - r
    dm = dm.reshape((n_cells, -1), order='F')
    m = map(x) * dm
    y = m * m
    z = np.sum(y, axis=-1)
    a = np.sqrt(z)

    t = (G.T @ W.T @ (W @ G @ a)) / a
    gm = 2 * m * t[:, None]
    # above equivalent to
    # gm = 2 * sp.diags((G.T @ W.T @ (W.T @ G @ a)) / a) @ m
    return map_deriv(x) @ gm.reshape(-1, order='F')
```

domfournier:
> So these derivatives are still not completely correct: I adjusted my previous example to include mappings
> 
> ```
> def f(x, map, r):
>     dm = x - r
>     dm = dm.reshape((n_cells, -1), order='F')
>     m = map(x) * dm
>     y = m * m
>     z = np.sum(y, axis=-1)
>     a = np.sqrt(z)
>     t = W @ G @ a
>     u = t * t
>     v = sum(u)
>     return v
> 
> def fp(x, map, r):
>     dm = x - r
>     dm = dm.reshape((n_cells, -1), order='F')
>     m = map(x) * dm
>     y = m * m
>     z = np.sum(y, axis=-1)
>     a = np.sqrt(z)
> 
>     t = (G.T @ W.T @ (W @ G @ a)) / a
>     gm = 2 * m * t[:, None]
>     # above equivalent to
>     # gm = 2 * sp.diags((G.T @ W.T @ (W.T @ G @ a)) / a) @ m
>     return map_deriv(x) @ gm.reshape(-1, order='F')
> ```

Thanks @jcapriot , that seems to do it. Ready for another pass.

SimPEG/regularization/vector.py
```diff
@@ -941,16 +943,99 @@ def f_m_deriv(self, m) -> csr_matrix:
...
         Returns
         -------
-        scipy.sparse.csr_matrix
-            The derivative of the regularization kernel function.
+        numpy.ndarray
+            The derivative of the first-order smoothness regularization.
         """
-        return sp.block_diag([self.cell_gradient] * self.n_comp) @ self.mapping.deriv(
-            self._delta_m(m)
+        amplitude = self.amplitude(m) + 1e16
```
jcapriot, line 949: Did you mean 1E-16?

domfournier:
So yea, the latest change fix the issue with the derivative test, but it also makes the problem very hard to solve/unstable. I think we should just close this in favor of #1329

PR #1314, opened 2023-12-30
Remove code duplication for time and frequency stitched

domfournier:
Proposed changes to PR #1300 to reduce code duplication and complexity.

Main changes:

1- Consolidate the methods of time and frequency `Simulation1DLayeredStitched` onto base class.
2- Storage of Hankel coefficients in dataclass container
3- Storage of Cole-Cole parameters in dataclass container
4- Review of new regularization and mesh class (to be continued)


SimPEG/electromagnetics/base_1d.py
```diff
@@ -17,9 +17,66 @@
...
+    :param c: Cole-Cole frequency dependency
+    :param chi: DC magnetic susceptibility
+    :param tau1: Lower bound for log-uniform distribution of time-relaxation constants
+    :param tau2: Upper bound for log-uniform distribution of time-relaxation constants
+    """
+
+    eta: Union[np.ndarray, float] = 0.0
+    tau: Union[np.ndarray, float] = 1.0
+    c: Union[np.ndarray, float] = 0.5
+    chi: Union[np.ndarray, float] = 0.0
```
domfournier, line 76: @sgkang I found a few places where we had `dchi` and `chi`. I am guessing they are different, and would have different defaults. Can you validate this?

SimPEG/electromagnetics/base_1d.py
```diff
@@ -17,9 +17,66 @@
...
+    :param inv_lambs: Inverse map of unique Hankel transform parameters
+    :param W: Array of weights
+    """
+
+    C0s: np.ndarray
+    C1s: np.ndarray
+    lambs: np.ndarray
+    unique_lambs: np.ndarray
+    inv_lambs: np.ndarray
+    W: sp.csr_matrix
```
domfournier, line 57: We should probably try to use more explicit names for those, but that's out of my area. Feel free to rename them and change the description if needed.

SimPEG/electromagnetics/base_1d.py
```diff
@@ -17,9 +17,66 @@
...
+    "run_em1d_simulation",
+    "OutputType",
+    "ColeColeParameters",
+    "HankelCoefficients",
+]
+
+
+class OutputType(Enum):
+    FORWARD = 1
+    SENSITIVITY = 2
```
domfournier, line 34: Don't have to keep this design with Enum, but generally more robust than string comparison. Move the responsibility to the user to provide valid inputs.

SimPEG/electromagnetics/base_1d.py
```diff
@@ -83,19 +125,16 @@ def __init__(
...
         h=None,
         hMap=None,
-        eta=0.0,
-        tau=1.0,
-        c=0.5,
-        dchi=0.0,
-        tau1=1.0e-10,
-        tau2=10.0,
+        cole_cole_parameters: ColeColeParameters | None = None,
         hankel_filter="key_101_2009",
```
domfournier, line 129: We could eventually also use an `Enum` here to restrict to the shortlist of valid inputs.

SimPEG/electromagnetics/base_1d.py
```diff
@@ -299,19 +366,25 @@ def compute_complex_mu(self, frequencies):
...
-                tau1 = np.tile(self.tau1.reshape([-1, 1]), (1, n_frequency))
-                tau2 = np.tile(self.tau2.reshape([-1, 1]), (1, n_frequency))
+                dchi = np.tile(
+                    self.cole_cole_parameters.chi.reshape([-1, 1]), (1, n_frequency)
+                )
+                tau1 = np.tile(
+                    self.cole_cole_parameters.tau1.reshape([-1, 1]), (1, n_frequency)
+                )
+                tau2 = np.tile(
+                    self.cole_cole_parameters.tau2.reshape([-1, 1]), (1, n_frequency)
```
domfournier, line 386: Not sure we actually need to tile the values. Can't we just pass the same vector to all frequencies?

PR #1320, opened 2024-01-05
Add `weights_keys` method to `BaseRegularization`

### LINKED ISSUE #1318: ENH: Add method to expose weights keys in regularizations
### Proposed new feature or change:

Would be nice to add a public method to regularization classes that exposes the keys of the `_weights` dictionary, so every array in it can be accessed without the need to know all the keys.

This new method should live in `BaseRegularization` and could look like this:

```python
@property
def weights_keys(self) -> list[str]:
    """Return keys of cell weights."""
    return list(self._weights.keys())
```

---

This idea came up after the SimPEG Meeting of 2024-01-03.

santisoler:
#### Summary

Add a new `weights_keys` method to `BaseRegularization` that exposes the keys of the private `_weights` attribute. Add tests for the new feature.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fixes #1318

### DIFF
--- a/SimPEG/regularization/base.py
+++ b/SimPEG/regularization/base.py
@@ -55,7 +55,7 @@ class BaseRegularization(BaseObjectiveFunction):
         mapping: maps.IdentityMap | None = None,
         reference_model: np.ndarray | None = None,
         units: str | None = None,
-        weights: dict | np.ndarray | None = None,
+        weights: dict | None = None,
         **kwargs,
     ):
         if isinstance(mesh, BaseMesh):
--- a/SimPEG/regularization/base.py
+++ b/SimPEG/regularization/base.py
@@ -55,7 +55,7 @@ class BaseRegularization(BaseObjectiveFunction):
         mapping: maps.IdentityMap | None = None,
         reference_model: np.ndarray | None = None,
         units: str | None = None,
-        weights: dict | None = None,
+        weights: dict | np.ndarray | None = None,
         **kwargs,
     ):
         if isinstance(mesh, BaseMesh):
@@ -382,6 +382,13 @@ class BaseRegularization(BaseObjectiveFunction):
             )
             self._weights[key] = values
         self._W = None
+
+    @property
+    def weights_keys(self) -> list[str]:
+        """
+        Return the keys for the existing cell weights
+        """
+        return list(self._weights.keys())
 
     def remove_weights(self, key):
         """Removes the weights for the key provided.
--- a/tests/base/test_regularization.py
+++ b/tests/base/test_regularization.py
@@ -6,7 +6,13 @@ import inspect
 
 import discretize
 from SimPEG import maps, objective_function, regularization, utils
-from SimPEG.regularization import BaseRegularization, WeightedLeastSquares
+from SimPEG.regularization import (
+    BaseRegularization,
+    WeightedLeastSquares,
+    Smallness,
+    SmoothnessFirstOrder,
+    SmoothnessSecondOrder,
+)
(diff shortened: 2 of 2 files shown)


santisoler:
@thibaut-kobold, would you like to review this PR?

I saw you were curious in the Issue, that's why I asked. There's no rush to do so, feel free to take your time.

santisoler:
Also, @domfournier, you might be interested in taking a look at this one.

SimPEG/regularization/base.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         mapping: maps.IdentityMap | None = None,
         reference_model: np.ndarray | None = None,
         units: str | None = None,
-        weights: dict | None = None,
+        weights: dict | np.ndarray | None = None,
```
domfournier, line 58: I don't think it should ever be an ndarrray. It can be right now just to accommodate the deprecation of `cell_weights`. We should probably change the init below and make it more explicit that "cell_weights" got converted to an entry of the dict.

Suggesting moving L109-110 inside the deprecation block at L98

tests/base/test_regularization.py
```diff
@@ -657,6 +663,67 @@ class Dummy:
             regularization.parent = invalid_parent
 
 
+class TestWeightsKeys:
```
domfournier, line 666: Thanks for the tests, that's great.

domfournier (commented):
Thanks for this. See comment about typing.

SimPEG/regularization/base.py
santisoler: I see. I added that type hint because the constructor supports weights as `ndarray`. If we actually don't want this and we want users to be passing dicts to it, then we need to go through a deprecation process: if I move L109-110 inside the deprecation block in L98, then we'll break a lot of exiting code (even our own docs). In case we want to do so we would need to raise warnings for one or two releases and then deprecate it. 

I wouldn't add deprecation warnings in this PR since it's unrelated to the new feature it's being added. Would be nice to open an issue for it.

domfournier: Hmm I don't see anywhere `weights=vector` being used anywhere. Pretty joe and I did a deep cleanup of those. But sure, don't need to move the lines on this PR. Let's just not change the type hint.

santisoler: There are a few places that pass either `weights` or `cell_weights` as arrays:

https://github.com/simpeg/simpeg/blob/e3cd6ed869d5e49ab817b45f1fec037ac638a68f/examples/08-vrm/plot_inv_vrm_eq.py#L199

https://github.com/simpeg/simpeg/blob/e3cd6ed869d5e49ab817b45f1fec037ac638a68f/tests/base/test_regularization.py#L282-L286

https://github.com/simpeg/simpeg/blob/e3cd6ed869d5e49ab817b45f1fec037ac638a68f/tests/base/test_regularization.py#L320-L322

But besides that, I agree: we shouldn't add the type hint and start a deprecation process for `weights` as arrays.

domfournier: Right, first one is `cell_weights`, that's fine, but the tests could be fixed. But that's the thing, I don't think we need to deprecate weights as array since it was never intended. But you're the judge.

santisoler: Passing `cell_weights` is the same as passing `weights` + a warning, so it would break that. But understand your point. So probably we could just remove it.

I'm merging this PR then and open a new one for your suggestion of allowing `weights` only as dicts or None.

PR #1321, opened 2024-01-10
Magnetic simulation with Choclo as engine

santisoler:
#### Summary

Add new implementation of the magnetic simulation using the integral formulation based on [Choclo](https://www.fatiando.org/choclo) and Numba. Allow users to run the simulation with Choclo as engine through a new `engine` argument of the constructor of the simulation class. Add `numba_parallel` argument to control if the simulation should be run in parallel or in a single thread. Move common methods and properties of the gravity and magnetic simulation to the base class for potential field simulations. Add new `potential_fields/_numba_utils.py` file with functions that are needed both in the magnetic and gravity Numba implementations. Refactor and extend tests for the magnetic simulation, including tests that assert the accuracy of the new implementation against analytic solutions. Add new tests for the base class for potential field simulations. Update examples using Choclo as engine, include admonitions instructing how to use it.

#### PR Checklist
      to a Pull Request

#### Additional information

**Todo:**

- [x] Move the `engine` and `numba_parallel` arguments to the base class.
- [x] Make the new mag simulation to properly use the `engine` and `numba_parallel` attributes of the parent.
- [x] Add check for `n_processes` when `engine=="choclo"` in mag simulation.
- [x] Make tests for `engine`, `numba_parallel` and `engine` + `sensitivity_path` for the base class (not only for the child classes).
- [x] Improve docstrings: explain the different options users have to choose a different engine and parallelization options.
- [x] Implement storing the sensitivity matrix in disk


SimPEG/potential_fields/magnetics/_numba_functions.py
```diff
@@ -0,0 +1,525 @@
...
+        Array with the locations of the receivers
+    nodes : (n_active_nodes, 3) array
+        Array with the location of the mesh nodes.
+    sensitivity_matrix : (n_receivers, n_active_nodes) array
+        Empty 2d array where the sensitivity matrix elements will be filled.
+        This could be a preallocated empty array or a slice of it.
+    cell_nodes : (n_active_cells, 8) array
+        Array of integers, where each row contains the indices of the nodes for
+        each active cell in the mesh.
+    regional_field : (3,) array
```
jcapriot, line 53: Can you enable it to have a regional field that can be different in each cell? The current forward operator supports this, (though we don't have anything that directly uses it). My hope was to add the option for different sources besides a uniform field (e.g. a static magnetic dipole source, or if it's a very large survey).

This is also the reason the projection direction for `tmi` data was treated separately.

SimPEG/potential_fields/magnetics/_numba_functions.py
```diff
@@ -0,0 +1,525 @@
...
+        field (uniform background field).
+    kernel_x, kernel_y, kernel_z : callable
+        Kernels used to compute the desired magnetic component. For example,
+        for computing bx we need to use ``kernel_x=kernel_ee``,
+        ``kernel_y=kernel_en``, ``kernel_z=kernel_eu``.
+    constant_factor : float
+        Constant factor that will be used to multiply each element of the
+        sensitivity matrix.
+    scalar_model : bool
+        If True, the sensitivity matrix is build to work with scalar models
```
jcapriot, line 64:
```suggestion
        If True, the sensitivity matrix is built to work with scalar models
```

SimPEG/potential_fields/magnetics/simulation.py
```diff
@@ -36,7 +56,9 @@ def __init__(
         chiMap=None,
         model_type="scalar",
         is_amplitude_data=False,
-        **kwargs
+        engine="geoana",
+        choclo_parallel=True,
```
jcapriot, line 60: I think these could get moved to the `Base` class correct? (as `numba_parallel` to match the gravity simulation)

SimPEG/potential_fields/magnetics/_numba_functions.py
```diff
@@ -0,0 +1,525 @@
...
+        (susceptibilities).
+        If False, the sensitivity matrix is build to work with vector models
```
jcapriot, line 66:
```suggestion
        If False, the sensitivity matrix is built to work with vector models
```

jcapriot (commented):
Just a few notes while you're working on this.

SimPEG/potential_fields/magnetics/_numba_functions.py
santisoler: I see. That sounds like a nice feature to have. I think it could be a little cumbersome to add it in this PR, specially since we currently don't have a test or example for that runs on the original simulation class. But I would be totally on board of implementing it in a future PR. Would you like to open an issue to capture these thoughts?

SimPEG/potential_fields/magnetics/_numba_functions.py
santisoler: Thanks!

SimPEG/potential_fields/magnetics/simulation.py
santisoler: Yes! We should move it to the base class. I just added it to the todo list in the PR description. I will also try to move the sanity check for those parameters to the base class so we have only one of them.

santisoler:
> Just a few notes while you're working on this.

Thanks @jcapriot for the early review!

SimPEG/potential_fields/base.py
```diff
@@ -188,6 +200,54 @@ def n_processes(self, value):
...
+    def numba_parallel(self) -> bool:
+        """
+        Run simulation in parallel or single-threaded when using Numba.
+
+        If True, the simulation will run in parallel. If False, it will
+        run in serial.
+
+        ..important::
+
+            If ``engine`` is not ``"choclo"`` this argument will be ignored.
```
santisoler, line 239:
```suggestion
            If ``engine`` is not ``"choclo"`` this property will be ignored.
```

santisoler:
@thibaut-kobold @jedman I've been working towards finishing this PR. I think it's ready to be reviewed. Would you like to take a look at it, give it a try and review it?

thibaut-kobold:
@santisoler 
question / feature request: Will it work with `SimulationEquivalentSourceLayer`? That'd be awesome.

I believe `SimulationEquivalentSourceLayer` work with a 2D mesh to which you set the top and bottom along topo through the simulation object (rather than the mesh). It seems this does not work with `choclo` at the moment (at least not on my end).

SimPEG/potential_fields/base.py
```diff
@@ -73,6 +78,8 @@ def __init__(
         store_sensitivities="ram",
         n_processes=1,
         sensitivity_dtype=np.float32,
+        engine="geoana",
+        numba_parallel=True,
```
jedman, line 82: @santisoler can you explain why we want the `numba_parallel` flag here? I know it determines which version of forward and sensitivity calculation we use (*_parallel or *_serial) but I think it's confusing to control numba parallelism in this way (especially in combination with options `engine = "geoana"` and `n_processes = 1`). I don't think there is ever going to be a performance advantage to `numba_parallel = False`, and if you want to ensure the computation only uses a subset of available threads there are other ways to control that https://numba.pydata.org/numba-doc/latest/user/threading-layer.html

tests/pf/test_base_pf_simulation.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        Test error after choosing "choclo" as engine but not being installed
+        """
+        monkeypatch.setattr(SimPEG.potential_fields.base, "choclo", None)
+        engine = "choclo"
+        msg = "The choclo package couldn't be found."
+        with pytest.raises(ImportError, match=msg):
+            mock_simulation_class(tensor_mesh, engine=engine)
+
+    def test_sensitivy_path_as_dir(self, tensor_mesh, mock_simulation_class, tmpdir):
```
jedman, line 87:
```suggestion
    def test_sensitivity_path_as_dir(self, tensor_mesh, mock_simulation_class, tmpdir):
```

santisoler: Nice catch! Thanks!

santisoler:
> @santisoler question / feature request: Will it work with `SimulationEquivalentSourceLayer`? That'd be awesome.
> 
> I believe `SimulationEquivalentSourceLayer` work with a 2D mesh to which you set the top and bottom along topo through the simulation object (rather than the mesh). It seems this does not work with `choclo` at the moment (at least not on my end).

Thanks for pointing it out. It wasn't my goal on this PR to make sure that this implementation would work with the `SimulationEquivalentSourceLayer`. We should enable that! Though I'm hesitant to tackle that feature in this PR: it's already a little bit large, and since "geoana" is the default engine at the moment, users won't face any breaking change after this PR gets merged.

Would you like to open an issue requesting that feature? It would also be great if you can include a minimal example and the error you're facing when trying to use `SimulationEquivalentSourceLayer` with the Choclo implementation of the magnetic simulation.

SimPEG/potential_fields/base.py
santisoler: Thanks @jedman for the question. The `numba_parallel` flag was moved from the gravity simulation class to the potential fields base simulation class since both the gravity and the magnetic one will use it. This was suggested by @jcapriot in a previous comment and I think it's a good call.

Through that parameter we are deciding if we would like to execute the forward modelling/building the sensitivity matrix functions in parallel or in a single thread. With that simple flag we can control the parallelism _only_ for those methods. The way you point to in Numba's docs applies globally. The goal of the flag is to give more flexibility to control how the simulation should run.

This flexibility might come handy when these simulations are part of a higher order workflow that can be run in parallel, while the simulations will be constrained to a single thread.

The reason to create a separate argument for it different than `n_processes` is to avoid confusion: with `numba_parallel` you can either switch parallelization on or off, but not control in how many threads it should run.

The name of the argument was discussed in #1285, and the reason behind it was to make it clear the difference between the parallelization for the simulation using Numba vs the parallelization of the simulation using multiprocessing.

Would you approach this differently?

jedman: I totally agree that the base class is the right place to manage parallel execution, but I still think the right way to do this is to use numba's threading control-- if you are running a bunch of BasePFSimulations in a multiprocessing context and you only want numba to use a single thread in each process, then you can set `NUMBA_NUM_THREADS=1` (or use `numba.set_num_threads`). See the example here https://numba.pydata.org/numba-doc/latest/user/threading-layer.html#example-of-limiting-the-number-of-threads

And if the BasePFSimulations are themselves being instantiated in numba threads, you can still control the overall parallelism with numba's threading control. I can't think of a situation where that wouldn't do what you want... 

If it were only up to me, I think the right choice for usability and perofrmance settle on a single "engine" that can be parallelized (or not) with numba, rather than continuing to support multiprocessing, which does not seem to offer a clear advantage. I understand there is a desire to avoid adding to numba to the requirements and maintain backward compatibility, however. 

In any case, if the plan is to keep all these possible code paths, then at a minimum we should update the docstring with an extensive explanation of all the possibilities, and which parameters are related to which other parameters (and when they are ignored). For example, my understanding is that when `engine=choclo` the `n_processes` parameter is ignored

santisoler: > I totally agree that the base class is the right place to manage parallel execution, but I still think the right way to do this is to use numba's threading control-- if you are running a bunch of BasePFSimulations in a multiprocessing context and you only want numba to use a single thread in each process, then you can set NUMBA_NUM_THREADS=1 (or use numba.set_num_threads). See the example here https://numba.pydata.org/numba-doc/latest/user/threading-layer.html#example-of-limiting-the-number-of-threads
> 
> And if the BasePFSimulations are themselves being instantiated in numba threads, you can still control the overall parallelism with numba's threading control. I can't think of a situation where that wouldn't do what you want...

You're right that in that scenario setting the env variable would do just fine. The problem I see with it is that you're changing the Numba behaviour globally. So if another process that uses Numba's parallelization is being used in that workflow, that configuration will change both the PF simulation and this other process.

That's why I mention that whole goal of having `numba_parallel` is to increase the flexibility.

> If it were only up to me, I think the right choice for usability and perofrmance settle on a single "engine" that can be parallelized (or not) with numba, rather than continuing to support multiprocessing, which does not seem to offer a clear advantage. I understand there is a desire to avoid adding to numba to the requirements and maintain backward compatibility, however.

After seeing the improved performance of the Numba-based simulations, we have the future goal of switching to that one permanently in the future. At this stage, since these simulations are among the most used ones by SimPEG users, we considered to take the conservative path of not making breaking changes and allow users to experiment with the new implementation and report bugs they find (as the one that @thibaut-kobold reported in #1373). We aim to make the switch in the future, after we release these implementations.

> In any case, if the plan is to keep all these possible code paths, then at a minimum we should update the docstring with an extensive explanation of all the possibilities, and which parameters are related to which other parameters (and when they are ignored). For example, my understanding is that when engine=choclo the n_processes parameter is ignored

I agree. I probably oversaw the docstrings here, and they should be improved. I'll add more details to them so the different behaviors of the simulation class are well explained. And yes, if `engine="choclo"`, then `n_processes` gets ignored. Users will even get a warning if they are passing both to the simulations' constructors:
https://github.com/simpeg/simpeg/blob/4516cc38439fa2b2defd3539bf991fc3aeefc979/SimPEG/potential_fields/magnetics/simulation.py#L75-L83

simpeg/potential_fields/base.py
```diff
@@ -188,6 +210,54 @@ def n_processes(self, value):
...
+
+    @property
+    def numba_parallel(self) -> bool:
+        """
+        Run simulation in parallel or single-threaded when using Numba.
+
+        If True, the simulation will run in parallel. If False, it will
+        run in serial.
+
+        ..important::
```
santisoler, line 247: Fix rst syntax
```suggestion
        .. important::
```

simpeg/potential_fields/magnetics/_numba_functions.py
```diff
@@ -0,0 +1,537 @@
...
+    regional_field,
+    constant_factor,
+    scalar_model,
+):
+    """
+    Fill the sensitivity matrix for TMI
+
+    This function should be used with a `numba.jit` decorator, for example:
+
+    ..code::
```
santisoler, line 157:
```suggestion
    ..  code::
```

PR #1329, opened 2024-01-14
Simpeg vector update

johnweis0480:
#### Summary
Update test_regularization.py to test vector smoothness derivatives. Update AmplitudeSmoothnessFirstOrder in vector.py so that the call matches the derivative.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/SimPEG/regularization/vector.py
+++ b/SimPEG/regularization/vector.py
@@ -917,10 +917,9 @@ class AmplitudeSmoothnessFirstOrder(SparseSmoothness, BaseAmplitude):
         numpy.ndarray
             The regularization kernel function evaluated for the model provided.
         """
-        fm = (
-            sp.block_diag([self.cell_gradient] * self.n_comp)
-            * (self.mapping * self._delta_m(m))
-        ).reshape((-1, self.n_comp), order="F")
+        fm = self.cell_gradient * (self.mapping * self._delta_m(m)).reshape(
+            (self.regularization_mesh.nC, self.n_comp), order="F"
+        )
 
         return np.linalg.norm(fm, axis=1)
 
--- a/SimPEG/regularization/vector.py
+++ b/SimPEG/regularization/vector.py
@@ -917,9 +917,10 @@ class AmplitudeSmoothnessFirstOrder(SparseSmoothness, BaseAmplitude):
         numpy.ndarray
             The regularization kernel function evaluated for the model provided.
         """
-        fm = (sp.block_diag([self.cell_gradient] * self.n_comp) * (self.mapping * self._delta_m(m))).reshape(
-            (-1, self.n_comp), order="F"
-        )
+        fm = (
+            sp.block_diag([self.cell_gradient] * self.n_comp)
+            * (self.mapping * self._delta_m(m))
+        ).reshape((-1, self.n_comp), order="F")
 
         return np.linalg.norm(fm, axis=1)
 
--- a/SimPEG/regularization/vector.py
+++ b/SimPEG/regularization/vector.py
@@ -917,9 +917,11 @@ class AmplitudeSmoothnessFirstOrder(SparseSmoothness, BaseAmplitude):
         numpy.ndarray
             The regularization kernel function evaluated for the model provided.
         """
-        a = self.amplitude(m)
+        fm = (sp.block_diag([self.cell_gradient] * self.n_comp) * (self.mapping * self._delta_m(m))).reshape(
+            (-1, self.n_comp), order="F"
+        )
 
-        return self.cell_gradient @ a
+        return np.linalg.norm(fm, axis=1)
 
     def f_m_deriv(self, m) -> csr_matrix:
         r"""Derivative of the regularization kernel function.
--- a/SimPEG/regularization/base.py
+++ b/SimPEG/regularization/base.py
@@ -383,6 +383,13 @@ class BaseRegularization(BaseObjectiveFunction):
             self._weights[key] = values
         self._W = None
 
+    @property
+    def weights_keys(self) -> list[str]:
+        """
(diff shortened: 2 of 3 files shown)


domfournier (approved):
Wow, is that really all we need to switch over? 

For my own education, how would the previous implementation differ in terms of derivatives? Almost looks like the old way would just be factoring the gradient operator out of the square root, but end up at the same place - unless I am missing something.

Anyway, I like it. I will take it for a spin this week.

johnweis0480:
@domfournier Sounds, good. I don't think the performance should change very much if at all, as the derivative implementation that matches the new call will in general decrease the loss of the old call as well, which is why it was working decently. On the other hand, if the derivative was changed to match the old call, there would only be constraint on the smoothness of the magnitude of magnetization and none on the direction, so the performance is a lot worse in my experience.

domfournier:
That was my question. What do the derivative should look like on the old call. What did I get wrong..

SimPEG/regularization/vector.py
```diff
@@ -917,9 +917,12 @@ def f_m(self, m):
         numpy.ndarray
             The regularization kernel function evaluated for the model provided.
         """
-        a = self.amplitude(m)
+        fm = (
+            sp.block_diag([self.cell_gradient] * self.n_comp)
```
jcapriot, line 921: Don’t do a block diag here (it is slow), reshape the `a` matrix into a (-1, 3) array and then multiply it by cell_gradient.

jcapriot:
> That was my question. What do the derivative should look like on the old call. What did I get wrong..

Hopefully this little snippet of the correct derivative should highlight what went wrong:

```
def f(x):
    # this is a very simply version of what amplitude smoothness currently measures
    y = x * x
    z = np.sum(y, axis=-1)
    a = np.sqrt(z)
    t = G @ a
    u = t * t
    v = sum(u)
    return v

def fp(x):
    # This is what the complete derivative looks like
    y = x * x
    z = np.sum(y, axis=-1)
    a = np.sqrt(z)
    t = 1 / a  * ((G.T @ G) @ a)  # this line is where it went wrong, in the amplitude regularization G=I
                                                # thus this whole line goes to 1. Here we can't cancel out the a term.
    gx = 2 * x * t[:, None]
    return gx
```

jcapriot:
Essentially this turns it into the JTV operator with an L2 norm instead of an L1 norm. Should we change the name of this since it is no longer technically measuring the smoothness of the vector amplitude? Maybe `VectorVariation`?

lheagy:
Thanks @johnweis0480! This looks great

@jcapriot: I don't think that the Total Variation is quite the right model to be thinking about here. 

Here we are doing (but soon we can drop the 1/2!):  

image

For TV, the square root is on the outside of everything (e.g. from Haber & Gazit, 2013)

image

so the square root would have to be moved outside of the sum for it to be total variation. 

I would be fine with leaving the naming as-is

jcapriot:
@lheagy 
It actually is similar. Starting from what you have:
Consider your $\phi_{smooth}$ for $p=1$, $w_j=1$, $\alpha_j=1$ (and no $\frac{1}{2}$), expanding all of the j and k indices it becomes:
image

The key is that in the definition of JTV from (Haber & Grazit) the $|\nabla m|$ and $|\nabla s|$ are not the amplitude of the gradients, but the absolute value of the gradients. For joint total variation applied to the vector components (and discretizing it):
image

P.S. It should be obvious, but I left the volume terms out of each of them.

jcapriot:
Hmm... actually I have now convinced myself otherwise for the L1 case, they are not the same there, for some reason I was getting caught up on there being a difference between the absolute value of a vector and the magnitude of a vector, when in fact there is no difference. I forgot there was an averaging operation that couples together the individual components of each vector before taking the square root.

Just so I don't forget:
image

I still feel like they are somewhat related to each other. For an L2 style norm for both of them though they work out to be remarkably similar. L2 norm styled Joint Total Variation (take the squared value of JTV before the sum over elements) ( and in the averaging matrix $A_{vx} = A_{vy} = A_{vz} = c$)
image

Which I believe works out to being the same as the $\phi_{smooth}$ for $p=2$.

jcapriot:
Wait, for L2 would this suggestion do anything differently than just a standard smoothness operator applied to each of the vector components?

lheagy:
Thanks for sharing your thought process @jcapriot. I still think that trying to make a connection with TV is potentially quite confusing. TV is defined using the absolute value (e.g. [Ascher et al, 2006](https://www.proquest.com/scholarly-journals/on-effective-methods-implicit-piecewise-smooth/docview/921133084/se-2?accountid=14656), [Holtzman Gazit, 2013](https://www.proquest.com/scholarly-journals/model-fusion-joint-inversion/docview/1441227824/se-2?accountid=14656). If the absolute value is changed to a 2 norm, it is no longer Total Varition. So I think it is worth keeping these distinct.

lheagy (approved):
Thanks for this @johnweis0480! And I appreciate you adding the test too. This looks good to go

jcapriot:
Yeah, comparing the squared TV kinda defeats the purpose...

I still think it should be renamed. It would now regularize the direction and the amplitude. I think the name implies it only regularizes on the amplitude.

domfournier:
This PR should close #1313 and resolve #1301

lheagy:
Given that this pr is a small pr that fixes a bug #1301, and does not change the current behaviour of the class in an inversion (the gradients are what influence the direction & solution, the value of the objective function itself is only used to estimate a beta and printed to the screen), I think we should merge it as is. 

Renaming would introduce the need for deprecations, etc. We can certainly chat further about this, but I think we should do so in an issue and not change the scope of this current pr.

PR #1330, opened 2024-01-16
Bump versions of flake8 and black and pin flake plugins

santisoler:
#### Summary

Bump versions of `black`, `flake8` and flake plugins to their latest available versions in PyPI and in conda-forge. Pin versions of `flake8` and its plugins in the `requirements_style.txt`, `environment_test.yml` and `pre-commit` configuration files. This prevents inconsistencies between the Style Checks run in Azure and the ones run locally through `pre-commit` or `make flake`.

#### PR Checklist
* [ ] If this is a work in progress PR, set as a Draft PR
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html).
* [ ] Added [tests](https://docs.simpeg.xyz/content/getting_started/practices.html#testing) to verify changes to the code.
* [ ] Added necessary documentation to any new functions/classes following the
      expect [style](https://docs.simpeg.xyz/content/getting_started/practices.html#documentation).
* [ ] Marked as ready for review (if this is was a draft PR), and converted 
      to a Pull Request
* [ ] Tagged ``@simpeg/simpeg-developers`` when ready for review.

PR #1332, opened 2024-01-17
fix failing pgi tests (factor of half PR)

thibaut-kobold:
fix failing pgi tests in https://github.com/simpeg/simpeg/pull/1326 (I do not have write persmission to the main simpeg repo with this account)

Note: For the record, my opinion about this has not changed from what I expressed in https://github.com/simpeg/simpeg/issues/763, but happy to help as I appear to be in minority.

### DIFF
--- a/tests/base/test_pgi_regularization.py
+++ b/tests/base/test_pgi_regularization.py
@@ -4,6 +4,7 @@ import discretize
 import numpy as np
 from pymatsolver import SolverLU
 from scipy.stats import multivariate_normal
+
 from SimPEG import regularization
 from SimPEG.maps import Wires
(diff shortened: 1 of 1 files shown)


santisoler (approved):
Thanks @thibaut-kobold for these fixes! The code style failing is not your fault, I already fixed this issue in #1330, so after I merge this into `factor-of-half` it will surely pass.

One small note: I would use `np.testing.assert_allclose` instead of `assert_almost_equal`, following Numpy's recommendations: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_almost_equal.html

But don't worry about that, I can apply those changes in the other PR.

I'm merging this! :rocket:

PR #1334, opened 2024-01-21
Fix implementation of coterminal function

### LINKED ISSUE #1333: Unintended behaviour of `SimPEG.utils.coterminal`?
## Question

If we pass angles equal to $-\pi$ or $\pi$ we'll get $\pi$ or $-\pi$, respectively. This means that the function will return floats with different signs for this pair of coterminal angles ($-\pi$ and $\pi$ are coterminal angles).

For example:

```python
import numpy as np
from SimPEG.utils import coterminal

angles = np.array([-np.pi, np.pi])
print(coterminal(angles))
```
```
[ 3.14159265 -3.14159265]
```

This may cause issues when trying to regularize such angles: if one cell has a vector with angle $-\pi$, and the next one has an angle of $\pi$, then a smoothness regularization on angles will introduce a penalty (because the difference between these two floats is non-zero), while it shouldn't because there's no change in angles between the two cells.

**Is this the intended behaviour?**

## Idea

Wouldn't be better if the function returns angles in the interval $(-\pi, \pi]$ ?
In that case, passing $-\pi$ and $\pi$ would return the same value for both of them. Therefore it should have the following behaviour for the previous example:

```python
import numpy as np
from SimPEG.utils import coterminal

angles = np.array([-np.pi, np.pi])
print(coterminal(angles))
```
```
[ 3.14159265 3.14159265]
```

## If we decide to go forward with this

If we want to change the behaviour of this function I think we should:

1. Modify the implementation of `SimPEG.utils.coterminal` accordingly.
2. Improve the docstrings to reflect the new change. Probably the $\gamma$ angle could be defined as belonging to $(-\pi, \pi]$.
3. Add test functions that check the function under these circumstances:
    - Passing $-\pi$, $\pi$, $3\pi$, $-3\pi$ should return the same angle: $\pi$
    - Are angles within the interval not being modified?
    - Are angles outisde the interval being correctly brought to that interval?

domfournier:
Fix the implementation of the `coterminal` function: use the `mod` operator to ensure that the functions returns the correct coterminal angle defined in $[-\pi, \pi)$. Add more tests.

Fixes #1333

---

PR to address issue #1333 , where the sign of coterminal angles is not preserved. Also fixes potential issues with angle > $2 \pi$

Running the example/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh

Before
old

After
new

### DIFF
--- a/SimPEG/utils/mat_utils.py
+++ b/SimPEG/utils/mat_utils.py
@@ -409,7 +409,7 @@ def coterminal(theta):
         Coterminal angles
 
     """
-    coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
+    coterminal = (theta + np.pi) % (2 * np.pi) - np.pi
     return coterminal
 
 
--- a/SimPEG/utils/mat_utils.py
+++ b/SimPEG/utils/mat_utils.py
@@ -396,7 +396,7 @@ def coterminal(theta):
         \theta = 2\pi N + \gamma
 
     and *N* is an integer, the function returns the value of :math:`\gamma`.
-    The coterminal angle :math:`\gamma` is within the range :math:`[-\pi , \pi]`.
+    The coterminal angle :math:`\gamma` is within the range :math:`[-\pi , \pi)`.
 
     Parameters
     ----------
--- a/SimPEG/utils/mat_utils.py
+++ b/SimPEG/utils/mat_utils.py
@@ -409,12 +409,8 @@ def coterminal(theta):
         Coterminal angles
 
     """
-    sub = theta[np.abs(theta) >= np.pi]
-    sub = -np.sign(sub) * (2 * np.pi - np.abs(sub))
-
-    theta[np.abs(theta) >= np.pi] = sub
-
-    return theta
+    coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
+    return coterminal
 
 
 def define_plane_from_points(xyz1, xyz2, xyz3):
--- a/tests/base/test_utils.py
+++ b/tests/base/test_utils.py
@@ -363,8 +363,8 @@ class TestCoterminalAngle:
 
     @pytest.mark.parametrize(
         "coterminal_angle",
-        (0, np.pi / 2, np.pi, -np.pi / 2),
-        ids=("0", "pi/2", "pi", "-pi/2"),
+        (0, np.pi / 2, -np.pi, -np.pi / 2),
+        ids=("0", "pi/2", "-pi", "-pi/2"),
     )
     def test_right_angles(self, coterminal_angle):
         """
--- a/tests/base/test_utils.py
+++ b/tests/base/test_utils.py
@@ -344,14 +344,45 @@ class TestDownload(unittest.TestCase):
         shutil.rmtree(os.path.expanduser("./test_url"))
 
 
-@pytest.mark.parametrize(
-    "angle", [np.pi, -np.pi, 3 * np.pi, 1.25 * np.pi, -1.25 * np.pi]
-)
-def test_coterminal(angle):
-    coangle = coterminal(angle)
-    assert np.abs(coangle) < np.pi
-    if coangle != 0:
-        assert np.sign(coterminal(angle)) == np.sign(angle)
+class TestCoterminalAngle:
+    """
+    Tests for the coterminal function
+    """
+
+    @pytest.mark.parametrize(
+        "coterminal_angle",
+        (1 / 4 * np.pi, 3 / 4 * np.pi, -3 / 4 * np.pi, -1 / 4 * np.pi),
+        ids=("pi/4", "3/4 pi", "-3/4 pi", "-pi/4"),
+    )
+    def test_angles_in_quadrants(self, coterminal_angle):
+        """
+        Test coterminal for angles in each quadrant
+        """
+        angles = np.array([2 * n * np.pi + coterminal_angle for n in range(-3, 4)])
+        np.testing.assert_allclose(coterminal(angles), coterminal_angle)
+
+    @pytest.mark.parametrize(
+        "coterminal_angle",
+        (0, np.pi / 2, np.pi, -np.pi / 2),
+        ids=("0", "pi/2", "pi", "-pi/2"),
+    )
+    def test_right_angles(self, coterminal_angle):
+        """
+        Test coterminal for right angles
+        """
+        angles = np.array([2 * n * np.pi + coterminal_angle for n in range(-3, 4)])
+        np.testing.assert_allclose(coterminal(angles), coterminal_angle)
+
+    @pytest.mark.parametrize(
+        "angle",
+        [np.pi, -np.pi, 3 * np.pi, 1.25 * np.pi, -1.25 * np.pi],
+        ids=("pi", "-pi", "3 pi", "1.25 pi", "-1.25 pi"),
+    )
+    def test_sign_coterminal(self, angle):
+        coangle = coterminal(angle)
+        assert np.abs(coangle) < np.pi
+        if coangle != 0:
+            assert np.sign(coterminal(angle)) == np.sign(angle)
 
 
 if __name__ == "__main__":
--- a/tests/base/test_utils.py
+++ b/tests/base/test_utils.py
@@ -1,4 +1,5 @@
 import unittest
+import pytest
 import numpy as np
 import scipy.sparse as sp
 import os
@@ -22,6 +23,7 @@ from SimPEG.utils import (
     Counter,
     download,
     surface2ind_topo,
+    coterminal,
 )
 import discretize
 
@@ -340,6 +342,16 @@ class TestDownload(unittest.TestCase):
         # clean up
         shutil.rmtree(os.path.expanduser("./test_urls"))
         shutil.rmtree(os.path.expanduser("./test_url"))
+
+
+@pytest.mark.parametrize(
+    "angle", [np.pi, -np.pi, 3 * np.pi, 1.25 * np.pi, -1.25 * np.pi]
+)
+def test_coterminal(angle):
+    coangle = coterminal(angle)
+    assert np.abs(coangle) < np.pi
+    if coangle != 0:
+        assert np.sign(coterminal(angle)) == np.sign(angle)
 
 
 if __name__ == "__main__":
--- a/tests/base/test_utils.py
+++ b/tests/base/test_utils.py
@@ -373,17 +373,6 @@ class TestCoterminalAngle:
         angles = np.array([2 * n * np.pi + coterminal_angle for n in range(-3, 4)])
         np.testing.assert_allclose(coterminal(angles), coterminal_angle)
 
-    @pytest.mark.parametrize(
-        "angle",
-        [np.pi, -np.pi, 3 * np.pi, 1.25 * np.pi, -1.25 * np.pi],
-        ids=("pi", "-pi", "3 pi", "1.25 pi", "-1.25 pi"),
-    )
-    def test_sign_coterminal(self, angle):
-        coangle = coterminal(angle)
-        assert np.abs(coangle) < np.pi
-        if coangle != 0:
-            assert np.sign(coterminal(angle)) == np.sign(angle)
-
 
 if __name__ == "__main__":
     unittest.main()


tests/base/test_utils.py
```diff
@@ -342,5 +344,46 @@ def test_downloads(self):
...
+        np.testing.assert_allclose(coterminal(angles), coterminal_angle)
+
+    @pytest.mark.parametrize(
+        "angle",
+        [np.pi, -np.pi, 3 * np.pi, 1.25 * np.pi, -1.25 * np.pi],
+        ids=("pi", "-pi", "3 pi", "1.25 pi", "-1.25 pi"),
+    )
+    def test_sign_coterminal(self, angle):
+        coangle = coterminal(angle)
+        assert np.abs(coangle) < np.pi
```
santisoler, line 383: This is not always true for every angle in the parametrization. For `np.pi` and `np.pi` we expect the absolute value of their respective coterminal angles be exactly `np.pi` for both.

tests/base/test_utils.py
```diff
@@ -342,5 +344,46 @@ def test_downloads(self):
...
+        if coangle != 0:
```
santisoler, line 384: For the angles in the parametrization, we shouldn't expect a coterminal angle of zero. If we get so, this test will pass without making any noise, hiding a possible bug.

tests/base/test_utils.py
```diff
@@ -342,5 +344,46 @@ def test_downloads(self):
...
+            assert np.sign(coterminal(angle)) == np.sign(angle)
```
santisoler, line 385: This is not true for every angle in the parametrization. For example $\pi$ and $-\pi$ are coterminal angles and the function should return $-\pi$ for both, therefore the sign of the coterminal angle won't match the one of $\pi$.

SimPEG/utils/mat_utils.py
```diff
@@ -409,12 +409,8 @@ def coterminal(theta):
...
 
     """
-    sub = theta[np.abs(theta) >= np.pi]
-    sub = -np.sign(sub) * (2 * np.pi - np.abs(sub))
-
-    theta[np.abs(theta) >= np.pi] = sub
-
-    return theta
+    coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
+    return coterminal
```
santisoler, line 413: I think we should move forward with this implementation:
```suggestion
    return (theta + np.pi) % (2 * np.pi) - np.pi
```

santisoler (changes requested):
Thanks @domfournier for opening this PR!

I've just pushed some more tests for the function. With the current bugfix some of them fail. I think we should move forward with @jcapriot's suggestion of:

```python
return (theta + np.pi) % (2 * np.pi) - np.pi
```

This will return, for any angle in $(-\inf, \inf)$ its coterminal angle in $[-\pi, \pi)$.

We use a similar one in Harmonica for adjusting the [longitude angles for tesseroids](https://github.com/fatiando/harmonica/blob/bbaf5a191c7a3ccf9835a2d3b565be26a3500d26/harmonica/_forward/_tesseroid_utils.py#L473-L502).

I left a few comments regarding the test function. I think it needs some work. Let me know what do you think.

Before we merge this we should update the docstring, reflecting the new change.

jcapriot:
Here's what the output of the functions for comparison:
image

jcapriot:
I'm also running some very simple regularization tests to get to the bottom of what we actually should be doing during regularization... trying to get a good way to present them.

jcapriot:
```python
from SimPEG.utils import coterminal as coterminal1
import numpy as np
import matplotlib.pyplot as plt
import discretize
from scipy.optimize import line_search

def coterminal1334(theta):
    coterminal = np.sign(theta) * (np.abs(theta) % np.pi)
    return coterminal

def coterminaljcapriot(theta):
    return (theta + np.pi) % (2 * np.pi) - np.pi

def undo_overrotate(theta):
    # this actually mostly
    x, y = np.cos(theta), np.sin(theta)
    theta = np.arctan2(y, x)
    return theta
```


```python
mesh = discretize.TensorMesh([[1]*8])
cell_gradient = (-mesh.face_divergence.T).toarray()
# cell_gradient is the [-1, 1] stencil (since dx = 1) with Dirichlet boundaries on both sides
# we know the minimum of this function should be the reference model ... let's see if we get there...

class AngleSmoothness():
    def __init__(self, m_ref, coterminal_func):

        self.m_ref = m_ref
        self.coterminal_func = coterminal_func

    def __call__(self, m):
        m = undo_overrotate(m) # ensures input model is on the [-pi, pi) range
        dm = (m - self.m_ref)
        d_alpha = cell_gradient @ dm
        d_alpha = self.coterminal_func(d_alpha)
        return d_alpha.T @ d_alpha

    def deriv(self, m):
        m = undo_overrotate(m) # ensures input model is on the [-pi, pi) range
        dm = (m - self.m_ref)
        d_alpha = cell_gradient @ dm
        d_alpha = self.coterminal_func(d_alpha)
        g_dm = 2 * (cell_gradient.T @ d_alpha)
        return g_dm

    def deriv2(self, m):
        gg_d_alpha = 2 * cell_gradient
        gg_dm = cell_gradient.T @ gg_d_alpha
        return gg_dm
```


```python
m_ref = np.zeros(mesh.n_cells)

# create a starting model that will have differences beyond the [-pi, pi] range.
value = 3 *np.pi / 4
angles = np.tile([value, -value], 4)
plt.plot(mesh.cell_centers, angles)
```




    [<matplotlib.lines.Line2D at 0x1aeb3e11250>]




    
output_2_1
    



```python
phi1 = AngleSmoothness(m_ref, coterminal1)
phi2 = AngleSmoothness(m_ref, coterminal1334)
phi3 = AngleSmoothness(m_ref, coterminaljcapriot)

phi1(angles), phi2(angles), phi3(angles)
```




    (28.375112653131907, 28.375112653131907, 28.375112653131907)




```python
def do_step(phi, m):
    f_val = phi(m)
    grad = phi.deriv(m)
    H = phi.deriv2(m)
    dm = np.linalg.solve(H, -grad)
    info = line_search(phi, phi.deriv, m, dm, maxiter=100)
    step = info[0]
    if step is None:
        print("local minimum")
        return m, 1
    m_p1 = undo_overrotate(m + step * dm)
    f_val_next = phi(m_p1)
    print(f_val, f_val_next, step)
    return m_p1, 0
```


```python
m1 = angles.copy()
for i in range(10):
    m1, info = do_step(phi1, m1)
    if info != 0:
        break
```

    28.375112653131907 4.386490844928603 1.0
    local minimum
    


```python
m2 = angles.copy()
for i in range(10):
    m2, info = do_step(phi2, m2)
    if info != 0:
        break
```

    28.375112653131907 8.74616812491446 0.47044917257683216
    8.74616812491446 5.999179246311323 0.1994407283953379
    5.999179246311323 4.742131644100865 0.13768135179261087
    4.742131644100865 4.009427145428162 0.10612543407660989
    4.009427145428162 2.7350752054674072 0.25
    2.7350752054674072 2.018252239239482 0.25
    local minimum
    


```python
m3 = angles.copy()
for i in range(10):
    m3, info = do_step(phi3, m3)
    if info != 0:
        break
```

    28.375112653131907 4.386490844928603 1.0
    4.386490844928603 4.386490844928602 1.0
    local minimum
    


```python
plt.plot(m1, '-.x', label='current')
plt.plot(m2, '--', label='#1333')
plt.plot(m3, '*', label='modulus')
plt.legend()
```
output_8_1

jcapriot:
Oddly enough.... I think the actual way we should be doing this... is Not using coterminal on the derivative evaluation (but still using it on the function evaluation, and ensuring the model itself remains on the $[-\pi, \pi]$ interval).

Think about it... if we're looking at the difference between two angles, $\alpha_1 = -3\pi/4$ and $\alpha_2 = 3\pi/4$.
The difference of the two is:
$\Delta\alpha = \alpha_1 - \alpha_2 = -3\pi/2$

$\Delta\alpha$ has a coterminal angle of $\pi/2$

Now, if we wanted to minimize the difference (so subtract the gradient), sure we could subtract the coterminal difference from $\alpha_1$ and get an angle that is coterminal with $\alpha_2$. I.E. ($coterminal(\alpha_1 + coterminal(\Delta\alpha)) = \alpha_2$, but we could just as easily have done: $\alpha_1 + \Delta\alpha$ to get to the same angle, without using or caring about coterminal angles.

jcapriot:
This is with removing the coterminal calculation from the derivative:
image
image

jcapriot:
> Basically, there are two ways to get to the same angle, so why not use the one without coterminal evaluations and remove all those nasty local minimums.

I do foresee at least one issue with my suggestion though, if we don't take a complete step, the function evaluation will increase for small step lengths in that direction.

domfournier:
> ```python
> return (theta + np.pi) % (2 * np.pi) - np.pi
> ```

Well, you guys can battle it out. But right now, the implementation does what we want, but the modulo operation above does not.

image

tests/base/test_utils.py
domfournier: We are probably just getting hung up on the word `coterminal` - which we can rename. The goal is to not get stuck on large difference in angles as we wrap around quadrants.

tests/base/test_utils.py
domfournier: Again, same issue with naming. Two vectors that are pointing in opposite direction (e.g. -pi/2 and another at pi/2) are generally just vector near vertical, so we son't want to overly penalize them in the direction. So the current "cotermina" will return 0.

tests/base/test_utils.py
domfournier: Don't agree.

tests/base/test_utils.py
jcapriot: No, I don't think you're argument is valid there. Vectors pointing in opposite directions should be maximally penalized.

jcapriot: vectors pointing at $-\pi$ and $\pi$ should be the minimum penalty.

santisoler:
I think you're right @domfournier. We are probably not thinking about the same definition for "coterminal angles". We should probably start with that haha.

## Coterminal angles

From my understanding, two angles $\alpha_1$ and $\alpha_2$ are _coterminal_ if 

$$ \exists n \in \mathbb{Z} / \alpha_1 = \alpha_2 + 2\pi n $$

With this definition in mind, we can create a mapping between $(-\infty, \infty)$ to any half-bounded interval of length $2\pi$. For example, map every angle in $(-\infty, \infty)$ to a unique angle in $[-\pi, \pi)$.

By reading the docstring of `coterminal`, I understand that for any given angle the function returns its corresponding coterminal angle in the interval $[-\pi, \pi]$. Here is where I make the note that this interval already includes two coterminal angles: $-\pi$ and $\pi$, that's why I opened the issue. With @jcapriot's implementation using the mod the function will always return an angle in $[-\pi, \pi)$.

I still think this function should be fixed because it's has an unwanted behaviour, as described in #1333.

## Regularizations

Our ultimate goal is to properly define the smoothness regularization for angles, regardless if we use or not the `coterminal` function.

I haven't played too much with those, therefore I'm sure that you and @jcapriot have a clearer idea on how to implement this. But for me it's crucial if we agree how the penalization should be for any pair of angles.

Are we imposing a positive constrain in the magnitude of the vectors? Or their magnitude could be either positive or negative?

Two angles pointing in opposite directions ($\alpha_1 = \alpha_2 + \pi$) should have a minimum penalty or a maximum penalty?

santisoler:
> Oddly enough.... I think the actual way we should be doing this... is Not using coterminal on the derivative evaluation (but still using it on the function evaluation, and ensuring the model itself remains on the [−π,π] interval).
> 
> Think about it... if we're looking at the difference between two angles, α1=−3π/4 and α2=3π/4. The difference of the two is: Δα=α1−α2=−3π/2
> 
> Δα has a coterminal angle of π/2
> 
> Now, if we wanted to minimize the difference (so subtract the gradient), sure we could subtract the coterminal difference from α1 and get an angle that is coterminal with α2. I.E. (coterminal(α1−coterminal(Δα))=α2, but we could just as easily have done: α1−Δα to get to the same angle, without using or caring about coterminal angles.

I agree that the coterminal function is needed in the smallness regularization.

And yes, probably for the smoothness we don't need to compute coterminal angles. As I said before, I think we need a function to compute the difference between two angles (any pair of angles in $(-\pi, \pi)$) that is lower or equal than $\pi$. For example:

```python
def angles_diff(theta_1: float, theta_2: float):
    diff = np.abs(theta_1 - theta_2) % (2 * np.pi)
    if diff > np.pi:
        return 2 * np.pi - diff
    return diff
```

(this is only for two floats, we should implement it for arrays).

But it will depend on how the smoothness regularization is defined.

jcapriot:
> I think you're right @domfournier. We are probably not thinking about the same definition for "coterminal angles". We should probably start with that haha.
> 
> ## Coterminal angles
> From my understanding, two angles α1 and α2 are _coterminal_ if
> 
> ∃n∈Z/α1=α2+2πn
> 
> With this definition in mind, we can create a mapping between (−∞,∞) to any half-bounded interval of length 2π. For example, map every angle in (−∞,∞) to a unique angle in [−π,π).
> 
> By reading the docstring of `coterminal`, I understand that for any given angle the function returns its corresponding coterminal angle in the interval [−π,π]. Here is where I make the note that this interval already includes two coterminal angles: −π and π, that's why I opened the issue. With @jcapriot's implementation using the mod the function will always return an angle in [−π,π).
> 
> I still think this function should be fixed because it's has an unwanted behaviour, as described in #1333.
> 
> ## Regularizations
> Our ultimate goal is to properly define the smoothness regularization for angles, regardless if we use or not the `coterminal` function.
> 
> I haven't played too much with those, therefore I'm sure that you and @jcapriot have a clearer idea on how to implement this. But for me it's crucial if we agree how the penalization should be for any pair of angles.
> 
> Are we imposing a positive constrain in the magnitude of the vectors? Or their magnitude could be either positive or negative?
> 
> Two angles pointing in opposite directions (α1=α2+π) should have a minimum penalty or a maximum penalty?

Currently:
* there is a positivity constraint on the amplitude
* The models are always re-projected onto a positive amplitude, horizontal angles $\in[-\pi,\pi]$, vertical angle $\in[0, \pi]$ after taking a step (why a "coterminal" calculation doesn't happen on the model coming in), This is not the clearest in the code, but it happens in a directive: `ProjectSphericalBounds`.
* Smallness is therefore measured in this space.
* The smoothness penalizes the coterminal angle of the difference of two angles, divided by the distance between them. (A co-terminal gradient if you will).

domfournier:
I obviously can't disagree with your definition of coterminal @santisoler - it is exactly that. 
I might be shifting the goal post here, but I am just not convinced that this is the formulation we need to make the spherical inversion run smoother. I finally managed to reconcile my runs with joe's (issue with random beta start), and after using the pure definition of coterminal as per `(theta + np.pi) % (2 * np.pi) - np.pi` it appears more unstable (as unstable as what's on main currently). Just from experimentation here, it looks like the expression on this PR fares a bit better at all angles of magnetization. It would be more like a "supplementary acute" angle at this point. From intuition, the penalty would take less importance when the vectors are near vertical, yet may have very different "declination angles". So we are not penalizing for polar opposites. 

For example, below are results for the example with a vertical magnetization

Using true coterminal

joe_vert


Using proposed acute angle

1331_vert

Kind of a worst case scenario for sure, but we can get this pretty easily, unfortunetly.

domfournier:
Aight I am not going to fight it no mo. Let's go with joe's. More to be researched...

jcapriot:
@domfournier I believe what you're suggesting might best be taken into account as an inclination dependent weighting of the smoothness of declination angles, maybe:

```
W = sin(inclination)
```

Which would give it a maximum penalty at inclination = $\pi/2$ and zero penalty at $0$ and $\pi$

santisoler (approved):
Looks good to me. I'll merge this! 🚀 

Thanks @domfournier for tackling this!

PR #1335, opened 2024-01-22
Implement base objective function classes as abstract ones

santisoler:
#### Summary

Draft for redesigning base classes for objective functions, data misfits and
regularizations as [abstract
classes](https://docs.python.org/3/library/abc.html).

#### PR Checklist
      to a Pull Request

Simplify the `BaseObjectiveFunction`:
- Decorate `__init__`, `__call__`, `deriv` and `deriv2` as `abstractmethods`.
- Decorate `nP` as `abstractproperty`.
- Remove the `mapping`. Only regularizations needs it. Data misfits rely on
  the simulation to handle mapping operations.

Simplify the `ComboObjectiveFunction`:
- Remove the `W`, it was overwritten by child classes, and the implementation
  only applied when the objective functions used and L2 norm.
- Simplify how it checks if the child objective functions need to get `fields`
  in their methods or not. Don't rely on every objective function to have
  `has_fields` (it doesn't make sense for regularization to have this
  attribute).

Remove `L2ObjectiveFunction`. It was only use to create the `BaseDataMisfit`
class, forcing it to be only an L2 class. Moreover, most of its methods were
being overwritten by the `BaseDataMisfit` and `L2DataMisfit`, so it didn't have
a clear purpose.

Make `BaseDataMisfit` and `BaseRegularization` abstract classes:
- `BaseDataMisfit` require children to implement `__call__`, `deriv` and
  `deriv2`.
- `BaseRegularization` require children to implement `f_m` and `f_m_deriv`

### DIFF
--- a/simpeg/data_misfit.py
+++ b/simpeg/data_misfit.py
@@ -183,7 +183,7 @@ class BaseDataMisfit(BaseObjectiveFunction):
 
         Where :math:`\mathbf{d}_\text{obs}` is the observed data vector and
         :math:`\mathbf{d}_\text{pred}` is the predicted data vector for a model
-        vector :math:`\mathbf{m}`, this function computes the data residual:
+        vector :math:`\mathbf{m}`, this method computes the data residual:
 
         .. math::
             \mathbf{r} = \mathbf{d}_\text{pred} - \mathbf{d}_\text{obs}


santisoler:
I noticed by running some tests on the drafted code that some regularization classes like `JointTotalVariation`, `CrossGradient` and `LinearCorrespondence` don't implement the `f_m` and/or the `f_m_deriv` methods. They are all childs of `BaseSimilarityMeasure` which overrides some methods of the `BaseRegularization` class.

I think we should revisit the design of these classes so we avoid violating [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

- What methods of the `BaseRegularization` are required by the `BaseSimilarityMeasure`?
- Which ones should be overwritten?
- `BaseSimilarityMeasure` should also be an abstract class. What methods and properties should be abstract?

santisoler:
After discussing with @lheagy, we decided not to refactor the `BaseRegularization` class in this PR. That branch of the hierarchy tree deserves some more thoughts and trying to tackle it here would create a massive PR. Lets focus on the `BaseObjectiveFunction` and in the `DataMisfit` branch of the tree.

PR #1337, opened 2024-01-24
Exponential Sinusoids Simulation

lheagy:
### Summary
- define the kernel functions at cell nodes rather than at cell centers 
- also move the __init__ from the middle of the class to the top of the class definition

### Reasoning 
- kernel functions should be defined on nodes as we think of them being piecewise linear (rather than discontinuous as we talk about at cell centers). This makes it more in line with other physical simulations
- this then also introduces techniques for discrete integration (eg Gaussian quadrature) 
- both of these things are important for teaching purposes

### Notes
- [x] there is one test where we manually check the output of the data. This will fail currently. We should decide whether we want to keep this test and update it or simply remove it.

### DIFF
--- a/SimPEG/simulation.py
+++ b/SimPEG/simulation.py
@@ -668,7 +668,6 @@ class ExponentialSinusoidSimulation(LinearSimulation):
             G_nodes = np.empty((self.mesh.n_nodes, self.n_kernels))
 
             for i in range(self.n_kernels):
-                print(self.g(i).shape)
                 G_nodes[:, i] = self.g(i)
 
             self._G = (self.mesh.average_node_to_cell @ G_nodes).T @ sdiag(
--- a/SimPEG/simulation.py
+++ b/SimPEG/simulation.py
@@ -665,12 +665,13 @@ class ExponentialSinusoidSimulation(LinearSimulation):
         Matrix whose rows are the kernel functions
         """
         if getattr(self, "_G", None) is None:
-            G = np.empty((self.mesh.nC, self.n_kernels))
+            G_nodes = np.empty((self.mesh.n_nodes, self.n_kernels))
 
             for i in range(self.n_kernels):
-                G[:, i] = self.g(i)
+                print(self.g(i).shape)
+                G_nodes[:, i] = self.g(i)
 
-            self._G = (
-                sdiag(self.mesh.cell_volumes) @ (self.mesh.average_node_to_cell @ G).T
+            self._G = (self.mesh.average_node_to_cell @ G_nodes).T @ sdiag(
+                self.mesh.cell_volumes
             )
         return self._G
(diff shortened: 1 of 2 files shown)


SimPEG/simulation.py
```diff
@@ -567,6 +567,14 @@ class ExponentialSinusoidSimulation(LinearSimulation):
         \int_x e^{p j_k x} \cos(\pi q j_k x) \quad, j_k \in [j_0, ..., j_n]
     """
 
+    def __init__(self, n_kernels=20, p=-0.25, q=0.25, j0=0.0, jn=60.0, **kwargs):
```
lheagy, line 570: the only change here is that I moved the init

jcapriot:
Would it make sense to instead define an option on the initialization akin to:

```
integrate : {'mid', 'trap', 'exact'}
```

and then just have the kernel functions do the correct thing?

lheagy:
> Would it make sense to instead define an option on the initialization akin to:
> 
> ```
> quadrature : {'mid', 'trap', None}
>     quadrature rule to apply to the kernel integral, None uses analytic solution.
> ```
> 
> and then just have the kernel functions do the correct thing?

Thanks for this suggestion @jcapriot! In my opinion though, I would be supportive of being opinionated on this example and sticking to one option. There are a couple of reasons for this: (1) defining the kernels on nodes is very much in line with the finite volume approach we take throughout SimPEG, (2) this example is generally the first that we have students work with, so keeping it simple reduces opportunities for confusion. 

For example, when I am teaching this, here are the slides I am using to introduce it 

image

image

So here, I think less is more.

lheagy:
Does anyone have concerns about us going ahead with this? I would love to be able to point to it from my class in the next couple of weeks if possible.

lheagy:
👋 @simpeg/simpeg-admins: if someone has a moment to review this, I would appreciate it! I would love to be able to point to this in the class I am teaching on Thursday.

SimPEG/simulation.py
```diff
@@ -665,10 +665,13 @@ def G(self):
         Matrix whose rows are the kernel functions
         """
         if getattr(self, "_G", None) is None:
-            G = np.empty((self.n_kernels, self.mesh.nC))
+            G_nodes = np.empty((self.mesh.n_nodes, self.n_kernels))
 
             for i in range(self.n_kernels):
-                G[i, :] = self.g(i) * self.mesh.h[0]
+                print(self.g(i).shape)
```
jcapriot, line 671: Looks to be a spurious print statement here

jcapriot (changes requested):
Just one small change.

Also, if you want to add a tiny bit to the docstrings saying that it uses Trapezoidal rule to compute the kernel integral, I wouldn’t say no.

lheagy:
Good call, many thanks @jcapriot! I removed the print statement and updated the docstring for the class.

jcapriot (approved):
looks good!

PR #1342, opened 2024-01-28
ISSUE-1341: Set parent of objective functions

### LINKED ISSUE #1341: Unassigned parent attributes in Regularizations
### Describe the issue:

I believe the structure of simpeg inheritance is causing issues. I have been trying to reproduce old codes from Simpeg 0.18.0 and do not believe the argument for 'gradient_type' is the Sparse regularization is working. It looks like this is due to inheritance and setters in the current simpeg, especially with the 'parent' attribute.

It seems that a lot of these classes that rely on complex and deep inheritance have had issues and could potentially be simplified. If this is not the case, it would be nice to put in some tests to make sure that these classes have the intended behavior.

### Reproducable code example:

```python
reg = regularization.VectorAmplitude(
    mesh,
    mapping=idenMap,
    active_cells=actv,
    reference_model_in_smooth=True,
    norms=[0.0, 2.0, 2.0, 2.0],
    gradient_type="total",
)
```

### Error message:

_No response_

### Runtime information:

Not applicable

### Context for the issue:

_No response_

domfournier:
#### Summary
Assign the parent on the list of  `ComboObjectiveFunction.objfcts`, such that a child function can navigate up the tree and interact with its siblings. 

We could eventually apply the same mechanics to the `BaseObjectiveFunction`, such that the misfits can also have parental relationships, but that's beyond scope.

#### Reference issue
Closes #1341 

The parent on the objective function was never set, and prevented the Sparse amplitude to apply the `gradient_type = "total"`

### DIFF
--- a/SimPEG/objective_function.py
+++ b/SimPEG/objective_function.py
@@ -388,9 +388,6 @@ class ComboObjectiveFunction(BaseObjectiveFunction):
 
         super().__init__(nP=nP)
 
-        for fct in objfcts:
-            fct.parent = self
-
         self.objfcts = objfcts
         self._multipliers = multipliers
         self._unpack_on_add = unpack_on_add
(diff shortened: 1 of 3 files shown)


jcapriot:
If we do want all of the `ObjectiveFunction`s to have access to this `parent` when inside a `ComboObjectiveFunction`, you should move the validator that currently exists on `regularization.base.BaseRegularization` to the `BaseObjectiveFunction`.

I'm still not quite sold it's the way to go, but at least it would solve #1341

domfournier:
Ok, as a compromise, I can move it down level to the WeightedLeastSquares. So at least the change is contained to the reg functions.

jcapriot (approved):
looks good!

domfournier:
Thanks @jcapriot

PR #1343, opened 2024-01-29
Ravel instead of flatten

thibaut-kobold:
`ravel` will avoid returning a new copy of the array; since a new array is already being created w/ toarray this is safe.

Thanks @jedman for  bringing this to my attention.


domfournier (approved):
Fine by me.

PR #1347, opened 2024-02-03
Upgrades to static EM simulations.

dccowan:
This pull request is to add functionality that was deemed useful after the AgroGeo24 workshop. In this pull request, we aim to add the following functionality:

* Linear anisotropy to 2.5D DC/IP simulation classes. This includes the forward problem and derivatives that can be used to develop and inversion framework.
* A 2.5D frequency-domain spectral IP simulation class. I suppose that we assume inductive effects are negligible and that the problem is effectively static. We cannot simply parameterize the IP effects using a Cole-Cole model. A prominent IP effect are caused by electric double layer capacitance, which is not parameterized using a Cole-Cole model. And user-defined parameterizations may be desired.

### DIFF
--- a/SimPEG/electromagnetics/static/resistivity/simulation_2d.py
+++ b/SimPEG/electromagnetics/static/resistivity/simulation_2d.py
@@ -1,4 +1,5 @@
 import numpy as np
+from scipy import sparse as sp
 from scipy.optimize import minimize
 import warnings
 
@@ -24,8 +25,30 @@ from discretize.utils import make_boundary_bool
 
 
 class BaseDCSimulation2D(BaseElectricalPDESimulation):
-    """
-    Base 2.5D DC problem
+    r"""Base 2.5D DC resistivity simulation class.
+
+    Parameters
+    ----------
+    mesh : discretize.BaseMesh
+        The mesh.
+    survey : None, Survey
+        The DC resisitivity survey.
+    nkz : int
+        Number of evaluations of the 2D problem in the wave domain.
+    storeJ : bool
+        Whether to construct and store the sensitivity matrix.
+    miniaturize : bool
+        If ``True``, we compute the fields for each unique source electrode location.
+        We avoid computing the fields for repeated electrode locations and the
+        fields for dipole sources can be constructed using superposition.
+    do_trap : bool
+        Use trap method to find an optimum set of quadrature points and weights
+        for how many evaluations of the 2D problem are required in the wave domain.
+    fix_Jmatrix : bool
+        Whether to fix the sensitivity matrix during Newton iterations.
+    surface_faces : None, numpy.ndarray of bool
+        Array defining which faces to interpret as surfaces of the Neumann boundary.
+
     """
 
     fieldsPair = Fields2D  # SimPEG.EM.Static.Fields_2D
@@ -37,7 +60,7 @@ class BaseDCSimulation2D(BaseElectricalPDESimulation):
         self,
         mesh,
         survey=None,
-        nky=11,
+        nkz=11,
         storeJ=False,
         miniaturize=False,
         do_trap=False,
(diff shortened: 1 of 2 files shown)


jcapriot:
I would recommend leaving the number of wave numbers in the perpendicular horizontal component as `nky` as is. I think `nkz` might cause people to assume that it is doing some sort of vertical integral.

jcapriot:
Also, if you don’t mind, please split the two suggestion’s up into two separate PR’s.

SimPEG/electromagnetics/static/resistivity/simulation_2d.py
```diff
@@ -648,25 +734,61 @@ def bc_type(self, value):
...
+        if getattr(self, "_MeSigma2D", None) is None:
+            if len(self.sigma) == self.mesh.nC:
+                self._MeSigma2D = self.MeSigma
+            elif len(self.sigma) == 3 * mesh.nC:
+                self._MeSigma2D = self.mesh.self.mesh.get_edge_inner_product(
+                    model=self.sigma[:2*self.mesh.nC]
+                )
+            else:
+                raise NotImplementedError(
+                    "Only isotropic and linear isotropic conductivities implemented."
```
jcapriot, line 749: This formulation should actually be able to handle all anisotropic conductivities (as we don’t need to invert the matrix). Let me double check what the tensor reduces to.

dccowan:
> I would recommend leaving the number of wave numbers in the perpendicular horizontal component as `nky` as is. I think `nkz` might cause people to assume that it is doing some sort of vertical integral.

Then perhaps we need to consider a third option. When we generate meshes for the 2.5D problem, and when we ask for things like cell center locations, the coordinates are x (along-line) and y (elevation). This is also the case when defining 2.5D DC/IP surveys. So 'nky' is not the number along the y-direction. Could we have a property like 'n_wave_coefficients'?

dccowan:
@jcapriot what do you think about defining new properties like 'MeSigma2D' and 'MnSigma2D' as opposed to overriding 'MeSigma' and 'MnSigma'? If we add the new properties, I will need to ensure they are cleared upon a model update.

SimPEG/electromagnetics/static/resistivity/simulation_2d.py
jcapriot: Oh I just meant in the future it can support more than just axial anisotropy, compared to the cell centered one that would only support axial.

jcapriot:
> > I would recommend leaving the number of wave numbers in the perpendicular horizontal component as `nky` as is. I think `nkz` might cause people to assume that it is doing some sort of vertical integral.
> 
> Then perhaps we need to consider a third option. When we generate meshes for the 2.5D problem, and when we ask for things like cell center locations, the coordinates are x (along-line) and y (elevation). This is also the case when defining 2.5D DC/IP surveys. So 'nky' is not the number along the y-direction. Could we have a property like 'n_wave_coefficients'?

We create meshes with x1 and x2 dimensions that correspond to x and z. The only properties of meshes that specifically refer to dimensions are on a `TensorMesh`. (`cell_centers_x, _y`,  etc. and `nodes_x, _y`). Most other properties don’t make reference to a name for that dimension anymore.

Also, making it the way you’ve suggested would mean that for 2.5D the anisotropy tensor would be ordered: along-line, elevation, cross-line, and for 3D it would be along-line, cross-line, elevation. So simply for consistencies sake we should continue referring to the cross line dimension as y.

jcapriot:
> @jcapriot what do you think about defining new properties like 'MeSigma2D' and 'MnSigma2D' as opposed to overriding 'MeSigma' and 'MnSigma'? If we add the new properties, I will need to ensure they are cleared upon a model update.

Leave the properties named the same, in fact I would just call the `super` methods when we don’t have anisotropy there.

PR #1327, opened 2024-01-14
Parse directives

domfournier:
**This PR is for discussion and planning only**

Proposed restructuring of the directives under different modules. The goal is to reduce code duplication across classes and overall improve code quality. After agreeing on a structure, moving/refactoring individual directives would happen one at a time to facilitate the review process.

The structure currently proposed is as follow:

# base
The base classes
 - InversionDirective
 - DirectiveList
 
# maps
Any directive that modify the inversion model itself.

 - ProjectSphericalBounds
 
# optimization
Directives that alter the overall objective function

 - UpdatePreconditioner
 - BetaSchedule

# misfit
Directives triggered by a change in phi_d

 - TargetMisfit
 - MultiTargetMisfits
 
# pgi
All PGI stuff

- PGI_UpdateParameters
- PGI_BetaAlphaSchedule
- PGI_AddMrefInSmooth,

# regularization
Directives that change the regularization function, either in terms of weights or scaling
- AlphasSmoothEstimate_ByEig
- Update_IRLS
- UpdateSensitivityWeights
- Update_Wj,

# save
All saving directive mechanisms (need major cleanup)

 - SaveEveryIteration
 - SaveModelEveryIteration
 - SaveOutputEveryIteration
 - SaveOutputDictEveryIteration
 - SimilarityMeasureSaveOutputEveryIteration

# joint

- SimilarityMeasureInversionDirective
- MovingAndMultiTargetStopping
- PairedBetaSchedule
- JointScalingSchedule
- ScalingMultipleDataMisfits_ByEig

# tradeoff_estimator
Anything that has to do with computing the tradeoff beta

 - BetaEstimateMaxDerivative
 - BetaEstimate_ByEig
 - PairedBetaEstimate_ByEig

Let's chat.

PR #1349, opened 2024-02-04
Irls refactor

domfournier:
#### Summary

Refactoring of the UpdateIRLS directive used to control the `Sparse` regularization. 

- Attribute name changes to snake_case format
- Remove unused attributes and change to instance attributes.
- Move responsibility for the scaling of spherical parameters into a new `SphericalDomain` directive.
- Move responsibility for the cooling schedule to a `BetaSchedule` directive.

#### PR Checklist
* [x ] If this is a work in progress PR, set as a Draft PR
      to a Pull Request
* [x ] Tagged ``@simpeg/simpeg-developers`` when ready for review.

#### Reference issue
Related to restructuring proposed on #1327 

Reduces complexity of current `Update_IRLS` class + PEP8 standard. 

#### Additional information
Clean up deprecation warnings on #1472


domfournier:
Sorry for the messy commit history. Pylint went a bit nuts on the imports.

santisoler:
> Sorry for the messy commit history. Pycharm went a bit nuts on the imports.

No worries! We can [Squash and Merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits) and clean it up later.

I see a few tests failing. Do you want me to review it now @domfournier? Or are you still working on it?

domfournier:
Hi @santisoler , yes let me fix those. I can ping you when I am done with it. Errors are mostly related to examples using deprecated arguments, which I should handle more gracefully right here.

Are we ok with updating the tests/examples for suppress the warnings after, on a separate PR?

santisoler:
> Hi @santisoler , yes let me fix those. I can ping you when I am done with it. Errors are mostly related to examples using deprecated arguments, which I should handle more gracefully right here.

Sounds good!

> Are we ok with updating the tests/examples for suppress the warnings after, on a separate PR?

It would be ok to update the examples here as well. But we can totally handle them in a separate PR if you feel we need to change a lot of them.

santisoler (commented):
Hi @domfournier! I went through a first pass on this PR. Before I dig deeper into it I wanted to bring up a few things.

**Deprecations**

Since we are writing a new class for this directive, I think we should implement the new interface for it without worrying about supporting the old arguments. In a similar way that the old magnetics' `SourceField` required the `parameters` argument, while the new `UniformBackgroundField` requires the user to pass the `amplitude`, `inclination` and `declination`.

So, in this case I would:
1. Keep the code of `Update_IRLS` as it is, but leave the decorator for raising a FutureWarning.
2. Implement the new `UpdateIRLS` without any support for old arguments. It should take `verbose` (no `silent`), `cooling_factor` and `cooling_rate` (not `coolingFactor` or `coolingRate`), etc.

**Nested structure for directives**

I noticed that this class stores a `BetaSchedule` directive inside it. That basically creates a nested structure for the directive list, where some directives contain another directives. I think this could create some confusion since the only assumption we make on directives is that they are executed in the order in which they appear in the directive list. I would try to avoid this structure (more on that later).

**Modifying the directive list from one of the directives**

At the moment the directives have a lot of power: they can modify targets, multipliers, weights, GMMs (on PGI), etc. But we can think that all these changes are carried out _downward_ through the inversion process, and in an order specified by the user in the directive list. This class is now able to edit the directive list if some directive class is missing. We can think about these changes to be _sidewise_. 

Although I appreciate that we would be warning users about it, I think this provides the directives too much power. Moreover, the `BetaSchedule` and the `SphericalDomain` directives are added to the end and start of the directive list, respectively, which are not usually the desired position for those.

I would avoid modifying the directive list, and keep the `validate` method to validations only.

For the beta schedule, I think we could implement a cooling schedule inside the `UpdateIRLS` directive without the need for it to have its own `BetaSchedule` directive and without the need to add one to the directive list.

For the spherical domain, the directive takes only 15 lines. Maybe we could make those lines as part of the `UpdateIRLS` and add a flag to its constructor allowing users to switch the update of the spherical parameters on/off. Something like:
```python
class UpdateIRLS:
    def __init__(self, ..., update_spherical_params=False):
        ...
        self.update_spherical_params = update_spherical_params

    def endIter(self):
        ...
        if self.update_spherical_params:
            self._update_spherical_scaling()
        ...

    def _update_spherical_scaling(self):
        """
        Add an 'angle_scale' to the list of weights on the angle regularization for the
        different block of models to account for units of radian and SI.
        """
        ...
```
Let me know what do you think, I'm open to ideas!

**Minor changes**

I just deleted the old `SimPEG/directives/__init__.py` since we don't need it.

I also left a few comments below. Check them out and let me know what do you think.

---

Looking forward to continue the conversation on this. Let me know if you need any help or clarification of my comments.

simpeg/directives/directives.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, inversion=None, dmisfit=None, reg=None, verbose=False, **kwar
             raise TypeError(f"'{key}' property has been removed. Please use 'verbose'.")
         self.inversion = inversion
         self.dmisfit = dmisfit
-        self.reg = reg
+        self.reg: ComboObjectiveFunction | None = reg
```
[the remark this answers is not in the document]
domfournier: Fair enough, I thought it was guaranteed that `self.reg` was always a Combo since it is enforced by the `inverse_problem.reg` (and default if not provided). But that's not true if we let the user input it.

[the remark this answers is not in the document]
domfournier: Done

domfournier:
@santisoler I reverted the deletion of the init. You will have to explain to me why we don't need those anymore.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,578 @@
...
+    ----------
+
+    beta_search: Proceed with a beta search step if the target misfit is outside the tolerance.
+    misfit_tolerance: Tolerance for the target misfit.
+    chifact_start: Starting chi factor for the IRLS iterations.
+    chifact_target: Target chi factor for the IRLS iterations.
+    cooling_factor: Factor to cool the IRLS threshold epsilon.
+    f_min_change: Minimum change in the regularization function to continue the IRLS iterations.
+    max_irls_iterations: Maximum number of IRLS iterations.
+    percentile: Percentile of the function values used to determine the initial IRLS threshold.
```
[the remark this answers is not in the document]
domfournier: It's going to duplicate with typehint, but makes sense for consistancy.

simpeg/directives/__init__.py
```diff
@@ -125,6 +126,16 @@
     PGI_AddMrefInSmooth,
 )
 
+from ._regularization import UpdateIRLS
+
+
+@deprecate_class(removal_version="0.22.0", error=False)
```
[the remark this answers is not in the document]
domfournier: Done

simpeg/directives/__init__.py
```diff
@@ -125,6 +126,16 @@
...
+class Update_IRLS(UpdateIRLS):
+    """Deprecated class, replaced by Smallness."""
```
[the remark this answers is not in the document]
domfournier: Brought back the class under `directives` for now

examples/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh.py
```diff
@@ -350,7 +350,6 @@
     minGNiter=1,
     beta_tol=0.5,
     coolingRate=1,
-    coolEps_q=True,
```
santisoler, line 353: If we plan to keep this example using `Update_IRLS`, we shouldn't need to delete this line (at least for now), right?

simpeg/directives/directives.py
```diff
@@ -140,7 +142,7 @@ def opt(self):
         return self.invProb.opt
 
     @property
-    def reg(self):
+    def reg(self) -> ComboObjectiveFunction:
```
santisoler, line 145: Since the regularizations can either be a combo or an objective function, I think we should update these type hints:
```suggestion
    def reg(self) -> BaseObjectiveFunction:
```

simpeg/directives/directives.py
```diff
@@ -164,7 +166,7 @@ def reg(self, value):
         self._reg = value
 
     @property
-    def dmisfit(self):
+    def dmisfit(self) -> ComboObjectiveFunction:
```
santisoler, line 169:
```suggestion
    def dmisfit(self) -> BaseObjectiveFunction:
```

simpeg/directives/directives.py
```diff
@@ -1963,6 +1965,7 @@ def endIter(self):
         self.outDict[self.opt.iter] = iterDict
 
 
+@deprecate_class(removal_version="0.23.0", error=False)
```
santisoler, line 1968: 👍🏼

simpeg/inverse_problem.py
```diff
@@ -96,7 +96,7 @@ def counter(self, value):
         self._counter = value
 
     @property
-    def dmisfit(self):
+    def dmisfit(self) -> ComboObjectiveFunction:
```
santisoler, line 99:
```suggestion
    def dmisfit(self) -> BaseObjectiveFunction:
```

simpeg/inverse_problem.py
```diff
@@ -113,7 +113,7 @@ def dmisfit(self, value):
         self._dmisfit = value
 
     @property
-    def reg(self):
+    def reg(self) -> ComboObjectiveFunction:
```
santisoler, line 116:
```suggestion
    def reg(self) -> BaseObjectiveFunction:
```

tutorials/07-fdem/plot_inv_1_em1dfm.py
```diff
@@ -271,7 +271,7 @@
 
 # Directive for the IRLS
 update_IRLS = directives.Update_IRLS(
-    max_irls_iterations=30, minGNiter=1, coolEpsFact=1.5, update_beta=True
+    max_irls_iterations=30, minGNiter=1, coolEpsFact=1.5
 )
```
santisoler, line 275: Shouldn't we need to make this example to use the new class? Otherwise I'd keep this example untouched (until we update it).

tutorials/08-tdem/plot_inv_1_em1dtm.py
```diff
@@ -260,7 +260,7 @@
 
 # Directives for the IRLS
 update_IRLS = directives.Update_IRLS(
-    max_irls_iterations=30, minGNiter=1, coolEpsFact=1.5, update_beta=True
+    max_irls_iterations=30, minGNiter=1, coolEpsFact=1.5
 )
```
santisoler, line 264: Same for this example

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+    mode: Literal[1, 2] = 1
+
+
+class UpdateIRLS(InversionDirective):
+    """
+    Directive to control the IRLS iterations for regularization.Sparse.
+
+    Parameters
+    ----------
+
```
santisoler, line 42:
```suggestion
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+    def misfit_tolerance(self, value):
+        self._misfit_tolerance = validate_integer("misfit_tolerance", value, min_val=0)
+
+    beta_tol = deprecate_property(
+        misfit_tolerance,
+        "beta_tol",
+        "UpdateIRLS.misfit_tolerance",
+        future_warn=True,
+        removal_version="0.22.0",
+    )
```
santisoler, line 138: Since this is a new class, I don't think we should worry about maintaining deprecated properties from the old class (`Update_IRLS`). Let's just move forward with the properties we want this class to have and how we want it to behave. If anyone needs backward compatibility, we'll still provide the `Update_IRLS` class for a while.
```suggestion
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+    def percentile(self, value):
+        self._percentile = validate_integer("percentile", value, min_val=0, max_val=100)
+
+    prctile = deprecate_property(
+        percentile,
+        "prctile",
+        "UpdateIRLS.percentile",
+        future_warn=True,
+        removal_version="0.22.0",
+    )
```
santisoler, line 155: Same with this one.
```suggestion
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+    @property
+    def iterStart(self) -> int:
+        """Iteration number when the IRLS process started."""
+
+        warnings.warn(
+            "The `iterStart` property is deprecated. Use `metrics.start_irls_iter` instead.",
+            stacklevel=2,
+        )
+
+        return self.metrics.start_irls_iter
```
santisoler, line 166:
```suggestion
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+        self.percentile: int = percentile
+
+        if "silent" in kwargs:
+            warnings.warn(
+                "The `silent` keyword argument is deprecated. "
+                "Use `verbose` instead.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            kwargs["verbose"] = not kwargs.pop("silent")
```
santisoler, line 90: Similar thing with the arguments. I don't think we should support a deprecated `silent` argument. Let's just keep with `verbose`. BTW, would be nice to include `verbose` as an optional argument in the constructor (and include it in the docstring), so it's more discoverable than having to pass it through kwargs.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+
+class UpdateIRLS(InversionDirective):
+    """
+    Directive to control the IRLS iterations for regularization.Sparse.
+
+    Parameters
+    ----------
+
+    beta_search: bool
+        Proceed with a beta search step if the target misfit is outside the tolerance.
```
santisoler, line 44: `beta_search` is not an argument of the constructor. Should docstrings be updated or the constructor itself?

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,482 @@
...
+    def __init__(
+        self,
+        beta_schedule: BetaSchedule | None = None,
+        chifact_start: float = 1.0,
+        chifact_target: float = 1.0,
+        cooling_factor: float = 1.2,
+        f_min_change: float = 1e-2,
+        max_irls_iterations: int = 20,
+        misfit_tolerance: float = 1e-1,
+        percentile: int = 100,
```
santisoler, line 70: Updating the type hint according to the docstring. Is it right?
```suggestion
        percentile: float = 100,
```

santisoler (changes requested):
Thanks for addressing my previous comments, @domfournier. Just went through a second pass over this PR. I think it's getting in good shape. I left a few comments below.

My major considerations with this PR are:
* The `UpdateIRLS` contains a `BetaSchedule` directive, generating a nested structure of directives. I'd like it better if the `UpdateIRLS` implements its own beta scheduling as required instead of relying on a underlying directive object.
* Missing tests. The new class is not being tested. We should add tests for it. We would also need to keep the old tests for the deprecated `Update_IRLS` class. We'll remove those when we delete that class though.

BTW, are you planning to update the examples using `Update_IRLS` in a separate PR, right?

Let me know if you need any clarification on my comments or any help with this!

domfournier:
> Thanks for addressing my previous comments, @domfournier. Just went through a second pass over this PR. I think it's getting in good shape. I left a few comments below.
> 
> My major considerations with this PR are:
> 
> * The `UpdateIRLS` contains a `BetaSchedule` directive, generating a nested structure of directives. I'd like it better if the `UpdateIRLS` implements its own beta scheduling as required instead of relying on a underlying directive object.

I would push back on this one. That is how it was before, and just ended up duplicating a lot of code. I kind of like the idea of re-using the mechanics of BetaSchedule without exposing it to the user. I don't really see the harm in that?

> * Missing tests. The new class is not being tested. We should add tests for it. We would also need to keep the old tests for the deprecated `Update_IRLS` class. We'll remove those when we delete that class though.

Yes can do.

> 
> BTW, are you planning to update the examples using `Update_IRLS` in a separate PR, right?

Next PR: #1472 
> 
> Let me know if you need any clarification on my comments or any help with this!

examples/03-magnetics/plot_inv_mag_MVI_Sparse_TreeMesh.py
domfournier: True

simpeg/inverse_problem.py
domfournier: Not here, since the setter is enforcing Combo

simpeg/inverse_problem.py
domfournier: Same, also enforced.

tutorials/08-tdem/plot_inv_1_em1dtm.py
domfournier: Had to do this in the previous implementation as the old class was getting overloaded with the new old. Reverting.

domfournier:
@santisoler ready for round 3

simpeg/inverse_problem.py
santisoler: Right! My bad. I still think these properties should be more general, but we shouldn't tackle it in this PR. Let's leave your type hints then.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+    percentile: int
+        Percentile of the function values used to determine the initial IRLS threshold.
+    verbose: bool
+        Print information to the screen.
+    """
+
+    def __init__(
+        self,
+        coolingRate: int = 1,
+        coolingFactor: float = 2.0,
```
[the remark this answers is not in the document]
santisoler: Moreover, since we are only inheriting `BetaSchedule` to use the `coolingFactor` and `coolingRate` properties, wouldn't be better to just implement those here (for `cooling_factor` and `cooling_rate`)? I know we would be repeating some code, but I think it worth for making the code cleaner: we are overriding the `endIter` method which is the core of the `BetaSchedule` directive, so the inheritance is only due to reuse of two properties.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+            )
+
+        beta_schedule = [
+            d for d in directive_list if isinstance(d, BetaSchedule) and d != self
+        ]
+
+        assert len(beta_schedule) == 0, (
+            "Beta scheduling is handled by the `UpdateIRLS` directive."
+            "Remove the redundant `BetaSchedule` from your list of directives.",
+        )
```
santisoler, line 340: We could replace these assertion errors for explicit error raising. Also, the check for an empty list could be simplified to `if not beta_schedule`.
```suggestion
        if not beta_schedule:
            raise TypeError(
            "Beta scheduling is handled by the `UpdateIRLS` directive."
            "Remove the redundant `BetaSchedule` from your list of directives.",
        )
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+        Add an 'angle_scale' to the list of weights on the angle regularization for the
+        different block of models to account for units of radian and SI.
+        """
+        # TODO Need to establish a clearer connection between the regularizations
+        # and the model blocks. There is an assumption here that the first
+        # regularization controls the amplitude.
+        max_p = []
+        for reg in self.reg.objfcts[0].objfcts:
+            f_m = abs(reg.f_m(self.invProb.model))
+            max_p += [np.max(f_m)]
```
santisoler, line 416: Minor change to use append here.
```suggestion
            max_p.append(np.max(f_m))
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+
+        max_p = np.asarray(max_p).max()
```
santisoler, line 418: No need to cast into array.
```suggestion
        max_p = max(max_p)
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+        if (
+            f_change < self.f_min_change
+            and self.metrics.irls_iteration_count > 1
+            and np.abs(
+                1.0
+                - self.invProb.phi_d / self.misfit_from_chi_factor(self.chifact_target)
+            )
+            < self.misfit_tolerance
+        ):
+            print("Minimum decrease in regularization." + "End of IRLS")
```
santisoler, line 382: Should we add a `if self.verbose` here?
```suggestion
            if self.verbose:
                print("Minimum decrease in regularization." + "End of IRLS")
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+            if isinstance(reg, (Sparse, BaseSparse)):
+                reg.model = self.invProb.model
+                phim_new += reg(reg.model)
+
+        # Check for maximum number of IRLS cycles
+        if self.metrics.irls_iteration_count == self.max_irls_iterations:
+            if self.verbose:
+                print(
+                    "Reach maximum number of IRLS cycles:"
+                    + " {0:d}".format(self.max_irls_iterations)
```
santisoler, line 366: Small change here to use f-string.
```suggestion
                    "Reach maximum number of IRLS cycles:"
                    + f" {self.max_irls_iterations:d}"
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    Data class to store metrics used by the IRLS algorithm.
+
+    Parameters
+    ----------
+
+    input_norms: List of norms temporarily stored during the initialization.
+    irls_iteration_count: Number of IRLS iterations.
+    start_irls_iter: Iteration number when the IRLS process started.
+    f_old: Previous value of the regularization function.
```
santisoler, line 24: Update docstring to numpy docstyle.
```suggestion
    input_norms : list of floats
        List of norms temporarily stored during the initialization.
    irls_iteration_count : int
        Number of IRLS iterations.
    start_irls_iter : int or None
        Iteration number when the IRLS process started.
    f_old : float
        Previous value of the regularization function.
```

santisoler (changes requested):
Thanks @domfournier for addressing my previous comments. Sorry about the thing with the type hint, I didn't notice that the setter was enforcing the data misfit and regularizations to be combos. I don't quite like that behavior, but we shouldn't change that in this PR.

I left a few minor suggestions below. The two major things I would do is to reconsider this class as a child of `BetaSchedule` (see more in comment below), and replace the assertion errors in the `validate` method for error raising (other than assertions).

Also, we need to add more tests to this new class. Specially to the core methods: `validate`, `endIter`, `start_irls` and `stopping_criteria`. And also to the `SphericalDomain` class.

domfournier:
@santisoler What's the next step?

domfournier:
Would be great to move this forward

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,458 @@
...
+        Target chi factor for the IRLS iterations.
+    irls_cooling_factor: float
+        Factor to cool the IRLS threshold epsilon.
+    f_min_change: float
+        Minimum change in the regularization function to continue the IRLS iterations.
+    max_irls_iterations: int
+        Maximum number of IRLS iterations.
+    misfit_tolerance: float
+        Tolerance for the target misfit.
+    percentile: int
```
santisoler, line 58: Small change on the type of the `percentile` argument:
```suggestion
    percentile: float
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,458 @@
...
+            f_change < self.f_min_change
+            and self.metrics.irls_iteration_count > 1
+            and np.abs(
+                1.0
+                - self.invProb.phi_d / self.misfit_from_chi_factor(self.chifact_target)
+            )
+            < self.misfit_tolerance
+        ):
+            if self.verbose:
+                print("Minimum decrease in regularization." + "End of IRLS")
```
santisoler, line 413: Minor change here to simplify the string.
```suggestion
                print("Minimum decrease in regularization. End of IRLS")
```

.pre-commit-config.yaml
```diff
@@ -3,12 +3,12 @@ repos:
     rev: 24.3.0
     hooks:
       - id: black
-        language_version: python3.11
+        language_version: python3.10
```
santisoler, line 6: We upgraded the python version used by precommit in #1482. I wouldn't downgrade it to 3.10 here. Is there any reason for this change? If not, feel free to restore these lines to 3.11:
```suggestion
        language_version: python3.11
```

.pre-commit-config.yaml
```diff
@@ -3,12 +3,12 @@ repos:
...
   - repo: https://github.com/pycqa/flake8
     rev: 7.0.0
     hooks:
       - id: flake8
-        language_version: python3.11
+        language_version: python3.10
```
santisoler, line 11:
```suggestion
        language_version: python3.11
```

santisoler (changes requested):
Hi @domfournier. Sorry for the delay on this review. I added a few minor suggestions below, nothing critical.

I think this is getting in good shape. As I mentioned before, I think this PR needs more tests, specially to test the core methods: `validate`, `endIter`, `start_irls` and `stopping_criteria`; and also to the `SphericalDomain` class.

I see that codecov is raising warnings on lack of coverage on some of the new lines. They can be used as guide to extend the tests.

Let me know if you need any help with those!

domfournier:
Aight @santisoler , ready for another round.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,494 @@
...
+        """
+        # After reaching target misfit with l2-norm, switch to IRLS (mode:2)
+        if (
+            self.metrics.start_irls_iter is None
+            and self.invProb.phi_d < self.misfit_from_chi_factor(self.chifact_start)
+        ):
+            self.start_irls()
+
+        # Check if misfit is within the tolerance, otherwise scale beta
+        self.adjust_cooling_schedule()
```
santisoler, line 272: I'm a little bit confused about this line. In the deprecated `Update_IRLS`, the `cooling_factor` was only updated when *mode 2* was reached. But in the new class this line is being run after every iteration, regardless if the IRLS process started or not. This means that while performing the L2 iterations we are updating the cooling factor. Is this the desired behaviour?

I would also update the comment, since it doesn't reflect what the line is doing: the `adjust_cooling_schedule` doesn't scale beta, it only updates the `cooling_factor`.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,494 @@
...
+
+    @cooling_factor.setter
+    def cooling_factor(self, value):
+        self._cooling_factor = validate_float(
+            "cooling_factor", value, min_val=0.0, inclusive_min=False
+        )
+
+    @property
+    def cooling_rate(self):
+        """Cool after this number of iterations.
```
santisoler, line 178: Minor update to make this property more clear.
```suggestion
        """Cool beta after this number of iterations.
```

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,494 @@
...
+        ):
+            self.start_irls()
+
+        # Check if misfit is within the tolerance, otherwise scale beta
+        self.adjust_cooling_schedule()
+
+        # Only update after GN iterations
+        if (
+            self.metrics.start_irls_iter is not None
+            and (self.opt.iter - self.metrics.start_irls_iter) % self.cooling_rate == 0
```
santisoler, line 277: From what I understood from the docstrings, the `cooling_rate` was used to control how often we update `beta`. But here it's used to also control how often we update the `irls_threshold`s. Is this the desired behaviour? If that's so, could we make it more explicit in the docs?

santisoler (commented):
@domfournier! I left some comments below. Feel free to check them out and let me know what do you think.

After addressing these, I think this should be ready to be merged. I noticed that there are a lot of tests that use the old `Update_IRLS` class. Were you planning to update those in a separate PR?

simpeg/directives/_regularization.py
domfournier: Yes you are right, missing a piece of logic there. Just pushed a fix.

simpeg/directives/_regularization.py
```diff
@@ -0,0 +1,494 @@
...
+        chi_factor : float
+            Chi factor to compute the target misfit from.
+        """
+        value = 0
+        if isinstance(self.survey, list):
+            for survey in self.survey:
+                value += survey.nD * chi_factor
+
+        else:
+            value = self.survey.nD * chi_factor
```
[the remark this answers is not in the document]
domfournier: I am agnostic. Sort of beyond the scope of this PR though...

santisoler:
@domfournier, I just pushed a few changes to this branch. Most of them minor things or documentation-related stuff.

* I added the new classes to the API reference.
* I changed the default values for the `input_norms` in the dataclass to None, since we cannot be sure that we'll always have four objective functions. Moreover, the default value was a tuple, but the type hint was stating it as a list. The issue is that we cannot set mutable variables (lists for example) as default variables. I think having it set to None is the best compromise.
* I bumped the simpeg version in which we'll remove the `Update_IRLS` class to `v0.24.0`.


I have one more concern about the new class. Currently, it's a child of `BetaSchedule`, therefore it inherits the `coolingFactor` and `coolingRate` properties (although they are not being used) and the `endIter` is being overwritten. So I don't see the point of that inheritance. Moreover, I see that this inheritance could lead to issues. For example, if you define the directive as:

```python
irls_directive = directives.UpdateIRLS(coolingFactor=3)
```

you won't get any error, since the `coolingFactor` is taken to the `**kwargs` and the `irls_directive.coolingFactor` is set by the `BetaSchedule.__init__`. Nonetheless, the `irls_directive.cooling_factor` (the one that it'll actually be used by the directive) is set as the default value. So, the user might think that they are changing one of the parameters of the directive, while in fact the value of `3` they are using doesn't have any impact in the inversion.

If we make `UpdateIRLS` a child of  `InversionDirective` we would avoid that, because users will get an error after trying to pass `coolingFactor` to the constructor.

What do you think?

---

PS: One of the tests in `test_directives.py` is actually using `coolingFactor`. We might want to fix it:
https://github.com/simpeg/simpeg/blob/e9d86229c326a9e259c6d85b882c3090fdf93e58/tests/base/test_directives.py#L276-L286

domfournier:
I think that you should just PR on this one with changes you want. It's getting pretty discouraging to continue working on this when the target keeps on moving.

domfournier:
Iteration 7. Can we get a merge @lheagy @jcapriot @santisoler

santisoler (approved):
Looks good to me!

domfournier:
@santisoler Not sure why the second set of codecov is stuck. Might have to force merge.

PR #1361, opened 2024-02-17
Set the model when calling `getJ` in DC and SIP simulations

### LINKED ISSUE #1358: BUG: Logic for getJ
### Describe the issue:

Right now, we cache `Jmatrix` in a lot of places (which is good for saving computation!) but we are not careful about checking if the `m` that is provided to `getJ` is indeed consistent with `self.model` and therefore with the `Jmatrix` that is returned. This can lead to some confusing behaviour 

For example, if you are performing a sensitivity analysis, you could get yourself into a confusing spot 

```python
model1 = halfspace 
model2 = model_with_structure 

sim = dc.Simulation2DNodal(mesh, survey, storeJ=True)  # because I later want to run an inversion 

J1 = sim.getJ(model1) 
J2 = sim.getJ(model2)  # this will return J1 -- wrong!
```

### Error message:

_No response_

### Runtime information:

latest SimPEG version 

### Context for the issue:

This is a conceptual bug. Although it could be resolved if we set `storeJ=False`, the behaviour is non-intuitive and is something that we should address. 

The issue lies here https://github.com/simpeg/simpeg/blob/716e221e83b986479037ae90ba9787f53d9b7334/SimPEG/electromagnetics/static/resistivity/simulation_2d.py#L267-L278

My suggestion, include a check that the model provided is basically the same as `self.model.` If this is computationally expensive enough that we don't want to be calling it every time, then we could also allow `m=None` and we return the cached J. Something like the following: 

```
def getJ(self, m=None, f=None):
    """
    Generate Full sensitivity matrix
    """
    if m is None:
        if self._Jmatrix is not None:
            return self._Jmatrix   # early return if no model provided and J has been created

    elif not np.all_close(m, self.model):
        self.model = m    # this will clear self._Jmatrix if it exists 

    if getattr(self, "_Jmatrix", None) is None:
        if self.verbose:
            print("Calculating J and storing")     
        if f is None:
            f = self.fields(m)
        self._Jmatrix = (self._Jtvec(self.model, v=None, f=f)).T
    return self._Jmatrix
```

Thanks @anna1963 for catching this! 

Note that I only highlighted this for DC, but it will be common to wherever we have `getJ`

lheagy:
#### Summary
make sure we set the model when calling getJ. This will ensure that if the model is changed, the sensitivity matrix is cleared (for nonlinear problems) and the sensitivity recomputed. This will close #1358 

#### PR Checklist
      to a Pull Request

#### Reference issue
Closes #1358


lheagy:
@simpeg/simpeg-developers: this is ready for review!

santisoler (changes requested):
Thanks for tackling this @lheagy.

The fixes are very straightforward, but since this is a bugfix, I would suggest adding some tests that would fail without them, so we can be sure we are not facing them again in the future.

Something like this:

```python
import pytest
import numpy as np

from discretize import TensorMesh
from SimPEG import utils
from SimPEG.maps import IdentityMap
from SimPEG.electromagnetics import resistivity as dc


@pytest.fixture
def mesh():
    """Sample mesh."""
    cs = 0.5
    npad = 11
    hx = [(cs, npad, -1.5), (cs, 15), (cs, npad, 1.5)]
    hy = [(cs, npad, -1.5), (cs, 15), (cs, npad, 1.5)]
    hz = [(cs, npad, -1.5), (cs, 15), (cs, npad, 1.5)]
    mesh = TensorMesh([hx, hy, hz], x0="CCC")
    return mesh


@pytest.fixture
def survey(mesh):
    """Sample survey."""
    # Set up survey parameters for numeric solution
    x = mesh.nodes_x[(mesh.nodes_x > -75.0) & (mesh.nodes_x < 75.0)]
    y = mesh.nodes_y[(mesh.nodes_y > -75.0) & (mesh.nodes_y < 75.0)]

    Aloc = np.r_[1.25, 0.0, 0.0]
    Bloc = np.r_[-1.25, 0.0, 0.0]
    M = utils.ndgrid(x - 25.0, y, np.r_[0.0])
    N = utils.ndgrid(x + 25.0, y, np.r_[0.0])

    rx = dc.receivers.Dipole(M, N)
    src = dc.sources.Dipole([rx], Aloc, Bloc)
    survey = dc.survey.Survey([src])
    return survey


def test_getj_model_assignment(mesh, survey, sigma):
    """
    Test model assignment when calling the Simulation3DNodal.getJ method
    """
    simulation = dc.simulation.Simulation3DNodal(
        mesh=mesh, survey=survey, sigmaMap=IdentityMap(mesh)
    )
    model_1 = np.ones(mesh.nC) * 1e-2
    model_2 = np.ones(mesh.nC) * 1e-1
    j_1 = simulation.getJ(model_1)
    j_2 = simulation.getJ(model_2)
    # Check if the two Js are different
    assert not np.allclose(j_1, j_2)
```

Tests like this should be included for each class that has been modified.

lheagy:
Many thanks @santisoler! These are great suggestions for the tests. If you have the time to add those, that would be awesome. I know it is a small thing, but I won't be able to get back to this for a bit. If not, no worries, I can try to find some time later in the week.

santisoler:
@lheagy, I've just pushed some tests for the DC simulations, plus one test for the SIP simulation. The DC tests are all passing, and they fail if we run them on `main`.

The test for SIP is falling into an error, but because I'm doing something wrong. Here I copy the traceback of the test failure:

```
=============================================== FAILURES ===============================================
________________________________ TestSIPSimulations.test_simulation_3d _________________________________

self = <tests.em.static.test_model_assignment.TestSIPSimulations object at 0x7fbf1f1e3810>
mesh_3d =
  TensorMesh: 8,820 cells

                      MESH EXTENT             CELL WIDTH      FACTOR
  dir    nC        mi...-262.50        262.50     25.00     25.00    1.00
   z     20       -500.00          0.00     25.00     25.00    1.00


survey_3d = <SimPEG.electromagnetics.static.spectral_induced_polarization.survey.Survey object at 0x7fbf24d8b5d0>

    def test_simulation_3d(self, mesh_3d, survey_3d):
        """
        Test model assignment on the ``getJ`` method of 3d simulations
        """
        wires = Wires(("eta", mesh_3d.nC), ("taui", mesh_3d.nC))
        sigma = np.ones(mesh_3d.nC) * 1e-2
        simulation = sip.Simulation3DNodal(
            mesh_3d,
            sigma=sigma,
            survey=survey_3d,
            etaMap=wires.eta,
            tauiMap=wires.taui,
        )
        model_1 = np.r_[sigma, 1.0 / sigma]
        model_2 = np.r_[sigma * 2, 1.0 / sigma]
        # Call `getJ` passing a model and check if it was correctly assigned
>       j_1 = simulation.getJ(model_1)

tests/em/static/test_model_assignment.py:161:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <SimPEG.electromagnetics.static.spectral_induced_polarization.simulation.Simulation3DNodal object at 0x7fbf1f222b50>
m = array([1.e-02, 1.e-02, 1.e-02, ..., 1.e+02, 1.e+02, 1.e+02])
f = <SimPEG.electromagnetics.static.resistivity.fields.Fields3DNodal object at 0x7fbf3613f750>

    def getJ(self, m, f=None):
        """
        Generate Full sensitivity matrix
        """
        self.model = m

        if self._Jmatrix is not None:
            return self._Jmatrix
        else:
            if self.verbose:
                print("Calculating J and storing")
            if f is None:
                f = self.fields(m)

            Jt = np.zeros(
>               (self._P.shape[1], int(self.survey.nD / self.survey.unique_times.size)),
                order="F",
            )
E           AttributeError: 'Simulation3DNodal' object has no attribute '_P'

SimPEG/electromagnetics/static/spectral_induced_polarization/simulation.py:373: AttributeError
```

Would you take a look at it and see why is this failing?

After that we would just need to write tests for the 2D SIP simulations. With that I think we should be able to merge this PR.

jcapriot:
It looks like `self._P` is only ever assigned if the user passes an `actind` array. There's actually a lot of logic that will probably not work for the SpectralInducedPolarization class when `storeJ==True`, and probably should be a separate PR fix.

santisoler:
I'm merging this. The tests for the SIP are currently failing, but decorated with `xfail`, so the tests pass. I'll open an Issue to keep track of the issues on that simulation.

Issue #1302, opened 2023-11-29
Maintenance: ditch classes/funcs/methods flagged for deprecation


santisoler:
### Proposed new feature or change:

We have several classes, methods, functions and arguments that were flagged for deprecation for 0.19 (and probably for 0.20).

We should remove them before releasing v0.21.

PR #1365, opened 2024-03-01
Remove deprecated regularization classes

santisoler:
#### Summary

Raise errors when constructing deprecated regularization classes. Add test that checks if those errors are being raised. Replace the removed regularization classes that were still being used in a few pieces of the code and in the tests. Use the new classes instead. Add removed classes to the `IGNORE_ME` list in `test_regularization.py` so other tests don't fail.

#### PR Checklist
      to a Pull Request

Part of the solution for #1302

#### Additional information

Most of these changes were cherry-picked from #1306

### DIFF
--- a/tests/base/test_regularization.py
+++ b/tests/base/test_regularization.py
@@ -793,5 +793,28 @@ class TestDeprecatedArguments:
             BaseRegularization(mesh, weights=weights, cell_weights=weights)
 
 
+class TestRemovedRegularizations:
+    """
+    Test if errors are raised after creating removed regularization classes.
+    """
+
+    @pytest.mark.parametrize(
+        "regularization_class",
+        (
+            regularization.Tikhonov,
+            regularization.Small,
+            regularization.SmoothDeriv,
+            regularization.SmoothDeriv2,
+            regularization.PGIwithNonlinearRelationshipsSmallness,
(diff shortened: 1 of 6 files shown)


tests/base/test_regularization.py
```diff
@@ -39,6 +39,16 @@
...
+    "PGIwithNonlinearRelationshipsSmallness",
+    "PGIwithRelationships",
+    "Simple",
+    "SimpleSmall",
+    "SimpleSmoothDeriv",
+    "Small",
+    "SmoothDeriv",
+    "SmoothDeriv2",
+    "Tikhonov",
 ]
```
jcapriot, line 52: I imagine this "IGNORE_ME" list is starting to include most of our regularizations.

jcapriot (approved):
looks good

tests/base/test_regularization.py
santisoler: I suspect so... I don't quite like this list. I would prefer a more "pro-active" approach of listing which classes we want to test, instead of listing the ones we are going to ignore. Moreover, that list is only used in some tests: is not that all test functions `test_regularization` will not test those classes.

I'll try to improve that in a separate PR since it's not critical for this one.

PR #1382, opened 2024-03-14
Speed up sphinx documentation building

jcapriot:
#### Summary
Stops the left sidebar from being completely expandable for every single item in the documentation.
https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/navigation.html#remove-reveal-buttons-for-sidebar-items

#### PR Checklist
      to a Pull Request

#### Reference issue
No number... but the docs take so long to build now....

Faster doc building (excluding the examples and tutorials)

#### Additional information
[](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/performance.html)

### DIFF
--- a/docs/_templates/autosummary/base.rst
+++ b/docs/_templates/autosummary/base.rst
@@ -1,0 +1,13 @@
+{% if objtype == 'property' %}
+:orphan:
+{% endif %}
+
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+{% if objtype == 'property' %}
+property
+{% endif %}
+
+.. auto{{ objtype }}:: {{ objname }}
--- a/docs/_templates/autosummary/base.rst
+++ b/docs/_templates/autosummary/base.rst
@@ -6,8 +6,4 @@
 
 .. currentmodule:: {{ module }}
 
-{% if objtype == 'property' %}
-property
-{% endif %}
-
 .. auto{{ objtype }}:: {{ objname }}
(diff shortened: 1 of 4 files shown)


jcapriot:
This significantly reduces the build time for the documentation (excluding the gallery). We should get this in quickly.

jcapriot:
Doc build test down from 1hr to 25 minutes

docs/_templates/autosummary/base.rst
```diff
@@ -0,0 +1,13 @@
...
+:orphan:
+{% endif %}
+
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+{% if objtype == 'property' %}
+property
+{% endif %}
```
santisoler, line 11: Is there a reason to add this? It adds a `<p>property</p>` after the `h1` heading and before the definition of the object:

image

santisoler (approved):
This is awesome @jcapriot. Loved to see the build time being reduced. The reason why this is faster is because it had to navigate the whole tree for each page it was building? Just curious.

I left a minor comment here. Besides that, feel free to merge 🚀

jcapriot:
Yeah, for every single page it had to link to basically… every single other page, so the html size (and time to build) grows exponentially with the number of pages.

It would build even faster if we didn’t include the inherited class members (about 13 minutes). We weren’t doing that before because it was making the doc build take… days…. I think we should include them now though because it makes it easier to look at each class without having to traverse the inheritance to see all of the inherited attributes.


We should also look into an automated way to populate the keyword arguments for the simulation __init__ documentation, but that’s a separate issue.

docs/_templates/autosummary/base.rst
jcapriot: I was just taking style from numpy, I’ll remove it for our needs.

PR #1390, opened 2024-03-25
Fix distance calculation in `convert_survey_3d_to_2d_lines`

kehrl-kobold:
#### Summary
Fix a bug in the distance calculation in `convert_survey_3d_to_2d_lines`. Distance is calculated along the survey line from the first point, `r0`. Currently, `r0[0]`, which is just the `x` component, is being subtracted from both the `x` and `y` components for all points to calculate the distance. Instead, `r0` should be subtracted. Update tests so they fail without the bugfix.

#### PR Checklist
      to a Pull Request

See above.


santisoler:
Hi @kehrl-kobold! Thanks a lot for opening this PR. From just reading the code, those changes look right. Did you run into this bug while using this function? Do you have any minimal example that can be used to reproduce the bug?

I'm asking this because it would be nice to add some tests checking that this bug has been fixed. Basically, a test function that should fail if we run it with the current code in `main`, but not with this fix. There's no need for you to write if you're too busy, we can help you with that.

Thanks again for your contribution, it's greatly appreciated!

thibaut-kobold:
@santisoler : just an FYI, the function is used in one tutorial: https://docs.simpeg.xyz/content/tutorials/05-dcr/plot_gen_3_3d_to_2d.html. We should at least rerun the tutorial with the update.

santisoler:
Thanks @thibaut-kobold! You're right, the tutorial will be rerun after we get the updated merged. And in the next release the updated docs will be deployed to the website.

I just ran a similar example to the one in the tutorial and can confirm that the fix is correctly setting the first coordinate of the south-north lines (the first coordinate in the 2D survey): instead of setting them to -650 or -1350 (as you can see in the final figures in that tutorial), it's setting them to zero, as described in the docstring.

So, a nice test for this would be to create a simple 3d DC survey with a west-east and a south-north lines, and check if the dipoles in the 2D survey are correctly, with the first coordinate of the first dipole being equal to zero.

Is this something you are interested in doing @kehrl-kobold?

kehrl-kobold:
I ran into the bug when setting up a rather complex survey in UTM coordinates, so it's definitely not a good example for a test case. I'm happy to set up a test case, but it may take me a bit of time. Would [this](https://github.com/simpeg/simpeg/blob/main/tests/em/static/test_DC_Utils.py) be the correct file for a test?

kehrl-kobold:
> I ran into the bug when setting up a rather complex survey in UTM coordinates, so it's definitely not a good example for a test case. I'm happy to set up a test case, but it may take me a bit of time. Would [this](https://github.com/simpeg/simpeg/blob/main/tests/em/static/test_DC_Utils.py) be the correct file for a test?

I just noticed your PR above with the test case. Thanks!

kehrl-kobold:
@santisoler I added the tests for `convert_survey_3d_to_2d_lines`. Let me know if you have any feedback and if there is anything else I can do to help get this merged. Thanks!

tests/em/static/test_DC_Utils.py
```diff
@@ -326,9 +336,14 @@ def test_convert_to_2d(self):
             survey_2d.locations_n,
         ]
 
+        # Coordinates should be roughly the same
         passed = np.allclose(loc3d[:, 0::2], loc2d)
         self.assertTrue(passed)
 
+        # Check that the first x-coordinate for electrode A is zero for both surveys
+        for i in range(0, len(survey_2d_list)):
+            self.assertEqual(survey_2d_list[i].locations_a[0, 0], 0)
```
santisoler, line 345: A minor improvement on the code style here, just to make it more Pythonic.
```suggestion
        for survey in survey_2d_list:
            self.assertEqual(survey.locations_a[0, 0], 0)
```

tests/em/static/test_DC_Utils.py
```diff
@@ -225,27 +225,36 @@ def setUp(self):
...
+            for ii in range(len(survey_type)):
+                source_list += utils.generate_dcip_sources_line(
+                    survey_type[ii],
+                    data_type,
+                    dimension_type,
+                    end_locations[i],
+                    0.0,
+                    num_rx_per_src,
+                    station_separation,
+                )
```
santisoler, line 249: Another minor change here, also to make it look more Pythonic. Nothing critical, just style.
```suggestion
        for end_location in end_locations:
            for survey_type_i in survey_type:
                source_list += utils.generate_dcip_sources_line(
                    survey_type_i,
                    data_type,
                    dimension_type,
                    end_location,
                    0.0,
                    num_rx_per_src,
                    station_separation,
                )
```

santisoler (approved):
Thanks a lot @kehrl-kobold for pushing these changes. I've run the tests with these new changes against the code in `main` and it fails, therefore the new tests are able to properly catch the bug. Nicely done.

I think this is ready to be merged. I just left two minor comments below, which are not critical for the PR, just some minor improvements to the style.

Thanks again for your contribution. Looking forward to merge this!

kehrl-kobold:
Thanks for the suggestions! I've made the updates.

santisoler:
Thanks @kehrl-kobold! I've just updated this branch to get the latest changes in `main`. I'll merge it after CI passes. Thanks again for spotting this bug and also fixing it!

santisoler:
Merged! 🚀 

Thanks again for your contribution @kehrl-kobold! Feel free to open a PR to add yourself to the `AUTHORS.rst` if you desire.

PR #1394, opened 2024-03-27
Use Numpy random number generator in codebase

### LINKED ISSUE #1290: ENH: Improve how random states are handled in directives
### Proposed new feature or change:

Raised by @lheagy in #1286, we should improve how random states (random seeds) are handled in directives, specially in the beta estimator directives. We should take the opportunity to move from the legacy Numpy's RandomState (using the `numpy.random.seed()` and `numpy.random.randn()` functions) to the new Numpy's [random number generators](https://numpy.org/doc/1.26/reference/random/index.html#module-numpy.random).

santisoler:
Replace the usage of legacy `numpy.random.seed()` in SimPEG codebase for the Numpy random number generator object that can be created through the `numpy.random.default_rng()` function. Add a new `typing` module that contain new type aliases, and create a new `RandomSeed` type alias for variables accepted by `numpy.random.default_rng()`. Extend the accepted values for `seed` arguments in several methods and functions to be `None` or `RandomSeed`. Apply these replacements to directives, `eigenvalue_by_power_iteration()`, `create_random_model()`, utility functions on `test_utils.py` and usage of `eigenvalue_by_power_iteration()` in tests. Update documentation and type hint of `make_synthetic_model()` allowing it to get a Numpy `Generator` as `random_seed`. Add a new documentation page for the new `typing` module.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fixes #1290

- Allow passing `np.random.Generator` to `make_synthetic_data`.
  Update the docs of `BaseSimulation.make_synthetic_data` allowing users to optionally pass a Numpy random number generator as the `random_seed`.
- Use Numpy rng in `eigenvalue_by_power_iteration`.
  Replace the use of `np.random.seed` for a Numpy random number generator. Allow users to pass a `np.random.Generator` as the `seed` argument.
- Use Numpy rng in directives.
  Extend the accepted types of `seed` attributes in directives to int, np.random.Generator or None. Replace the usage of `np.random.seed` for `np.random.default_rng` in directive methods. Explicitly pass instances of `Generator` to calls of `eigenvalue_by_power_iteration`.
- Use Numpy rng in `create_random_model`.
  Replace the `np.random.seed` for Numpy `Generator` in `create_random_model`. Fix check for the shape, turning it into a tuple only if it's a int (not a float).
- Pass `seed` to `eigenvalue_by_power_iteration` in `test_mat_utils.py`.
- Remove global random seed in `test_utils.py`.
- Use Numpy rng in `test_utils.random()` function.
  By default the seed is set to a value (and not None) because this function is used for testing purposes. We want to make sure that tests are run with reproducible random states.

#### TODO

- [x] Update SimPEG to simpeg in the API reference for the typing module? See comment below.


SimPEG/directives/directives.py
```diff
@@ -347,16 +348,18 @@ class BaseBetaEstimator(InversionDirective):
     ----------
     beta0_ratio : float
         Desired ratio between data misfit and model objective function at initial beta iteration.
-    seed : int, None
-        Seed used for random sampling.
+    seed : int, numpy.random.Generator or None, optional
```
jcapriot, line 351: Technically I think we should support the same inputs as https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng (and maybe point the user to this function for more details about the rng?)

SimPEG/directives/directives.py
```diff
@@ -386,13 +389,13 @@ def seed(self):
...
         -------
-        int
+        int, numpy.random.Generator or None
         """
         return self._seed
 
     @seed.setter
     def seed(self, value):
-        if value is not None:
+        if value is not None and not isinstance(value, np.random.Generator):
```
jcapriot, line 398: Maybe we should instead do,
```Python
try:
    rng = np.random.default_generator(value)
except TypeError as err:
    raise TypeError( f'Unable to initialize the random number generator with a {type(value).__name__}`) from err

self._rng = rng
self._seed = rng.bit_generator.seed_seq
```

Which will allow someone to both see the actual used seed (If `None` was passed), and re-use it for a future run.

jcapriot: Not sure what to do if someone were to pass something besides the default bit generator (which I believe is a `PCG64`).

jcapriot (commented):
A few comments for us to consider.

SimPEG/directives/directives.py
santisoler: Makes sense. I'll update the docstrings.

SimPEG/directives/directives.py
santisoler: I usually prefer having a proper check like `if not isisinstance(..., ...): ...` rather than using `try` `except`, but in this case I think your proposal it's way easier than implementing that check since the possible types of `seed` is a little bit complex.

But I'm not sure about storing the `self._seed` as the `seed_seq` of the generator. I think it could create the confusion of assuming that the random numbers generated in the class are done by setting a new rng with that seed and from the beginning of the random sequence.
While in reality, the random numbers are generated from the current random sequence in the generator that was passed. And in the end, we can still access that seed with `directive_obj.seed.bit_generator.seed_seq`.

I also think we should assign the passed value to `self._seed` even if it's a random number generator. This keeps the interface simple and less confusing than having something like:

```python
rng = np.random.default_rng(seed=42)
directive_obj = DirectiveClass(..., seed=rng)
print(directive_obj.seed)
```
```
SeedSequence(
    entropy=42,
)
```

What do you think?

SimPEG/directives/directives.py
```diff
@@ -347,16 +348,18 @@ class BaseBetaEstimator(InversionDirective):
...
+        ``numpy.random.default_rng``.
 
     """
 
     def __init__(
         self,
         beta0_ratio=1.0,
         n_pw_iter=4,
-        seed=None,
+        seed: int | np.random.Generator | None = None,
```
santisoler, line 362: Now I need to update these type hints to match the description in the docstrings. Should we keep type hints in this case or just remove them?

SimPEG/directives/directives.py
jcapriot: I'm more so looking for a way to restart (as it were) the process further down the line in a repeatable manner. So at the very least I think we should store the random number generator on the classes where appropriate. You can access the current state of the `BitGenerator` for a `Generator` as:
```python
generator.bit_generator.state
```

Which could then be used to recreate the `BitGenerator` at whatever state you were interested in.

Maybe it just doesn't make sense to store the "seed" property on classes at all when using generators. We would only allow a seed to be passed to the `__init__`.

SimPEG/directives/directives.py
jcapriot: it is rather large, it would end up being:
```Python
seed: None | int | np.typing.NDArray[np.int_] | np.random.SeedSequence, np.random.BitGenerator, np.random.Generator
```

SimPEG/directives/directives.py
santisoler: I think the capabilities for restarting the random state are already there. You can either reuse the same seed (as `int`), or redefine your Generator with the same seed and with the same state in the random sequence. For example:

```python
seed = 42
directive_obj = DirectiveClass(..., seed=seed)

# Run an inversion and use the directive
...

# Restart random state
directive_obj.seed = seed

# Reuse the directive after resetting the random state
...
```

And I think we should still store the `seed` as an attribute with the same value passed to the constructor, whether it is an `int` or a Generator, to simplify the interface as I pointed out in the last comment.

Nonetheless, I think your proposal would imply a change in the current behavior and/or interface of these classes, which I think is kind of out of the scope of this PR. I'd love to continue this discussion, but maybe we can find a compromise on this PR to move forward and then discuss how to improve the interface for these classses. Do you agree?

SimPEG/directives/directives.py
santisoler: I'm considering defining a [TypeAlias](https://typing.readthedocs.io/en/latest/spec/aliases.html#type-aliases) for the seeds, something like:

```python
RandomSeed: TypeAlias = (
    int
    | np.typing.NDArray[np.int_]
    | np.random.SeedSequence
    | np.random.BitGenerator
    | np.random.Generator
)
```

So the signature of the methods could look like:
```python
    seed: None | RandomSeed = None,
```

We could also define a `SimPEG.typing` module for aliases like this one. I would like to generate a plethora of these, but in some cases they could be useful.

What do you think?

SimPEG/directives/directives.py
jcapriot: That's a good point that the functionality is already there, although it's not directly from simpeg. I was mostly imagining the case where a user initializes the seed on a class instance with an integer, but then would want to access the created generator itself. I'm not saying we shouldn't store the seed, just was bringing it up as a potential option.

santisoler:
@jcapriot I added a new `typing` module with the `RandomSeed` type alias. I also added a new documentation page for it. I noticed that it was easier to just set `RandomSeed.__doc__` than the setup that Numpy has (that basically extends the `__doc__` in the `typing/__init__.py` including the rst bits for every new type alias. But let me know if you think it's enough.

I also updated how the setter methods in the directives check if the given value is valid, following your suggestion.

Is there anything else we would like to include in this PR?

SimPEG/typing/__init__.py
```diff
@@ -0,0 +1,61 @@
+"""
+=============================
+Typing (:mod:`simpeg.typing`)
```
santisoler, line 3:
```suggestion
Typing (:mod:`SimPEG.typing`)
```

SimPEG/typing/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+=============================
+
+This module provides additional `PEP 484 <https://peps.python.org/pep-0484/>`_
+type aliases used in ``simpeg``'s codebase.
```
santisoler, line 7:
```suggestion
type aliases used in ``SimPEG``'s codebase.
```

SimPEG/typing/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        npt.NDArray[np.int_],
+        np.random.SeedSequence,
+        np.random.BitGenerator,
+        np.random.Generator,
+    ]
+
+RandomSeed.__doc__ = """
+A ``typing.Union`` for random seeds and Numpy's random number generators.
+
+These type of variables can be used throughout ``simpeg`` to control random
```
santisoler, line 47:
```suggestion
These type of variables can be used throughout ``SimPEG`` to control random
```

santisoler (commented):
Use capitalization of SimPEG. In case we merge #1430 before this one, we can make the necessary changes and use the lowercase version for the package name.

santisoler:
I kept using `SimPEG` in the API reference while including the documentation for the new `typing` module in order to follow the current convention in `main`. I opened #1437 to discuss about updating the API reference to use `simpeg` instead. Based on that decision, this PR might need to be updated.

santisoler:
The newly added tests are failing due to the bug that should be fixed in #1460. Will wait for that PR to be merged before merging this one.

simpeg/directives/directives.py
```diff
@@ -347,16 +349,18 @@ class BaseBetaEstimator(InversionDirective):
     ----------
     beta0_ratio : float
         Desired ratio between data misfit and model objective function at initial beta iteration.
-    seed : int, None
-        Seed used for random sampling.
+    seed : {None, RandomSeed}, optional
```
lheagy, line 352: As we chatted about, I think it would be great to call this `random_seed` so it is clear to users that this is controlling the randomness, but no problem if this is addressed in a separate pr

lheagy (approved):
Thanks for this @santisoler! It will be great to be more methodical about how random numbers are treated in SimPEG. The one comment is about naming `random_seed` vs. `seed`, but as we chatted about, this is fine to do in a separate pr if you prefer

SimPEG/directives/directives.py
lheagy: There is a lot of power in the random number generators! For now, my 2 cents would be to keep it simple. I suspect that many users won't be aware of the workings of random number generators, so if you set a number and then ask for what you set and see something like 
```
SeedSequence(
    entropy=42,
)
```
that might be a bit of an intimidating animal to encounter 🐻 😄

simpeg/directives/directives.py
jcapriot: That sounds like a good name to me! that or just `rng` for random number generator?

lheagy: My vote would be for spelling it out `random_seed` (`rng` could easily be confused with `range` in some contexts)

santisoler: I agree with renaming it to `random_seed`. Is the name I used when I added the parameter to the [`make_synthetic_data`](https://docs.simpeg.xyz/content/api/generated/SimPEG.potential_fields.gravity.Simulation3DIntegral.make_synthetic_data.html#SimPEG.potential_fields.gravity.Simulation3DIntegral.make_synthetic_data) methods of simulations.

I'll open an issue for the renaming, so I can start working on that after this gets merged.

PR #1398, opened 2024-03-28
Minor adjustments to Sphinx configuration

santisoler:
#### Summary

Align navbar elements to the left to avoid wrapping text in nav items. Remove plausible line from the html template and set it through sphinx pydata theme.

#### PR Checklist
      to a Pull Request


jcapriot:
So the primary reason the navbar items are centered is to keep a somewhat consistent formatting with the base [simpeg.xyz](https://simpeg.xyz) landing page. Do you have an image of what the change looks like, and what the potential issue it solves is? I do get text wrapping depending on the size of the window, but eventually it gets small enough to where they are collapsed into a menu bar. Can we better control when that happens?

santisoler:
The issue is trying to solve is the wrapping of the text elements in the navbar. If I build the docs we currently have in `main` I get something that look like this:

image

By setting the alignment to the left, the text elements are displayed in a single line and also the text box is kept in the same line as the other elements. Here's what I get after building the docs with these changes:

image

jcapriot:
It seems to be an odd behavior from pydata-sphinx-theme. I see that they updated the format of the search button (which I agree with), but now the left element of the navbar doesn't shrink the same way as it used to (thus causing the text wrapping).

Before it would collapse the left element until there was no more room, and then start wrapping.

jcapriot:
I’d personally pause the navbar adjustment portion of this update for now, and let’s look into a way to fix it without changing the layout of the site too much.

santisoler:
The change that the `"navbar_align": "left"` option does is not to specify the width of the columns for the logo and for the navlinks (using Bootstrap grid). Running a diff between the `index.html` files spits out these two sections:

For the logo in the navbar:
```diff
<   <div class="col-lg-3 navbar-header-items__start">
---
>   <div class=" navbar-header-items__start">
```

For the nav items:
```diff
<   <div class="col-lg-9 navbar-header-items">
---
>   <div class=" navbar-header-items">
```

The nav items are still being centered though (as it can be seen in the second screenshot), so I don't think we are changing the layout so much. I personally prefer having a slim navbar since it's _sticky_ and to give readers more space for the content of the docs.

santisoler:
Sounds good! We should make a list of things to improve in the docs and aim to tackle those for the v0.22 release. I've updated the branch so we can merge it today.

PR #1414, opened 2024-04-10
Extend docstrings for FDEM and TDEM fields and 3D simulations

dccowan:
#### Summary
Complete the docstrings for 3D field and simulation classes for FDEM and TDEM modules. This includes
* providing sufficient background theory
* updating the formatted according to our current style guide

We also added appropriate __init__ functions. Some child classes had properties that should have been fixed.

**Summary for Squash and Merge**

Extend docstrings for the 3D field and simulation classes for FDEM and TDEM modules. Provide mathematical background for the implementations. Update format of docstrings following numpydoc style. Add a few constructor methods for some classes to fix some missing properties in child classes.

#### PR Checklist
* New tests not required.
      to a Pull Request


jcapriot (changes requested):
Hey Devon, thanks for putting all these together. In general it looks great and adds many of the missing details.

I've left many comments through the frequency domain fields and the first few simulations, but there's many common suggestions that I will list here instead of suggesting changes on all of them in every file, so please apply these sorts of rules to everywhere here.
* You can shorten many of the links, for example write ```:class:`.frequency_domain.Simulation3DElectricField` ``` instead of ```:py:class:`SimPEG.electromagnetics.frequency_domain.simulation.Simulation3DElectricField` ```. A few points I want to showcase from this example change:
    * When you prefix an item with a `.` sphinx will look for an item with a matching suffix. Only use a unique suffix.
    * The `:py` part of `:py:class:` is assumed within the python documentation so you don't need it.
    * Make sure to point the link to where the item actually lives in the documentation (`.frequency_domain.Simulation3DElectricField` and not `.frequency_domain.simulation.Simulation3DElectricField`).
* Be vigilant about using `n_cells`, `n_edges`, `n_faces`, `n_nodes` instead of `nC`, `nN`, `nE`, `nF` (and fix them if they were wrong outside of the changes you've made (but only in the changed files).
* There are two separate test functions in the formulation, one lives on edges and the other on faces.
* When a function returns multiple things, list them individually in the `Returns` section (similar to how parameters are listed). Don't refer to the returned type as a `tuple`.
* Make sure optional parameters are marked as optional (i.e. `storeJ` and `forward_only`).

dccowan:
@jcapriot we shouldn't bring this in until we rectify our definition of the source term. Right now, the definition of the magnetic and electric source terms according to the math doesn't universally line up with how the code defines the quantities. E.g. look at any of the methods for getRHS and you will see the definition for the source term doesn't match the variables in the method exactly.

jcapriot:
I was thinking more on this as well. I don’t like the way we are currently using permitivity in the frequency domain classes. 

We are currently using it mostly as a means to handle complex resistivity. It is included in the constitutive relationship between Electric field and current density as you have noted in the docs here (which is what made me think of it). As is well known, the “displacement currents” from maxwell’s 4th equation are not actually currents, so we ideally should not include it in the current density term in the field, or other places… etc.

Instead, we should just actually support complex conductivity.

dccowan:
> I was thinking more on this as well. I don’t like the way we are currently using permitivity in the frequency domain classes.
> 
> We are currently using it mostly as a means to handle complex resistivity. It is included in the constitutive relationship between Electric field and current density as you have noted in the docs here (which is what made me think of it). As is well known, the “displacement currents” from maxwell’s 4th equation are not actually currents, so we ideally should not include it in the current density term in the field, or other places… etc.
> 
> Instead, we should just actually support complex conductivity.

At this point, I'm documenting how the code actually works. Would it be sufficient to make a note in the documentation that future work will modify the code to accept complex conductivity? Or perhaps leave permittivity as an input argument, but remove it from any derivations?

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
 
 
 class Fields:
-    """Fancy Field Storage
+    r"""Base class for storing fields.
+
+    Fields classes are used to store the discrete field solution for a
+    corresponding simulation object; see :py:class:`SimPEG.simulation.BaseSimulation`.
+    Generally only one field solution (e.g. 'eSolution', 'phiSolution', 'bSolution') is stored.
```
santisoler, line 12:
```suggestion
    Generally only one field solution (e.g. ``'eSolution'``, ``'phiSolution'``, ``'bSolution'``) is stored.
```

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
...
+    However, it may be possible to extract multiple field types (e.g. 'e', 'b', 'j', 'h')
```
santisoler, line 13:
```suggestion
    However, it may be possible to extract multiple field types (e.g. ``'e'``, ``'b'``, ``'j'``, ``'h'``)
```

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
...
+    field types that can be extracted depend on the formulation used by the associated simulation.
+    See the example below to learn how fields are extracted from fields objects.
+
+    Parameters
+    ----------
+    simulation : SimPEG.simulation.BaseSimulation
+        The simulation object used to compute the discrete field solution.
+    knownFields : dict of {key: str}, optional
+        Dictionary defining the field solutions that are stored and where
+        on the mesh they are discretized. E.g. `{'eSolution': 'E', 'bSolution': 'F'}
```
santisoler, line 24:
```suggestion
        on the mesh they are discretized. E.g. ``{'eSolution': 'E', 'bSolution': 'F'}``
```

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
...
+        Set aliases to extract different field types from the field solutions that are
+        stored by the fields object. The ``key`` defines the name you would like to use
+        when extracting a given field type from the fields object. In order, the list
+        contains:
+
+        * the key for the known field solution that is used to compute the field type
+        * where the output field type lives {'CC', 'N', 'E', 'F'}
+        * the name of the method used to compute the output field.
+
+        E.g. {'b': ['eSolution', 'F', '_b']} is an alias that
```
santisoler, line 37:
```suggestion
        E.g. ``{'b': ['eSolution', 'F', '_b']}`` is an alias that
```

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
...
+        would allow you to extract a field type ('b') that lives on mesh faces ('F')
+        from the E-field solution ('eSolution') by calling a method ('_b').
```
santisoler, line 39:
```suggestion
        would allow you to extract a field type (``'b'``) that lives on mesh faces (``'F'``)
        from the E-field solution (``'eSolution'``) by calling a method (``'_b'``).
```

simpeg/fields.py
```diff
@@ -5,24 +5,71 @@
...
+    dtype : dtype or dict of {str : dtype}, optional
+        Set the Python data type for each numerical field solution that is stored in
+        the fields object. E.g. ``float``, ``complex``,
+        {'eSolution': ``complex``, 'bSolution': ``complex``}.
```
santisoler, line 43:
```suggestion
        ``{'eSolution': complex, 'bSolution': complex}``.
```

simpeg/fields.py
```diff
@@ -61,39 +109,41 @@ def simulation(self, value):
...
-        {'e': 'E', 'phi': 'CC'}
+        Dictionary defining the field solutions that are stored and where
+        on the mesh they are discretized. The ``key`` defines the name
+        of the field solution that is stored, and a ``str`` defines where
+        on the mesh the stored field solution is discretized. The
+        ``str`` must be one of {'CC', 'N', 'E', 'F'}.
 
-        Would represent that the `e` field and `phi` fields are known, and they are
-        located on the mesh edges and cell centers, respectively.
+        E.g. `{'eSolution': 'E', 'bSolution': 'F'}
```
santisoler, line 120:
```suggestion
        E.g. ``{'eSolution': 'E', 'bSolution': 'F'}``
```

simpeg/fields.py
```diff
@@ -106,28 +156,53 @@ def aliasFields(self):
 
     @property
     def dtype(self):
-        """The data type of the storage matrix
+        """Python data type(s) used to store the fields.
+
+        the Python data type for each numerical field solution that is stored in
+        the fields object. E.g. ``float``, ``complex``, {'eSolution': complex, 'bSolution': complex}.
```
santisoler, line 162:
```suggestion
        the fields object. E.g. ``float``, ``complex``, ``{'eSolution': complex, 'bSolution': complex}``.
```

simpeg/fields.py
```diff
@@ -273,21 +348,73 @@ def __contains__(self, other):
...
 
 class TimeFields(Fields):
-    """Fancy Field Storage for time domain problems
-    .. code:: python
+    r"""Base class for storing TDEM fields.
+
+    ``TimeFields`` is a base class for storing discrete field solutions for simulations
+    that use discrete time-stepping; see :py:class:`SimPEG.simulation.BaseTimeSimulation`.
+    Generally only one field solution (e.g. 'eSolution', 'phiSolution', 'bSolution') is stored.
+    However, it may be possible to extract multiple field types (e.g. 'e', 'b', 'j', 'h')
```
santisoler, line 356:
```suggestion
    Generally only one field solution (e.g. ``'eSolution'``, ``'phiSolution'``, ``'bSolution'``) is stored.
    However, it may be possible to extract multiple field types (e.g. ``'e'``, ``'b'``, ``'j'``, ``'h'``)
```

simpeg/fields.py
```diff
@@ -273,21 +348,73 @@ def __contains__(self, other):
...
+    field types that can be extracted depend on the formulation used by the associated simulation.
+    See the example below to learn how fields are extracted from fields objects.
+
+    Parameters
+    ----------
+    simulation : SimPEG.simulation.BaseTimeSimulation
+        The simulation object used to compute the discrete field solution.
+    knownFields : dict of {key: str}, optional
+        Dictionary defining the field solutions that are stored and where
+        on the mesh they are discretized. E.g. `{'eSolution': 'E', 'bSolution': 'F'}
```
santisoler, line 367:
```suggestion
        on the mesh they are discretized. E.g. ``{'eSolution': 'E', 'bSolution': 'F'}``
```

simpeg/fields.py
```diff
@@ -61,39 +109,41 @@ def simulation(self, value):
...
 
-        >>> fields.aliasFields
-        {'b': ['e', 'F', '_e']}
+        * the key for the known field solution that is used to compute the field type
+        * where the output field type lives {'CC', 'N', 'E', 'F'}
+        * the name of the method used to compute the output field.
 
-        Would represent that the `e` field and `phi` fields are known, and they are
-        located on the mesh edges and cell centers, respectively.
+        E.g. {'b': ['eSolution', 'F', '_b']} is an alias that
```
santisoler, line 144:
```suggestion
        E.g. ``{'b': ['eSolution', 'F', '_b']}`` is an alias that
```

simpeg/fields.py
```diff
@@ -273,21 +348,73 @@ def __contains__(self, other):
...
+        Set aliases to extract different field types from the field solutions that are
+        stored by the fields object. The ``key`` defines the name you would like to use
+        when extracting a given field type from the fields object. In order, the list
+        contains:
+
+        * the key for the known field solution that is used to compute the field type
+        * where the output field type lives {'CC', 'N', 'E', 'F'}
+        * the name of the method used to compute the output field.
+
+        E.g. {'b': ['eSolution', 'F', '_b']} is an alias that
```
santisoler, line 380:
```suggestion
        E.g. ``{'b': ['eSolution', 'F', '_b']}`` is an alias that
```

simpeg/fields.py
```diff
@@ -273,21 +348,73 @@ def __contains__(self, other):
...
+        would allow you to extract a field type ('b') that lives on mesh faces ('F')
+        from the E-field solution ('eSolution') by calling a method ('_b').
+    dtype : dtype or dict of {str : dtype}, optional
+        Set the Python data type for each numerical field solution that is stored in
+        the fields object. E.g. ``float``, ``complex``, {'eSolution': complex, 'bSolution': complex}.
```
santisoler, line 385:
```suggestion
        the fields object. E.g. ``float``, ``complex``, ``{'eSolution': complex, 'bSolution': complex}``.
```

santisoler (approved):
Great work @dccowan! I pushed a few minor changes to the rst syntax and highlighted a few more in the comments below. I think we should merge this!

simpeg/electromagnetics/time_domain/simulation.py
```diff
@@ -936,68 +1570,275 @@ def getRHS(self, tInd):
...
+        else:
+            return self.MeSigmaDeriv(-u, Grad * v, adjoint)
+
+    # def clean(self):
+    #     """
+    #     Clean factors
+    #     """
+    #     if self.Adcinv is not None:
+    #         self.Adcinv.clean()
+
```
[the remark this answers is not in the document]
dccowan: I think this is being handled somewhere else. Will remove.

simpeg/electromagnetics/time_domain/fields.py
```diff
@@ -273,19 +346,61 @@ def _dhdtDeriv_m(self, tInd, src, v, adjoint=False):
...
+    """
+
+    def __init__(self, simulation):
+        super().__init__(simulation=simulation)
+        self._knownFields = {"eSolution": "E"}
+        self._aliasFields = {
+            "e": ["eSolution", "E", "_e"],
+            "j": ["eSolution", "E", "_j"],
+            "b": ["eSolution", "F", "_b"],
+            # 'h': ['eSolution', 'F', '_h'],
```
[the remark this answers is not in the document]
dccowan: Maybe we do the same as we have for _b. That is, it throws a not implemented error. h and b can't be accessed from the e-field solution.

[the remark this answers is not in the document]
santisoler: Sounds good. Since this PR is docstring-only, I would keep this change to a different PR. I also noticed there are some commented lines in the `_b` method, which should also be removed.

santisoler:
Thanks for applying the suggested changes. Would you open a few Issues to remind us to keep forward with the discussion that @jcapriot opened regarding handling complex conductivity, and to clean up some of the commented lines I mentioned? We don't need to start working on them asap, but it would be nice to have them opened so we can retake them in the future.

I'm going to merge this! 🚀

PR #1421, opened 2024-04-11
Simplify interface of UniformBackgroundField

### LINKED ISSUE #1418: Remove default arguments for UniformBackgroundField?
In the meeting of 2024-04-10 we discussed about modifying the interface of the constructor of the `UniformBackgroundField` class.

One idea that floated around was to make the `amplitude`, `inclination` and `declination` arguments as required arguments, removing the default values they already take.

Another idea was to actually remove the `kwargs` from the constructor, since they are not being used.

Related to https://github.com/simpeg/user-tutorials/issues/12

santisoler:
#### Summary

Remove the `**kwargs` from the constructor of `UniformBackgroundField`. Make `receiver_list`, `amplitude`, `inclination` and `declination` as required arguments without any default value. Include type hints in the signature of the constructor. Add check for type of receivers to the constructor of `UniformBackgroundField`. Add test that checks the value of `UniformBackgroundField.b0`, passing `receiver_list` as an actual list of point receivers and as `None`. Add test that checks if the error is raised after passing invalid receiver type. Minor improvements to the docstring of the class.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fixes #1418


jcapriot (changes requested):
I would recommend either overloading the parent's `receiver_list.setter` or just completely overloading the property

If you just want to overload the setter:
```python
    @BaseSrc.receiver_list.setter
    def receiver_list(self, value):
        self._receiver_list = validate_list_of_types(
            "source_list", new_list, Point, ensure_unique=True
        )
```

or if you want to overload the getter and setter (probably to make the `receiver_list` property have a slightly more relevant doctoring).

```python
    @property
    def receiver_list(self):
        """List of receivers associated with the survey

        Returns
        -------
        list of .magnetic.Point
            List of magnetic receivers associated with the survey
        """
        return self._receiver_list

    @receiver_list.setter
    def receiver_list(self, value):
        self._receiver_list = validate_list_of_types(
            "source_list", new_list, Point, ensure_unique=True
        )
```

santisoler:
Sounds good. I agree that would be better, indeed. One thing that is not very clear in the `validate_list_of_types` is that it doesn't fail if the passed object is not a list, but just a single object of the valid `class_type`. Minor thing, but probably worth documenting.

tests/utils/test_io_utils.py
```diff
@@ -256,7 +256,9 @@ def setUp(self):
         self.std = std
 
         rx2 = magnetics.receivers.Point(xyz, components="tmi")
-        src_bad = magnetics.sources.UniformBackgroundField([rx, rx2])
+        src_bad = magnetics.sources.UniformBackgroundField(
+            receiver_list=[rx1, rx2], amplitude=50_000, inclination=90, declination=0
```
santisoler, line 260:
```suggestion
            receiver_list=[rx, rx2], amplitude=50_000, inclination=90, declination=0
```

jcapriot:
So the only awkward thing left here is that `receivers_list` was an optional parameter that defaulted to `None` which is a valid thing, (mostly in the context of the PDE solution where you might be interested in the fields themselves but have no receivers). Now it would require explicitly setting `receivers_list=None` to obtain the same behavior.

santisoler:
True. I think it's a good compromise. An alternative would be to define the receiver list as an optional argument. I don't have strong feelings toward any of these options.

jcapriot:
> True. I think it's a good compromise. An alternative would be to define the receiver list as an optional argument. I don't have strong feelings toward any of these options.

The thing is, you can't make the `receiver_list` optional without changing the order of the arguments, which is not an option. So the only real options are to either make everything required (which is technically a breaking change), or to just leave them all as keyword arguments. I think the part just removing the `**kwargs` is enough to make it behave better.

santisoler:
I wouldn't like to change order of arguments, that's why I decided to make `receiver_list` as required from the start.

I also agree with the idea of not setting default values for the background field. Setting them as required arguments force users to consider which values should the pass to define the background field. I think that's a good thing.

I don't quite like setting the values as None and then raising an error after them. I think it'll create some confusion, will require us to add checks and tests for those checks, and is not very clean. Moreover, changing a default value is also a breaking change, so we are not avoiding that either.

That's why I think having `receiver_list` as required is a good compromise: we'll force users to pass the values for the background field, with the side effect of having to pass the receiver list as well.

lheagy (approved):
Looks good to me!

PR #1410, opened 2024-04-08
Add new Issue template for making a release

santisoler:
#### Summary

Add a new issue template for making a new release, containing a checklist of
the steps needed to make a new release.

#### PR Checklist
* [ ] If this is a work in progress PR, set as a Draft PR
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html).
* [ ] Added [tests](https://docs.simpeg.xyz/content/getting_started/practices.html#testing) to verify changes to the code.
* [ ] Added necessary documentation to any new functions/classes following the
      expect [style](https://docs.simpeg.xyz/content/getting_started/practices.html#documentation).
* [ ] Marked as ready for review (if this is was a draft PR), and converted
      to a Pull Request
* [ ] Tagged ``@simpeg/simpeg-developers`` when ready for review.

PR #1428, opened 2024-04-16
Add version switcher to Sphinx docs

santisoler:
#### Summary

Add a version switcher to the navbar of the docs, using the built in one in PyData Sphinx theme. Point to the experimental `versions.json` file located in `doctest.simpeg.xyz` that should serve the files in [simpeg/simpeg-doctest](https://github.com/simpeg/simpeg-doctest). Add experimental staged to Azure Pipelines that pushes docs to the `simpeg-doctest` repository. One is triggered nightly and deploys the latest docs in `main` to the `dev` branch in `simpeg-doctest` and updates the `dev` submodule in `gh-pages`. The other one is triggered after a release and pushes the new version of the docs to `gh-pages` branch in `simpeg-doctest` onto a new folder, while also updating the `latest` link.

#### PR Checklist
      to a Pull Request

#### Additional information

TODO:

- [x] Add task in the release checklist (#1410) to update the `versions.json` file
- [x] Build and deploy docs after push to `main`. Docs should be pushed to replace the content of the `dev` folder where the docs are being hosted.
- [x] Add a checkout step that also syncs tags: see #1444 for reference


azure-pipelines.yml
```diff
@@ -8,12 +8,22 @@ trigger:
     include:
     - '*'
 
+schedules:
+- cron: "0 8 * *" # trigger cron job every day at 08:00 AM GMT
```
santisoler, line 12:
```suggestion
- cron: "0 8 * * *" # trigger cron job every day at 08:00 AM GMT
```

santisoler: Fix cron syntax

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
             git push
           displayName: Push documentation to simpeg-docs
           env:
-            GH_TOKEN: $(gh.token)
\ No newline at end of file
+            GH_TOKEN: $(gh.token)
+
+- stage: Deploy dev docs to simpeg-doctest (experimental)
```
santisoler, line 156:
```suggestion
- stage: Deploy_dev_docs_experimental
```

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
...
+            # Make the push quiet just in case there is anything that could
+            # leak sensitive information.
+            echo -e "\nPushing changes to simpeg/simpeg-doctest."
+            git push -fq origin gh-pages 2>&1 >/dev/null
+            echo -e "\nFinished uploading generated files."
+          displayName: Push documentation to simpeg-doctest
+          env:
+            GH_TOKEN: $(gh.token)
+
+- stage: Deploy release docs to simpeg-doctest (experimental)
```
santisoler, line 242:
```suggestion
- stage: Deploy_release_docs_experimental
```

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
...
-            GH_TOKEN: $(gh.token)
\ No newline at end of file
+            GH_TOKEN: $(gh.token)
+
+- stage: Deploy dev docs to simpeg-doctest (experimental)
+  dependsOn: Testing
+  condition: eq(variables['Build.Reason'], 'Schedule')  # run only scheduled triggers
+  jobs:
+    - job:
+      displayName: Deploy docs to simpeg-doctest (experimental)
```
santisoler, line 161:
```suggestion
      displayName: Deploy dev docs to simpeg-doctest (experimental)
```

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
...
+          displayName: Push documentation to simpeg-doctest
+          env:
+            GH_TOKEN: $(gh.token)
+
+- stage: Deploy release docs to simpeg-doctest (experimental)
+  dependsOn: Testing
+  condition: startsWith(variables['build.sourceBranch'], 'refs/tags/')
+  jobs:
+    - job:
+      displayName: Deploy docs to simpeg-doctest (experimental)
```
santisoler, line 247:
```suggestion
      displayName: Deploy release docs to simpeg-doctest (experimental)
```

jcapriot:
Since this is changing the format of the documentation page (and I imagine you might start seeing text wrapping soon), I think it's also appropriate to consider a few things:

* Do we really need the youtube link on the navbar on every page, or is a link on the main page sufficient?
* Is a link back to `simpeg.xyz` in the center navbar useful? (Probably not)
* Is the `contact` link to Mattermost in the center navbar useful? (Definitely leave the icon link)
* Should we move the search bar to the left menu (i.e. https://numpy.org/doc/stable/index.html)?
* Consider renaming the items in the center with one word aliases (consider these as suggestions)?
    * `API Reference` -> `API`
    * `Release Notes` -> `changelog`
    * `User Tutorials` -> `Learn`

ghwilliams:
Great.
It is a headache when the documentation is not synchronized with the source code. Many times, I got lost trying to reproduce an example from an open-source code just because the documentation was created for an old, outdated version. If the documentation is tested against the corresponding version of the software and kept recorded together with each release, it will make learning SimPEG a much more enjoyable experience.

santisoler:
@jcapriot would you like to take a look at this? Specially to the scripts for deploying the docs.

I added those to experiment and see if the setup works properly on `simpeg-doctest`. If we are happy with it, at some point we could update the pipelines and switch them to `simpeg-docs` and make it the official deployment process.

Also, I was considering hosting the docs in a `gh-pages` orphan branch in this repo, instead of having a separate repo for them. That would also ease the process of implementing this new workflow: we can just have these new deployment staged pushing to `gh-pages` and then flip the switch in GitHub Pages configuration. What do you think?

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
...
+          displayName: Build package
+
+        - script: |
+            source "${HOME}/conda/etc/profile.d/conda.sh"
+            conda activate simpeg-test
+            export KMP_WARNINGS=0
+            make -C docs html
+          displayName: Building documentation
+
+        # Upload dev build of the docs to gh-pages branch in simpeg/simpeg-docs
```
santisoler, line 206:
```suggestion
        # Upload dev build of the docs to gh-pages branch in simpeg/simpeg-doctest
```

azure-pipelines.yml
```diff
@@ -141,4 +151,177 @@ stages:
...
```
santisoler, line 292:
```suggestion
        # Upload dev build of the docs to gh-pages branch in simpeg/simpeg-doctest
```

jcapriot (commented):
Hey @santisoler in general the azure scripts look good, However, do you know if there's a way to serve the doc versions from different branches? (I don't think there's a way in github pages, but there could be something else.) 

My thought is to avoid doing the git-clone on the entire documentation website (as it will continually get larger and larger with each release). Instead just commit it's own documentation to a specific branch.

jcapriot:
> Hey @santisoler in general the azure scripts look good, However, do you know if there's a way to serve the doc versions from different branches? (I don't think there's a way in github pages, but there could be something else.)
> 
> My thought is to avoid doing the git-clone on the entire documentation website (as it will continually get larger and larger with each release). Instead just commit it's own documentation to a specific branch.

Maybe through using git's submodules with github pages might work...

santisoler:
I think one can only serve one GH page per repo. Basically, we can only configure GH Pages on a single branch.

While reading your first comment I thought the same: git submodules. But I'm not sure if GH Pages follow the submodules when "generating" the Pages. Every time we push a change to a branch that is serving a GH Page, an Action triggers that "generates the update". I think it basically checks if it needs to run Jekyll or not, and then make the html files available through their webserver. I'm not sure if in that process they also follow all submodules and include those html files.

jcapriot:
> I think one can only serve one GH page per repo. Basically, we can only configure GH Pages on a single branch.
> 
> While reading your first comment I thought the same: git submodules. But I'm not sure if GH Pages follow the submodules when "generating" the Pages. Every time we push a change to a branch that is serving a GH Page, an Action triggers that "generates the update". I think it basically checks if it needs to run Jekyll or not, and then make the html files available through their webserver. I'm not sure if in that process they also follow all submodules and include those html files.

It looks like it might? https://docs.github.com/en/pages/getting-started-with-github-pages/using-submodules-with-github-pages

santisoler:
Awesome! Thanks for pointing to the docs.

I'll give it a try in [simpeg-doctest](https://github.com/simpeg/simpeg-doctest) and if we like it I'll update the pipeline here.

jcapriot:
> Ok. I've tried yesterday setting up a `main` branch in `simpeg-doctest` that includes submodules `dev` and `v0.21.1` that points to branches with the same name in that repo. I've configured the Pages to serve from `main`. You can see how it renders by going to https://doctest.simpeg.xyz.
> 
> The CSS of the pages are not being read. What's curious is that the `custom.css` file cannot be found in https://doctest.simpeg.xyz/v0.21.1/_static/css/custom.css. I'm curious why this is happening...

Looks like there is no `.nojekyll` file there. That might be causing an issue.

santisoler:
That was it. Thanks for spotting that!

santisoler:
I've updated the scripts for deploying to simpeg-doctest. Would love to have a second pair of eyes over there.

One question that I have now regarding submodules is if gh pages get updated when there are changes in one of the submodules. For example, if we push changes to the `dev` branch, do they trigger an update in the Pages so the served docs are updated?

jcapriot:
> I've updated the scripts for deploying to simpeg-doctest. Would love to have a second pair of eyes over there.
> 
> One question that I have now regarding submodules is if gh pages get updated when there are changes in one of the submodules. For example, if we push changes to the `dev` branch, do they trigger an update in the Pages so the served docs are updated?

As you can kinda see from the main branch, the submodules are tied to specific commit hashes. So a change to those branches will not immediately propagate to the `main` documentation.

The process for the dev docs would be:

- Pull the `dev` branch of the docs
      - commit & push the `dev` changes
- Switch to the `main` branch
      - update the git submodule for `dev` to point to the recent commit.
      - commit & push this back to `main`.
 
For release docs we would have two options, One being just relying on tags to dev:
- Pull the `dev` branch
      - commit the current changes
      - tag this commit
      - push the changes and the tag
- Switch to `main`
    - update `dev` like before
    - create a new submodule specifically pointing to that `tagged` commit of `dev` (but named according to the version).
    - push the change back to main.
 
or if we want to actually have them as separate branches of the repo, replace the "tag this commit" operation with "create a new branch from this commit", and then follow the same logic.

Fundamentally the submodules do not point to branches, they point to commits.

jcapriot:
I think I'm in favor of the `tagged` approach as it keeps the entire repo itself cleaner.

santisoler:
That makes sense. I forgot we need to update the submodules as well.

Regarding your proposal, I think it would be better to have versions in their own branches. Mainly because in case we need to apply changes to old docs (change analytics, add a banner, etc), it would be a lot easier than trying to insert commits in the history of the single `dev` branch.

santisoler:
I'm rethinking the deployment process. I think having the dev docs in a separate branch is great to avoid a commit history that will grow to large. But the documentation for releases ideally won't change in time. So having all the tagged docs in separate branches makes the process more complex without adding any benefit.

So, what if we keep the dev docs in the `dev` branch, update the submodule in the `gh-pages` branch, but commit the release docs to folders in the same `gh-pages` branch?

azure-pipelines.yml
```diff
@@ -141,4 +151,192 @@ stages:
...
+            # Remove all files
+            shopt -s dotglob  # configure bash to include dotfiles in * globs
+            export GLOBIGNORE=".git"  # ignore .git directory in glob
+            git rm -rf * # remove all files
+            # Copy the built docs to a new dev folder
+            cp -r $BUILD_SOURCESDIRECTORY/docs/_build/html/* -t .
+            # Commit the new docs. Amend to avoid having a very large history.
+            git add .
+            message="Azure CI deploy dev from ${commit}"
+            echo -e "\nAmmending last commit:"
```
santisoler, line 225:
```suggestion
            echo -e "\nAmending last commit:"
```

.github/ISSUE_TEMPLATE/release-checklist.md
```diff
@@ -111,6 +110,24 @@ Pull Requests
...
 
+### Add new version to version switcher
+
+Edit the `docs/_static/versions.json` file and:
+
+- [ ] Add an entry for the new version.
+- [ ] Move the line with `"name":` to the new entry (so the new version is set
+  as the _latest_ one).
+- [ ] Update the version number in the `"name":` line.
+- [ ] Run `cat docs/_static/versions.json| python -m json.tool > /dev/null` to
```
santisoler, line 121:
```suggestion
- [ ] Run `cat docs/_static/versions.json | python -m json.tool > /dev/null` to
```

santisoler:
I'm merging this one. I'm quite happy with the setup. It's still experimental so I'll have to open another PR to update the repo where docs should be pushed and the location of the `versions.json` file before making this the default way to deploy our docs. In order to check that the new pipeline works properly, I need to try it out. Hope everything works smoothly (famous last words) 🤞🏼

PR #1430, opened 2024-04-19
Move to a PEP8 compliant package name.

jcapriot:
#### Summary
Change package name from `SimPEG` to `simpeg`

Makes the package name PEP8 compliant.

#### Additional information:
This is mostly just a suggestion PR as it's something we've always considered, but is likely to be a big change effecting everyone. To minimize this, it also then installs a `SimPEG.py` stub file as well. In my testing the stub file correctly redirects all `SimPEG` imports to `simpeg`, and will print the message only on the first import of `SimPEG`. So I hope this will provide the change in the most painless way possible to end users. I have not investigated any other sorts of impacts (to IDEs, documentation, etc.)

** Note: ** This won't affect how people obtain the package as that was already handled by pypi and conda-forge (e.g. `pip install simpeg` was already lowercased, as well as `conda install -c conda-forge simpeg`)


prisae:
I think it would be great to change to a pep8 name! One question: It looks like you replaced all and every «SimPEG» by «simpeg», also in docs etc. But the "name" of the Project would still be SimPEG, right? And just the package would be simpeg. Just as usually it is written NumPy, SciPy, etc... In that regards maybe less change has to be done than is currently in the PR.

rowanc1:
I would concur with @prisae that we follow the conventions set by NumPy (`numpy`), and don't change the name/capitalization of the project name, just how you import it. That is, the name referred to on the website, docs, papers is `SimPEG` and is imported using the lowercase variant: `import simpeg`, `pip install simpeg`, `simpeg.xyz` etc.

jcapriot:
> I think it would be great to change to a pep8 name! One question: It looks like you replaced all and every «SimPEG» by «simpeg», also in docs etc. But the "name" of the Project would still be SimPEG, right? And just the package would be simpeg. Just as usually it is written NumPy, SciPy, etc... In that regards maybe less change has to be done than is currently in the PR.

Sounds right, I’ll undo those changes.

jcapriot:
Ok, it looks like I've separated everywhere it's referred to by it's project name vs it's package name.

docs/README.md
```diff
@@ -1,8 +1,8 @@
-# simpeg-docs
+# SimPEG-docs
```
santisoler, line 1: This title should be lowercase since it refers to the `simpeg-docs` repo.
```suggestion
# simpeg-docs
```

setup.py
```diff
@@ -27,8 +27,10 @@
 with open("README.rst") as f:
     LONG_DESCRIPTION = "".join(f.readlines())
 
+print(find_packages(exclude=["tests*", "examples*", "tutorials*"]))
+
```
santisoler, line 31: Do we want to add this print?

santisoler (changes requested):
I really like this! Thanks @jcapriot for the work behind this.

I agree with the convention that @prisae proposed of having `simpeg` as the package and SimPEG as the project and the community.

I just left two minor comments. Looking forward to see this merged!

setup.py
jcapriot: Not at all, just a little leftover debugging from the 400+ changed files

lheagy:
Thanks for kicking this off @jcapriot! I am supportive of this too, and I am glad it is possible to deprecate cleanly.  Also agreed on keeping the project name SimPEG, but the code `simpeg`. For depracating, I think we will need to leave the possibility for both imports for quite a while. This will hit everyone who uses SimPEG, so I would suggest we leave it for quite a few versions (0.30.0??) until `import SimPEG` is no longer supported

jcapriot:
> Thanks for kicking this off @jcapriot! I am supportive of this too, and I am glad it is possible to deprecate cleanly. Also agreed on keeping the project name SimPEG, but the code `simpeg`. For depracating, I think we will need to leave the possibility for both imports for quite a while. This will hit everyone who uses SimPEG, so I would suggest we leave it for quite a few versions (0.30.0??) until `import SimPEG` is no longer supported

My thought would be to abandon it at 1.0.0.

Only thing I want to do is to just double check the new script gets packaged up into the wheel correctly before merging it.

santisoler:
uuuuhuuu! Awesome to see this merged! 🚀

jcapriot:
Yeah, I suspect there might be some issues that pop up… I think mostly related to people swapping between branches before/after this merge.

PR #1444, opened 2024-05-01
Fix links to source code in documentation pages

santisoler:
#### Summary

Update the `linkcode_resolve` function in Sphinx configuration file so the links to the source code of each function, class or method points to the actual code corresponding to that version, and not always to the code in `main`. Update some Azure configurations: checkout simpeg repo without shallow depth and fetching tags. Clean up the working directory before installing SimPEG so the version doesn't have an extra hash.

#### PR Checklist
      to a Pull Request

This PR fixes a current issue in SimPEG docs: links to source code lead to the current code in the `main` branch, which is not the code that corresponds to the documentation. This inconsistency could lead to confusion among users: it will lead them to think they are running the code they are reading while they are not.

#### Additional information

TODO:
- [x] Remove hardcoded tag before merging
- [x] Revert [58a6d94](https://github.com/simpeg/simpeg/pull/1444/commits/58a6d941a1d44f74408fa6b46ddb367e4fd835ee)


santisoler:
I'm going to hard code the `github_version` to the latest release and check the built docs (Azure artifact) to see if the change works as expected. If it does I'll revert that commit so we can merge this PR.

jcapriot:
OH... I think i know what you're running into. Look at this line from the build: https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4022&view=logs&j=a3da1655-dbd3-5466-5850-e4f85983e337&t=b6fcda75-f2ce-5ec9-7c70-5cf16be9995c&l=45

The sources that CI pulls don't include tags, so it won't get an accurate version number when it's installed.
The tagged commits due get this information.

santisoler:
Nice catch! Thanks for noticing it! Fetching tags is also super relevant for #1428.

santisoler:
I'm trying to dig more into the issue with tags. Checking out the repo with tags works (https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4027&view=logs&jobId=50e2b49d-b8ca-5aaf-b730-aab43fbdd612&j=e13373ee-b201-51bd-a7eb-56842f187dd4&t=d8226405-3c7c-58b3-ab29-76fc135fa198), but the simpeg version is not correctly set (https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4027&view=logs&jobId=50e2b49d-b8ca-5aaf-b730-aab43fbdd612&j=e13373ee-b201-51bd-a7eb-56842f187dd4&t=9de72094-40d6-5eb3-386e-721ab4adb7b6).

BTW, apparently I'm still used to type SimPEG haha. Need to retrain muscle memory.

santisoler:
[It worked](https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4030&view=logs&j=e13373ee-b201-51bd-a7eb-56842f187dd4&t=9de72094-40d6-5eb3-386e-721ab4adb7b6) 🎉

jcapriot:
Great. Another thing is that in the testing scripts it does directly modify a tracked file (It appends a python version to the `environment_test.yml` file for if we want to test on different versions of python.) So there will always be the extra git hash on the testing version. We could probably adjust this step to copy the `environment_test.yml` into a temporary file, then modify that temporary file with the python version instead. That way it won't modify the tracked files.

.azure-pipelines/matrix.yml
```diff
@@ -31,10 +44,14 @@ jobs:
           - script: |
               source "${HOME}/conda/etc/profile.d/conda.sh"
               source "${HOME}/conda/etc/profile.d/mamba.sh"
-              echo "  - python="${{ py_vers }} >> environment_test.yml
-              mamba env create -f environment_test.yml
+              cp environment_test.yml environment_test_with_pyversion.yml
+              echo "  - python="$(python.version) >> environment_test_with_pyversion.yml
+              mamba env create -f environment_test_with_pyversion.yml
+              rm environment_test_with_pyversion.yml
```
jcapriot, line 50: I just fixed it myself

santisoler:
Ok, I totally misundertood the documentation of pytest theme (https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/source-buttons.html). The changes to the `html_context` are intended for the Edit Source Code buttons, not for the links to the source code of the classes, functions and methods in the API.

Those links are configured through the `linkcode_resolve` function, which I updated to use the right tag.

santisoler:
Now the source links are working properly!

bitmap

That link won't work because the doc pages with `simpeg` as the package don't exist yet. But in the next release they will.

I'm removing the hardcoded tag, and after that I think this should be ready!

santisoler:
@jcapriot, I've been seeing that the `ubuntu-latest_3.8_tests/base tests/flow ...` workflow gets stuck in this PR. I've run it a few times already (re-running failed jobs, updating the branch) and it always gets stuck by the end of running the tests. I've only seen it in this PR.

Might it be related to the the fact that I disabled the shallow depth while checking out the repo?

jcapriot:
It's possible it's some race condition happening with the multiprocessing tests. There are some broken pipe messages in those tests when they pass it looks like.

santisoler:
Right! Nice catch. I'll wait for #1464 to fix it then.

jcapriot:
I hope that fixes it, (it at least fixes the broken pipe messages).

jcapriot:
Well 💩, looks like it’s still racing away.

santisoler:
I've just disabled pytest's capture of stdout and stderr so we can get more idea of what's going on... let's wait to see if there is some information in those messages.

jcapriot:
Not sure why but on this branch, the mappings for the dask meta aren't being sent to the correct workers:
https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4144&view=logs&j=894b861d-556c-5901-7578-c933b75ce341&t=19b313e8-909c-53b5-c60c-1097c881cd09&l=4252


Whereas on main they are: https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4141&view=logs&j=894b861d-556c-5901-7578-c933b75ce341&t=ed1e0f4a-533f-5064-832c-7c51df369f6b&l=548

jcapriot:
> Not sure why but on this branch, the mappings for the dask meta aren't being sent to the correct workers: https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4144&view=logs&j=894b861d-556c-5901-7578-c933b75ce341&t=19b313e8-909c-53b5-c60c-1097c881cd09&l=4252
> 
> Whereas on main they are: https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4141&view=logs&j=894b861d-556c-5901-7578-c933b75ce341&t=ed1e0f4a-533f-5064-832c-7c51df369f6b&l=548

I can reproduce this locally on my machine, give me a minute to do some debugging.

jcapriot:
Took me a bit, but #1469 should fix the race condition.

.azure-pipelines/matrix.yml
```diff
@@ -31,10 +40,14 @@ jobs:
           - script: |
               source "${HOME}/conda/etc/profile.d/conda.sh"
               source "${HOME}/conda/etc/profile.d/mamba.sh"
-              echo "  - python="${{ py_vers }} >> environment_test.yml
-              mamba env create -f environment_test.yml
+              cp environment_test.yml environment_test_with_pyversion.yml
+              echo "  - python="$(python.version) >> environment_test_with_pyversion.yml
```
santisoler, line 44: @jcapriot, I'm going back to use the `py_vers` variable for creating the environment so we can keep running these tests on Python 3.8. I'm aware that we should fix that race condition that is happening in Python 3.12, but I think we can do it in #1469. And actually, I prefer using the `py_vers` variable than relying on the default Python version that Azure ships.
```suggestion
              echo "  - python="{{ py_vers }} >> environment_test_with_pyversion.yml
```

.azure-pipelines/matrix.yml
```diff
@@ -31,10 +40,14 @@ jobs:
           - script: |
               source "${HOME}/conda/etc/profile.d/conda.sh"
               source "${HOME}/conda/etc/profile.d/mamba.sh"
-              echo "  - python="${{ py_vers }} >> environment_test.yml
-              mamba env create -f environment_test.yml
+              cp environment_test.yml environment_test_with_pyversion.yml
+              echo "  - python="{{ py_vers }} >> environment_test_with_pyversion.yml
```
santisoler, line 44:
```suggestion
              echo "  - python="${{ py_vers }} >> environment_test_with_pyversion.yml
```

.azure-pipelines/matrix.yml
jcapriot: Yeah, I meant to have it do that in the first place. I had already made that fix in #1469 as well.

santisoler:
Merging this now that CI passes! Thanks @jcapriot!

PR #1454, opened 2024-05-07
Augmented receivers for airborne NSEM

dccowan:
#### Summary
Add/modify NSEM receiver classes to simulate various AirMT data types, both theoretical and ones used in practice. Data types include:

* Admittance tensor data
* ZTEM data (all fields measured at same location before)
* Quasi-impedance data
* MobileMT data

#### PR Checklist
      to a Pull Request


dccowan:
The **getP** method used by all receiver classes is visible by the user. This should be a private method.

dccowan:
We should think of how we want inheritance to work for these classes. All receivers use the getP method. All but Tipper data has locations_e and locations_h. Admittance and Tipper use the same components and orientations.

santisoler:
Review request received 😎! I'll get into this in the following days.

simpeg/electromagnetics/natural_source/__init__.py
```diff
@@ -23,8 +23,11 @@
 .. autosummary::
   :toctree: generated/
 
+  receivers.PointImpedance
   receivers.PointNaturalSource
   receivers.Point3DTipper
+  receivers.Point3DAdmittance
+  receivers.Point3DMobileMT
```
lheagy, line 30: A thought on naming... for the MobileMT and elsewhere where we now have different base station and measurement locations, we might want to move away from `Point`

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -468,28 +776,318 @@ def __init__(
...
+                        )
+                else:
+                    raise Exception("incorrect size of list, must be length of 1 or 2")
+
+                self._locations_bs = locations[0]
+                locations = locations[-1]
+            elif isinstance(locations, np.ndarray):
+                self._locations_bs = locations
+            else:
+                raise Exception("locations need to be either a list or numpy array")
```
santisoler, line 1027: It's better to specify the type of exception instead of using `Exception`.
```suggestion
                raise TypeError("locations need to be either a list or numpy array")
```

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -468,28 +776,318 @@ def __init__(
...
+        input argument, the receiver can be set to compute the following:
+        - 'real': Real component of the tipper (unitless)
+        - 'imag': Imaginary component of the tipper (unitless)
+    locations_bs : (n_loc, n_dim) numpy.ndarray, optional
+        Locations for remote horizontal magnetic field measurements (e.g. base station).
+        Must be same shape as `locations`.
+    locations_e : (n_loc, n_dim) numpy.ndarray, optional
+        Deprecated property. To be removed in simpeg 0.20.0.
+    locations_h : (n_loc, n_dim) numpy.ndarray, optional
+        Deprecated property. To be removed in simpeg 0.20.0.
```
santisoler, line 985: I'm confused about these two parameters marked as deprecated. Is there a reason why they should be deprecated from now on?

If we do need to deprecate them, I'm hesitant to include the deprecation in this PR that is mostly about introducing new features. I would open a separate PR just to deprecate these arguments. BTW, I think these arguments should be removed at least in simpeg 0.23.0.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -627,3 +1188,25 @@ def evalDeriv(self, src, mesh, f, du_dm_v=None, v=None, adjoint=False):
         if adjoint:
             return imp_deriv
         return getattr(imp_deriv, self.component)
+
+
+@deprecate_class(removal_version="0.20.0", error=False)
+class PointNaturalSource(PointImpedance):
+    """This class is deprecated and will be removed in simpeg v0.20.0.
+    Please use :class:`.PointImpedance`."""
```
santisoler, line 1196: If we deprecate this function we should remove it at least in simpeg v0.23.
```suggestion
    """This class is deprecated and will be removed in simpeg v0.23.0.
    Please use :class:`.PointImpedance`."""
```

tests/em/nsem/forward/test_Simulation3D_vs_Analytic_pytest.py
```diff
@@ -0,0 +1,206 @@
...
+    analytic_solution = np.hstack(
+        [
+            get_analytic_halfspace_solution(sigma_hs, f, survey_type, component)
+            for f in frequencies
+        ]
+    )
+    analytic_solution = np.repeat(analytic_solution, n_locations)
+
+    # print(np.c_[numeric_solution, analytic_solution])
+
```
santisoler, line 200: We don't need this commented line.
```suggestion
```

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -66,16 +146,271 @@ def __init__(
...
+
+        this method computes and returns the derivative:
+
+        .. math::
+            \dfrac{\partial \mathbf{d}}{\partial \mathbf{u}} =
+            \dfrac{\partial [ \mathbf{P} (\mathbf{u}) ]}{\partial \mathbf{u}}
+
+        Parameters
+        ----------
+        str : .frequency_domain.sources.BaseFDEMSrc
```
santisoler, line 300: Minor typo:
```suggestion
        src : .frequency_domain.sources.BaseFDEMSrc
```

santisoler (changes requested):
Thanks @dccowan for the work done in this PR! It has a good shape!

I made a first pass and left a few comments below. Nothing too critical.

A few broad comments around the whole PR:
- I think we should probably use a © character when mentioning systems developed by companies in the docstrings. Just to make it clear that those can be trademarked systems.
- I noticed that a lot of errors are being raised with the general `Exception`. It's always recommended to specify the type of error and to avoid the usage of the general `Exception`. For most of the errors I saw here, `ValueError` and a `TypeError` fit well. I haven't commented all the appearances, just a single one.
- Since this PR is mostly about introducing new features, I'm a little bit hesitant to deprecate classes, properties and arguments here. I'm tempted to open a different PR for those changes. Is there a strong reason for us to deprecate bits here?


Let me know if I need to clarify some of my comments or if you need help with anything.

Great job with the tests!

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -66,16 +146,271 @@ def __init__(
...
+        Returns
+        -------
+        (n_data, dim) numpy.ndarray
+            Location for the magnetic field measurements for each datum.
+        """
+        return self._locations_h
+
+    def _eval_mobilemt(self, src, mesh, f):
+        if mesh.dim < 3:
+            raise NotImplementedError("MobileMT receiver not implemented for dim < 3.")
```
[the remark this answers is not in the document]
dccowan: Better yet, we should test it as part of the setter for the survey object attached to the simulation.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -66,16 +146,271 @@ def __init__(
...
+        Parameters
+        ----------
+        str : .frequency_domain.sources.BaseFDEMSrc
+            The NSEM source.
+        mesh : discretize.TensorMesh
+            Mesh on which the discretize solution is obtained.
+        f : .frequency_domain.fields.FieldsFDEM
+            NSEM fields object for the source.
+        du_dm_v : None,
+            Supply pre-computed derivative?
```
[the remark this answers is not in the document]
dccowan: I don't think is used right now. It was there previously and I decided not to touch it.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -66,16 +146,271 @@ def __init__(
...
+        """Compute receiver data from the discrete field solution.
+
+        Parameters
+        ----------
+        src : .frequency_domain.sources.BaseFDEMSrc
+            NSEM source.
+        mesh : discretize.TensorMesh
+            Mesh on which the discretize solution is obtained.
+        f : .frequency_domain.fields.FieldsFDEM
+            NSEM fields object of the source.
```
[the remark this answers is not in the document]
dccowan: Added

santisoler:
Idea that came up during last meeting (2024-06-05): don't use trademarked systems in the class names.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -1,50 +1,134 @@
...
+    else:
+        raise ValueError("Field type {} unrecognized. Use 'e' or 'h'".format(field))
+
+    P = mesh.get_interpolation_matrix(locs, projected_grid)
+    if rx.storeProjections:
+        rx._Ps[(mesh, stash_name, field)] = P
+    return P
+
+
+class MobileMT(BaseRx):
```
lheagy, line 85: I agree with the comment from @domfournier about not using trademark names as classes. What do you think about instead using `ApparentConductivity` as the class name (and removing reference to specific vendors in the docs)?

dccowan: Ya, ApparentConductivity seems fine.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -434,26 +720,48 @@ def evalDeriv(self, src, mesh, f, du_dm_v=None, v=None, adjoint=False):
...
+    locations_h : (n_loc, n_dim) numpy.ndarray
+        Locations where the magnetic fields are measured.
+        Must be same size as `locations_e`.
     """
 
     def __init__(
         self,
-        locations,
-        orientation="zx",
+        locations=None,
```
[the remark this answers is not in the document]
dccowan: It's because you can use either 'locations', or 'locations_e' and 'locations_h'. You will get an error is everything is None

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -1,42 +1,122 @@
...
+        aren't reused.
+    """
+    # Inherited method from BaseRx projecting to self.locations
+    if mesh.dim < 3:
+        return rx.getP(mesh, projected_grid)
+
+    if is_tipper_bs:
+        stash_name = projected_grid + "_bs"
+    else:
+        stash_name = projected_grid
```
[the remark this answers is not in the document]
dccowan: This happens behind the scenes. So I don't think it's too important. I decided to make this function hidden so that I could use it universally for all receiver classes.  The important part is ensuring you can simulate AFMAG data or ZTEM data. I decided to make this function hidden so that I could use it universally for all receiver classes.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -434,26 +720,48 @@ def evalDeriv(self, src, mesh, f, du_dm_v=None, v=None, adjoint=False):
...
+        Locations where the magnetic fields are measured.
+        Must be same size as `locations_e`.
     """
 
     def __init__(
         self,
-        locations,
-        orientation="zx",
+        locations=None,
+        orientation="xy",
```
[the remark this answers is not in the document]
dccowan: This might be outdated. This corresponds to the Admittance class, which is new.

dccowan:
@santisoler PR is ready for final review.

simpeg/electromagnetics/natural_source/receivers.py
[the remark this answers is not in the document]
santisoler: Makes sense!

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -1,50 +1,133 @@
...
+        locs = rx.locations_e
+    elif field == "h":
+        if isinstance(rx, ApparentConductivity):
+            locs = rx.locations_h
+        elif is_tipper_bs:
+            locs = rx.locations_bs
+        else:
+            locs = rx.locations
+    else:
+        raise ValueError("Field type {} unrecognized. Use 'e' or 'h'".format(field))
```
santisoler, line 77: Small change to use f-string here.
```suggestion
        raise ValueError(f"Field type {field} unrecognized. Use 'e' or 'h'")
```

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -407,89 +690,421 @@ def eval(self, src, mesh, f, return_complex=False):  # noqa: A003
...
+                raise ValueError("locations need to be either a list or numpy array")
+
+        super().__init__(locations)
+
+        if (locations_e is not None) or (locations_h is not None):
+            warnings.warn(
+                (
+                    "'locations_e' and 'locations_h' are deprecated properties that are unused by the Tipper class.",
+                    "Receiver locations are set using 'locations' (and 'locations_bs').",
+                    "These property will be removed in simpeg v.0.23.0.",
```
santisoler, line 1053: Fix typo
```suggestion
                    "These properties will be removed in simpeg v.0.23.0.",
```

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -407,89 +690,421 @@ def eval(self, src, mesh, f, return_complex=False):  # noqa: A003
...
+                ),
+                DeprecationWarning,
```
santisoler, line 1055: A `FutureWarning` is more suited for this type of warnings. `DeprecationWarning`s are [ignored by default](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) and intended to other developers. But we want these warnings to reach users if they are passing deprecated attributes.
```suggestion
                FutureWarning,
```

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -1,50 +1,133 @@
...
 from ...survey import BaseRx
 
 
 def _alpha(src):
     return 1 / (2 * np.pi * mu_0 * src.frequency)
 
 
-class PointNaturalSource(BaseRx):
-    """Point receiver class for magnetotelluric simulations.
+# THIS SHOULD BE HIDDEN METHOD
```
santisoler, line 16: I don't think we need this comment anymore.
```suggestion
```

santisoler (changes requested):
Looking great @dccowan! I left a few minor suggestions below. I think this is quite ready to be merged.

My only doubt is related to the addition of deprecated arguments, like `locations_e` and `locations_h` in `Tipper`. Since this is a new class with a new constructor that overrides the one of its parent, is there any reason we want to add arguments that we'll have to remove in the next release? Note: if you remove the arguments, feel free to ignore my comments related to the warnings I made below.

Related to this point: what if we leave the old code for the classes we are going to deprecate in the next release (`Point3DTipper` and `PointNaturalSource`)? This way we don't need to support deprecated arguments in the new classes.

Let me know what do you think.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -1,18 +1,1200 @@
...
+
+        Parameters
+        ----------
+        src : .frequency_domain.sources.BaseFDEMSrc
+            The NSEM source.
+        mesh : discretize.TensorMesh
+            Mesh on which the discretize solution is obtained.
+        f : simpeg.electromagnetics.frequency_domain.fields.FieldsFDEM
+            NSEM fields object for the source.
+        du_dm_v : None,
```
santisoler, line 309:
```suggestion
        du_dm_v : None, optional
```

santisoler (changes requested):
Looking great. I just gave it another pass and left a few comments below.

I think this is getting in good shape. The one thing I would add is a few more tests to cover the situations that aren't being tested (checkout the codecov report for more details), and also add a few tests that checks if the FutureWarnings are raised for the deprecated classes.

Let me know if you need any help with those.

jcapriot:
I’ll try to take a look through this in the next few days myself as well, please wait for merging.

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -81,6 +82,22 @@ def survey(self):
     def survey(self, value):
         if value is not None:
             value = validate_type("survey", value, Survey, cast=False)
+            # Check for acceptable receiver types
+            is_valid = np.all(
+                [
+                    np.all(
+                        [
+                            type(rx_ii) in (Impedance, PointNaturalSource)
```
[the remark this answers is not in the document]
dccowan: No, child classes of Impedance and PointNaturalSource cannot be used. The simulation is not for Tipper data.

jcapriot (changes requested):
Generally, I think the inheritance of these classes could use a decent amount of work.

From what I can see the `Impedence` class is just a renaming of the `PointNaturalSource`. But it doesn't make sense that this is now a subclass of `ApparentConductivity`.

It seems like this is asking for a `BaseNaturalSourceRx` receiver class that handles common items (i.e. setting `locations`, `locations_e`, and `locations_h`), and implement the `_get_projection` method on that class.

Then that would give us,
```python
class BaseNaturalSourceRx(BaseRx):
    def __init__(self, locations=None, locations_e=None, locations_h=None):
         ...

    def _get_projection(self, mesh, projected_grid, field="e", is_tipper_bs=False)
class ApparentConductivity(BaseNaturalSourceRx):
    ...

class Impedance(BaseNaturalSourceRx):
    ...

class Admittance(Impedance):
    ...

class Tipper(BaseNaturalSourceRx):
    def __init__(self, locations_e, locations_h):
       super().__init__(locations_e=locations_e, locations_h=locations_h)
        ...

```

Then some more implementation requests:
* I think you should re-use the `locations_e` and `locations_h` argument for the `Tipper` class in this inheritance scheme , instead of defining a new one that effectively does the same thing (which should simplify the _get_projections method a bit).

* make the `PointNatrualSource` and the `Point3DTipper` classes explicitly a deprecated version of the new versions using the `deprecate_class` decorator:
```python
@deprecate_class(removal_version='0.24.0')
class PointNaturalSource(Impedance):
     pass

@deprecate_class(removal_version='0.24.0')
class Point3DTipper(Tipper):
    pass

```

dccowan:
> Generally, I think the inheritance of these classes could use a decent amount of work.
> 
> From what I can see the `Impedence` class is just a renaming of the `PointNaturalSource`. But it doesn't make sense that this is now a subclass of `ApparentConductivity`.
> 
> It seems like this is asking for a `BaseNaturalSourceRx` receiver class that handles common items (i.e. setting `locations`, `locations_e`, and `locations_h`), and implement the `_get_projection` method on that class.
> 
> Then that would give us,
> 
> ```python
> class BaseNaturalSourceRx(BaseRx):
>     def __init__(self, locations=None, locations_e=None, locations_h=None):
>          ...
> 
>     def _get_projection(self, mesh, projected_grid, field="e", is_tipper_bs=False):
>         ...
> 
> class ApparentConductivity(BaseNaturalSourceRx):
>     ...
> 
> class Impedance(BaseNaturalSourceRx):
>     ...
> 
> class Admittance(Impedance):
>     ...
> 
> class Tipper(BaseNaturalSourceRx): #< I don't think the signature of this class needs to be changed...
>     def __init__(
>         self,
>         locations,
>         orientation="zx",
>         component="real",
>         locations_e=None,
>         locations_h=None,):
>        super().__init__(locations=locations, locations_e=locations_e, locations_h=locations_h)
>         ...
> ```
> 
> Then some more implementation requests:
> 
> * I think you should re-use the `locations_e` and `locations_h` argument for the `Tipper` class in this inheritance scheme , instead of defining a new one that effectively does the same thing (which should simplify the _get_projections method a bit).
> * If you want to get the orientation setter to be common, you might be able to define the setter on the `BaseNaturalSourceRx` to use a variable that is defined at the class level of the implemented versions.
> * The new Tipper class just adds support for more orientations in the E-field measurement correct? I think I see the one line difference in there.
> * make the `PointNatrualSource` and the `Point3DTipper` classes explicitly a deprecated version of the new versions using the `deprecate_class` decorator:
> 
> ```python
> @deprecate_class(removal_version='0.24.0')
> class PointNaturalSource(Impedance):
>      pass
> 
> @deprecate_class(removal_version='0.24.0')
> class Point3DTipper(Tipper):
>     pass
> ```

I had done something like you are suggesting previously. But we decided to define classes the way they are not for several reasons.

* There are no E-field measurements for tipper data. 'locations_e' should not be an input argument. Also, 'locations_h' made sense for the Point3DTipper class because it could only simulate AFMAG data. But for the general tipper data allowed by the new class, horizontal H-fields can be instead measured at a base station. So I did not want the Tipper receiver class to have 'locations_h'.
* The orientations allowed for admittance and tipper are not the same as impedance. I wanted to make a base receiver class, but the combinations of orientation and location arguments don't universally overlap.
* In the interim, there may be a more compact way to organize the receiver classes. But this was seen as the easiest way to remove the original receiver classes when the time comes.

jcapriot:
> > Generally, I think the inheritance of these classes could use a decent amount of work.
> > From what I can see the `Impedence` class is just a renaming of the `PointNaturalSource`. But it doesn't make sense that this is now a subclass of `ApparentConductivity`.
> > It seems like this is asking for a `BaseNaturalSourceRx` receiver class that handles common items (i.e. setting `locations`, `locations_e`, and `locations_h`), and implement the `_get_projection` method on that class.
> > Then that would give us,
> > ```python
> > class BaseNaturalSourceRx(BaseRx):
> >     def __init__(self, locations=None, locations_e=None, locations_h=None):
> >          ...
> > 
> >     def _get_projection(self, mesh, projected_grid, field="e", is_tipper_bs=False):
> >         ...
> > 
> > class ApparentConductivity(BaseNaturalSourceRx):
> >     ...
> > 
> > class Impedance(BaseNaturalSourceRx):
> >     ...
> > 
> > class Admittance(Impedance):
> >     ...
> > 
> > class Tipper(BaseNaturalSourceRx): #< I don't think the signature of this class needs to be changed...
> >     def __init__(
> >         self,
> >         locations,
> >         orientation="zx",
> >         component="real",
> >         locations_e=None,
> >         locations_h=None,):
> >        super().__init__(locations=locations, locations_e=locations_e, locations_h=locations_h)
> >         ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then some more implementation requests:
> > 
> > * I think you should re-use the `locations_e` and `locations_h` argument for the `Tipper` class in this inheritance scheme , instead of defining a new one that effectively does the same thing (which should simplify the _get_projections method a bit).
> > * If you want to get the orientation setter to be common, you might be able to define the setter on the `BaseNaturalSourceRx` to use a variable that is defined at the class level of the implemented versions.
> > * The new Tipper class just adds support for more orientations in the E-field measurement correct? I think I see the one line difference in there.
> > * make the `PointNatrualSource` and the `Point3DTipper` classes explicitly a deprecated version of the new versions using the `deprecate_class` decorator:
> > 
> > ```python
> > @deprecate_class(removal_version='0.24.0')
> > class PointNaturalSource(Impedance):
> >      pass
> > 
> > @deprecate_class(removal_version='0.24.0')
> > class Point3DTipper(Tipper):
> >     pass
> > ```
> 
> I had done something like you are suggesting previously. But we decided to define classes the way they are not for several reasons.
> 
> * There are no E-field measurements for tipper data. 'locations_e' should not be an input argument. Also, 'locations_h' made sense for the Point3DTipper class because it could only simulate AFMAG data. But for the general tipper data allowed by the new class, horizontal H-fields can be instead measured at a base station. So I did not want the Tipper receiver class to have 'locations_h'.
> * The orientations allowed for admittance and tipper are not the same as impedance. I wanted to make a base receiver class, but the combinations of orientation and location arguments don't universally overlap.
> * In the interim, there may be a more compact way to organize the receiver classes. But this was seen as the easiest way to remove the original receiver classes when the time comes.

I didn't see any of that discussion here on inheritance, if it wasn't then I don't agree with what is here.

Is 'ApparentConductivity` the best name for the other class as well? It's a very specific apparent conductivity measurement, and I'd imagine it could confuse people who have traditional apparent conductivity measurements.

Either way, I think you're limiting yourself too much trying to stick with the way the classes were written before. There could simply be a base class that accepts a `locations` argument, then each subclass would further expand upon it.

jcapriot:
Also, since these are new classes, I don’t think there’s any reason to continue supporting two different way of setting locations for the receivers. Just make `locations` a required argument and let it accept either `(n_locs, dim) or (2, n_locs, dim) array_like`. If you want you can still add getters with relevant names to the subclasses.

dccowan:
> Also, since these are new classes, I don’t think there’s any reason to continue supporting two different way of passing locations for the receivers. Just make `locations` a required argument and let it accept either `(n_locs, dim) or (2, n_locs, dim) array_like`. If you want you can still add getters with relevant names to the subclasses.

I agree with this. But I am strongly advocating for locations_e locations_h type arguments instead of an all purpose locations argument that can have different dimensions. I think this ensures the user clearly understand where the fields are being measured.

dccowan:
> > > Generally, I think the inheritance of these classes could use a decent amount of work.
> > > From what I can see the `Impedence` class is just a renaming of the `PointNaturalSource`. But it doesn't make sense that this is now a subclass of `ApparentConductivity`.
> > > It seems like this is asking for a `BaseNaturalSourceRx` receiver class that handles common items (i.e. setting `locations`, `locations_e`, and `locations_h`), and implement the `_get_projection` method on that class.
> > > Then that would give us,
> > > ```python
> > > class BaseNaturalSourceRx(BaseRx):
> > >     def __init__(self, locations=None, locations_e=None, locations_h=None):
> > >          ...
> > > 
> > >     def _get_projection(self, mesh, projected_grid, field="e", is_tipper_bs=False):
> > >         ...
> > > 
> > > class ApparentConductivity(BaseNaturalSourceRx):
> > >     ...
> > > 
> > > class Impedance(BaseNaturalSourceRx):
> > >     ...
> > > 
> > > class Admittance(Impedance):
> > >     ...
> > > 
> > > class Tipper(BaseNaturalSourceRx): #< I don't think the signature of this class needs to be changed...
> > >     def __init__(
> > >         self,
> > >         locations,
> > >         orientation="zx",
> > >         component="real",
> > >         locations_e=None,
> > >         locations_h=None,):
> > >        super().__init__(locations=locations, locations_e=locations_e, locations_h=locations_h)
> > >         ...
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then some more implementation requests:
> > > 
> > > * I think you should re-use the `locations_e` and `locations_h` argument for the `Tipper` class in this inheritance scheme , instead of defining a new one that effectively does the same thing (which should simplify the _get_projections method a bit).
> > > * If you want to get the orientation setter to be common, you might be able to define the setter on the `BaseNaturalSourceRx` to use a variable that is defined at the class level of the implemented versions.
> > > * The new Tipper class just adds support for more orientations in the E-field measurement correct? I think I see the one line difference in there.
> > > * make the `PointNatrualSource` and the `Point3DTipper` classes explicitly a deprecated version of the new versions using the `deprecate_class` decorator:
> > > 
> > > ```python
> > > @deprecate_class(removal_version='0.24.0')
> > > class PointNaturalSource(Impedance):
> > >      pass
> > > 
> > > @deprecate_class(removal_version='0.24.0')
> > > class Point3DTipper(Tipper):
> > >     pass
> > > ```
> > 
> > 
> > I had done something like you are suggesting previously. But we decided to define classes the way they are not for several reasons.
> > 
> > * There are no E-field measurements for tipper data. 'locations_e' should not be an input argument. Also, 'locations_h' made sense for the Point3DTipper class because it could only simulate AFMAG data. But for the general tipper data allowed by the new class, horizontal H-fields can be instead measured at a base station. So I did not want the Tipper receiver class to have 'locations_h'.
> > * The orientations allowed for admittance and tipper are not the same as impedance. I wanted to make a base receiver class, but the combinations of orientation and location arguments don't universally overlap.
> > * In the interim, there may be a more compact way to organize the receiver classes. But this was seen as the easiest way to remove the original receiver classes when the time comes.
> 
> I didn't see any of that discussion here on inheritance, if it wasn't then I don't agree with what is here.
> 
> Is 'ApparentConductivity` the best name for the other class as well? It's a very specific apparent conductivity measurement, and I'd imagine it could confuse people who have traditional apparent conductivity measurements.
> 
> Either way, I think you're limiting yourself too much trying to stick with the way the classes were written before. There could simply be a base class that accepts a `locations` argument, then each subclass would further expand upon it.

'ApparentConductivity' was called 'MobileMT' but a discussion was had and we wanted to avoid any system-specific names. What about 'PseudoConductivity'?

I don't need to make another base class if the only unifying proper is 'locations'. That's what we're inheriting from ...survey.BaseRx.

jcapriot:
> > Also, since these are new classes, I don’t think there’s any reason to continue supporting two different way of passing locations for the receivers. Just make `locations` a required argument and let it accept either `(n_locs, dim) or (2, n_locs, dim) array_like`. If you want you can still add getters with relevant names to the subclasses.
> 
> I agree with this. But I am strongly advocating for locations_e locations_h type arguments instead of an all purpose locations argument that can have different dimensions. I think this ensures the user clearly understand where the fields are being measured.

Then options for the call signatures of these classes when defining locations would be:

```python

class Impedance(BaseNaturalSourceRx):

    # option 1
    def __init__(locations):
        '''locations : (n_loc, n_dim) or (2, n_loc, n_dim) array_like'''

    # option 2
    def __init__(location_e, location_h=None):
        '''
        location_e : (n_loc, dim) array_like
        location_h : (n_loc, dim) array_like, optional
            Defaults to the same locations as `location_e`
        '''

```
Either way I'm happy, but I do agree that the second option is probably clearer. They should still both be able to inherit from a `BaseNaturalSourceRx` receiver class that would store this information in its `locations` (think of how the `Dipole` receiver works in DC).

> I don't need to make another base class if the only unifying proper is 'locations'. That's what we're inheriting from ...survey.BaseRx.

The class would need to be able to accept locations as a `(2, n_loc, n_dim) array_like` instead of an `(n_loc, n_dim) array_like`, which would necessitate an extension.

lheagy:
> 'ApparentConductivity' was called 'MobileMT' but a discussion was had and we wanted to avoid any system-specific names. What about 'PseudoConductivity'?

What do you think about `TotalFieldApparentConductivity` (It is a bit long, but is descriptive). This distinguishes it from an apparent conductivity computed from an impedance and still refers to the data as `ApparentConductivity`, which is how it is referred to in the literature. I think it would be good to keep ApparentConductivity in the name because that is how the data are referred to, and it is looking at total E field (and we may want to extend this to give the option to the user to either use the determinant of **H** or |**H**| down the line.

jcapriot:
> > 'ApparentConductivity' was called 'MobileMT' but a discussion was had and we wanted to avoid any system-specific names. What about 'PseudoConductivity'?
> 
> What do you think about `TotalFieldApparentConductivity` (It is a bit long, but is descriptive). This distinguishes it from an apparent conductivity computed from an impedance and still refers to the data as `ApparentConductivity`, which is how it is referred to in the literature. I think it would be good to keep ApparentConductivity in the name because that is how the data are referred to, and it is looking at total E field (and we may want to extend this to give the option to the user to either use the determinant of **H** or |**H**| down the line.

Essentially it is defined using the ratio of the magnitudes of the Electric and Magnetic fields, correct? That sounds fine to me then. Maybe, `ApparentConductivityMagnitude` (just as another suggestion not that I like it any better). It's such an odd quantity to name appropriately in the context of all of the other types of MT data.

dccowan:
@dccowan on the hook for improved derivative tests for tippers. Essentially using a more appropriate model, otherwise some tipper components can be zero.

dccowan:
> @dccowan on the hook for improved derivative tests for tippers. Essentially using a more appropriate model, otherwise some tipper components can be zero.

A small block model was added to ensure non-zero tipper anomalies in tests. Tipper components xx, xy, yx, yy are 1 or 0 unless the fields are measured at different locations. This was addressed to ensure appropriate tipper data for tests.

simpeg/electromagnetics/natural_source/receivers.py
```diff
@@ -569,61 +789,586 @@ def _eval_tipper_deriv(self, src, mesh, f, du_dm_v=None, v=None, adjoint=False):
...
+        if return_complex:
+            temp = self.component
+            self.component = "complex"
+            out = super().eval(src, mesh, f)
+            self.component = temp
+        else:
+            out = super().eval(src, mesh, f)
+        return out
+
+    # locations = property(lambda self: self._locations[0], Impedance.locations.fset)
```
dccowan, line 1312: This line to return the locations that were defined for the soon to be deprecated class breaks a lot of tests. By adding this test, the inherited method which returns the projection matrices doesn't work properly.

dccowan:
Adjoint test for Simulation2DElectricField passes to almost machine precision. This corresponds to data derived from the Zxy impedance. Adjoint test for Simulation2DMagneticField only passes to relative precision of about 1e-7. This corresponds to data derived from the Zyx impedance. This issue is not due to the functionality in the PR. The same would be found for the functionality of main branch. Also, this issue does not occur when simulating the same data using the 3D simulation class.

dccowan:
The ``getJ`` method for Simulation2DMagneticField fails.

dccowan:
> Adjoint test for Simulation2DElectricField passes to almost machine precision. This corresponds to data derived from the Zxy impedance. Adjoint test for Simulation2DMagneticField only passes to relative precision of about 1e-7. This corresponds to data derived from the Zyx impedance. This issue is not due to the functionality in the PR. The same would be found for the functionality of main branch. Also, this issue does not occur when simulating the same data using the 3D simulation class.

Will be handled in separate PR. See issue #1540

dccowan:
> The `getJ` method for Simulation2DMagneticField fails.

This extends to all 1D and 2D simulation classes. This will be handled in a separate PR. See issue #1541

jcapriot:
@santisoler More codecov issues. Not sure why it's so hit or miss on success rate anymore.

jcapriot:
Also, This should be good to go now assuming everything still passes.

santisoler:
> @santisoler More codecov issues. Not sure why it's so hit or miss on success rate anymore.

I saw that a few days ago too. But I also saw that codedov was failing to report status (not only in this repo), so I thought there might be an integration issue between codecov and github.

These failures are annoying and hard to track down... let's cross fingers for now and see what happens...

For the future, I would like to continue working on #1537 so we could download the raw coverage reports locally and compare them with the output of codecov, just to check where the issue is coming from.

PR #1392, opened 2024-03-26
Check inputs for converting 3d surveys to 2d lines

santisoler:
#### Summary

Add checks to validate the `survey` and `lineID` inputs in the `convert_survey_3d_to_2d_lines` function. Add tests to check if errors are raised after passing invalid arguments. Improve the docstring of the function.

#### PR Checklist
* [ ] If this is a work in progress PR, set as a Draft PR
* [ ] Linted my code according to the [style guides](https://docs.simpeg.xyz/content/getting_started/contributing/code-style.html).
* [ ] Added [tests](https://docs.simpeg.xyz/content/getting_started/practices.html#testing) to verify changes to the code.
* [ ] Added necessary documentation to any new functions/classes following the
      expect [style](https://docs.simpeg.xyz/content/getting_started/practices.html#documentation).
* [ ] Marked as ready for review (if this is was a draft PR), and converted
      to a Pull Request
* [ ] Tagged ``@simpeg/simpeg-developers`` when ready for review.

#### Reference issue

Noticed the need to add these checks while reviewing #1390

PR #1460, opened 2024-05-13
Fix bug on arguments of beta estimator directives

santisoler:
#### Summary

Fix bug in the constructor of beta estimator directive classes: the `seed` argument was being passed as a positional argument to the parent's constructor, which would assign it to the `n_pw_iter` attribute instead to the `seed`, leading to unwanted behaviours. Pass the arguments to the constructor through keywords to make it more explicit and less prone to errors. Add tests that would fail without this bugfix. Remove the `n_pw_iter` and the `method` arguments from the constructor of `BaseBetaEstimator`.

#### PR Checklist
      to a Pull Request

#### Reference issue

Related to #1290

### DIFF
--- a/tests/base/test_directives.py
+++ b/tests/base/test_directives.py
@@ -397,5 +397,33 @@ class TestUpdateSensitivityNormalization:
             d_temp.normalization_method = normalization_method
 
 
+class TestBetaEstimatorArguments:
+    """
+    Test if arguments are assigned in beta estimator directives.
+    """
+
+    def test_beta_estimate_by_eig(self):
+        """Test on directives.BetaEstimate_ByEig."""
+        beta0_ratio = 3.0
+        n_pw_iter = 3
+        seed = 42
+        directive = directives.BetaEstimate_ByEig(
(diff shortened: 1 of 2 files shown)


tests/base/test_directives.py
```diff
@@ -397,5 +397,33 @@ def test_normalization_method_setter_invalid(self, normalization_method):
             d_temp.normalization_method = normalization_method
 
 
+class TestBetaEstimatorArguments:
+    """
+    Test if arguments are assigned in beta estimator directives.
```
santisoler, line 402:
```suggestion
    Test if arguments are assigned in beta estimator directives.
    
    These tests catch the bug described and fixed in #1460.
```

santisoler:
@dccowan would you like to take a look at this PR? While working on extending the tests for #1392, I noticed that the `seed` argument was not being correctly passed to the parent's constructor, therefore the `seed` for these two directives wasn't being set even if that was the user's desire.

santisoler:
I'm merging this so we can unblock some other PRs.

PR #1466, opened 2024-05-15
Replace deprecated refine_tree_xyz across tests

santisoler:
#### Summary

Replace usage of the deprecated `refine_tree_xyz` function from discretize for their latest equivalent methods of the `TreeMesh`: `refine_surface` and `refine_points`. Apply these replacements across simpeg tests.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/tests/pf/test_mag_MVI_Octree.py
+++ b/tests/pf/test_mag_MVI_Octree.py
@@ -11,7 +11,7 @@ from simpeg import (
 )
 
 
-from discretize.utils import mesh_builder_xyz, refine_tree_xyz, active_from_xyz
+from discretize.utils import mesh_builder_xyz, active_from_xyz
 import numpy as np
 from simpeg.potential_fields import magnetics as mag
 import shutil
@@ -61,8 +61,8 @@ class MVIProblemTest(unittest.TestCase):
         mesh = mesh_builder_xyz(
             xyzLoc, h, padding_distance=padDist, depth_core=100, mesh_type="tree"
         )
-        mesh = refine_tree_xyz(
-            mesh, topo, method="surface", octree_levels=[4, 4], finalize=True
+        mesh.refine_surface(
+            topo, method="surface", padding_cells_by_level=[4, 4], finalize=True
         )
         self.mesh = mesh
         # Define an active cells from topo
--- a/tests/pf/test_mag_inversion_linear_Octree.py
+++ b/tests/pf/test_mag_inversion_linear_Octree.py
@@ -2,7 +2,7 @@ import shutil
 import unittest
 import numpy as np
 
-from discretize.utils import mesh_builder_xyz, refine_tree_xyz, active_from_xyz
+from discretize.utils import mesh_builder_xyz, active_from_xyz
 from simpeg import (
     directives,
     maps,
@@ -71,14 +71,7 @@ class MagInvLinProblemTest(unittest.TestCase):
             mesh_type="TREE",
         )
 
-        self.mesh = refine_tree_xyz(
-            self.mesh,
-            topo,
-            method="surface",
-            octree_levels=nCpad,
-            octree_levels_padding=nCpad,
-            finalize=True,
-        )
+        self.mesh.refine_surface(topo, padding_cells_by_level=nCpad, finalize=True)
 
         # Define an active cells from topo
         actv = active_from_xyz(self.mesh, topo)
--- a/tests/pf/test_pf_quadtree_inversion_linear.py
+++ b/tests/pf/test_pf_quadtree_inversion_linear.py
@@ -3,7 +3,7 @@ import unittest
 
 import numpy as np
 from discretize import TensorMesh
-from discretize.utils import mesh_builder_xyz, mkvc, refine_tree_xyz
+from discretize.utils import mesh_builder_xyz, mkvc
 
 from simpeg import (
     data_misfit,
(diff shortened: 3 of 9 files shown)


tests/pf/test_mag_MVI_Octree.py
```diff
@@ -61,8 +61,8 @@ def setUp(self):
         mesh = mesh_builder_xyz(
             xyzLoc, h, padding_distance=padDist, depth_core=100, mesh_type="tree"
         )
-        mesh = refine_tree_xyz(
-            mesh, topo, method="surface", octree_levels=[4, 4], finalize=True
+        mesh.refine_surface(
+            topo, method="surface", padding_cells_by_level=[4, 4], finalize=True
```
santisoler, line 65:
```suggestion
            topo, padding_cells_by_level=[4, 4], finalize=True
```

tests/pf/test_pf_quadtree_inversion_linear.py
```diff
@@ -50,20 +50,15 @@ def create_mesh(h, padDist, nCpad, topo, data, min_height=1.0):
...
                 mesh_type="TREE",
             )
 
-            self.mesh = refine_tree_xyz(
-                mesh,
-                data[:, :2],
-                method="radial",
-                octree_levels=nCpad,
-                octree_levels_padding=nCpad,
-                finalize=True,
```
santisoler, line 66: @jcapriot, is there a replacement for `octree_levels_padding` in the new `refine_points` method? With the current change to the code, the method is not generating the same mesh as the original one.

jcapriot: Actually for a one-to-one here, I think you need to do:
`padding_cells_by_level=[nCpad]`. The updated methods are bit more clear in what the arguments do compared to documentation of the old one.

jcapriot: Well, it's been a while since I've looked at this, but I think you had the translation correctly to start. Like I said originally, the `octree_levels_padding` was unused for the radial option. The `refine_tree_xyz` function for radial basically looked at the `octree_levels` argument, then turned it into an array, iterated through it starting at the maximum level and created a ball around each point that was 

```Python
radius = 0
for i_level, n_pad in enumerate(octree_levels):
    radius += mesh.h[0].min() * n_pad * 2**(mesh.max_level - i_level) 
```

now it does:
``Python
radius = 0
h_min = np.max([h.min() for h in self.h])
for i_level, n_pad in enumerate(octree_levels):
    radius += h_min * n_pad * 2**(mesh.max_level - i_level)
```
So now it guarantees at least that many padding cells in all directions. So if the mesh had unequal cell sizes, it likely changed slightly.

Comment (santisoler) [reply tests/pf/test_pf_quadtree_inversion_linear.py:66]: using `padding_cells_by_level=[nCpad]` falls into an error because `nCpad` is already a list. `padding_cells_by_level=nCpad` works fine, and today I'm noticing they are generating the same mesh. I'll give it one more try and see what happens. Ignore this comments for now 😅

File: tests/pf/test_mag_inversion_linear_Octree.py [hunk 3]
```diff
@@ -71,14 +71,7 @@ def setUp(self):
...
         )
 
-        self.mesh = refine_tree_xyz(
-            self.mesh,
-            topo,
-            method="surface",
-            octree_levels=nCpad,
-            octree_levels_padding=nCpad,
-            finalize=True,
-        )
```
santisoler, line 81: Ok, my bad. The issue is not with radial methods, but with surface methods. When passing `octree_levels_padding`, the refined meshes are different between the old code and the new code.

Minimal example:
```python
import numpy as np
from simpeg import utils
from discretize.utils import mesh_builder_xyz, refine_tree_xyz

# First we need to define the direction of the inducing field
# As a simple case, we pick a vertical inducing field of magnitude
# 50,000nT.
# From old convention, field orientation is given as an
# azimuth from North (positive clockwise)
# and dip from the horizontal (positive downward).
h0_amplitude, h0_inclination, h0_declination = (50000.0, 90.0, 0.0)

# Create a mesh
h = [5, 5, 5]
padDist = np.ones((3, 2)) * 100
nCpad = [2, 4, 2]

# Create grid of points for topography
# Lets create a simple Gaussian topo and set the active cells
[xx, yy] = np.meshgrid(np.linspace(-200.0, 200.0, 50), np.linspace(-200.0, 200.0, 50))

b = 100
A = 50
zz = A * np.exp(-0.5 * ((xx / b) ** 2.0 + (yy / b) ** 2.0))

# We would usually load a topofile
topo = np.c_[utils.mkvc(xx), utils.mkvc(yy), utils.mkvc(zz)]

# Create and array of observation points
xr = np.linspace(-100.0, 100.0, 20)
yr = np.linspace(-100.0, 100.0, 20)
X, Y = np.meshgrid(xr, yr)
Z = A * np.exp(-0.5 * ((X / b) ** 2.0 + (Y / b) ** 2.0)) + 5

xyzLoc = np.c_[utils.mkvc(X.T), utils.mkvc(Y.T), utils.mkvc(Z.T)]

# Build mesh with deprecated functions
old_mesh = mesh_builder_xyz(
    xyzLoc,
    h,
    padding_distance=padDist,
    mesh_type="TREE",
)
old_mesh = refine_tree_xyz(
    old_mesh,
    topo,
    method="surface",
    octree_levels=nCpad,
    octree_levels_padding=nCpad,
    finalize=True,
)

# Build mesh with new methods
new_mesh = mesh_builder_xyz(
    xyzLoc,
    h,
    padding_distance=padDist,
    mesh_type="TREE",
)

new_mesh.refine_surface(topo, padding_cells_by_level=nCpad, finalize=True)

print(f"old_mesh.n_cells: {old_mesh.n_cells}")
print(f"new_mesh.n_cells: {new_mesh.n_cells}")
```
```
old_mesh.n_cells: 39624
new_mesh.n_cells: 57516
```

PR #1469, opened 2024-05-20
Dask races

jcapriot:
#### Summary
Updates for `dask>2024.2.1`

#### PR Checklist
      to a Pull Request

A few things:
1) dask recently updated how things items are hashed to create a more deterministic hash for items. For objects it will change depending on the state of the objects. `Simulation` objects are mutable things, thus their hash would change overtime for the same object. For the `DaskMetaSimulation`, we want a single simulation's hash (also a single map's hash) to be constant in time, so this PR adds uuid properties to each of those base classes that are then registered to be used by dask to hash the objects.

2) This deterministic hashing lead to a race condition in the error testing. When objects are scattered, dereferenced, then scattered again. The garbage collector would destroy the scattered objects dereferenced from the first call `after` the second scatter, thus creating a case where we are trying to access a Canceled Future.

3) This is also now a bit more rigorous about only using a single worker for each simulation operation when the simulations inadvertently live on multiple workers, by selecting the worker that had fewer simulations assigned to it.

#### Additional information

See https://github.com/dask/distributed/issues/8576 for more information about the race condition.

This also implements one of the updates in #1444 regarding the testing environment script so that we are sure we are creating the correct environment.


jcapriot:
Ok, this bit of maintenance seems ready.
It does add a private `_uuid` to Simulations and Maps that we can use to uniquely identify an instance of those objects, that is not tied to their state (needed for recent versions of dask).

jcapriot:
As this PR mostly changes the meta simulations which are marked as works in progress, it should be safe to merge now.

PR #1472, opened 2024-05-25
Update tests and examples to use the new `UpdateIRLS`

domfournier:
**Squash and merge summary**

Update tests, examples and tutorials that were using the deprecated `Update_IRLS` so they use the new `UpdateIRLS` directive class.

_Fix_ deprecation warnings from #1349


santisoler:
@domfournier, please ping me when this PR is ready for a review.

santisoler (commented):
@domfournier, please ping me when this PR is ready for a review.

domfournier:
Reaady for review @santisoler

simpeg/directives/_regularization.py
```diff
@@ -345,16 +345,19 @@ def start_irls(self):
...
 
-        if any(lin_precond_ind) and lin_precond_ind.index(True) < self_ind:
-            raise AssertionError(
-                "The directive 'UpdatePreconditioner' must be after Update_IRLS "
-                "in the directiveList"
-            )
+        if any(lin_precond_ind):
+            if lin_precond_ind.index(True) < self_ind:
+                raise AssertionError(
+                    "The directive 'UpdatePreconditioner' must be after Update_IRLS "
```
santisoler, line 358:
```suggestion
                    "The directive 'UpdatePreconditioner' must be after UpdateIRLS "
```

tests/base/test_directives.py
```diff
@@ -21,19 +21,23 @@ class directivesValidation(unittest.TestCase):
     def test_validation_pass(self):
         betaest = directives.BetaEstimate_ByEig()
 
-        IRLS = directives.Update_IRLS(f_min_change=1e-4, minGNiter=3, beta_tol=1e-2)
+        IRLS = directives.Update_IRLS(f_min_change=1e-4, beta_tol=1e-2)
```
santisoler, line 24: Is there a reason why we don't want to replace the `Update_IRLS` for `UpdateIRLS` in this test file? If there's no particular reason, I would apply the following change:

```suggestion
        IRLS = directives.UpdateIRLS(f_min_change=1e-4, beta_tol=1e-2)
```

PS: remember to update the other appearances of `Update_IRLS` in this file if you accept this suggestion.

tests/pf/test_mag_inversion_linear.py
```diff
@@ -115,28 +115,42 @@ def setUp(self):
...
-        #                grid=True, clim=(0, 0.02))
+        # midx = int(self.mesh.shape_cells[0] / 2)
+        # self.mesh.plot_slice(
+        #     self.actvMap * self.model,
+        #     ax=ax,
+        #     normal="Y",
+        #     ind=midx,
+        #     grid=True,
+        #     clim=(0, 0.02),
+        # )
```
santisoler, line 153: Is there a good reason to keep these commented lines? They are not run while running tests and therefore we are not keeping track that they are currently working as expected. What do you think?

santisoler (changes requested):
Thanks @domfournier for these updates. I left a few comments below, mostly minor things. Feel free to comment them.

When updating code throughout the entire library I find `git grep` super useful. For example, to list all the appearances of `Update_IRLS` you can run `git grep Update_IRLS` and get a list of the files where they appear along with a preview of the line of the occurrence.

tests/base/test_directives.py
domfournier: Well I duplicated the tests in the previous PR, and it depends if you want to keep testing the old class before we deprecate it completely...

tests/pf/test_mag_inversion_linear.py
domfournier: Yea for sure, I just kept them around because it is helpful to check the model when test fails. I am easy either way.

domfournier:
@santisoler See replies. I can do the mods depending on your answers.

tests/base/test_directives.py
santisoler: I see. I agree we should keep the tests for `Update_IRLS` until we remove it. But I don't understand why we want to remove the `minGNiter` in these tests. I think we should keep the tests for `Update_IRLS` untouched for now to ensure the directive keeps working as it used to.

tests/pf/test_mag_inversion_linear.py
santisoler: I see. What if we add these plots as another test function/method that we decorate with `@pytest.mark.skip`? This way we won't be execute them when running the tests, but at least the code is not going to be commented out.

What do you think?

tests/base/test_directives.py
domfournier: Yep good point. It's probably just a mix up. Reverting.

tests/pf/test_mag_inversion_linear.py
domfournier: Done

domfournier:
@santisoler Changes made. Good for another round.

domfournier:
Side note, it looks like codecov is including the test files in its measure?

jcapriot:
> Side note, it looks like codecov is including the test files in its measure? Like coverage of the codebase by the tests + test files themselves?

Yep! That way you can see that the tests you wrote are actually ran!

domfournier:
Thought it would be mostly obvious since coverage requires the tests to run, but whatever. The decrease in coverage here is due to the new chunks under @pytest.mark.skip

santisoler:
Looks good, @domfournier. And yes, don't worry about codecov failing. I just pushed some small changes restoring `minGNiter` in the tests that still use `Update_IRLS`. After CI passes I'm merging this one.

santisoler (approved):

PR #1480, opened 2024-06-06
Reorganize the maps submodule

santisoler:
#### Summary

Remove the `simpeg/maps.py` file and create a new `simpeg/maps` directory. Add private files for each category of map classes inside the new submodule. Make these classes public through the `simpeg/maps/__init__.py`.

#### PR Checklist
      to a Pull Request


santisoler:
@lheagy would you take a look at this. I'd like you to check if the organization of the map classes seems correct or if there's any modification that we should make. There's no need for you to go over every line of the new files, I just moved these classes to their corresponding private .py file. For a summary of the organization I proposed, check out the `maps/__init__.py`: https://github.com/simpeg/simpeg/blob/reorganize-maps/simpeg/maps/__init__.py

simpeg/maps/__init__.py
```diff
@@ -0,0 +1,35 @@
+from ._base import (
+    ComboMap,
+    IdentityMap,
+    LinearMap,
+    Projection,
+    SelfConsistentEffectiveMedium,
```
lheagy, line 6: This one is mesh agnostic and is pretty specialized, so I don't think it should be in the `base`

simpeg/maps/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    SphericalSystem,
+    SumMap,
+    Wires,
+)
+from ._clustering import PolynomialPetroClusterMap
+from ._injection import Mesh2Mesh, InjectActiveCells
+from ._mesh_agnostic import (
```
lheagy, line 13: rather than `mesh agnostic`, what do you think of `property_transforms` (e.g. these are one-to-one mappings that map from a property from one space to another)?

simpeg/maps/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+    BaseParametric,
+    ParametricBlock,
+    ParametricBlockInLayer,
+    ParametricCasingAndLayer,
+    ParametricCircleMap,
+    ParametricEllipsoid,
+    ParametricLayer,
+    ParametricPolyMap,
+    ParametricSplineMap,
+    TileMap,
```
lheagy, line 33: The TileMap is a bit of an odd one here, how about we put it with the `base`

lheagy (commented):
Thanks for taking care of this @santisoler! This will make the code much easier to work with. There are just a couple of suggestions for locations and the naming of `mesh_agnostic`. See what you think!

jcapriot:
I think this might be a good time to introduce a `BaseMapping` class that serves as the base class for maps (instead of `IdentityMap` that is used like that currently).

PR #1481, opened 2024-06-07
Split Azure Pipelines configuration into multiple files

santisoler:
Split the stages from `azure-pipelines.yml` into multiple files under the `.ci/azure` directory. Strip deployment to PyPI from the deployment of the docs into different stages. Add new job to upload each version merged to `main` to TestPyPI. Configure the `setuptools_scm` local version scheme so we can modify it before building the wheels for TestPyPI. Refactor the new docs pipeline: have a single build job and two jobs for deploying the docs to `simpeg-doctest`, one after a release and one for nightly builds. While building the sdist (before pushing to PyPI), rely on `build` to install build dependencies listed in `pyproject.toml` in isolation.


jcapriot:
Just FYI, you’ll need to always map an azure secret variable to an environment variable if you want to use it (look at how the current pipeline uses the GitHub token, and the twine variables)

.ci/azure/pypi.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - script: |
+          python setup.py sdist
+        displayName: "Create source distribution for simpeg"
+
+      - script: |
+          twine upload --repository testpypi dist/*
+        displayName: "Upload to Test PyPI"
+        env:
+          TWINE_USERNAME: $(test.twine.username)
```
jcapriot, line 30:
```suggestion
          TWINE_USERNAME: $(twine.username)
```

jcapriot: Or just write out `__token__`

santisoler: Ohh, only for the username. Got it.

jcapriot: The username is literally just `__token__` (it’s not even a secret variable on azure)

santisoler: I see. Thanks for the tip.

santisoler:
I think this is getting in shape. I was wondering if we could strip the doc building process from the tests. Is there any reason why would like to keep the `test_docs.py` file and still run them in CI? 

I would like to keep checking that docs build properly in CI, but we can trigger a docs build as part of the "Docs" stage  with a `html-noplot` target (so we don't build the gallery) on every PR.

Would love to hear thoughts about this.

jcapriot:
The `test_docs` runs it in a slightly different way that (I think) captures coverage.

jcapriot:
What I’d love to do is make use of the output of the the gallery tests somehow, store them as artifacts during the build, and then merge their artifacts during the doc build instead of re-running them sequential during the full build.

santisoler:
> What I’d love to do is make use of the output of the the gallery tests somehow, store them as artifacts during the build, and then merge their artifacts during the doc build instead of re-running them sequential during the full build.

If we only run them once (not including docs builds in the tests), then we wouldn't need to worry about storing results as artifacts. But I like your idea of reducing the CI times.

santisoler:
@jcapriot, would you like to take a look at this? I think this is ready to be merged. I'll leave the changes for the new deployment of the docs for another PR that I'll start after we merge this one.

jcapriot:
> @jcapriot, would you like to take a look at this? I think this is ready to be merged. I'll leave the changes for the new deployment of the docs for another PR that I'll start after we merge this one.

I will, also can you take a look at #1482, that should be ready as well. It's just a matter of which one is easier to pull in first, then deal with the conflicts. I suspect #1482 will probably come in first?

.ci/azure/pypi.yml
```diff
@@ -0,0 +1,72 @@
+jobs:
```
jcapriot, line 1: Should merge these two jobs into one and then make the two deploy steps conditional

jcapriot (changes requested):
Just one suggestion to reduce duplications in the pypi deploy file.

.ci/azure/old-docs.yml
```diff
@@ -0,0 +1,93 @@
...
+          pip install -e .
+        displayName: Build package
+
+      - script: |
+          source "${HOME}/conda/etc/profile.d/conda.sh"
+          conda activate simpeg-test
+          python -c "import simpeg; print(simpeg.__version__)"
+        displayName: Check SimPEG version
+
+      # DON'T PUSH TO PYPI HERE (we have another stage for that)
```
jcapriot, line 57: can clean these up too

.ci/azure/pypi.yml
santisoler: I've merged the two deployment jobs into a single one with conditional steps. What I also did is split the build and the deployment into different jobs. And now we can build on every push and check if the build is correct.

.ci/azure/pypi.yml
```diff
@@ -0,0 +1,83 @@
...
+        fetchTags: true
+        displayName: "Checkout repository (including tags)"
+
+      - task: UsePythonVersion@0
+        inputs:
+          versionSpec: "3.9"
+        displayName: "Setup Python"
+
+      - bash: |
+          pip install setuptools>=61 setuptools_scm wheel twine
```
jcapriot, line 19: I think you'll also need `build` here

santisoler: Right!

santisoler:
@jcapriot, I think this is ready to be merged. Would you take a second look at it?

.ci/azure/pypi.yml
```diff
@@ -0,0 +1,83 @@
...
+      - bash: |
+          # Change setuptools-scm local_scheme to "no-local-version" so the
+          # local part of the version isn't included, making the version string
+          # compatible with Test PyPI. Only do this when building for TestPyPI.
+          sed --in-place 's/node-and-date/no-local-version/g' pyproject.toml
+        condition: not(startsWith(variables['build.sourceBranch'], 'refs/tags/'))
+        displayName: "Configure local_scheme (except on release)"
+
+      - bash: |
+          python -m build --sdist .
```
santisoler, line 31: The biggest change I made after updating this branch once the `pyproject.toml` was merged was to change how we build the sdist. Instead of installing the build dependencies (and pin the versions here in the yml), I just install `build` and let it install the required build dependencies in isolation. This also solved an issue I saw when trying to push to TestPyPI, where the sdist didn't have the right simpeg version. I think it was related to incompatible versions of `setuptools` and `setuptools_scm`.

.ci/azure/old-docs.yml
```diff
@@ -0,0 +1,65 @@
...
+          GH_NAME: $(gh.name)
+          GH_EMAIL: $(gh.email)
+
+      - bash: echo "##vso[task.prependpath]$CONDA/bin"
+        displayName: Add conda to PATH
+
+      - bash: .ci/azure/setup_env.sh
+        displayName: Setup SimPEG environment
+
+      - script: |
```
jcapriot, line 33: Should remove this step since it uploads to pypi in the other file.

jcapriot (changes requested):
Just a little bit more cleanup.

.ci/azure/old-docs.yml
santisoler: Right! I forgot to update the new version of the `old-docs.yml` as well 🤦🏼‍♂️

Issue #1356, opened 2024-02-09
DOC: Solvers


jcapriot:
### Issue with current documentation:

We do not discuss choosing solvers at all in the documentation.

### Idea or request for content:

We should create a page that describes choosing solvers for the PDE simulations, and discuss some of the different types of solvers that might be available.

I imagine this would probably go in the User Guides section.

PR #1482, opened 2024-06-08
Pyproject.toml

jcapriot:
#### Summary
Moves simpeg to a pyproject.toml based project

#### PR Checklist
      to a Pull Request

removes setup.py in favor of a pyproject.toml file, and moves as much tool configuration options to that file as possible.

#### Additional information
also does a bit of repository cleanup.
* updates mailmap
* removes extra bits from makefile
* moves common ci commands to their own folder
* stop downloading `mambaforge` (just use azure's own condo install which in turn uses libmamba)

Other things to consider in future PRs:
* [] Update the doc deploy steps to be shell scripts.


jcapriot:
I'd like to move the deployment doc building to actual shell scripts as well, but we can do that again. This is also relevant to #1481.

.ci/requirements_style.txt
jcapriot, line 1: I don't like having to have this file still hanging around. However we need work around the fact that there is no way for pip to install only the dependencies. There is a decent amount of talk about this as an option, but still not a set way to do it

I think an option in the meantime is to create a script that installs just these dependencies by parsing `pyproject.toml` to put the style requirements into a `requirements_style.txt` file.

.ci/environment_test.yml
jcapriot, line 1: We should really consider pinning versions of all of these packages so we have a consistent testing environment. Potentially maintain a different file for each python version? (This would also then allow us to cache the `conda` environment on azure pipelines to save a few minutes on each individual build.

.ci/requirements_style.txt
santisoler: Something like [`dependente`](https://github.com/fatiando/dependente)?

jcapriot: I was trying to avoid having to download anything outside the standard library to get it working. It's simple enough here to parse the pyproject with tomllib to get what we need. (I've updated this already)

pyproject.toml
```diff
@@ -0,0 +1,248 @@
...
+Repository = 'http://github.com/simpeg/simpeg.git'
+
+[project.optional-dependencies]
+dask = ["dask", "zarr", "fsspec>=0.3.3"]
+choclo = ["choclo"]
+reporting = ["scooby"]
+plotting = ["plotly"]
+all = [
+    "simpeg[dask,choclo,plotting,reporting]"
+] # all optional *runtime* dependenies (not related to development)
```
santisoler, line 60:
```suggestion
] # all optional *runtime* dependencies (not related to development)
```

jcapriot:
> Looking great! I just fixed a small typo.
> 
> My only concern is about moving the environment file to `.ci`. I personally rely on it as an easy way to have a development environment every time I work on simpeg. And we also instruct contributors to set up their environment by creating one from this file (https://docs.simpeg.xyz/content/getting_started/contributing/setting-up-environment.html).
> 
> I would keep an `environment.yml` file in the root of the directory. We can also keep the `.ci/environment_test.yml` if we want to avoid installing unneeded packages in Azure.
> 
> Regardless of the decision we take, I think we should make sure that the getting started docs are updated after these changes.

Yeah I wasn't completely sure on it either. My hope was to avoid having a ton of duplicated "requirement" like files that we have to maintain every time a dependency is updated.

jcapriot:
In general I think we need to consider several questions if we do want to have some `environment.yml` files:

1) If we have an `environment.yml` file, what should be in it? SimPEG requirements? SimPEG dev requirements? Testing requirements? Should these be multiple files? Do we want to have multiple `environment.yml` for each of these cases? (i.e. `environment.yml`-> SimPEG running requirements, `environment_dev.yml` -> a standard development environment for SimPEG?, `environment_test.yml` the environment used for running tests on CI.)

3) Where should we put them (especially if there are multiple)? Probably in an `envs/` directory is my guess.

4) Is there anything we can do to make sure they stay consistent with the requirements in `pyproject.toml`? I suspect we can setup a GitHub action on pull requests to make sure all of the requirements are consistent in all of the places they would be specified (multiple `environment.yml` files, `.pre-commit-config.yaml`, etc.)? (i.e. if an equality is specified in `pyproject.toml` it is consistent throughout. If there are minimums in `pyproject.toml` make sure the version specified in `environment.ymls` are consistent?

santisoler:
> In general I think we need to consider several questions if we do want to have some `environment.yml` files:
> 
>     1. If we have an `environment.yml` file, what should be in it? SimPEG requirements? SimPEG dev requirements? Testing requirements? Should these be multiple files? Do we want to have multiple `environment.yml` for each of these cases? (i.e. `environment.yml`-> SimPEG running requirements, `environment_dev.yml` -> a standard development environment for SimPEG?, `environment_test.yml` the environment used for running tests on CI.)
> 
>     2. Where should we put them (especially if there are multiple)? Probably in an `envs/` directory is my guess.

I would just keep a single `environment.yml` file so contributors can have a complete environment with all the requirements perform tasks: check style, run tests, build docs, build sources. Having multiple environments for each one of these tasks could be cumbersome. I would go with something like [pandas have](https://github.com/pandas-dev/pandas/blob/main/environment.yml): a "global" `environment.yml` file in the root of the directory.

Besides that main `environment.yml`, we could have a particular file for running the tests on CI. I'm fine with having that one (or those ones if we need more than one of them) in `.ci`. Having them in `envs` also sounds good. I don't have strong opinions for those.

>     3. Is there anything we can do to make sure they stay consistent with the requirements in `pyproject.toml`? I suspect we can setup a GitHub action on pull requests to make sure all of the requirements are consistent in all of the places they would be specified (multiple `environment.yml` files, `.pre-commit-config.yaml`, etc.)? (i.e. if an equality is specified in `pyproject.toml` it is consistent throughout. If there are minimums in `pyproject.toml` make sure the version specified in `environment.ymls` are consistent?

That sounds like a nice idea: a checker for versions. I personally would like to be able to easily run it locally, through the Makefile for example, and then Actions just run that target/script.

environment.yml
```diff
@@ -2,48 +2,57 @@ name: simpeg-test
...
   - conda-forge
 dependencies:
+# dependencies
   - numpy>=1.20
   - scipy>=1.8
   - scikit-learn>=1.2
-  - pymatsolver>=0.2
-  - matplotlib
+  - pymatsolver-base>=0.2
+  - matplotlib-base
```
jcapriot, line 10: make these dependent upon base version of the package.
* `matplotlib-base` avoids downloading the back QT package. This is also what would be installed if you did `conda install simpeg`.
* `pymatsolver-base`  makes it platform independent. `pymatsolver` on `conda` explicitly includes `pydiso` as a requirement due to legacy reasons which is not available on all platforms.

environment.yml
```diff
@@ -2,48 +2,57 @@ name: simpeg-test
...
   - discretize>=0.10
   - geoana>=0.5.0
   - empymod>=2.0.0
-  - setuptools_scm
   - pandas
+
+# solver
+# uncomment the next line if you are on an intel platform
+#  - pydiso # if on intel pc
```
jcapriot, line 18: Unfortunately there is not way to specify a conditional dependance in an `environment.yml` file. So this line is commented out with some notes to keep it platform independent.

Hopefully someone sees this before using the environment.yml file. #1356

environment.yml
```diff
@@ -2,48 +2,57 @@ name: simpeg-test
...
+  - flake8-pyproject==1.2.3
   - flake8-bugbear==23.12.2
   - flake8-builtins==2.2.0
   - flake8-mutable==1.2.0
   - flake8-rst-docstrings==0.3.0
   - flake8-docstrings==1.7.0
+
+# recommended
+  - jupyter-lab
+  - pyvista
```
jcapriot, line 58: These aren't necessarily required by anything in `simpeg`, but it feels like a good thing to add?

santisoler (commented):
Looking great @jcapriot! I just left two comments below.

I created the new environment (replacing `jupyter-lab` for `jupyter`) and ran `make flake`. I got the following error:
```
flake8 --version
7.0.0 (Flake8-pyproject: 1.2.3, flake8-bugbear: 23.12.2, flake8-builtins: 2.2.0,
flake8-docstrings: 1.7.0, flake8-mutable: 1.2.0, flake8-rst-docstrings: 0.3.0, mccabe:
0.7.0, pycodestyle: 2.11.1, pyflakes: 3.2.0) CPython 3.12.3 on Linux
flake8  simpeg examples tutorials tests
simpeg/meta/dask_sim.py:250:77: E225 missing whitespace around operator
make: *** [Makefile:29: flake] Error 1
```

pre-commit is also catching it (if I tried to commit a minor change to that line). I'm curious why the style checks on CI are not catching it. Do you experience the same locally?

jcapriot:
> Looking great @jcapriot! I just left two comments below.
> 
> I created the new environment (replacing `jupyter-lab` for `jupyter`) and ran `make flake`. I got the following error:
> 
> ```
> flake8 --version
> 7.0.0 (Flake8-pyproject: 1.2.3, flake8-bugbear: 23.12.2, flake8-builtins: 2.2.0,
> flake8-docstrings: 1.7.0, flake8-mutable: 1.2.0, flake8-rst-docstrings: 0.3.0, mccabe:
> 0.7.0, pycodestyle: 2.11.1, pyflakes: 3.2.0) CPython 3.12.3 on Linux
> flake8  simpeg examples tutorials tests
> simpeg/meta/dask_sim.py:250:77: E225 missing whitespace around operator
> make: *** [Makefile:29: flake] Error 1
> ```
> 
> pre-commit is also catching it (if I tried to commit a minor change to that line). I'm curious why the style checks on CI are not catching it. Do you experience the same locally?

I didn't see anything like that locally. Maybe it's a difference due to python 3.11 and 3.12? That's the only difference I see on the CI tested version.

environment.yml
```diff
@@ -1,49 +1,58 @@
...
   - flake8==7.0.0
+  - flake8-pyproject==1.2.3
   - flake8-bugbear==23.12.2
   - flake8-builtins==2.2.0
   - flake8-mutable==1.2.0
   - flake8-rst-docstrings==0.3.0
   - flake8-docstrings==1.7.0
+
+# recommended
+  - jupyter-lab
```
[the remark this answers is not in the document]
jcapriot: that's fine

jcapriot:
Okay, once CI passes I'm going to merge this.

PR #1486, opened 2024-06-12
Fix contour colors in gravity plot in User Guide

williamjsdavis:
#### Summary
Correcting contour colors in User Guide: Gravity: plot 1a

**Squash and Merge Summary**

Correcting contour colors in a gravity plot in the User Guide for a gravity plot: correctly use the color limits both in the contours and while building the custom color bar.

#### Issue found

In the last figure of the User Guide: Gravity: plot 1a, the plotted contour colors do not correspond to the correct gravity anomaly value indicated in the colorbar on the right.

Relevant section:
https://docs.simpeg.xyz/content/tutorials/03-gravity/plot_1a_gravity_anomaly.html#simulation-gravity-anomaly-data-on-tensor-mesh

### Identifying the problem

There appears to me a mismatch between the plotted contour colors of gravity anomaly and the colorbar on the right.

Currently, the figure displays like this:
gravity-original

However it should plot like this:
gravity-fixed

The mismatch can be viewed in a more noticeable form if we change the model to only have a positive density anomaly:

altered-density

In this new scenario, the original plotting script in the documentation produces a highly misleading figure:

altered-gravity-original

Whereas it should actually plot like this:

altered-gravity-fixed

#### Diagnosing the problem

The error arrises because the color limits were mistakenly left out of the call to `plot2Ddata`, in the following line:

https://github.com/simpeg/simpeg/blob/cbd200b6774b201f2e6f60cda8b35b4118d512cd/tutorials/03-gravity/plot_1a_gravity_anomaly.py#L218

#### My fix for the problem

The corrected call to `plot2Ddata` should be:

```python
v_max = np.max(np.abs(dpred))
plot2Ddata(receiver_list[0].locations, dpred, clim=(-v_max, v_max), ax=ax1, contourOpts={"cmap": "bwr"})
```

#### Reference issue

No issue to reference.

My change fixes the plotted contours in the last figure of the User Guide: Gravity: plot 1a. In the previous implementation the contour colors did not correspond to the correct gravity anomaly value indicated in the colorbar on the right. In my new implementation, the contour colors now correspond to the correct gravity anomaly value indicated in the colorbar on the right.

#### Additional information

There may be more instances in the documentation where this error is present. My change here is the only one I've found so far.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/tutorials/03-gravity/plot_1a_gravity_anomaly.py
+++ b/tutorials/03-gravity/plot_1a_gravity_anomaly.py
@@ -214,14 +214,16 @@ dpred = simulation.dpred(model)
 # Plot
 fig = plt.figure(figsize=(7, 5))
 
+v_max = np.max(np.abs(dpred))
+
 ax1 = fig.add_axes([0.1, 0.1, 0.75, 0.85])
-plot2Ddata(receiver_list[0].locations, dpred, ax=ax1, contourOpts={"cmap": "bwr"})
+plot2Ddata(receiver_list[0].locations, dpred, clim=(-v_max,v_max), ax=ax1, contourOpts={"cmap": "bwr"})
 ax1.set_title("Gravity Anomaly (Z-component)")
 ax1.set_xlabel("x (m)")
 ax1.set_ylabel("y (m)")
 
 ax2 = fig.add_axes([0.82, 0.1, 0.03, 0.85])
-norm = mpl.colors.Normalize(vmin=-np.max(np.abs(dpred)), vmax=np.max(np.abs(dpred)))
+norm = mpl.colors.Normalize(vmin=-v_max, vmax=v_max)
 cbar = mpl.colorbar.ColorbarBase(
     ax2, norm=norm, orientation="vertical", cmap=mpl.cm.bwr, format="%.1e"
 )
--- a/tutorials/03-gravity/plot_1a_gravity_anomaly.py
+++ b/tutorials/03-gravity/plot_1a_gravity_anomaly.py
@@ -217,7 +217,13 @@ fig = plt.figure(figsize=(7, 5))
 v_max = np.max(np.abs(dpred))
 
 ax1 = fig.add_axes([0.1, 0.1, 0.75, 0.85])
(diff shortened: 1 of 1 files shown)


williamjsdavis:
Apologies for taking a few attempts to correct those formatting errors; I thought I had configured my linter correctly but apparently not! @simpeg/simpeg-developers this is ready for review

santisoler:
Hi @williamjsdavis! Nice catch! Thanks a lot for reporting this and also providing a solution to it.

Your code solves this perfectly so I'm merging this :rocket:

Just as a comment to it, I don't quite like the initial approach of building a new colorbar for the plot. I think that's a potential source of errors (as you pointed out here) and I think for the long run we could simplify this plot snippet to something like:

```python
dpred = simulation.dpred(model)

# Plot
maxabs = np.max(np.abs(dpred))
fig, ax = plt.subplots(figsize=(7, 5))

contours, _ = plot2Ddata(
    receiver_list[0].locations,
    dpred,
    ax=ax,
    contourOpts={"cmap": "bwr"},
    clim=(-maxabs, maxabs),
)
ax.set_title("Gravity Anomaly (Z-component)")
ax.set_xlabel("x (m)")
ax.set_ylabel("y (m)")
plt.colorbar(contours, ax=ax, label="mGal")

plt.show()
```

But that's for a future PR.

santisoler (approved):

Wafer233:
Sorry, I am new for simpeg. But I have a small question for the plot in this tutorial.
Since the density of block is -0.2, which is smaller than the background, I think the gravity anomaly supposed be lower than 0, which is totally inverse with the plot. 
And by reading your comments, the plot since to be right, I would appreciate it if you can tell me why. T.T

Wafer233:
> Sorry, I am new for simpeg. But I have a small question for the plot in this tutorial. Since the density of block is -0.2, which is smaller than the background, I think the gravity anomaly supposed be lower than 0, which is totally inverse with the plot. And by reading your comments, the plot since to be right, I would appreciate it if you can tell me why. T.T

fig 1
I think maybe it would display like this.

jcapriot:
`SimPEG` consistently uses a right handed coordinate system with z positive upwards.

PR #1489, opened 2024-06-18
Experiment with removing test_docs

santisoler:
#### Summary

Experiment with removing `test_docs` from the test suite and always build the
docs in the Docs stage (in Azure). Run a `html-noplot` and a `linkcheck-noplot`
on every PR.

#### PR Checklist
      to a Pull Request


jcapriot:
What is the goal in not running them during “test”. From what I see, this is just moving when the docs are tested to after the other tests are completed.

santisoler:
From a conceptual perspective, I don't like including docs building as part of the test suite. From a practical point of view, since we already have a Docs stage, we can use it as the single place where we build the docs and upload the built docs as artifact.

Moreover, `test_docs` runs Python code that calls Sphinx through its API, but when we build the docs for publishing we use the targets in `docs/Makefile`. I think we can stick with a single way of checking that docs build properly.

Lastly, I noticed that `test_docs` don't fail if the linkcheck fails. Running `make -C docs linkcheck-html` does fail: https://dev.azure.com/simpeg/simpeg/_build/results?buildId=4433&view=logs&j=f5e0b957-a0bb-542d-63f3-48c18f3b89ce&t=6b3e234a-fb10-5df9-0d8b-35a3152b01c6.
So, `test_docs` is not currently ensuring that all links in the docs work properly.

Lastly, requiring test to pass before building the docs would reduce the computation load we put in Azure. If test are not passing there's no much sense on building the docs.

I started this PR as a prototype for these ideas. We can continue the discussion.

Issue #1429, opened 2024-04-17
ENH: Make explicit the Public and Private members of the simulation classes.


jcapriot:
### Proposed new feature or change:

I want to use this as a building block to Identify the bare bones items that should (and therefore shouldn't be) public methods/properties of SimPEG classes. For now I think it would be most beneficial to focus on `Simulation` classes, but should certainly be thought about for every other class as well.

My first thoughts for public items would be:

* `BaseSimulation`:
  * methods:
        * `dpred(m, f=None) -> ndarray`
        * `fields(m) -> fields_object`
        * `Jvec(m, v, f=None) -> ndarray`
        * `Jtvec(m, v, f=None)-> ndarray`
        * `make_synthetic_data(m, relative_error=0.5, noise_floor=0, f=None, add_noise=False, random_seed=None) -> Data object`
        * `getJ(m, f=None) -> 2D array_interface thing (could be sparse, as long as it supports the `array_interface`)
        * `getJtJdiag(m, f=None, W=None)` ndarray
  * properties:
        * `survey -> Survey`
        * `model -> ndarray`
Another thing to consider, should we set `m=None` as default for the `dpred`, `fields`, and `make_synthetic_data` forward arguments, so it is more explicit they can be ran without setting a model? It already works to pass `m=None` to them now in these cases, and it already errors when `m=None` and a physical property map has been set.

* PhysicaPropertySimulations(potentially consider be reformatted as mixin classes, not directly inheriting PDE simulations).
    * properties:
        * the physical property `sigma`/`conductivity` (once it's renamed to not be the confusing symbol)
        * the physical property map `sigmaMap`/`conductivity_map`

* BasePDESimulation:
    * methods:
        * `getA(m, *args) -> Sparse Matrix` (maybe it could be `getA(m, src)` or `getA(m, src, *args)` ... usually though when A is dependent upon something besides the model, it is specific to that simulation each simulation in addition (source frequency, spatial wavenumber. The signature `getA(m, *args)` wouldn't be used, but just as a template)
        * `getAinv(m, *args)-> Solver object` (should take same arguments as `getA` whatever those will be)
        * `getRHS(m, source)-> ndarray`
     * properties:
        * `mesh`

Should probably make this class be the one to add the (private! ) methods for mass matrices for any `props.Properties` (using a subclass of the props.HasModel` metaclass.

* LinearSimulations:
  * methods:
[...]

PR #1511, opened 2024-07-31
Default solver

jcapriot:
#### Summary
Adds methodology to explicitly set a default solver for simpeg to use, and explicitly lists a priority to try to import them.

#### PR Checklist
      to a Pull Request

There are too many explicit uses of `Pardiso` in the simpeg tests and examples making them unrunnable on systems that do not implement `Pardiso`.

It is also unclear to the user which solver is chosen when using default, thus if the user implicitly uses a default solver in a simulation, or asks for the default solver, a `DefaultSolverWarning` is issued saying explicitly what solver is used. This is a new Warning class that the end user can choose to silence if needed.

This also removes the re-implemented Solver wrappers and instead rely on `pymatsolver`'s versions of those wrappers.

#### Additional information
Plenty of room to adjust the implementation here (thus the draft PR). Once functionality is finalized I'll add tests.


jcapriot:
There's a lot of file changes, mostly just removing the `try ... except ...` imports of `Pardiso`. The relevant implementation is in `simpeg/utils/solver_utils.py` and in `simpeg/simulation.py`

santisoler (commented):
I like this approach @jcapriot! And I agree with the priority list for solvers. I left one minor correction to one of the docstrings.

Now that I see the changes you made to the `LinearSimulation`s, I would like to retake the conversation about removing the `solver` from the `BaseSimulation` since not all of them need to have one. We can work on that in a different PR though, but I'd like to hear your thoughts about it.

jcapriot:
> I like this approach @jcapriot! And I agree with the priority list for solvers. I left one minor correction to one of the docstrings.
> 
> Now that I see the changes you made to the `LinearSimulation`s, I would like to retake the conversation about removing the `solver` from the `BaseSimulation` since not all of them need to have one. We can work on that in a different PR though, but I'd like to hear your thoughts about it.

I completely agree that `solver` and `solver_opts` should not be on the `BaseSimulation`. I think it should be on the `BasePDESimulation`. (Relevant to issue: #1429)

tests/em/nsem/inversion/test_complex_resistivity.py
```diff
@@ -32,7 +32,7 @@ def setUp(self):
 
         # create background conductivity model
         sigma_back = 1e-2
-        sigma_background = np.zeros(mesh.nC) * sigma_back
+        sigma_background = np.ones(mesh.nC) * sigma_back
```
jcapriot, line 35: I just wanted to put an extra emphasis on this change here...

The original tests were using a 0 conductivity background model... I'm surprised they passed before this. The solver updates identified this flaw. Since fixing this to actually use the intended background conductivity, *most* of the tests still pass, except for the `phase_xx` derivatives and `phase_yy` derivatives. I'm not sure why those fail, yet all other tests pass (including derivatives for other phase components, all the amplitude components, and all of the real/imaginary components) so I suspect it is due to the model producing phases that are near the wraparound of the complex angle argument. I don't believe it's warranted to try to fix that in this PR, and it might even be fixed already by #1454, so I've marked them as expected failures for now and I'll open an issue regarding these tests.

jcapriot (commented):
Just flagging this newly caught error in the nsem tests for visibility

jcapriot:
@simpeg/simpeg-developers This is ready to move in!

santisoler (approved):
Looks great. I love the automatic behavior and the removal of those `try` and `except` in examples and tutorials.

My two cents on this would be to think about making the new module more discoverable. If a user wants to get the `get_default_solver` function needs to import:
```python
from simpeg.utils.solver_utils import get_default_solver
```
We need to navigate through two modules whose names are kind of redundant: e.g. `utils`, `solver_utils`, `get_default_solver`.

So, what if we expose this module a little bit  more. I tend to conceive the `utils` modules as a place to store functions that are not a core portion of the package or functions that help some tedious tasks for other functions or user-facing code. Since solvers are an important piece of the numerical simulations and the inversion process, maybe it worth creating a `simpeg.solver` module where the two new functions live along with all the available solvers. So we could import stuff from it:

```python
from simpeg.solver import get_default_solver
```

What do you think?

Regardless of our thoughts on that, I think this PR is ready and I think we can merge it right away.

jcapriot:
I think that's fine to move these into a new `simpeg.solvers` module! (But let's do it in another PR eventually)

tests/em/nsem/inversion/test_complex_resistivity.py
lheagy: Nice catch @jcapriot! 

I suspect that we might not have a strong enough 3D structure to be generating large on-diagonal components, but we can handle that elsewhere

PR #1513, opened 2024-08-07
Fix misuse of the `requires` decorator in `code_utils.py`

prisae:
Use the `discretize`'s `requires` to decorate `mem_profile_class`, instead of `simpeg.utils.code_utils.requires`, since the former is intended to be used for modules, while the latter should be used for attributes.

I completely missed in #1104 , that

- `simpeg.utils.code_utils.requires` for attributes is, and not like
- `discretize.utils.code_utils.requires` for modules.

Fixing this by using the discretize version for the `mem_profile_class` function.

@simpeg/simpeg-developers - ready for review

### DIFF
--- a/simpeg/utils/code_utils.py
+++ b/simpeg/utils/code_utils.py
@@ -5,6 +5,7 @@ from functools import wraps
 import warnings
 
 from discretize.utils import as_array_n_by_dim  # noqa: F401
+from discretize.utils import requires as module_requires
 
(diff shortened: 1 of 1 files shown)


santisoler (approved):
Nice catch @prisae! I'll merge after CI pass. Codecov will probably fail since the `except` branch in the `try` won't be tested, but it should be fine.

PR #1514, opened 2024-08-08
Make `pandas` & `scikit-learn` optional dependencies

prisae:
#### Summary

Set `sklearn` and `pandas` as optional dependencies in `pyproject.toml`. Make use of `discretize`'s `requires` utils function to decorate the functions and methods that require either `pandas` or `scikit-learn` to be installed, specifically in `simpeg.electromagnetics.static.resisitivity.IO` and PGI-related classes.

`sklearn` and `pandas` are comparably heavy dependency, but they are only used in one specific place each:

- `pandas`: `simpeg/electromagnetics/static/resistivity/IODC.py`;
- `sklearn`: `simpeg/utils/pgi_utils.py`.

As such, they are potentially only used by a fraction of the SimPEG users.

This PR implements them as soft dependencies, in the same way as `matplotlib` is a soft dependency in `discretize`.

#### PR Checklist
      to a Pull Request

An installation of SimPEG will not install `sklearn` nor `pandas`. If a function/class is called that requires one of these packages, the following message is given:
```
Missing dependencies: ['sklearn'].
Not running `__init__`.
```

If we want to make this message more explicit (e.g., with instructions how to install it), we'll have to adjust the `requires` function in `discretize`.

#### Comments
- I did not add any tests - I guess the functionality of `requires` is tested in `discretize`. However, I can add some if wanted.
- Not sure if it should be mentioned at the appropriate places in the documentation.

### DIFF
--- a/.ci/environment_test.yml
+++ b/.ci/environment_test.yml
@@ -4,13 +4,11 @@ channels:
 dependencies:
   - numpy>=1.20
   - scipy>=1.8
-  - scikit-learn>=1.2
   - pymatsolver-base>=0.2
   - matplotlib-base
   - discretize>=0.10
   - geoana>=0.5.0
   - empymod>=2.0.0
-  - pandas
 
 # Solver
   - pydiso
@@ -22,6 +20,8 @@ dependencies:
   - choclo
   - scooby
   - plotly
+  - scikit-learn>=1.2
+  - pandas
 
 # documentation building
   - sphinx
--- a/environment.yml
+++ b/environment.yml
@@ -6,13 +6,11 @@ dependencies:
   - python=3.11
   - numpy>=1.20
   - scipy>=1.8
-  - scikit-learn>=1.2
   - pymatsolver-base>=0.2
   - matplotlib-base
   - discretize>=0.10
   - geoana>=0.5.0
   - empymod>=2.0.0
-  - pandas
 
 # solver
 # uncomment the next line if you are on an intel platform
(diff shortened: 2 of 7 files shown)


santisoler:
Thanks @prisae for opening this. I think the PR is in good shape to be merged.

One comment I have is that it would be better if the `requires` decorator throws an `ImportError` or `ModuleNotFoundError` if the optional dependency is missing. I think that could improve the user experience. For example, let's say you have a preexisting code that was saving the `IO` as a csv with `write_to_csv`. If in your new environment you forgot to install `pandas`, then the same code will run without failing, although the csv file won't be created and the user will only get one printed message in the stdout. This comment is outside the scope of this PR, but maybe this is something we could add to the decorator in discretize. What do you think?

One caveat of having errors being raised, is that we should test if every module can be imported without having optional dependencies installed. Is not a big deal, but I would like to have that automatically running before the conda-forge tests run a similar one after a release.

prisae:
One reason against throwing an error is that errors frighten the un-experienced user, and they are often not able to properly read them (in my experience). This is why I opted for an informative message instead of an error. But surely something that can be discussed.

santisoler:
Makes sense! I'm merging this now. We can keep discussing that in discretize. Thanks again for this @prisae!

prisae:
Thanks @santisoler!
Are you in charge of the conda-recipe? This could be adjusted accordingly.

santisoler:
Yes, I can push to the feedstock repo. I'll update the recipe after the next release.

PR #1517, opened 2024-08-09
Pass `rtol` to SciPy solvers for SciPy>=1.12

### LINKED ISSUE #1516: BUG: SciPy 1.14 removed the `tol` argument of `scipy.sparse.linalg.***`
### Describe the issue:

SciPy 1.14 removed the `tol` argument of `scipy.sparse.linalg.{bcg,bicstab,cg,cgs,gcrotmk, mres,lgmres,minres,qmr,tfqmr}` in favour of `rtol`. Furthermore, the default value of `atol` for these functions has changed to `0.0`. (It was deprecated in SciPy 1.12.)

This will need some changes in the code base.

```python
Any code example that uses a scipy.sparse.linalg-solver.
```

### Error message:

```shell
/simpeg/simpeg/optimization.py:1072: UserWarning: tol is not a valid keyword for cg and will be ignored
  Hinv = SolverICG(
```
```


### Runtime information:

SciPy >= 1.14

### Context for the issue:

_No response_

prisae:
#### Summary

Pass `rtol` or `tol` to SciPy solvers depending on the available version of SciPy: for `scipy>=1.12` use `rtol`, otherwise use `tol`. SciPy 1.12 changed `tol` to `rtol` for `scipy.sparse.linalg...`, and removed `tol` in SciPy 1.14.

See #1516

Just a simple attempt, maybe you have other ideas, feel free to close this one in favour of another solution!

#### PR Checklist

@simpeg/simpeg-developers - ready for review

#### Reference issue
Closes #1516

### DIFF
--- a/.ci/environment_test.yml
+++ b/.ci/environment_test.yml
@@ -11,7 +11,6 @@ dependencies:
   - geoana>=0.5.0
   - empymod>=2.0.0
   - pandas
-  - packaging
 
 # Solver
   - pydiso
--- a/.ci/environment_test.yml
+++ b/.ci/environment_test.yml
@@ -11,6 +11,7 @@ dependencies:
   - geoana>=0.5.0
   - empymod>=2.0.0
   - pandas
+  - packaging
 
 # Solver
   - pydiso
--- a/environment.yml
+++ b/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - geoana>=0.5.0
   - empymod>=2.0.0
   - pandas
+  - packaging
 
 # solver
 # uncomment the next line if you are on an intel platform
--- a/environment.yml
+++ b/environment.yml
@@ -13,7 +13,6 @@ dependencies:
   - geoana>=0.5.0
   - empymod>=2.0.0
   - pandas
-  - packaging
 
 # solver
 # uncomment the next line if you are on an intel platform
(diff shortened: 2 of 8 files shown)


santisoler:
Thanks @prisae for opening this. I think you approach should work. I just pushed a few changes, feel free to comment on those or just push corrections to them:

1. I used `packaging.version.parse()` to check if the installed version of scipy is greater or equal to 1.12. I think this is a more robust check than comparing only to the values for the minor release.
2. I applied some changes to the code style: uppercase the global variable for the scipy version flag, define the dictionary `inp` only once and use a predefined variable for the tolerance key.

Let me know what do you think!

cc @jcapriot. Would you agree on adding `packaging` as a required dependency?

prisae:
Not sure if it is worth to add a dependency for an issue that will go away soon again (as soon as the minimum SciPy version is. But no hard feelings in this respect, either is fine!

jcapriot:
I honestly wouldn’t worry much about the formatting or details on this, as it will all get removed in #1511 as these will rely on the pymatsolver versions.

prisae:
Awesome! (This speaks against adding an additional dependency, IMHO.)

Do you think #1511 will come soon enough so we abandon this one, or should we bring this in as a temporary quick fix?

santisoler:
Makes sense that including `packaging` as a dependency is a little bit overkill for this quickfix. I just replaced it by a simple snippet that compares the major and minor parts of the versions.

I think we could merge this so we have it until we start using a newer version of `pymatsolver`.

prisae:
LGTM, thanks @santisoler!

Issue #1420, opened 2024-04-10
Discussion regarding `kwargs`


santisoler:
I think it would be nice to continue the discussion regarding the usage of `**kwargs` in class constructors and their role in inheritance.

In past meetings I've expressed my desire to avoid the usage of `**kwargs` arguments in class constructors unless we find a very good reason to do so. I think there are a few different reasons to do so:

1. Allowing `kwargs` on constructors increases complexity of class abstractions. Passing arguments as kwargs through the inheritance tree can have hard-to-track behaviors that depend mostly on the order in which those arguments are being passed. I've experienced these behaviors while refactoring the base classes in #1241.
1. The `kwargs` could possible lead to misbehavior. 
    - If the passed arguments are ignored, the class will behave in an unexpected way without complain, leading to a silent misbehavior (as seen in https://github.com/simpeg/user-tutorials/issues/12).
    - If the class assigns every `kwarg` as attribute, then the class is a container of everything, breaking the sense of abstraction (as the `BaseObjectiveFunction` was doing before https://github.com/simpeg/simpeg/pull/1241/files#diff-57dc55ee40104c9badf8c244fea415e8d563ccf942370e706505f4a155dc2cc3L36)  
1. Interfacing functions with `kwargs` is hard. Users cannot rely on autocompletion or type checking tools when passing arguments through `kwargs`. Moreover, as a developer, the possible arguments that could be passed to `kwargs` are usually overlooked, generating incomplete docstrings. Moreover, asking required arguments to be passed as `kwargs` can create objects that will fail down the line. A good example of this are some simulations like [gravity.Simulation3DIntegral](http://docs.simpeg.xyz/content/api/generated/SimPEG.potential_fields.gravity.Simulation3DIntegral.html#SimPEG.potential_fields.gravity.Simulation3DIntegral), that can be created without passing a `survey`, but they will ultimately fail to run `dpred()`.
1. Keeping `kwargs` might lead to more maintenance work while deprecating arguments. If a function accepts `kwargs`, we need to write an error raising if a deprecated argument is being passed through `kwargs`. When to remove that error is not a trivial question (since we don't want to suddenly allow users to pass legacy arguments).

I think most of these issues can be solved by removing `**kwargs` from the signatures of methods that don't need them, and explicitly ask for the required arguments either as positional or optional (with a default value).

santisoler:
[...]

PR #1518, opened 2024-08-16
Class parameter documentation inheritance

jcapriot:
#### Summary
Adds the ability to inherit information from a parent class's documentation and correspondingly update the class's call signature.

#### PR Checklist
      to a Pull Request

#### Reference issue
Relevant to #1420

##### Documentation implementation

Adds a mechanism to inherit documentation descriptions from parent classes. This avoids us having to copy and paste documentation code for each parameter into every single subclass. For example, we can write the description of the `mesh` parameter in one place, then make use of that in every child class. Or, we can access all of the other parameters that are less likely to be used from a parent class. It also then updates the call signature of the class's `__init__` function to match the description in the documentation.

 As of right now, my proposed syntax for this is demonstrated in the few classes I've added this to.

The docstring for `LinearSimulation` as coded:
https://github.com/jcapriot/simpeg/blob/17f934d7b25e7669dd483788f50f366b63ad224c/simpeg/simulation.py#L717-L758

Essentially the `%(....)` triggers a replacement on the docstring when the `@doc_inherit()` decorator is attached to the class. Specifically a `%(super.item)` triggers a lookup for the description of an argument named `item` in the next class that describes `item` in the class's inheritance tree. `%(module.ClassName.item)` would trigger a lookup for `item` specifically in the importable class `module.ClassName`, and `%(super.*)` would trigger an inclusion of every parameter from the class's inheritance tree (that is not already documented on this class already (you could also trigger this from a specific class as `%(module.ClassName.*)`). There is also an option to specifically exclude parameters from the star include operation, as seen in the `LinearSimulation` excluding the `solver` and `solver_opts` parameters.

After the replacements the `LinearSimulation` docstring looks like:
```
"""Linear forward simulation class.

The ``LinearSimulation`` class is used to define forward simulations of the form:

.. math::
    \mathbf{d} = \mathbf{G \, f}(\mathbf{m})

where :math:`\mathbf{m}` are the model parameters, :math:`\mathbf{f}` is a
mapping operator (optional) from the model space to a user-defined parameter space,
:math:`\mathbf{d}` is the predicted data vector, and :math:`\mathbf{G}` is an
``(n_data, n_param)`` linear operator.

The ``LinearSimulation`` class is generally used as a base class that is inherited by
other simulation classes within SimPEG. However, it can be used directly as a
simulation class if the :py:attr:`G` property is used to set the linear forward
operator directly.

By default, we assume the mapping operator :math:`\mathbf{f}` is the identity map,
and that the forward simulation reduces to:

.. math::
    \mathbf{d} = \mathbf{G \, m}

Parameters
----------
mesh : discretize.base.BaseMesh, optional
    Mesh on which the forward problem is discretized.
model_map : simpeg.maps.BaseMap
    Mapping from the model parameters to vector that the linear operator acts on.
G : (n_data, n_param) numpy.ndarray or scipy.sparse.csr_matrx
    The linear operator. For a ``model_map`` that maps within the same vector space
    (e.g. the identity map), the dimension ``n_param`` equals the number of model parameters.
    If not, the dimension ``n_param`` of the linear operator will depend on the mapping.
model : (n_m,) array_like, optional
    The parameter model, often used to descibe the model in an inversion.
    If there are any physical property maps assigned, those physical properties
    will be linked to this model through the map, and accessing them will require
    a model to be set.
survey : simpeg.survey.BaseSurvey, optional
    The survey for the simulation.
sensitivity_path : str, optional
    Path to directory where sensitivity file is stored.
counter : None or simpeg.utils.Counter
    SimPEG ``Counter`` object to store iterations and run-times.
verbose : bool, default: False
    Verbose progress printout.
"""
```

Because we have included a `%(super.*)` lookup, it doesn't include a `**kwargs` description.

##### Call signatures
Updating the documentation is nice... But it doesn't enable tab completion on most editors, or type inference, and it doesn't make the default values for optional arguments shown in the call signature. But that's the key there, you can specifically set a function's `__signature__` property to change it's call signature.
So the next bit of this replacement appropriately updates the functions call signature to add the items that were included from elsewhere (that were not already part of it's call signature). This parameters are only allowed as keyword arguments, and are marked as such when added to the call signature. They are added in the order they are added to the documentation.

Now inspecting the loaded `LinearSimulation` class in a jupyter notebook we see it's description as:
image

Tab completion in a notebook shows all of the added parameters:
image

The generated documentation page:

#### Additional information

This relies on the classes implementing the `numpydoc` style doc strings. It will search for descriptions in parents that are described in either the `Paramaters` or `Other Parameters` sections.

This spirit of this was inspired by `matplotlib`, where they automatically populate their docstrings with the multitude of parameters that their classes take.

### DIFF
--- a/simpeg/base/pde_simulation.py
+++ b/simpeg/base/pde_simulation.py
@@ -4,6 +4,7 @@ from discretize.utils import Zero, TensorType
 from ..simulation import BaseSimulation
 from .. import props
 from scipy.constants import mu_0
+from ..utils.doc_utils import doc_inherit
 
 
 def __inner_mat_mul_op(M, u, v=None, adjoint=False):
@@ -491,7 +492,23 @@ class BasePDESimulation(BaseSimulation):
 
 @with_property_mass_matrices("sigma")
 @with_property_mass_matrices("rho")
+@doc_inherit()
 class BaseElectricalPDESimulation(BasePDESimulation):
+    """A simulation containing the electrical physical properties sigma and rho.
+
+    Parameters
+    ----------
+    %(super.mesh)
+    sigma, rho : (mesh.n_cells) array_like, optional
+        Conductivity and resitivity properties. These are linked to each
+        other as inverses and at most one can be set.
+    sigmaMap, rhoMap : simpeg.IdentityMap, optional
+        Relationship between the `model` and conductivity or resistivity.
+        These are linked to each other as inverses and at most one can be set.
+    **kwargs
+        keyword arguments passed to the parent class.
+    """
+
     sigma, sigmaMap, sigmaDeriv = props.Invertible("Electrical conductivity (S/m)")
     rho, rhoMap, rhoDeriv = props.Invertible("Electrical resistivity (Ohm m)")
     props.Reciprocal(sigma, rho)
@@ -500,8 +517,12 @@ class BaseElectricalPDESimulation(BasePDESimulation):
         self, mesh, sigma=None, sigmaMap=None, rho=None, rhoMap=None, **kwargs
     ):
         super().__init__(mesh=mesh, **kwargs)
+        if sigma and rho:
+            raise TypeError("Cannot set both sigma and rho.")
         self.sigma = sigma
         self.rho = rho
+        if sigmaMap and rhoMap:
+            raise TypeError("Cannot set both sigmaMap and rhoMap.")
         self.sigmaMap = sigmaMap
         self.rhoMap = rhoMap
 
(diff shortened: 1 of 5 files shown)


jcapriot:
I've been thinking a bit more about where this functionality should live, and it seems like it really should be it's own package, mostly because I want to use its functionality independent of `SimPEG`. Any thoughts on a name?

`docerator` has been living in my mind as a portmanteau of documentation and generator (and it sounds like decorator).

Issue #1121, opened 2022-10-26
Active Cells names


jcapriot:
We need to standardize the name we use to refer to active cell index arrays throughout SimPEG. In some areas it is referred to as `ind_active`, in other `actInd`, and some newer ones as `active_cells`.

I would favor switching all of them to `active_cells`.

thibaut-kobold:
`active_cells` sounds good to me.

santisoler:
Now we have `active_cells` in VRM and potential field simulations, and in maps.  The missing places are:

* the `spectral_ip_mappings` function,
* the `drapeTopotoLoc` function.

Which I don't think they are widely used.

PR #1520, opened 2024-08-23
Replace `ind_active` for `active_cells` in pf simulations

santisoler:
#### Summary

Replace the `ind_active` argument in potential field simulation classes for `active_cells`. Add tests that check the corresponding warnings and errors are being raised. Update example, tutorials and tests so they use `active_cells` instead of `ind_active`.

#### PR Checklist
      to a Pull Request

#### Reference issue

Part of #1121


jcapriot:
Thinking a little forward here…

Probably best practice to actually leave `ind_active` as a keyword argument in `__init__` (but move it to the end). Then, leave it in the docstring (at the appropriate location), but mark it as deprecated in the parameter’s description.

I know it’s not how we’ve done it elsewhere, but I think we should start doing it that way.

simpeg/potential_fields/base.py
```diff
@@ -92,9 +94,25 @@ def __init__(
         # If deprecated property set with kwargs
         if "actInd" in kwargs:
             raise AttributeError(
-                "actInd was removed in SimPEG 0.17.0, please use ind_active"
+                "actInd was removed in SimPEG 0.17.0, please use 'active_cells'"
```
jcapriot, line 97: We can probably just remove this check now, been a few years since 0.17.0

santisoler:
> Thinking a little forward here…
> 
> Probably best practice to actually leave `ind_active` as a keyword argument in `__init__` (but move it to the end). Then, leave it in the docstring (at the appropriate location), but mark it as deprecated in the parameter’s description.
> 
> I know it’s not how we’ve done it elsewhere, but I think we should start doing it that way.

Makes sense! I agree that we should start doing it that way.

santisoler:
@jcapriot should we merge this one?

PR #1523, opened 2024-08-27
Fix validate_ndarray_with_shape

prisae:
#### Summary

The current implementation of `simpeg.utils.code_utils.validate_ndarray_with_shape` works for numbers or lists, but fails for `np.ndarray`'s.

Example
```python
import numpy as np
from simpeg.utils.code_utils import validate_ndarray_with_shape
``` 

First a list, all as expected, it returns a complex numpy array.
```python
validate_ndarray_with_shape('Test 1', [1+1j, 2+2j], '*', (float, complex))
```
```
array([1.+1.j, 2.+2.j])
```

Now directly for a complex  numpy array: It casts the array to the first provided dtypes, here floats, and raises a warning if that casting has some side-effects, here that it casts complex values to real values. Not what was wanted.
```python
validate_ndarray_with_shape('Test 1', np.array([1+1j, 2+2j]), '*', (float, complex))
```
```
/simpeg/utils/code_utils.py:1044: ComplexWarning: Casting complex values to real discards the imaginary part
  var = np.asarray(var, dtype=dtype)
array([1., 2.])
```

#### PR Checklist
      to a Pull Request

@simpeg/simpeg-developers - small fix, should be ready.

### DIFF
--- a/simpeg/utils/code_utils.py
+++ b/simpeg/utils/code_utils.py
@@ -1036,7 +1036,7 @@ def validate_ndarray_with_shape(property_name, var, shape=None, dtype=float):
     for dtype in dtypes:
         try:
             if isinstance(var, np.ndarray):
-                var = var.astype(dtype, casting='safe', copy=False)
+                var = var.astype(dtype, casting="safe", copy=False)
             else:
                 var = np.asarray(var, dtype=dtype)
             bad_type = False
--- a/simpeg/utils/code_utils.py
+++ b/simpeg/utils/code_utils.py
@@ -1033,20 +1033,17 @@ def validate_ndarray_with_shape(property_name, var, shape=None, dtype=float):
         dtypes = (dtype,)
     else:
         dtypes = dtype
-
-    # If var is already a numpy array of an allowed dtype, all good
-    if isinstance(var, np.ndarray) and var.dtype in dtypes:
-        bad_type = False
-    # Else, loop over dtypes and try to cast.
-    else:
-        for dtype in dtypes:
-            try:
+    for dtype in dtypes:
+        try:
+            if isinstance(var, np.ndarray):
+                var = var.astype(dtype, casting='safe', copy=False)
+            else:
                 var = np.asarray(var, dtype=dtype)
-                bad_type = False
-                break
-            except (TypeError, ValueError) as err:
-                bad_type = True
-                raised_err = err
+            bad_type = False
+            break
+        except (TypeError, ValueError) as err:
+            bad_type = True
+            raised_err = err
 
     if bad_type:
         raise TypeError(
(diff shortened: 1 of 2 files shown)


jcapriot:
I think the bug fix here is that we need to catch the `ComplexWarning` that numpy issues when it casts the complex numbers to a `float`.

prisae:
Your suggestion would work for my use case, yes.

But I think the way I tried to fix it here is better (IMHO) / more general. Lets say you input a real array, but your dtypes are `(complex, float)`, then it will always cast it to complex values (without any warning). Isn't it better to leave existing numpy arrays as they are, if they are one of the acceptable dtypes.

jcapriot:
That goes against the documented behavior of the function and would imply it has a different behavior for numpy array inputs than list or other iterate inputs.

specifically:
```
    dtype : class, or tuple of class, optional
        The data type for the array. I.e. float, int, complex, bool, etc.
        Validated from left to right if a tuple.
```

prisae:
OK. If you prefer that then I will adjust it accordingly then.


>  would imply it has a different behavior for numpy array 

Indeed, it would mean that. Numpy arrays are the only type that actually have a `dtype`, and any input is cast into a numpy array within the function, so this different behaviour would make sense for me. It would mean to not change something that is already set. All other inputs (lists, tuples, just numbers) are cast into a numpy array within the function.

jcapriot:
If it is already an ndarray, I think it’s best to use its `astype(type, casting='safe’, copy=False)` method.

prisae:
> it’s best to use its `astype(type, casting='safe’, copy=False)` method

That could work, not sure which one to prefer (the suggested method or astype), I guess we would have to test some edge cases.

But to be honest, I only see my use case (complex data) within `simpeg.data.Data().dobs` - so I truly lack the bigger picture (for what else is it used) and the history (how did `properties` use to handle this). Whatever works is fine with me.

I leave it for a while as it is, maybe others chime in or you settled on the preferred method, which I can (or you) implement.

Thanks for thinking along!

prisae:
I did some testing now, and I like your suggestion @jcapriot! What do you think of the current implementation?

prisae:
(Why does Codecov check the coverage of the `tests/`-directory?)

jcapriot:
> (Why does Codecov check the coverage of the `tests/`-directory?)

To ensure tests are ran.

prisae:
Codecov complains that the three test lines are not themselves tested - so I can happily ignore that I assume?

prisae:
Got it, because the test failed. Sorry for the fuss.

jcapriot (approved):
Looks good now. Not sure why codecov didn't update it's comments, but it shows everything tested now.

jcapriot:
> I did some testing now, and I like your suggestion @jcapriot! What do you think of the current implementation?

Yeah, I think the way that `safe` keyword works is a good description of the intended functionality of this validator. If it can be safely cast to one of the listed types (without loosing information) it will do so from left to right. That's why I wouldn't recommend passing `(complex, float)` as the acceptable types, because `complex` includes `float` so it will always return a `complex` type.

PR #1524, opened 2024-08-27
Ensure misfit is purely real valued

prisae:
#### Summary

The misfit is always real valued, even for complex data, as it is the dot-product of the complex-conjugate value with the value itself. However, if the data is complex, `vdot(data, data)` will be a complex number with zero complex part (`x+0j`). This PR ensures it gives back a real-only value (`x`).

(Originally fixed from @sgkang in one of the many abandoned SimPEG(emg3d) pull requests.)

#### PR Checklist
      to a Pull Request

@simpeg/simpeg-developers


santisoler:
I totally agree that the data misfits should return real values when called. Nonetheless, I think the responsibility of ensuring that should come from the simulations: the `residual` method of simulations should always return real values. Do you recall which simulations where returning complex values?

I also think we should add some tests that fail if we don't have this fix, just to make sure we have tests that catch this bug. If we decide to enforce the simulations to return the residuals as real values, then the tests should be against them and not against the `L2DataMisfit`.

What do you think?

prisae:
> the residual method of simulations should always return real values.

No, not if you have complex data, then the residual is still complex.

prisae:
I see what you mean. If the residual would return `absolute_residual=abs(complex_residual)`, `np.vdot(absolute_residual, absolute_residual)` would yield the same as with complex residuals with `np.vdot(complex_residual, complex_residual).real`.

I guess indeed that can be discussed. I am used to complex residual, but if the opinion is that the residual should be real-valued, then we can replace this PR with a PR that fixes the computation of residual for complex data within SimPEG.

prisae:
(However and again, slapping a `.real` onto the `vdot` here will not have any negative side effects and would it make work with real and complex valued residuals...)

prisae:
> Do you recall which simulations where returning complex values?

I use the `BaseFDEMSimulation`. But it comes basically from the `BaseSimulation`,

https://github.com/simpeg/simpeg/blob/5e4b583be3e5dafa3622dcff59894a2a02e5a304/simpeg/simulation.py#L435-L461

-> this returns a complex valued residual if `dobs` and `dpred` are complex values.

Therefore, we should either do `vdot(R, R).real`, or cast https://github.com/simpeg/simpeg/blob/main/simpeg/simulation.py#L461 to real. However, the latter option might have more unwanted side effects than the former.

santisoler:
I'm not very familiar with how the simulations and the data misfits were designed to behave with complex data. From [this example](https://docs.simpeg.xyz/latest/content/tutorials/07-fdem/plot_fwd_3_fem_3d.html) I see that the returned data is complex (there are real and imaginary parts of the Bz field), but the `dobs` array is filled with real values: half of it correspond to the real parts, half of it correspond to the imaginary parts. For this reason I assumed that the `dpred` method of simulations always return real values, and hence my recommendation of delegating the responsibility of returning real valued residuals to the simulations.
 
Nonetheless, if the simulations would return complex arrays, keeping just the real part when computing the residual wouldn't throw away all the misfits in the imaginary parts? Is this the desired behavior?

jcapriot:
I do not know a case where we currently support using complex values data in an inversion. Some of the FDEM codes support outputting data as complex values, but none of them support inversion.

prisae:
> Nonetheless, if the simulations would return complex arrays, keeping just the real part when computing the residual wouldn't throw away all the misfits in the imaginary parts? Is this the desired behavior?

I would much prefer to keep complex residuals, indeed.
- The misfit is purely real, `vdot(R, R)` has never a non-zero imaginary part, so you would not loose anything. Just using the real is how it is usually done (AFAIK) in inversions with complex data.
- If you prefer to keep the residuals real, it would have to be the `abs(residual)`. But I'd prefer the above option, keeping only the misfit real..

> I do not know a case where we currently support using complex values data in an inversion.

There probably isn't any documented, but it is possible due to the fact that, e.g., the `simpeg.data.Data.dobs` can be complex. Seogi and I use it for SimPEG(emg3d), that is where this is coming from. It is just to make it more general. 

(@sgkang - it was a change you proposed back on 26 April 2021, maybe you have some opinion on it?)

santisoler:
Thanks @prisae for the clarification. I think I wasn't understanding the issue very well. I agree that doing `np.vdot(residual, residual).real` won't loose any information, since the imaginary part of the value that `vdot` returns is always zero. Therefore, your suggested change makes sense. Now that I understand this better, I take back my suggestion of making simulations to always return real-valued residuals.

Now, I think the decision comes down to design:

* Should we support inversions with complex models?
* Should we support simulations with complex data?
* If we support these two, should the residuals be complex?
* Should data misfits always return real values when called?

I don't have strong feelings on the complex model and complex data side. I know that supporting them won't be too easy (we need to start thinking about complex data and complex uncertainties). But if there's a good case to support them, it's still achievable.

On the other side, I'm 100% positive that the data misfits should return real values when returned.

cc @dccowan (you might be interested in joining this conversation)

prisae:
> Should we support simulations with complex data?

AFAIK, that is a decision long made, as it is implemented for EM1D. See, e.g.,

https://github.com/simpeg/simpeg/blob/5e4b583be3e5dafa3622dcff59894a2a02e5a304/simpeg/electromagnetics/frequency_domain/receivers.py#L15-L17

And the data container permits complex values:

https://github.com/simpeg/simpeg/blob/5e4b583be3e5dafa3622dcff59894a2a02e5a304/simpeg/data.py#L129-L133

prisae:
bumping this one...?

santisoler:
Hi @prisae. After following the discussion, I think your suggested change makes sense. I think we should merge it.

Could we add a test for it though? Basically, a test that checks that the data misfit returns a real-valued array even if the residual is complex.

prisae:
Yes, I'll try to get a test in.

prisae:
I kept it as simple as possible. Passing in complex data to compare with would result in a complex valued misfit before this PR. Now we can assert that it is a real-valued misfit, which checks the implementation.

santisoler:
Thanks @prisae!! I'm merging this 🚀

santisoler (approved):

prisae:
Awesome, thanks @santisoler!

PR #1525, opened 2024-08-27
Add ElectricDipole

prisae:
#### Summary

Add an `ElectricDipole` to `simpeg/electromagnetics/frequency_domain/sources.py` similar to `MagDipole`.

Last piece for SimPEG(emg3d).

#### PR Checklist
      to a Pull Request

@simpeg/simpeg-developers - I think this is ready, feedback welcomed!


simpeg/electromagnetics/frequency_domain/sources.py
```diff
@@ -723,6 +723,93 @@ def bPrimary(self, simulation):
...
+    ----------
+    receiver_list : list of simpeg.electromagnetics.frequency_domain.receivers.BaseRx
+        A list of FDEM receivers
+    frequency : float
+        Source frequency
+    location : (dim) numpy.ndarray, default: numpy.r_[0., 0., 0.]
+        Source location.
+    strength : float, default: 1.0
+        Source strength (Am).
+    orientation : {'z', x', 'y'} or (dim) numpy.ndarray
```
santisoler, line 741: Following [numpydoc's style guide](https://numpydoc.readthedocs.io/en/latest/format.html#parameters), let's put the default value first:
```suggestion
    orientation : {'x', y', 'z'} or (dim) numpy.ndarray
```

santisoler (changes requested):
Thanks @prisae for opening this PR! It looks good to me! I just left one minor suggestion in the docstring.

simpeg/electromagnetics/frequency_domain/sources.py
prisae: Good point, done!

santisoler:
@dccowan, I think this PR is ready to be merged, but I'd like to hear your thoughts on it.

prisae:
I rewrote the part of `SimPEG(emg3d)` to use the RawVec-versions, which is more general for any possible source field. So I do not need this PR any longer.

However, it might still be valuable to have it. Indeed, it is just a special case of `LineCurrent`, so I changed it to be a subclass of `LineCurrent`. But because it is only a single point, it requires the `orientation`. Not sure if my implementation as such works.

Let me know if I shall close it, or if it is worth bringing it in.

santisoler:
Hi @prisae! During last week's SimPEG meeting, @lheagy mentioned that the `ElectricDipole` class won't work with SimPEG simulation as it is because it lacks the `s_m` and `s_e` methods. I'm not sure how useful would be to have this class in SimPEG since you don't need it anymore. Maybe @lheagy and @dccowan can comment on that.

prisae:
Since the latest push it does have the `s_e` and `s_m` methods, as it is now just a special case of `LineCurrent`. But I am not entirely sure if it works this way with `orientation`. However, in the corresponding `MagDipole` this seems to work. Anyway, I am fine either way!

santisoler:
Hi again @prisae! Ok, I was taking a look at this PR and there's something off about what this class should be and what it does. As it is defined, it's basically a `LineCurrent` with an `orientation` attribute. The main issue I see is that it takes a single point as `location` (according to the docs). But when we'll try to use this source in a simulation, it will fail because the line segment is expected to have at least two locations: the start and end of the segment.

For example, if I run the following example:

```python
import numpy as np
import simpeg
import simpeg.electromagnetics.frequency_domain as fdem


# Define two receivers: for the real and imaginary components
receiver_locations = np.array([10.0, 0.0, 30.0])
receivers = [
    fdem.receivers.PointMagneticFieldSecondary(
        locations=receiver_locations,
        orientation="z",
        data_type="ppm",
        component=component,
    )
    for component in ("real", "imag")
]

# Define an ElectricDipole as source
source_location = np.array([0.0, 0.0, 30.0])
source = fdem.sources.ElectricDipole(
    receiver_list=receivers,
    frequency=10e4,
    location=source_location,
    orientation="z",
    current=1.0,
)

survey = fdem.Survey([source])

# Define a simulation
layer_thicknesses = np.array([20.0, 40.0])
conductivity_map = simpeg.maps.IdentityMap()
simulation = fdem.Simulation1DLayered(
    survey=survey,
    thicknesses=layer_thicknesses,
    sigmaMap=conductivity_map,
)

# Define a layered Earth
layer_conductivities = np.array([0.1, 1.0, 0.1])

# Run forward
dpred = simulation.dpred(layer_conductivities)
```

I obtain the following error:

```
Traceback (most recent call last):
  File "/home/santi/git/simpeg/electric_dipole_simple.py", line 49, in <module>
    dpred = simulation.dpred(layer_conductivities)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/santi/git/simpeg/simpeg/electromagnetics/frequency_domain/simulation_1d.py", line 97, in dpred
    f = self.fields(m)
        ^^^^^^^^^^^^^^
  File "/home/santi/git/simpeg/simpeg/electromagnetics/frequency_domain/simulation_1d.py", line 107, in fields
    self._compute_coefficients()
  File "/home/santi/git/simpeg/simpeg/electromagnetics/frequency_domain/simulation_1d.py", line 70, in _compute_coefficients
    self._compute_hankel_coefficients()
  File "/home/santi/git/simpeg/simpeg/electromagnetics/base_1d.py", line 384, in _compute_hankel_coefficients
    h = src.location[2] - self.topo[-1]
        ~~~~~~~~~~~~^^^
IndexError: index 2 is out of bounds for axis 0 with size 1
```

I guess that in order to properly build this type of source we would need a length for the dipole, and assume that the `location` is the dipole's center. Along with the `orientation`, we could build a straight line segment that could be used a source.

I know that @lheagy has some considerations on doing so, mainly on how this dipole might get used, and how it might or might not be aligned with edges of the mesh. But I'll let her chime in 🙂 

On a side note, we should probably sanitize the input of `LineCurrent` and raise an earlier error if the `locations` is only a single point.

prisae:
Thanks for the insights and feedback @lheagy - I close this, one PR less!

PR #1528, opened 2024-09-03
Multiplier refactor

domfournier:
#### Summary
This PR intends to move the responsibility of `multiplier` to ObjectiveFunction, as a property of the class rather than a list stored on the parent Combo, such that weighting can easily be manipulated at the function level. This can greatly simplify the mechanics of directives that act on a specific objective (or combo) function. Scaling can be changed directly on that function, rather than having to look for its multiplier on the parent combo.

This would simplify/clean up existing modules such as pgi_directives, sim_directives that currently assume ordering or a set number of regularization terms.

I tried to keep the interaction from the outside the same (i.e. setting `multipliers` from the combo). 

As discussed

- Evaluating the objective function `fun(m)` does NOT include the multiplier
- Evaluating the ComboObjectiveFunction does include the sub multiplier but not its own multiplier

$\phi(m) =\sum \alpha_i \phi_i$

I think @santisoler it is ok if 

$\phi_A(m) + \phi_B(m) \neq \phi(m)$

Because $\phi_A + \phi_B$ becomes a Combo, and $\phi(m)$ now involves multipliers. It is much more in line with the usual definition of the global objective function. Which is not the same as evaluating the functions individually.

- Beta is now a multiplier stored on the top level reg. Misfit functions themselves have multipliers for future functionality.

#### PR Checklist
      to a Pull Request


tests/base/test_objective_function.py
```diff
@@ -352,7 +352,7 @@ def test_mul(self, unpack_on_add):
         combo_mul = 3.5 * combo
         assert len(combo_mul) == 1
         assert combo_mul.multipliers == [3.5]
-        assert combo_mul.objfcts == [combo]
+        # assert combo_mul.objfcts == [combo]
```
domfournier, line 355: We can not do straight equal anymore since the function gets copied and we end up with two different instances (not the same id)

tests/base/test_objective_function.py
```diff
@@ -385,7 +385,7 @@ def test_add_multiple_terms(self):
         combo = 1.1 * phi1 + 1.2 * phi2 + 1.3 * phi3
         assert len(combo) == 3
         assert combo.multipliers == [1.1, 1.2, 1.3]
-        assert combo.objfcts == [phi1, phi2, phi3]
+        # assert combo.objfcts == [phi1, phi2, phi3]
```
domfournier, line 388: Same here

domfournier:
@santisoler Guess we never managed to connect to talk about this. Feel free to write your suggestions here.

simpeg/objective_function.py
```diff
@@ -435,57 +447,61 @@ def multipliers(self):
...
+        for i, objfct in enumerate(self):
+            if objfct.multiplier == 0.0:  # don't evaluate the fct
                 continue
             if f is not None and objfct.has_fields:
                 aux = objfct.deriv(m, f=f[i])
             else:
                 aux = objfct.deriv(m)
             if not isinstance(aux, Zero):
-                g += multiplier * aux
+                g += objfct.multiplier * aux
```
domfournier, line 490: @dccowan @jcapriot @santisoler @lheagy 

The main change happens here, where the top combo asks for the function's multiplier, rather than storing a list of multipliers and zipping through both. So no change on the value of `function.call()` or `function.deriv()`. 

The main advantage is that directives can act on scaling functions directly, without having to worry about the order in the global combo. Beta simply becomes the `inv.reg.multiplier` also - all same mechanism for addition.

simpeg/objective_function.py
```diff
@@ -435,57 +447,61 @@ def multipliers(self):
...
+            return
+
         for multiplier in value:
             _validate_multiplier(multiplier)
+
         _check_length_objective_funcs_multipliers(self.objfcts, value)
-        self._multipliers = value
+
+        for val, obj in zip(value, self.objfcts):
+            obj.multiplier = val
```
domfournier, line 464: We can also keep the top-level `multipliers` assignment - they just get sent down the chain to their respective functions.

simpeg/inverse_problem.py
```diff
@@ -64,6 +64,7 @@ def beta(self):
     @beta.setter
     def beta(self, value):
         self._beta = validate_float("beta", value, min_val=0.0)
+        self.reg.multiplier = self._beta
```
domfournier, line 67: This could let us treat the functions in the global objective function as just a flat list of functions (`mult1 * misfit1 + mult2 * misfit2 + beta * multi_a * reg_a + beta * multi_b * reg_b + ...` )

simpeg/objective_function.py
domfournier: @santisoler So we could call a specific method instead to get the same value scaled. If that makes more sense to you guys. Something like
```
g += objfct.scaled_deriv(m)
```

or as @thibaut-kobold suggested
```
g += objfct.deriv(m, scaled=True)
```
and we leave the default behaviour as scaled=False

domfournier:
Got a much simpler idea. To be continued on #1542

domfournier:
Got a much simpler idea. To be continued on #1542

PR #1542, opened 2024-09-28
Handle multipliers on component of Combo

domfournier:


domfournier:
@simpeg/simpeg-developers Ready for review gang.

jcapriot:
Just did a quick parsing, so forgive me if this is not quite how it works,

I'm definitely in favor of handling multiplication of objective functions like this. I.E.
```
>>> obj = ...
>>> obj2 = 10 * obj
>>> type(obj2)
ScaledObjectiveFunction
```

But what seems odd to me is what happens when they get added (in an abstract sense, I know why it is done internally here...):
```
>>> obj1 = DataMisfit(...)
>>> obj2 = Regularization(...)
>>> obj3 = obj1 + 10 * obj2
>>> type(obj3)
ComboObjectiveFunction
>>> type(obj3[0])
ScaledObjectiveFunction
```
Why did `DataMisfit` need to get turned into a `ScaledObjectiveFunction`?

santisoler (changes requested):
Hi @domfournier!

Thanks for the updates on this PR.

I think there are a few things I would like to change regarding the mechanics of the `BaseObjectiveFunction`, the `ScaledObjectiveFunction` and the `ComboObjectiveFunction`.

## Combo as a collection of `BaseObjectiveFunction`s

Firstly, I agree with @jcapriot that we shouldn't be promoting classes when adding them together: the `ComboObjectiveFunction` should be a collection of `BaseObjectiveFunction`s. For example, the following assertions should not raise errors:

```python
phi_d = DataMisfit(...)
phi_m = Smallness(...)
beta = ...
phi = phi_d + beta * phi_m

assert phi[0] is phi_d
assert isinstance(phi[1], ScaledObjectiveFunction)
assert phi[1].function is phi_m
assert phi[1].multiplier == beta
```


## Nested `ScaledObjectiveFunction`s

Secondly, I think we should not nest `ScaledObjectiveFunction`s after multiplications. For example, let's say I have the following class:

```python
class MockPhi(BaseObjectiveFunction):

    def __init__(self):
        self._nP = 5
```

And I perform these operations:

```python
phi = MockPhi()
phi1 = 3.0 * phi
phi2 = 5.0 * phi1
```

With the current implementation, `phi2` becomes a `ScaledObjectiveFunction` with a multiplier of 5.0, that contains another `ScaledObjectiveFunction` with a multiplier of `3.0`. If we repeat such operations, we can create a deep nested structure. I think the desired behaviour would be to return a new `ScaledObjectiveFunction` with a new multiplier, like:

```python
phi = MockPhi()
phi1 = 3.0 * phi
phi2 = 5.0 * phi1
assert phi2.function is phi
assert phi2.multiplier == 3.0 * 5.0

# But we don't touch phi1 while creating phi2:
assert phi1 is not phi2
assert phi1.function is phi
assert phi1.multiplier == 3.0
```

To solve this, I think we should add a `__mul__` method to `ScaledObjectiveFunction`, like:

```python
def __mul__(self, scalar):
    return ScaledObjectiveFunction(self.function, self.multiplier * scalar)
```

## Some tests for these properties

I wrote a few tests that check these behaviours. We can totally use them for this PR, feel free to add them:

<details>
<summary>Sample tests for the new classes</summary>

```python
from simpeg.objective_function import (
    BaseObjectiveFunction,
    ComboObjectiveFunction,
    ScaledObjectiveFunction,
)


class MockPhi(BaseObjectiveFunction):

    def __init__(self):
        self._nP = 5


class TestObjectiveFunctionOperations:

    def test_add(self):
        phi1, phi2 = MockPhi(), MockPhi()
        phi = phi1 + phi2
        assert isinstance(phi, ComboObjectiveFunction)
        assert phi[0] is phi1
        assert phi[1] is phi2

    def test_multiple_additions(self):
        phi1, phi2, phi3 = MockPhi(), MockPhi(), MockPhi()
        phi = phi1 + phi2 + phi3
        assert isinstance(phi, ComboObjectiveFunction)
        assert phi[0] is phi1
        assert phi[1] is phi2
        assert phi[2] is phi3

    def test_multiplication(self):
        phi1 = MockPhi()
        phi = 3.14 * phi1
        assert isinstance(phi, ScaledObjectiveFunction)
        assert phi.multiplier == 3.14
        assert phi.function is phi1

    def test_multiple_multiplications(self):
        phi = MockPhi()
        phi1 = 3.14 * phi
        phi2 = 2.0 * phi1
        assert isinstance(phi1, ScaledObjectiveFunction)
        assert isinstance(phi2, ScaledObjectiveFunction)
        assert phi1 is not phi2
        assert phi1.multiplier == 3.14
        assert phi2.multiplier == 2.0 * 3.14
        assert phi1.function is phi
        assert phi2.function is phi

    def test_addition_and_multiplication(self):
        phi1, phi2, phi3 = MockPhi(), MockPhi(), MockPhi()
        alpha1, alpha2, alpha3 = 2.0, 3.1, 4.2
        phi = alpha1 * phi1 + alpha2 * phi2 + alpha3 * phi3
        assert isinstance(phi, ComboObjectiveFunction)
        for function in phi:
            assert isinstance(function, ScaledObjectiveFunction)

        for i, (alpha, function) in enumerate(
            zip((alpha1, alpha2, alpha3), (phi1, phi2, phi3))
        ):
            assert phi[i].multiplier == alpha
            assert phi[i].function is function
```

</details>

## Should we remove `unpack_on_add`?

Then there are other minor design decisions that we could take.

I've never quite liked the `unpack_on_add` mechanic I added, it always felt like a patch.

Now that we have `Combo` for additions and `Scaled` for multiplications, I think we could remove `unpack_on_add` and enforce these rules:

1. Adding an objective function to a combo, returns a new expanded combo (similar to how lists work). For example:
   ```python
   phi1, phi2 = MockPhi(), MockPhi()
   phi = phi1 + phi2

   # Add a phi3 to phi and return a new combo with phi1, phi2, phi3 as components
   phi3 = MockPhi()
   phi_new = phi + phi3

   assert isinstance(phi_new, ComboObjectiveFunction)
   assert phi_new is not phi
   assert len(phi_new) == 3
   assert phi_new[0] is phi1
   assert phi_new[1] is phi2
   assert phi_new[2] is phi3
   ```
1. Adding a `ScaledObjectiveFunction` never unpacks. This rule would allow us to group objective functions together, like:
   ```python
   phi_d = DataMisfit(...)
   phi_s, phi_x, phi_y = Smallness(...), Smoothness("x", ...), Smoothness("y", ...)
   phi_m = phi_s + 3.0 * phi_x + 6.0 * phi_y
   phi = phi_d + 1e-2 * phi_m

   assert len(phi) == 2
   assert phi[0] is phi_d
   assert isinstance(phi[1], ScaledObjectiveFunction)
   assert phi[1].function is phi_m
   assert phi[1].multiplier == 1e-2
   ```

If we have two combos, adding them together returns a new _expanded_ combo (by unpacking them):

```python
combo_a = phi1 + phi2
combo_b = phi3 + phi4
new_combo = combo_a + combo_b

assert len(new_combo) = 4
assert new_combo[0] is phi1
assert new_combo[1] is phi2
assert new_combo[2] is phi3
assert new_combo[3] is phi4
```

But, if we want to avoid that, we could multiply them by 1:

```python
combo_small = 1.0 * combo_a + 1.0 * combo_b

assert len(new_combo) = 2
```

This way, users could easily group or unpack combos without the need to modify the `unpack_on_add` attribute.

We can talk about this, but I'd be keen on relying on these mechanics rather than having a special argument (`unpack_on_add`) for the `Combo`.


## Combo should not know about multipliers

Following to the idea in the first paragraphs, I don't think that the `ComboObjectiveFunction` should know about multipliers. Multipliers might or might not exist for each one of the components, so I think it's better to leave handling of multipliers to the `ScaledObjectiveFunction` (if there are any within the components).

I would remove the `multipliers` argument in the constructor of `ComboObjectiveFunction`, and the `multipliers` property. We can also get rid of the `_check_length_objective_funcs_multipliers` function.


## Minor things

Then there are minor things I think we could improve.

For example, making the `ScaledObjectiveFunction.function` a read-only property by removing the setter. I think it's a good safeguard to ensure that once a `ScaledObjectiveFunction` gets defined, it'll always be tied to the same objective function. We could store it in a private `self._function` attribute that gets defined in the constructor, where we can check if the object is valid (moving the check we have in the setter to the constructor).

Another minor thing: remove the default value for the `multiplier` argument in `ScaledObjectiveFunction`. I think such class is not meant to be initialized by end users directly, but generated by operations through dunder methods. Therefore, I think we should explicitly ask for a multiplier when initializing it.

Lastly, we could make use of some type hints in the dunder methods. Like:

```python
def __mul__(self, scalar) -> ScaledObjectiveFunction:
    return ScaledObjectiveFunction(self, scalar)
```

---

Let me know what do you think about these points. I'm happy to help with whatever you need.

jcapriot:
@santisoler Thanks for putting that together! That's exactly how I would expect it to happen too.

domfournier:
The idea was to alway only deal with Scaled functions, so that we don't need to consistently check whether it is a scaled  function of combo. When evaluating `phi_d + beta*phi_m` we just always end up doing a recursive `multiplier * fun` addition. It also makes it more obvious (in my mind) that each function will have a multiplier. 

`phi = phi_d + 1e-2 * phi_m` => `a_1 * phi_1 + a_ * phi_2` 

and we don't need to check constantly and simplifies directives down the road to act on phi_d directly.

domfournier:
```
def __mul__(self, scalar):
    return ScaledObjectiveFunction(self.function, self.multiplier * scalar)
```
I like that, makes sense to not nest because of a multiplication.

domfournier:
re: `1- Adding an objective function to a combo, returns a new expanded combo`

I really don't see the point of adding that complexity. We always want to scale the inner functions of a combo. I don't know of a case where the individual functions wouldn't have a scaling.

domfournier:
If I understand you correctly, the push back from is that would like the ComboObjective function to behave like a iterable. We can probably come with a compromise such the __getItem__ of ComboObjective returns the input elements as given, but have a dedicated attribute that contains promote Scaled functions?

jcapriot:
> ```
> def __mul__(self, scalar):
>     return ScaledObjectiveFunction(self.function, self.multiplier * scalar)
> ```
> 
> I like that, makes sense to not nest because of a multiplication.

Actually... I don't think this is a good idea. Let's say you do (for some odd reason):

```
m2 = 2 * m1
m3 = 3 * m2
```
Then `m3` would actually have no reference to `m2` anywhere. And let's say you wanted to manipulate `m2`, `m3` would never know about it.

jcapriot:
https://gist.github.com/jcapriot/2ba82c109b664a1f02fc7d8b5e9f89b3 Is my current thinking of how adding objectives and multiplying objectives should work.

domfournier:
Example looks good @jcapriot - you want fancy with your prints.

I am not sure this is intuitive though
image

I would have expected a flat list.

Again, there are many ways to skin that cat, and I still not convinced that mixing types is cleaner way.

domfournier:
> I don't quite like this idea. If the Combo is an ordered collection of `ScaledObjectiveFunction`s, then its i-th element should always be a `ScaledObjectiveFunction`. Moreover, I don't think the Combo should store information about the "original" objective function (the one used in the add operation, the object before the promotion). I think Combos should be quite simple: just a collection of other objective functions with an iterable interface.

Fair enough 

> Moreover, going back to Zen of Python: `There should be one-- and preferably only one --obvious way to do it`. If Combo is an iterable, accessing its elements should be as straight forward as for any collection: indexing, or looping over it. The syntax for that is quite obvious for everyone in Python, and I don't think we should provide multiple ways of doing so.

 Agreed

> * Feel free to ignore the `unpack_on_add` comment. I would like to put more time to think about it, without blocking this PR.

Sounds like a separate issue - agreed

> * I would still recommend to remove the `multiplier` property from the Combo. It would be quite easy to get them with: `multipliers = [p.multiplier for p in phi]`, where `phi` is a Combo.

I am cool either way, but if we remove it, it will break backward compatibility. I had it in there so that we don't have to change much elsewhere. We might have to do more refactoring of other classes if we take it off.

> * I would also add more tests to these changes, like the `test_multiple_multiplications` one I pasted in the comment, and more tests regarding the mechanics of the promotion.

jcapriot:
> Example looks good @jcapriot - you want fancy with your prints.
> 
> I am not sure this is intuitive though image
> 
> I would have expected a flat list.
> 
> Again, there are many ways to skin that cat, and I still not convinced that mixing types is cleaner way.

Regarding the flat list, like I mentioned in the gist, there’s no detectable difference from the python interpreter between the following operations to create `f`:

```
f = x + y + z
f = (x + y) + z
# which is equivalently
g = x + y
f = g + z
```
Therefore you simply cannot infer the code writers intentions (maybe they wanted f to be a flat list of x y and z, maybe they wanted x and y to be grouped separately from z). Since there’s no way to distinguish between the above lines.

Doing anything like automatically unpacking them on addition, or rescaling on multiplication, runs the risk of loosing references to any intermediate results (which we again can’t distinguish between temporary variables in intermediate steps and assigned variables)

domfournier:
> Doing anything like automatically unpacking them on addition, or rescaling on multiplication, runs the risk of loosing references to any intermediate results (which we again can’t distinguish between temporary variables in intermediate steps and assigned variables)

I don't know man, sounds like just a personal preference. I don't see coersing type as being a huge deal, given that the original instance is still present, simply wrapped inside a ScaleFunction.function.

domfournier:
> Then `m3` would actually have no reference to `m2` anywhere. And let's say you wanted to manipulate `m2`, `m3` would never know about it.

I don't think practically we really need to keep track of all multiplications that have happened. In the case of a "scaled joint inversion" the multiplicators can change 10s of times, and if we don't  over write thee would end up with some very deep nesting.

domfournier:
@santisoler What do you think? Still go ahead and remove the reference to `multipliers`?

PR #1565, opened 2024-10-30
Only warn about default solver when set in simulations

### LINKED ISSUE #1563: warning with default solver
I appreciate the idea of wanting to alert users that a default solver is being used, but it is odd that if I am writing a tutorial or example and want to explicitly set the default solver, that I get a warning (e.g. in the following figure). A warning usually implies that I should change my code, but here, it is just information. For the purposes of a tutorial, I then opt not to use the handy functionality of getting a default solver because having a warning message pop up is distracting (and I don't want to be promoting that users should suppress all warnings). 

image

So, then the question is what to do. I would suggest that we remove the warning and instead include a statement in the inversion log about what solver is being used (we could also add this as a statement that is printed out when running a "verbose" simulation, but I know that we don't have that option implemented uniformly across all simulations at the moment). Totally open to other suggestions too, but I think that when a user explicitly calls `get_default_solver` that we shouldn't throw a warning at them.

santisoler:
#### Summary

Add a `warn=False` argument to the `get_default_solver()` function. Only when `warn=True`, the warning about the setting of the default solver is raised. This makes the function not to raise a warning when called with the default arguments. When default solvers are being used within simulations, call the function with `warn=True` to let users know about the solver being automatically selected. Update tests accordingly.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fix #1563

### DIFF
--- a/simpeg/potential_fields/magnetics/simulation.py
+++ b/simpeg/potential_fields/magnetics/simulation.py
@@ -1612,7 +1612,7 @@ def MagneticsDiffSecondaryInv(mesh, model, data, **kwargs):
 
     # Create an optimization program
     opt = optimization.InexactGaussNewton(maxIter=miter)
-    opt.bfgsH0 = get_default_solver()(sp.identity(model.nP), flag="D")
+    opt.bfgsH0 = get_default_solver(warn=True)(sp.identity(model.nP), flag="D")
     # Create a regularization program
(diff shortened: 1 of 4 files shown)


lheagy (approved):
Nice, thanks @santisoler!

jcapriot:
Looked good on my end too, thanks for taking care of this.

PR #1581, opened 2024-11-13
Have an option to take a step when the Linesearch breaks

lheagy:
#### Summary
As we chatted about in the [November 13 meeting (and earlier!)](https://curvenote.com/@simpeg/meeting-notes/11-13-2024/) for many types of non-linear problems it can be useful to allow the inversion to take a step even if the line search fails. This is a first cut at allowing that. I used the keyword `aggressive_stepping` for the flag, but happy to think about this with others' input.  

#### PR Checklist
      to a Pull Request


thibaut-kobold:
ChatGPT is telling me that such a line search method would be called a "non-monotone line search"; and there seems to be some litterature on the topic under that name.

simpeg/optimization.py
```diff
@@ -390,9 +393,12 @@ def evalFunction(x, return_g=False, return_H=False):
             p = self.scaleSearchDirection(self.searchDirection)
             xt, passLS = self.modifySearchDirection(p)
             if not passLS:
-                xt, caught = self.modifySearchDirectionBreak(p)
-                if not caught:
-                    return self.xc
+                if self.aggressive_stepping is False:
```
domfournier, line 396:
```suggestion
                if not self.aggressive_stepping:
```

domfournier (approved):

simpeg/optimization.py
jcapriot: Or `strict_decrease`? (Current behavior would is `True`

lheagy: or maybe `require_decrease` (with the default being `True`). That might be a touch more intuitive than `strict`?

thibaut-kobold: `require_decrease` sounds the better option to me.

lheagy: With loose consensus, I updated this to `require_decrease`. If folks are okay with that, then we can go with it

PR #1582, opened 2024-11-18
Move simulation solver from base simulation to PDE simulation

jcapriot:
#### Summary
Move `solver` attribute to the `BasePDESimulation`.

#### PR Checklist
* [x] If this is a work in progress PR, set as a Draft PR
* [x] Linted my code according to the [style guides](https://docs.simpeg.xyz/latest/content/getting_started/contributing/code-style.html).
* [x] Added [tests](https://docs.simpeg.xyz/latest/content/getting_started/contributing/testing.html) to verify changes to the code.
* [x] Added necessary documentation to any new functions/classes following the
      expect [style](https://docs.simpeg.xyz/latest/content/getting_started/contributing/documentation.html).
* [x] Marked as ready for review (if this is was a draft PR), and converted 
      to a Pull Request
* [x] Tagged ``@simpeg/simpeg-developers`` when ready for review.

#### What does this implement/fix?
Not all simulations need `solver`, leading to awkward setting of `solver=None` on `LinearSimulation`, and also on the layered 1D simulations. This move the `solver` attribute to the `BasePDESimulation` (the ones that actually need a solver).

PR #1585, opened 2024-11-18
Change location of `mesh` attribute

jcapriot:
#### Summary
Removes the `mesh` attribute of `BaseSimulation` as not all simulations require meshes.

#### PR Checklist
      to a Pull Request

Not all simulations require a `mesh`, so this makes it more explicit which simulations actually require it, and I've added further constraints on mesh types for simulations which are restricted on their mesh types allowed.

#### Additional Information
This branches off of #1582. so should be merged after that one.


simpeg/flow/richards/empirical.py
```diff
@@ -34,12 +36,19 @@ class NonLinearModel(props.HasModel):
     """A non linear model that has dependence on the fields and a model"""
 
     counter = None  #: A simpeg.utils.Counter object
-    mesh = None  #: A discretize Mesh
 
     def __init__(self, mesh, **kwargs):
         self.mesh = mesh
         super(NonLinearModel, self).__init__(**kwargs)
 
+    @property
```
jcapriot, line 44: I did this because it already accepted a mesh argument, but I don't think this is ever used anywhere, and this base class doesn't really do anything anyways and should likely just be removed.

santisoler (changes requested):
Looks great @jcapriot!

I've just have some suggestions to make the tests over PF simulations more clear, specially after the changes introduced in this PR. I pushed these suggestions to another branch, check them out in d51891d. Feel free to cherry-pick it if you like it.

PR #1588, opened 2024-11-25
Improve dot products in `PGIsmallness.__call__` and update docstring

johnweis0480:
#### Summary
Change the `__call__` method in PGI smallness to avoid for loops through numpy arrays. Fix the docstring to reflect that physical properties can still be correlated if `approx_eval=True` and `approx_gradient=True`.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/regularization/pgi.py
+++ b/simpeg/regularization/pgi.py
@@ -574,7 +574,8 @@ class PGIsmallness(Smallness):
             ) and not self.non_linear_relationships:
                 r = np.zeros_like(r0)
                 for i in range(self.gmm.n_components):
-                    r[membership == i] = r0[membership == i] * self.gmm.precisions_[i].T
+                    selection = membership == i
+                    r[selection] = r0[selection] * self.gmm.precisions_[i].T
                 r = mkvc(r)
 
             else:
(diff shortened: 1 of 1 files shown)


simpeg/regularization/pgi.py
```diff
@@ -70,10 +70,10 @@ class PGIsmallness(Smallness):
         ``tensor``, ``QuadTree`` or ``Octree`` meshes.
     approx_gradient : bool
         If ``True``, use the L2-approximation of the gradient by assuming
-        physical property values of different types are uncorrelated.
```
thibaut-kobold, line 73: Oh yeah I remember correcting that statement somewhere else (maybe directives).
Disclaimer: I did not write those docstring. But the fault lies with me if there are inaccurate (I should have wrote them in the first place). That is to say, the person who wrote those did to their best of their understanding, but you should feel free to improve it.

santisoler (approved):
Looks great! Thanks @johnweis0480 for pushing this. I'm merging this!

~I'm not quite sure if the word distinct is the right replacement for uncorrelated in this case. I'd like to hear @thibaut-kobold's opinion on that change.~

thibaut-kobold:
> Looks great! Thanks @johnweis0480 for pushing this.
> 
> I'm not quite sure if the word _distinct_ is the right replacement for _uncorrelated_ in this case. I'd like to hear @thibaut-kobold's opinion on that change.

I am fine with it.
"uncorrelated" was wrong (what does that even mean in this context?). "Distinct" is better description of when the L2 approximation is correct (when the physical properties distribution of each units are not overlapping).

santisoler:
haha I commented that before seeing your review (sorry!). Then I removed that line from my comment because you already provided an answer. Thanks for that super swift reply haha.

santisoler:
Restored and striked out!

Issue #1227, opened 2023-05-26
naming of physical properties 


lheagy:
Right now, we follow the convention that we are naming physical properties by their greek letter. This isn't necessarily obvious for users and could lead to some challenging scenarios. e.g. 
- dielectric permittivity is `epsilon`, but epsilon is also used in places to refer to a small number used as a threshold or stabilizer. 
- `rho` is also overloaded. It means density in the gravity simulations and resistivity in DC / EM  

I suggest we transition to calling properties by their physical properties. E.g.
- `density`
- `susceptibility`
- `conductivity` 
- `permeability` 
- `permittivity` 
- ...

PR #1591, opened 2024-12-10
Common physical props

jcapriot:
#### Summary
Move physical properties to base classes meant for inheritance.

#### PR Checklist
      to a Pull Request

#### Reference issue
Working towards #1429, and #1227 

Moving the physical properties to base classes to reduce duplication, and planning ahead for their future rename.
The goal for doing this here is to have a single location to place deprecations on the rename.


simpeg/electromagnetics/frequency_domain/simulation.py
```diff
@@ -64,24 +62,19 @@ class BaseFDEMSimulation(BaseEMSimulation):
     """
 
     fieldsPair = FieldsFDEM
-    permittivity = props.PhysicalProperty("Dielectric permittivity (F/m)")
 
     def __init__(
         self,
         mesh,
         survey=None,
+        *,
```
jcapriot, line 70: I've done this to several classes to make it more clear in the source code which parameters should only be set as keywords.

jcapriot:
@santisoler I think this is about ready now, I'd appreciate some feedback. I'll add documentation for the classes soon.

One of the things I'm still unsure about is what to do with the one off physical properties that are in unique classes (that are not necessarily physical properties, but more like parameterizations of physical properties,)

jcapriot:
I decided to move the parameters I was suspicious of to the common physical properties, as more than one simulation used them.
However, there is not (at the moment) a way to change a physical property's invertibility, for the classes where should be reused.

santisoler (approved):
Looks good! I think we should include the physical properties as arguments for the constructor of simulations as well if we are still going to require them for the simulation to run, but I guess you are planning to tackle that in #1593, right?

The other thing that looks a little bit weird is having the simulation classes as childs of physical property classes. I understand the technicality of it (so they inherit the properties), but from an abstraction point of view it kind of looks weird. I'm nit picking, and I wouldn't block this PR because of that, but I wanted to mention it.

simpeg/electromagnetics/base_1d.py
```diff
@@ -35,7 +37,9 @@
 ###############################################################################
 
 
-class BaseEM1DSimulation(BaseSimulation):
+class BaseEM1DSimulation(
+    BaseSimulation, ElectricalConductivity, MagneticPermeability, LayerThickness
```
lheagy, line 41: I am raising this as a point of conversation... inheriting 4 classes introduces a fair bit of complexity for someone to understand when learning the codebase. I appreciate that the physical property classes are lightweight, but none-the-less, it does create a fair bit to navigate through to understand that we are going through 4 classes to instantiate this. Is there any way we could simplify this? 

Having physical properties be `properties` of the class makes conceptual sense. Having a `simulation` inherit a `physical property` is a bit odd. I realize that it brings with it the benefits of not having to duplicate documentation. 

If you really don't see an alternative, then I understand. But I wanted to raise it as something to discuss, as part of the goal of SimPEG is as a learning tool, and this introduces a new layer of complexity.

jcapriot:
> I am raising this as a point of conversation... inheriting 4 classes introduces a fair bit of complexity for someone to understand when learning the codebase. I appreciate that the physical property classes are lightweight, but none-the-less, it does create a fair bit to navigate through to understand that we are going through 4 classes to instantiate this. Is there any way we could simplify this?
> 
> Having physical properties be `properties` of the class makes conceptual sense. Having a `simulation` inherit a `physical property` is a bit odd. I realize that it brings with it the benefits of not having to duplicate documentation.
> 
> If you really don't see an alternative, then I understand. But I wanted to raise it as something to discuss, as part of the goal of SimPEG is as a learning tool, and this introduces a new layer of complexity.

One of my overall goals is to "cleanup" the simulations.

There's a few points to I think this highlights for discussion,

First on the level of complexity of the codebase. I feel it's worth discussing where we should aim for, and there are a few different levels of people who interact with the code. I see at least three:
1) Someone who uses simpeg (should be easy and require hopefully basic understandings of python) I see this as the average user
2) Someone who wants to extend simpeg to suit their own needs (Moderate knowledge of python). Should be clear (not necessarily "easy", but at least straightforward).
3) Simpeg developers (Higher knowledge of python expected). We internally make use of many different aspects of python that require a higher level of understanding in the first place, things like: Multiple Inheritance, Metaclasses, Wrappers, etc.

I feel like the internals of `simpeg` should be straightforward to follow, but not necessarily simple. There are many great aspects of python that we get to take advantage of to reduce code duplication. Having less code to maintain is always helpful.

One of the biggest issues I currently see with our simulations is not that they have complicated inheritance, it's that they simply inherit too many properties from those parents that *most* people would never need to interact with. (`sigmaDeriv`? `getRHSDeriv`, `MccSigma`, etc.). All of those extra properties adds so much noise to the simulations making them hard to use, hard to read, and hard to extend.

Another issue is that the "base" items live all over the place. There are fundamental bits of Simulations, Physical Properties, Models, Sources, Receivers, Surveys, etc.) based items in multiple of the **first level** subpackages of `simpeg`. These are things that an average user would never interact with. I would be a proponent of these `base` classes into the `base` submodule and separating them a bit more than they already are.


Now moving onto this PR and my overall goals for unifying them here.
The biggest purpose of this is to reduce code and documentation duplication, things that would need to be duplicated if this wasn't in a common place:
* Many classes implement the same physical properties, and having them defined in one place makes it a little easier to document them throughout
* Any extra bits that would be common (like issuing the same warning on all simulations that would use permittivity). 
* We should also issue errors on instantiation if a user passes both `sigma` and `sigmaMap` (or if two reciprocal properties are passed).
* More forward looking
    * If we want to change the names of the properties on the classes to not be symbol based (e.g. `sigma` $\rightarrow$
       `conductivity`), it's much easier to do if those properties are only defined in one location. I believe it will be easier to
       deprecate it this way and then easier to *test* the deprecations this way.

Then this last bit is more of a suspicion of mine, but I hope it also makes sense. I would trust someone who wants to make their own simulation to inherit a class rather than to make sure they use the correct combination of `PhysicalProperty`, `Mapping`, `Derivatives`, and `Invertible` classes internally.

Some other alternatives:
* It might work to just have some predefined `PhysicalProperty`, `Mapping` and `Derivative` objects here instead that developers could grab.
    * That would get us some removal of code duplications, (Users could be a little more sure of themselves using a correct combination of `PhysicalProperty`, `Mapping` etc.) and make the attributes on the classes have the same documentation.
    * Don't get to re-use code for checking any other aspects about them though (Deprecations, errors, etc.).
 

So basically I believe moving them to common classes will reduce the overhead in the future.

simpeg/electromagnetics/frequency_domain/simulation.py
lheagy: could we add comments throughout where we use * in the code to help developers understand why this is here and what keywords it is meant to catch? That would be helpful for contributors who are learning the codebase by reading it

simpeg/flow/richards/empirical.py
```diff
@@ -26,96 +25,35 @@ def _partition_args(mesh, Hcond, Theta, hcond_args, theta_args, **kwargs):
...
+class Haverkamp_theta(WaterRetention):
 
     alpha, alphaMap, alphaDeriv = props.Invertible("")
 
     beta, betaMap, betaDeriv = props.Invertible("")
 
     def __init__(
         self,
-        mesh,
+        *args,
```
lheagy, line 42: similarly here, could we add a comment to help developers who are learning by reading the code?

lheagy (commented):
Thanks for pushing this forwards @jcapriot and apologies for taking soooo long to get back to you on this! I like the idea and I can see how this will simplify name changes in the future. My main comment is that if we could add a few comments for where the `*` is in the init to just explain what is happening, that would be helpful. A lot of folks learn SimPEG by reading the code, and it might not be obvious to all what is happening, so if we could help provide some guidance through comments, I think that would be helpful

santisoler:
Leaving here a few thoughts on this PR.

I like some ideas implemented here:

* I'm in favor of using the asterisk argument to force users to pass arguments in a more verbose way. We could explain how that argument works in a tutorial, and even leave a comment in the code to address @lheagy's comment regarding confusion created by its usage.
* I like the idea of having a single place where physical properties are defined and validated (e.g. raising error if resistivity and conductivity are passed at the same time).

But there are a few things I don't quite like:

* The idea that simulations inherit from physical properties. I understand how this enables the simulations to host less code, but from the perspective of OOP it sounds a bit weird: a common abstraction is that children classes are a particular case of the parent ones. Seeing that a simulation class inherits from a physical property could confuse people into thinking how these two work together and what's the difference between them. I'm aware of other abstractions for inheritance (like mixins for example), but those cases are usually explicit in the name of the inherited class. But in this particular case one could misunderstand that a "FDEM simulation is a `ElectricalConductivity`".
* Removing the accepted physical properties (and mappings) from the signature of the `__init__` methods. I think it doesn't hurt listing the physical properties a simulation can take in the signature of the `__init__` method, and the benefits are quite big: autocomplete in IDEs and notebooks can suggest arguments, easier to see the options in the signature without having to read the whole docstring, easier to understand the code.

**Some ideas**

What if instead of relying on inheritance, we still define those physical property classes in a different way and use them within the simulations. Since we are already using metaclasses for the definition of the physical properties, mappings, and derivatives, we could have specific physical property classes for each.

Roughly something like (also including the idea of merging the physical property and the mapping):

```python
class Simulation(BaseSimulation):
    _conductivity_class = ElectricalConductivity

    def __init__(self, mesh, survey, *, conductivity=None, resistivity=None):
        self._conductivity = self._conductivity_class(conductivity, resistivity)
```

The metaclass could define getter and setter for `conductivity` and `resistivity` to `Simulation`, and the `HasModel` could assign the `model` setter and getter. The `ElectricConductivity` class could be a child of an abstract `PhysicalProperty` class so it's easy to spot by the metaclass. The `ElectricConductivity` object could host either the values of conductivity or resistivity, or their mappings. It could also relate the two through a reciprocal operation and validate that users are not defining conductivity and resistivity at the same time.

I know there might be quite a lot of details around this kind of alternative solution, but just throwing this as an idea to avoid the inheritance. I'm open to other ideas and suggestions too!

PR #1592, opened 2024-12-11
Use zizmor to lint GitHub Actions workflows

santisoler:
Make use of `zizmor` to lint the GitHub Actions workflows under the `.github/workflows` directory for common security vulnerabilities. Add a new `check-actions` target in the `Makefile` that runs `zizmor` on those files. Add a new GitHub Action that runs `zizmor`. Add `zizmor` to the `environment.yml`.

#### PR Checklist
      to a Pull Request

#### Additional information

Inspired by: https://github.com/fatiando/community/issues/159

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,0 +1,35 @@
+# Lint GitHub Actions for common security issues using zizmor.
+# Docs: https://woodruffw.github.io/zizmor
+
+name: lint-actions
+
+# Only run on PRs and the main branch.
+# Pushes to branches will only trigger a run when a PR is opened.
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install zizmor
+        run: python install zizmor
+
+      - name: List installed packages
+        run: python -m pip freeze
+
+      - name: Lint GitHub Actions
+        run: make check-actions
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,16 @@
 STYLE_CHECK_FILES = simpeg examples tutorials tests
+GITHUB_ACTIONS=.github/workflows
 
-.PHONY: help docs check black flake
+.PHONY: help docs clean check black flake flake-all zizmor
 
 help:
 	@echo "Commands:"
 	@echo ""
-	@echo "  check       run code style and quality checks (black and flake8)"
-	@echo "  black       checks code style with black"
-	@echo "  flake       checks code style with flake8"
-	@echo "  flake-all   checks code style with flake8 (full set of rules)"
+	@echo "  check         run code style and quality checks (black and flake8)"
+	@echo "  black         checks code style with black"
+	@echo "  flake         checks code style with flake8"
+	@echo "  flake-all     checks code style with flake8 (full set of rules)"
+	@echo "  check-actions lint GitHub Actions workflows (with zizmor)"
 	@echo ""
 
 docs:
@@ -31,3 +33,6 @@ flake:
 flake-all:
 	flake8 --version
 	flake8 ${FLAKE8_OPTS} --ignore "" ${STYLE_CHECK_FILES}
+
+check-actions:
+	zizmor ${GITHUB_ACTIONS}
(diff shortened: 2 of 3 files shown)


jcapriot:
From what I've read, this was due to insecure steps in their `pull_request_target` action. I know when I set this up in #1427 I was very careful to ensure that no code in the PR source gets executed. Not to say that I couldn't have made a mistake, but that I was at least aware that it was something to avoid (running PR source code in a secure action), and I did my best to make sure that each action never executed uncontrolled code.

santisoler:
I agree that at the moment the usage of `pull_request_target` trigger does not imposes a security risk in our `pull_request.yml` workflow.

Nonetheless, I'm worried about future developers not being aware of this possible vulnerability. My initial idea was to use the `pull_request` trigger instead, but reviewdog needs writing permissions in this repo to be able to comment on PRs opened from forks. So, I commented the `pull_request_target` trigger so `zizmor` doesn't complain about it, and added a warning about running code from PR branches in that workflow.

Makefile
```diff
@@ -1,14 +1,16 @@
 STYLE_CHECK_FILES = simpeg examples tutorials tests
+GITHUB_ACTIONS=.github/workflows
 
-.PHONY: help docs check black flake
+.PHONY: help docs clean check black flake flake-all zizmor
```
santisoler, line 4:
```suggestion
.PHONY: help docs clean check black flake flake-all check-actions
```

.github/workflows/actions.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: python -m pip install zizmor
+
+      - name: List installed packages
+        run: python -m pip freeze
+
+      - name: Lint GitHub Actions
+        run: make check-actions
+        env:
+          # Set GH_TOKEN to allow zizmor to check online vulnerabilities
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jcapriot, line 36: Th GitHub token will not be (and should not be) available to PRs from external forks.

santisoler: According to GitHub docs on [Using secrets in a workflow](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow):

> With the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository.

For a `pull_request` trigger, the token provides read only privileges if the runner is triggered from a forked repo ([Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)).

So I think this is safe. Even if the PR is coming from a forked repo, the `GITHUB_TOKEN` will allow `zizmor` to check for online vulnerabilities.

This action is similar to the one in zizmor's docs: https://woodruffw.github.io/zizmor/usage/#use-in-github-actions

.github/workflows/pull_request.yml
```diff
@@ -1,5 +1,31 @@
...
+# =========
+# IMPORTANT
+# =========
+#
+# TL;DR:
+#     THIS ACTION SHOULD NOT RUN ANY CODE FROM THE PR BRANCH.
+#
+# This action is triggered after new events in Pull Requests (as the
+# `pull_request` trigger does), but this ones provides the workflow writing
+# permissions to the repo.
```
jcapriot, line 12: I would add a note clarifying that the version of this workflow file that is actually executed is the one on the pull request's TARGET branch, not the one on pull request's SOURCE. 

It's a subtle, but important difference, because it runs from the target branch (usually main), thus why it gets elevated permissions,

jcapriot: We should probably include a link to the original pr this was added in for further reference.

.github/workflows/pull_request.yml
```diff
@@ -1,5 +1,40 @@
...
+# For this particular workflow we need the writting permission in order for
+# reviewdog to publish comments in the PR.
+# zizmor will complain about the `pull_request_target` trigger, so we will
+# ignore it.
+#
+# Worth noting that the runner will execute the steps specified in the version
+# of this workflow file that lives in the **target branch** (usually `main`),
+# not the one in the Pull Request branch. This means that even if a contributor
+# opens a PR with a change to this file, the change won't be executed. This is
+# inteded to avoid third-party contributors to run custom code with high
```
jcapriot, line 29:
```suggestion
# intended to prevent third-party contributors from running custom code with high
```

jcapriot (changes requested):
Requesting some minor changes for grammar.

Also, give the `zizmor` github action a more descriptive file name than `actions.yml`.

jcapriot (approved):
looks good

Issue #1345, opened 2024-02-01
ENH: physical properties as functions?


jcapriot:
### Proposed new feature or change:

How should we handle physical properties that are themselves functions of parameters that are not the model (aka a physical property that is dependent on how it is observed).

As a concrete question:
How should we account for time (or frequency) dependent conductivity?

Clearly a simple solution for the frequency dependent case is to have a single simulation for each frequency we are interested in, (requiring much duplication of internally cached items) but I'm interested in a more general solution.

Say our conductivity is a function of time (as in there might be some induced polarization effects) and we want to simulate time-domain EM data. There is not a good way for us to do that at the moment directly (it would involve creating subclasses of the TDEM simulations, each with their own respective form of the time-dependent conductivity, similar to how the SIP class is specifically using a stretched exponential to handle the time-dependent conductivity).

What if instead of the physical properties being arrays (even in the case of defining a physical property map they evaluate to be arrays), we also allowed them to be functions of some tertiary variable that would then return the correct sigma array. 

i.e.
```Python
def sigma_func(omega):
    return sigma_0 * some_fancy_cole_cole_model(omega)

sim = SimulationTime(sigma=sigma_func)
```

Also somehow combine it with a map:

```Python
def sigma_map_func(m, time):
    return maps.Exp(m) * some_fancy_cole_cole_model(omega)
```

Maybe it might be better to elevate this to a dedicated class type for more reliable checking if the physical property is one of these proposed things:

```Python
class PropertyFunction:
    def __init__(self, function, ...)  # not quite sure what all would go here
        pass
    
    # might also need to define some derivatives of the function for certain simulations?
    # e.g. where previous time-independence was assumed to move conductivity outside a time derivative.
```

Function dependent mapping would likely require some sort of dedicated mapping class:
```Python
class FunctionMapping:
    
    def __init__(self, ...):


    def deriv_m(x, m):
        pass
    ...

# likely used as
sigma = mapping_func(omega, model)
sigma_m_deriv = mapping_func.deriv_m(omega, model)
sigma_t_deriv = mapping_func.deriv_x(omega, model)  # Is this necessary? (Probably)
```

I realize that this is probably a large project involving a lot of changes and selection cases. It would go into a lot of the fundamental mass matrix code and likely require some reformulations for previous simulations that assumed time-independence of a property, but that shouldn't necessarily be a reason to ignore this forever!

My intention would be to use this page as a springboard to discuss potential ideas details and roadblocks.

[...]

PR #1593, opened 2024-12-12
Set physical properties as maps

jcapriot:
#### Summary
Allows physical properties to be directly set as maps

#### PR Checklist
      to a Pull Request

#### Reference issue
More simulation cleanup

Currently in simpeg, we have two ways of interacting with a simulations physical properties.
1) Setting `sim.rho` to be an array
2) Setting `sim.rhoMap` to be a mapping, which means that accessing `sim.rho` will calculate `sim.rho` using the mapping from the model to the physical property.

I'm proposing unifying these two operations into only the physical property attribute itself. I.E. I want to be able to set `sim.rho = mapping`, and have it be functionally the same as interaction 2) above.

``` sim.rho = array```

or

```sim.rho = mapping```

which allows us to then deprecate `sim.rhoMap`.

#### Additional information

A few details:
* The `PhysicalProperty` class is now a subclass of `property`, allowing a more intuitive behavior than before.
    * Previously a `PhysicalProperty` is transformed into a `property` by the `PhysicalPropertyMeta` metaclass. This meant, there was not a good way to modify a physical properties getter or setters (See the awkward hijacking of `__getattr__` in the BaseMagneticPDE simulation).
* If a `PhysicalProperty` is not invertible, assigning a `mapping` is disallowed
* You can change the invertibility of a `PhysicalProperty` on a class-by-class basis.
* PhysicalProperty` derivatives are privatized,
* I think removing public access to the stored `mapping` and the `deriv` function will not impact end users, (for developers there is still a way to access the map, and I think it's a cleaner method for when we must modify the behavior, e.g. the IP class.)

The previous functionality of:
* `props.Mapping`
* `props.Derivative`
* `props.Invertible`
* `props.Reciprocal`

is now unified into `props.PhysicalProperty`

### CODE REVIEW

**Comment by lheagy** [comment]
Thanks for kicking this off Joe, and my apologies for not having time to jump in sooner on this idea. As I mentioned in our last SimPEG meeting, I am quite concerned about the potential confusion that could be introduced by collapsing the physical property (e.g. `rho`) with the mapping (e.g. `rhoMap`). There are a few reasons for my concerns. 

First, mappings are a challenging concept for people to grasp when they are first introduced to SimPEG. This is fair, it is an abstract concept. Right now, we can explain that if you set a `rhoMap` then `rho` is an invertible property in the inversion. If we combine the two, that statement becomes more complex. 

Importantly, this implementation also means that what `sim.rho` is depends on the state. If we instantiate `rho` as a mapping, then there are two possible states 
- if `sim.model = None`, then `sim.rho` is a mapping
- but, if `sim.model is not None` then `sim.rho` is a vector, which is the mapped physical property

I think this is dangerous. It is quite a common pattern to instantiate the simulation class and then access the mapping from there, 

e.g. 
```
sim = tdem.Simulation1D(thicknesses = thicknesses, sigmaMap=sigma_map)
sigma0 = sim.sigmaMap * starting model
```
and after the inversion 
```
sigma_recovered = sim.sigmaMap * mrec
```

This is fine in the current implementation, but if we combine the two, then this could lead to some unexpected behaviours 
e.g. 
``` 
sim = tdem.Simulation1D(thicknesses = thicknesses, sigma=sigma_map)
sigma0 = sim.sigma * starting model  # this is valid and will perform the mapping of our starting model
```
and after the inversion 
```
sigma_recovered = sim.sigma * mrec  # this now makes no sense! After the inversion, the simulation will have a model, so this is multiplying two vectors
```
This example is particularly dangerous if the mapping is a one-to-one mapping (e.g. like an exp map), because if the model is set on the simulation `sim.sigma` will be a vector of the same size as `mrec`. So the two quantities will be valid to multiply. However, the output of that multiplication is meaningless as it would be the mapped physical property times the unmapped model vector. This will be difficult to debug, and there is nothing we can do to guard against it because it would multiplying two numpy vectors. 

It is a common pattern amongst our users to access the mapping from the simulation. We even do this in our own tutorials (e.g. the [SEG NS 2024 tutorial](https://github.com/simpeg/segns-2024-tutorial/blob/main/notebooks/skytem-inversion.ipynb))
image

So I don't think we can reasonably ask our users to unlearn the pattern of accessing the mapping from the simulation. I think we are much better of staying explicit when defining simulations and that `propMap` is always a `Map` and `prop` is a vector defining the physical property.

**Comment by jcapriot** [comment]
The functionality I'm proposing here is a little bit different than what you've interpreted it there.

Basically these things would happen:

If you set `sigma` to something besides a map, it will be a not be dependent on the model.

```
>>> sim.sigma = [1,2,3]
>>> sim.sigma
[1, 2, 3]
```

If you set `sigma` as a map, it becomes a dynamic property dependent on `sim.model`:
```
>>> sim.model = [1, 2, 3]
>>> sim.sigma = maps.IdentityMap()
>>> sim.sigma
[1, 2, 3]
>>> sim.model = [3, 4, 5]
>>> sim.sigma
[1, 2, 3]
```
The only difference between this and what we have now is that we would currently have to write: `sim.sigmaMap = map` instead of  `sim.sigma = map` to achieve this behavior.

If you have not set a model then:
```
>>> sim.sigma = maps.IdentityMap()
>>> sim.sigma
AttributeError: Saying something about sigma was set as a mapping, and requires a model, but a model wasn't set
```

If you haven't given `sigma` a value at all (and it is not an optional property) then it would raise an attribute error.
```
>>> sim.sigma
AttributeError: You haven't given sigma a value nor a mapping...
```

-------

I suspect that `PhysicalProperty` attributes are the least understood aspects of `simpeg` simulations that users interact with everyday.

I do see your point about allowing users retrieve the assigned mapping to a simulation, for which this method would add a public method along the lines of `sim.get_property_mapping('sigma`).

One of the clunky things that I've found while writing `simpeg` is especially apparent on the IP, Spectral IP, and Self-Potential methods. You can't really change whether a `PhysicalProperty` is invertible on subclasses. These classes all subclass DC simulations (where `sigma` is invertible) yet `sigma` is not invertible on those child classes. Inheriting the DC sims reduce a significant amount of code duplication, but there's no way to remove `sigmaMap` from the class's attributes. This functionality makes it possible to do this.

The cleanest thing I like about this approach (which is something that I think we could change without doing this entire approach) Is that `PhysicalProperty` implements the `Descriptor` protocol (see https://docs.python.org/3/howto/descriptor.html for more details on descriptors). Right now we are basically using a `MetaClass` to mimic all of the functionalities of descriptors instead of just using descriptors. I know metaclasses took me a while to understand and I suspect it is also beyond the scope of many python developers. Implementing the `PhysicalProperty` items as descriptors would allow us to get rid of the `PhysicalPropertyMetaclass` altogether in the future, hopefully simplifying this bit of code.

Also, something I'd love to tackle in the future is to really extend the power of `PhysicalProperty` beyond just allowing for maps, i.e. #1345. So having `sigmaMap` could be limiting us!

jcapriot:
Closing in favor of working towards #1601

PR #1596, opened 2024-12-17
Coverage upload on failed test

jcapriot:
#### Summary
Make the coverage step run if the succeeded OR failed.

This makes the azure-pipeline codecov upload step run if the tests succeed or fail, as it can still be useful to see coverage analyses of failed tests.


jcapriot:
Still need to make the artifact upload also trigger

PR #1601, opened 2025-01-28
Parametrized physical properties

jcapriot:
#### Summary
Implements a parametrized interface to `PhysicalProperties`

#### PR Checklist
      to a Pull Request

#### Additional information
Continuing on from #1593


jcapriot:
I think the new interface is coming together, here is a summary of the changes:

New interface
---------------
On the constructor you can do:

```
# Initialize with a given value
sim = Simulation(sigma=[1,2,3])

# or initialize as a parametrization (mappings for now)
sim = Simulation(sigma=mapping)
```

Or changing it after creating a simulation:
```
# setting the sigma property as a value
sim.sigma = [1, 2, 3]

# parametrizing sigma
sim.parametrize(sigma=mapping)
```

When properties are `parametrized` they are dependent on the simulation's `model` attribute, and derivatives w.r.t. that property are then enabled for inversion. These physical properties also *require* a `model` and will error if the simulation doesn't have one assigned.

> A non-invertible property cannot be parametrize

Added methods
===========
These public methods are added to the `HasModel` class to work together with `PhysicalProperties`

* `HasModel.parametrize(**kwargs)`, the aforementioned item.
* `HasModel.parametrizations' a `ParametrizationList* ` storing the mapping used to parametrize each property
    * Accessible with attributes named after the `PhysicalProperty` class.
    
      ```
      map1 = ExpMap()
      sim.simulation(sigma = a[p1)
      # And it can be retrieved using 
     maps1 == sim.parametrizations.sigma
     ```

* `HasModel.physical_properties` A class method to return a dictionary of `attr:PhysicalProperty` of all the physical properties defined on a class.
* `HasModel.invertible_properties` A class method to return a dictionary of `attr:PhysicvvalProperty` of all the **invertible** physical properties defined on a class.

Derivatives
=======
Getting the derivative of a property is now handled through a function, as opposed to a property, allowing us to pass through multiplication vectors to the parametrizations

* `HasModel._prop_deriv(self, attr)`

 Unless anyone has strong feelings otherwise, I'm in favor of keeping this private to keep the public API cleaner.

I'll add more details in the future

jcapriot:
These are things that would be deprecated or removed with this PR.

Deprecations
--------------
On initialization
* `Simulation(sigmaMap=mapping`)` would be deprecated (in favor of the above interface.

Attribute Assignment:
* `sim.sigmaMap = mapping` would go to `sim.parametrize(sigma=mapping)`

Attribute Access:
* `mapping = sim.sigmaMap` would go to `sim.parametrizations.sigma`
* `deriv = sim.sigmaDeriv` is no longer an attribute, and is instead a function `sim._prop_deriv("sigma", m, v=None)`.
  This is basically a helper for doing `sim.parametrizations.sigma.deriv(m, v)`, that also accounts for reciprocals.

Removals
-----------
* `simpeg.props.Mapping`, `simpeg.props.Derivative`, `simpeg.props.Invertible`, `simpeg.props.Reciprocal`.  These items were used together with the MetaClass on HasModel to create all of the `property` attributes that we previously had. As these were used internally and will not work with the new interface, they will throw a SyntaxError if someone attempts to use them.q

examples/10-pgi/plot_inv_1_PGI_Linear_1D_joint_WithRelationships.py
```diff
@@ -112,12 +112,12 @@ def g(k):
 relatrive_error = 0.01
 noise_floor = 0.0
 
-prob1 = simulation.LinearSimulation(mesh, G=G, model_map=wires.m1)
+prob1 = simulation.LinearSimulation(G=G, model_map=wires.m1)
```
lheagy, line 115: similarly in this file, are these updates meant to be a part of this pr?

examples/10-pgi/plot_inv_0_PGI_Linear_1D.py
```diff
@@ -66,7 +66,7 @@ def g(k):
 mtrue[mesh.cell_centers_x > 0.6] = 0
 
 # simpeg problem and survey
-prob = simulation.LinearSimulation(mesh, G=G, model_map=maps.IdentityMap())
+prob = simulation.LinearSimulation(G=G, model_map=maps.IdentityMap())
```
lheagy, line 69: Is this change intended to be in this PR? I am not sure it is in this scope? or maybe I am missing something

examples/10-pgi/plot_inv_1_PGI_Linear_1D_joint_WithRelationships.py
lheagy, line 115: similarly here, are these meant to be a part of this pr? it seems a bit different in scope?

simpeg/props.py
```diff
@@ -1,228 +1,284 @@
...
+
+        # If I'm optional, return my default value
+        if self.optional:
+            return self.default
+        # If I had an optional reciprocal
+        if recip and recip.optional:
+            val = recip.default
+            # If it wasn't None, try to invert it.
+            if val is not None:
+                val = 1 / val
```
lheagy: do we need a check that this is a scalar (or at least isotropic)? This wouldn't work if anisotropic

simpeg/props.py
```diff
@@ -1,228 +1,284 @@
...
-                    scope.name, value, shape=scope.shape, dtype=scope.dtype
-                )
-                if value.ndim == 0:
-                    value = value.item()
+    def __set__(self, obj: "HasModel", value):
+        if self.fset is not None:
+            self.fset(obj, value)
+        else:
+            self._fset(obj, value)
+        obj._remove_parametrization(self.name)
```
lheagy, line 348: if I am following correctly, this would be hit if someone swapped out a mapping with a value or another mapping, is that right? If so, we might want to throw a warning? Changing something from an invertible parameter to not has some potentially involved implications

simpeg/props.py
```diff
@@ -328,36 +332,259 @@ def __new__(mcs, name, bases, classdict):
...
+        required = not prop1.optional and not prop2.optional
+        if required and inp1 is None and inp2 is None:
+            warnings.warn(
+                f"Setting both `{prop1.name}` and `{prop2.name}` to None for `{type(self).__name__}`'s required "
+                f"physical properties is deprecated behavior. This message will be changed to an error in simpeg "
+                f"version X.X",
+                FutureWarning,
+                stacklevel=3,
+                # f"`{type(self).__name__}` requires one of `{prop1.name}` or `{prop2.name}`"
+            )
```
lheagy, line 744: A question on warning / error handling, will the user first see that they need to change `sigmaMap=mapping` to `sigma=mapping`? If not, this warning might be confusing

simpeg/props.py
```diff
@@ -483,3 +708,140 @@ def model(self):
...
+            ) is not None:
+                return maps.ReciprocalMap() @ recip_map
+        return None
+
+    @prop_map.setter
+    def prop_map(self, value):
+        cls_name = type(self).__name__
+        warnings.warn(
+            f"Setting `{cls_name}.{old_map}` directly is deprecated. Instead register a parametrization with `{cls_name}.parametrize('{new_name}')`",
+            FutureWarning,
```
lheagy, line 1120: Can we also add that the user can set the mapping in the class instantiation via `prop=mapping`. I think that is the behaviour that we will want to encourage (I suspect most people will find the transition from `sigmaMap=mapping` --> `sigma=mapping` much easier than going through the `class.parameterize .....` approach

lheagy (commented):
Thanks for bringing this up again @jcapriot and sorry it took me so long to get back to this. I think this idea can work. I have just done a light pass to ask a few questions. Would you mind giving a demo again at an upcoming SimPEG meeting? It would be valuable to share this idea again with the group and get input from others. I think this could simplify things for users, but would appreciate chatting through implications for developers and people working to understand the code

simpeg/props.py
```diff
@@ -328,36 +332,259 @@ def __new__(mcs, name, bases, classdict):
...
             nested_modelers.update(getattr(parent, "_nested_modelers", set()))
 
-        newcls._all_map_names = map_names
         newcls._nested_modelers = nested_modelers
+        newcls._has_nested_models = len(nested_modelers) > 0
 
         return newcls
 
 
+class ParametrizationList:
```
santisoler, line 615: It might be nice to add a `__str__` method to this class, to easily discover the available parametrizations in the list.

santisoler, line 615: On this note, I was initially going to suggest to make this a child of `collections.UserDict` to inherit the interface of dictionaries. But maybe it will require even more work to ensure that the list is "read only".

simpeg/props.py
```diff
@@ -328,36 +332,259 @@ def __new__(mcs, name, bases, classdict):
...
+        self._fields = fields
+
+    def __getitem__(self, key):
+        return self._fields[key]
+
+    def __setattr__(self, key, value):
+        if key in self.__slots__:
+            super().__setattr__(key, value)
+        elif key in self._fields:
+            raise AttributeError(f"'Cannot set attribute '{key}'") from None
```
santisoler: We could potentially extend this error message, explaining that the `ParametrizationList` is meant to be read-only and we cannot modify the existing attribute.

simpeg/props.py
```diff
@@ -328,36 +332,259 @@ def __new__(mcs, name, bases, classdict):
...
+            Each attribute keyword is parametrized by the given relationship.
+
+        Examples
+        --------
+
+        >>> sim.parametrize(sigma=maps.ExpMap())
+
+        """
+        for attr, parametrization in kwargs.items():
+            if not isinstance(parametrization, maps.IdentityMap):
```
santisoler, line 787: I'm wondering if situations like these ones might be a good opportunity to user `typing.Protocol`s decorated with the `@runtime_checkable`.

I understand that we are currently using the `IdentityMap` as the base class for every map (something that it's weird, since all the other maps are not the identity one). So we could potentially make the distinction here that we expect the `parametrization` to be a mapping but not necessarily an `IdentityMap`.

simpeg/props.py
```diff
@@ -328,36 +332,259 @@ def __new__(mcs, name, bases, classdict):
...
+
+    def _remove_parametrization(self, attr):
+        """Remove attr's parametrization"""
+        # just silently succeed here as it is an internal method.
+        self.parametrizations._fields.pop(attr, None)
+
+    def _prop_deriv(self, attr):
+        # TODO Add support for adjoints here and on mapping derivatives
+        # TODO Add support for passing v to the parametrization
+        paramers = self.parametrizations
```
santisoler: Typo here? (if so, remember to update the variable name in the rest of the method)

```suggestion
        parameters = self.parametrizations
```

simpeg/props.py
```diff
@@ -403,16 +630,25 @@ def model(self):
...
 
     @model.setter
     def model(self, value):
         if value is not None:
             # check if I need a model
-            if not self.needs_model and not self._has_nested_models:
-                raise ValueError(
-                    "Cannot add model as there are no active mappings"
-                    ", choose from: ['{}']".format("', '".join(self._all_map_names))
+            paramers = self.parametrizations
```
santisoler: Same typo here?
```suggestion
            parameter = self.parametrizations
```

simpeg/base/pde_simulation.py
```diff
@@ -607,21 +613,78 @@ def MeI(self):
...
+
+    @rho.deleter
+    def rho(self):
+        prop = type(self).rho
+        if hasattr(self, prop.private_name):
+            delattr(self, prop.private_name)
+
+        for mat in self._clear_on_rho_update:
+            if hasattr(self, mat):
+                delattr(self, mat)
```
santisoler, line 672: Are we planning to keep these properties implemented in this way? Or should `HasModel` add these to the class based on the `PhysicalProperty`?

santisoler (commented):
I like the ideas in this PR! I think I mentioned this several times, but I would also like to merge the physical property and their mapping into a single argument in simulations.

I also like how the PR groups together all things related to a physical property into a single object (instead of having `sigma`, `sigmaMap`, `sigmaDeriv`, for example).

Another thing I quite like is how easy it would be to know the shape of the expected `model` directly from the parametrization list. This would be a great feature to have when using SimPEG simulations in the [new inversion framework](github.com/simpeg/inversion-design-ideas/), where the `DataMisfit` tries to get the `n_params` property from the `simulation` itself. Currently that's a bit of a challenge that varies across different simulations.

I'm leaving a few comments below so we can move forward with this.

## Error messages while setting physical property as mapping

I would improve the error message when trying to set a physical property as a mapping. For example, if I run `simulation.sigma = conductivity_mapping` I get the following error:

```
TypeError: '`Simulation3DNodal.sigma`' must be array_like with data type of <class 'float'>, got <class 'simpeg.maps._injection.InjectActiveCells'>
```

Maybe we could make this error clearer, and point users on using the new `simulation.parametrize(sigma=conductivity_model)` interface.


## Making `physical_properties` and `invertible_properties` properties?

WOuld it be possible to make the `physical_properties` and `invertible_properties` as (Python) properties of the class, so we could access them through:

```python
simulation.physical_properties
```

Instead of having to call them (as methods)?


## Updating deprecation admonitions

I noticed a few admonitions in docs warning users about deprecations of the old properties. Most of them have old versions of SimPEG. I would update them to the next release: v0.26.0.


## Connecting *exclusive* physical properties?

I think I briefly mentioned this idea in the last meeting, but I'll try to expand more on it here.

I was wondering if it would be best to group *exclusive* physical properties (exclusive in the sense of XOR, i.e. one can either pass one XOR the other) together in a single `PhysicalProperty`.

Just brainstorming, but something like:

```python
sigma = props.PhysicalProperty(
    "Electrical conductivity (S/m)",
    reciprocal=props.PhysicalProperty( "Electrical resistivity (Ohm m)"),
    shape=[(), (1,), ("mesh.n_cells",)],
)
```

The `HasModel` can inject the two properties and the setters for each one of them. This way we could avoid having to handle the two of them separately in a few different places.

I don't intend to change the public interface, just refactor some of the internals so two reciprocal physical properties are naturally connected to each other.

Is it a bad idea?

simpeg/props.py
jcapriot: That's a good idea.

simpeg/base/pde_simulation.py
jcapriot: Right now, it was mostly just a mechanism to invalidate the cache for those properties. I think it would best to update the caching mechanism on the HasModel class, but likely better for a separate PR.

jcapriot: Handling the shape can likely be done better by defining where properties exist ("mesh.cell_centers", "mesh.faces", etc.) once we add that capability..

simpeg/props.py
jcapriot: Can update it to be anisotropic aware. At most it would be a block operation of inverting 3x3 matrices.

jcapriot:
> I like the ideas in this PR! I think I mentioned this several times, but I would also like to merge the physical property and their mapping into a single argument in simulations.
> 
> I also like how the PR groups together all things related to a physical property into a single object (instead of having `sigma`, `sigmaMap`, `sigmaDeriv`, for example).
> 
> Another thing I quite like is how easy it would be to know the shape of the expected `model` directly from the parametrization list. This would be a great feature to have when using SimPEG simulations in the [new inversion framework](github.com/simpeg/inversion-design-ideas/), where the `DataMisfit` tries to get the `n_params` property from the `simulation` itself. Currently that's a bit of a challenge that varies across different simulations.
> 
> I'm leaving a few comments below so we can move forward with this.
> 
> ## Error messages while setting physical property as mapping
> I would improve the error message when trying to set a physical property as a mapping. For example, if I run `simulation.sigma = conductivity_mapping` I get the following error:
> 
> ```
> TypeError: '`Simulation3DNodal.sigma`' must be array_like with data type of <class 'float'>, got <class 'simpeg.maps._injection.InjectActiveCells'>
> ```
> 
> Maybe we could make this error clearer, and point users on using the new `simulation.parametrize(sigma=conductivity_model)` interface.
> 
> ## Making `physical_properties` and `invertible_properties` properties?
> WOuld it be possible to make the `physical_properties` and `invertible_properties` as (Python) properties of the class, so we could access them through:
> 
> ```python
> simulation.physical_properties
> ```
> 
> Instead of having to call them (as methods)?
> 
> ## Updating deprecation admonitions
> I noticed a few admonitions in docs warning users about deprecations of the old properties. Most of them have old versions of SimPEG. I would update them to the next release: v0.26.0.
> 
> ## Connecting _exclusive_ physical properties?
> I think I briefly mentioned this idea in the last meeting, but I'll try to expand more on it here.
> 
> I was wondering if it would be best to group _exclusive_ physical properties (exclusive in the sense of XOR, i.e. one can either pass one XOR the other) together in a single `PhysicalProperty`.
> 
> Just brainstorming, but something like:
> 
> ```python
> sigma = props.PhysicalProperty(
>     "Electrical conductivity (S/m)",
>     reciprocal=props.PhysicalProperty( "Electrical resistivity (Ohm m)"),
>     shape=[(), (1,), ("mesh.n_cells",)],
> )
> ```
> 
> The `HasModel` can inject the two properties and the setters for each one of them. This way we could avoid having to handle the two of them separately in a few different places.
> 
> I don't intend to change the public interface, just refactor some of the internals so two reciprocal physical properties are naturally connected to each other.
> 
> Is it a bad idea?

I'll try to touch on a few things:

Improved Error Messages: Sounds great!

Regarding making them accessible as properties of the class:
We can achieve that, but just remember`@properties` cannot exist together with `@classmethod`. They would have to be constructed as class variables by the meta class. Certainly it is possible, I just think I moved away from it here because I was trying to move as much as I could out of the meta class. But I think those two lists are reasonable things to do there.


Regarding closer linking of the XOR properties: 
I think I remember trying something like this, and the issue was that they could become unlinked from each other very easily. For example if a getter or setter gets overwritten on a class, it creates a *new* instance of a `PhysicalProperty` on that class (in the same way that overwriting a getter and setter for a property sets a new instance of a property on the class) , so its original pair would still point to the old item, and not the new one causing them to get unlinked from each other.

simpeg/props.py
jcapriot: I do agree it would be good to add them as a protocol check. I put the initials of it here, but it would take several changes in ‘maps.py’ to properly support it.

PR #1604, opened 2025-01-31
Minor improvements to UpdateIRLS stopping criteria

santisoler:
#### Summary

Apply some minor improvements to the stopping criteria method of `directives.UpdateIRLS`. Ditch the hard-coded `1e-12` value that gets added to the `f_old` to avoid division by zero: assign `f_change` to `np.inf` instead. Remove the absolute value computation while comparing `1 - chi_factor` with the `misfit_tolerance`. This way we avoid ending the inversion with a chi factor above the 1.0.

#### PR Checklist
      to a Pull Request

#### Additional information

**TODO**

- [ ] Enable users to use `np.abs` when checking for the chifactor within tolerance. Basically, allow users to stop the IRLS even if the chifact is slightly higher than the target.

### DIFF
--- a/simpeg/directives/_regularization.py
+++ b/simpeg/directives/_regularization.py
@@ -410,15 +410,13 @@ class UpdateIRLS(InversionDirective):
             else np.inf
         )
 
-        # Get target chi factor minus current chi factor
-        chi_factor_diff = 1.0 - self.invProb.phi_d / self.misfit_from_chi_factor(
-            self.chifact_target
-        )
-
+        # Get misfit_ratio as: 1 - (phi_d / phi_d_star)
+        phi_d_star = self.misfit_from_chi_factor(self.chifact_target)
+        misfit_ratio = 1.0 - self.invProb.phi_d / phi_d_star
         if (
             f_change < self.f_min_change
             and self.metrics.irls_iteration_count > 1
-            and (0 <= chi_factor_diff < self.misfit_tolerance)
+            and (0 <= misfit_ratio < self.misfit_tolerance)
         ):
             if self.verbose:
                 print("Minimum decrease in regularization. End of IRLS")
--- a/simpeg/directives/_regularization.py
+++ b/simpeg/directives/_regularization.py
@@ -403,17 +403,22 @@ class UpdateIRLS(InversionDirective):
                 )
             return True
 
-        # Check if the function has changed enough
-        f_change = np.abs(self.metrics.f_old - phim_new) / (self.metrics.f_old + 1e-12)
+        # Get how much the regularization has changed from previous value
+        f_change = (
+            np.abs(self.metrics.f_old - phim_new) / self.metrics.f_old
+            if self.metrics.f_old != 0.0
+            else np.inf
+        )
+
+        # Get target chi factor minus current chi factor
+        chi_factor_diff = 1.0 - self.invProb.phi_d / self.misfit_from_chi_factor(
+            self.chifact_target
+        )
 
         if (
             f_change < self.f_min_change
             and self.metrics.irls_iteration_count > 1
-            and np.abs(
-                1.0
-                - self.invProb.phi_d / self.misfit_from_chi_factor(self.chifact_target)
-            )
-            < self.misfit_tolerance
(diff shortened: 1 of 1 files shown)


santisoler:
@domfournier would you like to take a look at this?

Just some minor improvements to a couple of things I noticed in the stopping criteria for the IRLS.

* I didn't quite like the hardcoded 1e-12 value. A given regularization might have values close to that, so the assumption of 1e-12 being a small value could fall, and the algorithm wouldn't work as expected.
* I noticed that by using the abs when checking for the misfit tolerance, we would allow the inversion to finish with chi factors slightly greater than the target. For example, if `misfit_tolerance=0.1` and the current chi factor is 1.09, then `np.abs(1.0 - 1.09) < misfit_tolerance` will hold, and the inversion will finish (assuming the other conditions are also true).

Please, let me know what do you think.

simpeg/directives/_regularization.py
```diff
@@ -403,17 +403,22 @@ def stopping_criteria(self):
...
+        f_change = (
+            np.abs(self.metrics.f_old - phim_new) / self.metrics.f_old
+            if self.metrics.f_old != 0.0
+            else np.inf
+        )
+
+        # Get target chi factor minus current chi factor
+        chi_factor_diff = 1.0 - self.invProb.phi_d / self.misfit_from_chi_factor(
+            self.chifact_target
+        )
```
santisoler, line 416: One other thing: shouldn't the hard coded 1.0 be the `self.chifact_target`? Or should it always be 1.0?
```suggestion
        chi_factor_diff = self.chifact_target - self.invProb.phi_d / self.misfit_from_chi_factor(
            self.chifact_target
        )
```

domfournier: I dont think so? The 1 in there is equivalent as doing 

`(phi_d* - phi_d) / phi_d*`

meaning the relative deviation from the target. The function `self.misfit_from_chi_factor` converts the chifact to phi_d value.

simpeg/directives/_regularization.py
```diff
@@ -403,17 +403,22 @@ def stopping_criteria(self):
...
 
         if (
             f_change < self.f_min_change
             and self.metrics.irls_iteration_count > 1
-            and np.abs(
-                1.0
-                - self.invProb.phi_d / self.misfit_from_chi_factor(self.chifact_target)
-            )
-            < self.misfit_tolerance
+            and (0 <= chi_factor_diff < self.misfit_tolerance)
```
domfournier, line 421: The absolute value was intended to allow to be "within range" of the target. That is, it's ok if the misfit is slightly higher or lower than the target. 
Forcing it to be strictly smaller is a bit more restrictive.

simpeg/directives/_regularization.py
santisoler: You're right! My bad. I think what we can do is to rename the variable to make it reflect better what value it holds.


One minor thing: I get that by definition the chi factor is going to have the units of `phi_d / N`, but we always assume that the chi factor is unitless, meaning that `phi_d` is also unitless (because we are dividing by the uncertainties). This becomes quite clear when we look at the stopping criteria in `TargetMisfit` where we check if `phi_d < chifact * phi_d_star`, meaning that `chifact` is assumed to be always unitless. Nothing to take action on, but just wanted to mention it.

simpeg/directives/_regularization.py
santisoler: Right, not using the absolute value would make the stopping criteria more restrictive. The reason of the change is that it sounds unintuitive that it's possible to get a model that generates a chifact greater than the target.

Talking with @lheagy, one option would be to offer a flag to allow users to make the stopping criteria less restrictive by using the abs, but not using the abs by default. What do you think?

simpeg/directives/_regularization.py
domfournier: Yep, you are right - that would make phi_d already unitless.

simpeg/directives/_regularization.py
domfournier: I have no objection. The only risk I can foresee is an IRLS that runs longer, but ending below the target makes sense.
We can make the change, and revert the default later if we see instability.

PR #1605, opened 2025-02-04
Fix mappings in regularizations

santisoler:
#### Summary

Fix how the mapping is being used in `Smallness` regularization: regularize on `mapping(m) - mapping(m_ref)` instead of `mapping(m - m_ref)`. Update the documentation of `Smallness` to include details on how the mapping is being used. Add tests to check this intended behaviour.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/regularization/base.py
+++ b/simpeg/regularization/base.py
@@ -539,8 +539,9 @@ class Smallness(BaseRegularization):
         Boolean array defining the set of :py:class:`~.regularization.RegularizationMesh`
         cells that are active in the inversion. If ``None``, all cells are active.
     mapping : None, simpeg.maps.BaseMap
-        The mapping from the model parameters to the active cells in the inversion.
-        If ``None``, the mapping is the identity map.
+        The mapping function applied to the model parameters. Use a mapping to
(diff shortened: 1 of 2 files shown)


santisoler:
@jcapriot, I could keep working on this for the next release. What do you think about this idea of switching how the mapping gets applied in regs?

Regardless of that decision, I think we should update the docstrings, since they don't mention how the mappings get applied.

jcapriot:
> @jcapriot, I could keep working on this for the next release. What do you think about this idea of switching how the mapping gets applied in regs?
> 
> Regardless of that decision, I think we should update the docstrings, since they don't mention how the mappings get applied.

I think it would be fine to get this in. Having it act directly on the model values is a bit more intuitive.

santisoler:
Sounds good. I'll try to keep working on this in the following weeks. If at any point we find a roadblock, I will fall back to the idea of just updating the docstrings for now.

santisoler:
I decided to leave the amplitude regularizations our of the scope of this PR. I think we should redesign them a bit, since they not necessarily fit into the `f_m` and `f_m_deriv` pattern. Related to #1301

PR #1610, opened 2025-02-25
Christiansen and Auken (2012) DOI modules implementation.

ghwilliams:
#### Summary
This PR is a proposal for adding Depth Of Investigation code based on the implementation described in
Vest Christiansen, A., & Auken, E. (2012). A global measure for depth of investigation. Geophysics, 77.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/electromagnetics/time_domain/__init__.py
+++ b/simpeg/electromagnetics/time_domain/__init__.py
@@ -106,6 +106,13 @@ Base Classes
   fields.FieldsDerivativesEB
   fields.FieldsDerivativesHJ
 
+DOI subroutines
+===============
+
+.. autosummary::
+  :toctree: generated/
+
+  doi.doi_tdem_1d_layer_CA2012
 """
 
 from .simulation import (
@@ -127,3 +134,7 @@ from . import receivers
 
 from . import sources as Src
 from . import receivers as Rx
+
+from .doi import (
+    doi_tdem_1d_layer_CA2012,
+)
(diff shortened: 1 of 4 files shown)


craigmillernz:
In the refine layers function
`c_out[(z_out >= a) & (z_out <= b)] = c_in[i]`
should be
`c_out[(z_out > a) & (z_out <= b)] = c_in[i]`
otherwise the first layer is missed.

ghwilliams:
Thanks for catching this one. I'll look at it and make the necessary corrections.

craigmillernz:
Just an update for you, i've been in touch with Anders Christiansen and has given us some pointers.  We think it's easier to work in linear space (i.e. IdentityMap) and then convert to loglog for the jacobian using the chain rule.  Also the final summation should be of the loglog jacobian matrix (eqn2), not the thickness normalised sensitivity but we aren't quite sure how to do that so have gone back to Anders for more details.  We also adjusted the time gates to match what is in the paper and this makes a difference.  Should have more info next week...

ghwilliams:
> Just an update for you, i've been in touch with Anders Christiansen and has given us some pointers. We think it's easier to work in linear space (i.e. IdentityMap) and then convert to loglog for the jacobian using the chain rule. Also the final summation should be of the loglog jacobian matrix (eqn2), not the thickness normalised sensitivity, look closely at eqn 5, but we aren't quite sure how to do that so have gone back to Anders for more details. We also adjusted the time gates to match what is in the paper and this makes a difference. Should have more info next week...

Great! Once we have satisfactory test results, we can move forward with requesting a merge. The issue of whether or not to divide by the layer thicknesses is admittedly a bit confusing, but for the initial implementation, we should focus on matching the results from the paper as closely as possible.

PR #1614, opened 2025-02-26
Make `dobs` a required argument for `Data`

santisoler:
#### Summary

Make `dobs` a required argument for `Data` objects. Also, raise errors when trying to set `Data` properties `dobs`, `noise_floor`, `relative_error`, and `standard_deviation` with arrays that contain `nan`s or `inf`s. Add a new `raise_if_nans` utility function. Add tests to check if errors are raised in `Data`.

#### PR Checklist
      to a Pull Request


santisoler:
During yesterday's meeting we decided that would be better to make `dobs` a required argument that only takes an array (without nans or infs). To make that happen, we would need to move the indexing properties of the `Data` to the `Survey`, so users can build `dobs` arrays by parts before having to define the `Data` class itself.

PR #1616, opened 2025-02-28
Move indexing of flat arrays to Survey classes

santisoler:
#### Summary

Add a new `get_slice` method to `BaseSurvey` that mimics the behaviour of indexing `Data` objects. The method returns a `slice` object that can be used to slice flat arrays like data or uncertainty arrays. Optimize the implementation by caching slices rather than Numpy arrays with indices. Add tests for the new method. Use new method in `Data`'s `__getitem__` and `__setitem__` instead of the `indexing_dictionary`.

#### PR Checklist
      to a Pull Request

#### Additional information

This PR is part of the process of cleaning up the interface for `Data` objects. Moving the indexing functionalities to the `Survey` classes would allow users to build data arrays by part before having to define the `Data` object in advance. Check meeting notes of 2025-02-26 for more details on this discussion.

### DIFF
--- a/simpeg/data.py
+++ b/simpeg/data.py
@@ -328,12 +328,12 @@ class Data:
     ##########################
 
     def __setitem__(self, key, value):
-        index = self.index_dictionary[key[0]][key[1]]
-        self.dobs[index] = mkvc(value)
+        slice_obj = self.survey.get_slice(*key)
+        self.dobs[slice_obj] = mkvc(value)
 
     def __getitem__(self, key):
-        index = self.index_dictionary[key[0]][key[1]]
-        return self.dobs[index]
+        slice_obj = self.survey.get_slice(*key)
+        return self.dobs[slice_obj]
 
     def tovec(self):
         """Convert observed data to a vector
--- a/simpeg/survey.py
+++ b/simpeg/survey.py
@@ -1,4 +1,5 @@
 import numpy as np
+from numpy.typing import ArrayLike
 import scipy.sparse as sp
 
 import warnings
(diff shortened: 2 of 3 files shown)


santisoler:
Todo:

- [ ] work on `BaseSIPSimulation.Jtvec`: it casts `v` into its own `Data` class. The `Data` class overwrites the `index_dictionary` to index by source, receiver and time channel. Should we move that indexing to the `spectral_induced_polarization.Survey`?
- [ ] Check `BaseDCSimulation._Jtvec` and `BaseDCSimulation2D._Jtvec`.

santisoler:
I think this is ready to be merged. @jcapriot would you like to take a look? I'm curious about your thoughts on the `deprecated` decorator I added here.

simpeg/utils/code_utils.py
```diff
@@ -782,6 +783,76 @@ def dep_function(*args, **kwargs):
     return dep_function
 
 
+def _deprecated(msg: str, *, category=DeprecationWarning, stacklevel=1):
```
jcapriot, line 786: I don’t know if this is really saving any work.
All it’s doing is calling `warnings.warn` ahead of the function call. Why not just not call `warnings.warn` directly inside the function?

jcapriot: Ah I see, it comes from [pep702](https://peps.python.org/pep-0702/)

If you really want to get deprecated support for items from static type checkers, I’d say it would be better to just conditionally require `typing_extensions` for Python versions less than 3.13

santisoler: I think it's nice to start using `deprecated`, since it'll probably be the default way to deprecate objects in Python. It also comes with the feature that type checkers can detect used of deprecated objects.

I agree with your point. I'll add `type_extensions` as a dependency then so we can conditionally use their `deprecated` decorator.

simpeg/data.py
```diff
@@ -1,9 +1,18 @@
+import sys
 import numpy as np
 import warnings
 
 from .survey import BaseSurvey
 from .utils import mkvc, validate_ndarray_with_shape, validate_float, validate_type
 
+if sys.version_info >= (3, 13):
+    # warnings.deprecated available in Python >=3.13
+    from warnings import deprecated
```
jcapriot, line 10: You could instead just use a try… except 
ImportError block here too instead of comparing versions

Issue #1618, opened 2025-03-05
BUG: Potential fields simulations when "forward_only" is selected


santisoler:
### Describe the issue:

Currently, if we set `store_sensitivities="forward_only"` in the potential fields simulations, methods like `Jvec` would still try to build the sensitivity matrix (`G`), defeating the purpose of the `forward_only` option. If the sensitivity matrix cannot be stored in memory, this results in an insufficient memory error. 

> [!NOTE]
> This issue was raised by @thibaut-kobold on Mattermost, particularly for the magnetic simulation, although it still applies to the gravity one.

Methods like `Jvec`, `Jtvec` and `getJ` should respect user's decision not to build the sensitivity matrix, but still return the requested result.

### Reproducible code example:

```python
import numpy as np

from discretize import TensorMesh
from simpeg.utils import model_builder
from simpeg import maps
from simpeg.potential_fields import gravity


# Observation points
x, y = np.linspace(-20.0, 20.0, 15), np.linspace(-20.0, 20.0, 15)
x, y = np.meshgrid(x, y)
z = 5.0 * np.ones_like(x)
coordinates = np.vstack((x.ravel(), y.ravel(), z.ravel())).T
receivers = gravity.receivers.Point(coordinates, components="gz")
source_field = gravity.sources.SourceField(receiver_list=[receivers])
survey = gravity.survey.Survey(source_field)

# Mesh
dh = 5.0
hx = [(dh, 8)]
mesh = TensorMesh([hx, hx, hx], "CCN")

# Simulation
model_map = maps.IdentityMap(nP=mesh.n_cells)
simulation = gravity.simulation.Simulation3DIntegral(
    survey=survey,
    mesh=mesh,
    rhoMap=model_map,
    store_sensitivities="forward_only",
    engine="choclo",
)
assert simulation._G is None

# Define densities.
densities = 1e-10 * np.ones(mesh.n_cells)
ind_sphere = model_builder.get_indices_sphere(
    np.r_[0.0, 0.0, -20.0], 10.0, mesh.cell_centers
)
densities[ind_sphere] = 0.2

simulation.dpred(densities)  # run dpred first (see issue #1617)
assert simulation._G is None

# Try to compute Jvec
vector = densities
simulation.Jvec(densities, vector)

# After computing Jvec with forward_only, the sensitivity matrix should not
# have been formed
assert simulation._G is None  # the last assert
```

### Error message:

```shell
Traceback (most recent call last):
[...]

PR #1622, opened 2025-03-08
Implement `G` matrix as `LinearOperator` in gravity simulation

santisoler:
#### Summary

Make the `G` property in the gravity survey to be either a 2D Numpy array, or a `scipy.sparse.LinearOperator` if `store_sensitivities` is set to `"forward_only"`. The `LinearOperator` is capable of performing operations like `G @ m` and `G.T @ v` without allocating the full `G` matrix on memory or disk. This allows `Jvec` and `Jtvec` to be called and not allocate the full `G` matrix. The `getJ` method also returns either a 2D Numpy array or a `LinearOperator` formed by the product between `G` and the `rhoDeriv` as a `LinearOperator` as well. Reimplement the `getJtJdiag` method: allow it to get the diagonal without building `G` if `"forward_only"`, optimize the computation of the diagonal when `G` is fully formed through `einsum`s, and fix the cache of `gtg_diagonal` by checking the hash of the `W` argument. Deprecate the public `gtg_diagonal` property. Extend docstrings of relevant methods. Add tests for all the new features.

#### PR Checklist
      to a Pull Request

#### Reference issue

Related to #1618

(possibly fixes it?)

#### Additional information

TODO:
- [x] Test the Jvec and Jtvec between ram and forward_only, so we are sure the two results are correct. Don't just test them independently with a simpler version of the same thing.
- [x] Extend docs for `getJ` and `G`.
- [x] Extend docs for `Jvec` and `Jtvec`
- [x] Extend docs for `getJtJdiag`
- [x] Improve the caching mechanism of `_gtgdiag`: store the hash of the weights.
- [x] Deprecate the public `gtgdiag` property
- [x] Extend docs: explain better what happens with `forward_only` in the constructor.


simpeg/potential_fields/gravity/_numba_functions.py
```diff
@@ -163,6 +163,170 @@ def _sensitivity_gravity(
...
+                    receivers[i, 1],
+                    receivers[i, 2],
+                    nodes[j, 0],
+                    nodes[j, 1],
+                    nodes[j, 2],
+                    kernel_func,
+                )
+            # Compute sensitivity matrix elements for the i-th row, multiply them
+            # by the i-th element of the vector and add them to the running result
+            # array.
```
santisoler, line 316:
```suggestion
            # Compute sensitivity matrix elements for the i-th row, multiply them
            # by the i-th element of the vector and add them to the result array
            # for this thread.
```

simpeg/potential_fields/gravity/_numba_functions.py
```diff
@@ -163,6 +163,170 @@ def _sensitivity_gravity(
...
+    n_receivers = receivers.shape[0]
+    n_nodes = nodes.shape[0]
+    n_cells = cell_nodes.shape[0]
+    n_threads = get_num_threads()
+    # Estimate number of receivers that will be assigned to each thread
+    chunk_size = n_receivers // n_threads
+    # Allocate result arrays, one per thread
+    result_per_thread = np.zeros((n_threads, n_cells), dtype=result.dtype)
+    # Split the job per thread
+    for thread_index in prange(n_threads):
```
jcapriot, line 292: I would recommend not explicitly chunking it by receiver, let numba do it automatically. It can recognize the reduction operation on the out vector: https://numba.pydata.org/numba-doc/dev/user/parallel.html#explicit-parallel-loops

simpeg/potential_fields/gravity/_numba_functions.py
```diff
@@ -163,6 +163,170 @@ def _sensitivity_gravity(
...
+                result_per_thread[thread_index, k] += (
+                    constant_factor
+                    * vector[i]
+                    * kernels_in_nodes_to_cell(
+                        kernels,
+                        cell_nodes[k, :],
+                    )
+                )
+    # Reduce the operation and write into the result array
+    result[:] += np.sum(result_per_thread, axis=0)
```
jcapriot, line 327:
```suggestion
    n_receivers = receivers.shape[0]
    n_nodes = nodes.shape[0]
    n_cells = cell_nodes.shape[0]
    result[:] = 0
    # Evaluate kernel function on each node, for each receiver location
    for i in prange(n_receivers):
        # Allocate vector for kernels evaluated on mesh nodes
        kernels = np.empty(n_nodes)
        local_row = np.empty(n_cells)
        for j in range(n_nodes):
            kernels[j] = _evaluate_kernel(
                receivers[i, 0],
                receivers[i, 1],
                receivers[i, 2],
                nodes[j, 0],
                nodes[j, 1],
                nodes[j, 2],
                kernel_func,
            )
        # Compute fields from the kernel values
        for k in range(n_cells):
            local_row[k] += (
                constant_factor
                * vector[i]
                * kernels_in_nodes_to_cell(
                    kernels,
                    cell_nodes[k, :],
                )
            )
        result += local_row
```
This should get numba to pickup that result is a reduction operation, and cause it to handle it correctly in the prange

jcapriot: Just did some local tests myself too, relying on `numba` to do the reduction produces the expected timings on my end (meaning the parallel forward and adjoint operations both had similar timings), what kind of array sizes were you experimenting with?

Here's my timings of a simple example: [the gist](https://gist.github.com/jcapriot/cdbd76e0f35013c60a8f02601303baf4)

santisoler: I think I was working with small arrays on Friday. I'm running better benchmarks now and I don't see any difference between the two implementations:

```python
import time
import numpy as np
import discretize
from simpeg.potential_fields import gravity
from simpeg import maps
from simpeg.utils import model_builder


def get_simulation(**kwargs):
    # Observation points
    x = np.linspace(-20.0, 20.0, 40)
    x, y = np.meshgrid(x, x)
    z = 5.0 * np.ones_like(x)
    coordinates = np.vstack((x.ravel(), y.ravel(), z.ravel())).T
    receivers = gravity.receivers.Point(coordinates, components="gz")
    source_field = gravity.sources.SourceField(receiver_list=[receivers])
    survey = gravity.survey.Survey(source_field)

    # Mesh
    dh = 5.0
    hx = [(dh, 60)]
    mesh = discretize.TensorMesh([hx, hx, hx], "CCN")

    # Define densities
    densities = 1e-10 * np.ones(mesh.n_cells)
    ind_sphere = model_builder.get_indices_sphere(
        np.r_[0.0, 0.0, -20.0], 10.0, mesh.cell_centers
    )
    densities[ind_sphere] = 0.2

    # Define simulation
    mapping = maps.IdentityMap(nP=mesh.n_cells)
    default_kwargs = {
        "store_sensitivities": "ram",
        "engine": "choclo",
    }
    default_kwargs.update(kwargs)
    simulation = gravity.simulation.Simulation3DIntegral(
        survey=survey,
        mesh=mesh,
        rhoMap=mapping,
        **default_kwargs,
    )
    return simulation, densities


if __name__ == "__main__":
    simulation, densities = get_simulation(store_sensitivities="forward_only")
    simulation.model = densities # assign model so rhoDeriv gets defined (see #1621)
    print(f"n_data: {simulation.survey.nD}, n_params: {simulation.mesh.n_cells}")

    # Run once to compile Numba functions
    vector = np.random.default_rng(seed=42).uniform(size=simulation.survey.nD)
    simulation.Jtvec(densities, vector)

    # Benchmark
    start = time.perf_counter()
    simulation.Jtvec(densities, vector)
    end = time.perf_counter()
    print(f"Elapsed time: {end - start:2e}s")
```

With 4e601b8fca3d47087753ac3797fc9eac0611686b:
```
n_data: 1600, n_params: 216000
Elapsed time: 3.510344e+00s
```

After applying your suggestions:
```
n_data: 1600, n_params: 216000
Elapsed time: 3.660166e+00s
```

The time difference between the two is within the error.

Let's apply your suggestions then, since the code ends up being simpler.

simpeg/potential_fields/gravity/simulation.py
```diff
@@ -246,24 +250,39 @@ def fields(self, m):
...
         else:
-            diag = self._gtg_diagonal
-        return mkvc((sdiag(np.sqrt(diag)) @ self.rhoDeriv).power(2).sum(axis=0))
+            diagonal = np.zeros(self.nC)
+            for i in range(self.survey.nD):
+                g_row_sq = self.G[i, :] ** 2
+                diagonal += weights[i] * g_row_sq
+
+        # Multiply the diagonal by the derivative of the mapping
+        diagonal *= self.rhoDeriv.diagonal() ** 2
```
jcapriot, line 284: Don’t do this, it will fail for joint inversions.

simpeg/potential_fields/gravity/simulation.py
```diff
@@ -282,18 +282,45 @@ def getJtJdiag(self, m, W=None, f=None):
...
-            diagonal = np.zeros(self.nC)
+            gtg_diagonal = np.zeros(self.nC)
             for i in range(self.survey.nD):
                 g_row_sq = self.G[i, :] ** 2
-                diagonal += weights[i] * g_row_sq
-
-        # Multiply the diagonal by the derivative of the mapping
-        diagonal = mkvc((sdiag(np.sqrt(diagonal)) @ self.rhoDeriv).power(2).sum(axis=0))
-        return diagonal
+                gtg_diagonal += weights[i] * g_row_sq
```
domfournier, line 322: Is there something preventing you from using the einsum implementation?
```
        gtg_diag = np.asarray(
            np.einsum("i,ij,ij->j", W**2, self.G, self.G)
        )
```
Looping is going to be really slow for large problems.

simpeg/potential_fields/gravity/simulation.py
```diff
@@ -282,18 +282,45 @@ def getJtJdiag(self, m, W=None, f=None):
             else np.ones(self.survey.nD, dtype=np.float64)
         )
 
-        # Compute diagonal of G.T @ G
+        # Compute gtg (G.T @ W.T @ W @ G) if it's not cached, or if the
+        # weights are not the same.
```
domfournier, line 286: Should we move this to the docstrings?

santisoler: Yes, I agree. I'll probably add it under the Notes section.

santisoler:
I think this is ready for review, in case anyone wants to take a look at it.

simpeg/potential_fields/gravity/_numba_functions.py
```diff
@@ -163,6 +163,309 @@ def _sensitivity_gravity(
...
+    receivers,
+    nodes,
+    cell_nodes,
+    kernel_func,
+    constant_factor,
+    weights,
+    diagonal,
+):
+    """
+    Diagonal of ``G.T @ W.T @ W @ G`` without storing ``G``.
```
santisoler, line 177:
```suggestion
    Diagonal of ``G.T @ W.T @ W @ G`` without storing ``G``, in serial.
```

simpeg/potential_fields/gravity/_numba_functions.py
```diff
@@ -163,6 +163,309 @@ def _sensitivity_gravity(
...
+    result : (n_active_cells) numpy.ndarray
+        Running result array where the output of the dot product will be added to.
+
+    Notes
+    -----
+    This function is meant to be run in parallel.
+    This implementation instructs each thread to allocate their own array for
+    the current row of the sensitivity matrix. After computing the elements of
+    that row, it gets added to the running ``result`` array through a reduction
+    operation handled by Numba.
```
santisoler, line 431:
```suggestion
    operation handled by Numba.
    
    A serialized implementation of this function is available in
    ``_sensitivity_gravity_t_dot_v_serial``.
```

santisoler:
Leaving a note here on a thought I had: if we store `G` on disk as a memmap, the `getJ` will actually try to store the `J` matrix as an array in memory. I'm not sure what the desired behaviour should be in such case. Shall we have a tempfile for `J` so it's also a memmap? Would it be better if `J` is also a LinearOperator in that case?

This is kind of a niche situation which is not very relevant in the inversion process, so I think we can discuss it and keep it for a separate PR.

PR #1626, opened 2025-03-11
Update iteration print out

jcapriot:
#### Summary
Update the iteration write out order.

#### PR Checklist
      to a Pull Request

For the longest time, the iteration printers have written out mismatched values at each iteration, and also did not print the last value of the inversion.

This updates the write outs to happen after the line-search (matching the behavior of the save directives), but before the directives themselves alter the objective function.

I've also removed several un-needed notifications:
* setting bfgsH0
  * only set on BFGS subclasses and 
  * if the UpdatePreconditioner directive is not in the directive list (It handles the appoxHinv).
*  Removes the repeated notifications about no nans in the model, and instead raises an error if the model vector has any nans.

Writers:
* The iteration writer formatting logic was rather convoluted to me, so I refactored it a bit to make it a little easier to understand what it was doing...

#### Additional information
This is not meant to be anything final, but at least it's an interim improvement to the iteration write outs... Have not tested it on many different minimizers just yet.


jcapriot:
Also, I've added a block to the `BetaCooling` to not happen if the objective function is at it's maximum iteration, as that will cause it to stop on the next function evaluation.

Something else to consider:
We should really implement a directive sorting priority, and put any `TargetDataMisfit` like stopping directives at the front of the list....

jcapriot:
I think I've found a sort of "path of least resistance" to get better output printing from the cross-gradient like inversion, which wasn't great with these changes as it was an iteration behind as well. 

I've made `ComboObjectiveFunctions` stash the last evaluate value of all of their `objfcts` in a `_last_obj_vals` private attribute. So the printer will then assume that you want to print the last evaluated value in an objective function. Since all `dmisfit` and `reg` are `ComboObjectiveFunctions` in the cross-gradient like inversions this works and is equivalent to what was assumed before, just now it doesn't re-evaluate them using the current iterate and get's the last value from the objective function (usually from the last iteration of a line-search).

I don't particularly like the solution, but the others I tried were much too involved and usually required much more rewrites.

simpeg/typing/__init__.py
```diff
@@ -27,8 +30,7 @@
     np.random.BitGenerator,
     np.random.Generator,
 ]
-
-RandomSeed.__doc__ = """
```
jcapriot, line 31: So as it turns out, we shouldn't assign to `__doc__` here, as you are actually setting the docstring on `typing.Union`, because that is what `RandomSeed` is equal to.

`sphinx` is still able to pick up on the docstring written just after this though.

simpeg/inverse_problem.py
```diff
@@ -213,38 +219,32 @@ def startup(self, m0):
...
+                    using the default solver {solver.__name__} with the 'is_symmetric=True` option set.
                     """
-                        simpeg.InvProblem is setting bfgsH0 to the inverse of the eval2Deriv.
-                        ***Done using the default solver {} and no solver_opts.***
-                        """.format(
-                        solver.__name__
-                    )
-                )
-                solver_opts = {}
+                solver_opts = dict(is_symmetric=True)
```
jcapriot, line 245: Setting this unconditionally for all 2nd derivatives of a scalar function *should* be correct, as the matrix of second derivatives hast to be symmetric as df/dxdy = df/dydx. It's not necessarily positive definite though.

simpeg/objective_function.py
```diff
@@ -463,6 +465,8 @@ def __call__(self, m, f=None):
             else:
                 objective_func_value = objfct(m)
             fct += multiplier * objective_func_value
+            obj_vals.append(objective_func_value)
+        self._last_obj_vals = obj_vals
```
jcapriot, line 469: This is what I'm referring to when I say I don't particularly like this approach, but at least it's a private variable and only used in the similarity printers.

Another approach could be to make `BaseObjectiveFunction` have a metaclass that captures the output of the `__call__` method for all objective functions.

simpeg/typing/__init__.py
santisoler: Good to know! Nice catch!

PR #1633, opened 2025-03-20
Deprecate `components` in potential field surveys

### LINKED ISSUE #1609: BUG: Inconsistency between docstring and result of `simpeg.potential_fields.gravity.survey.Survey.components()`
### Describe the issue:

There is an inconsistency between the docstring of the `components` property of `simpeg.potential_fields.gravity.survey.Survey`, and the output of the function. The relevant lines of code are here:

https://github.com/simpeg/simpeg/blob/2e3689625225955cc338dadad13cb3cdb4310e8b/simpeg/potential_fields/gravity/survey.py#L77-L86

The docstring indicates the method returns the number of components at each receiver. However the function returns instead the list of strings representing the components of the first receiver.

There are a few cases where this causes unintended outputs.

### Unintended output 1: Expect an int, get an array of strings

This is the simplest case. We have 2 components defined for the receiver, so the expected output is 2. However, the list of strings representing the components of the first receiver, i.e., `['gx', 'gz']`.

```python
r = gravity.receivers.Point([0, 0, 0], ["gx", "gz"])

source_field = gravity.sources.SourceField(receiver_list=[r])

survey = gravity.survey.Survey(source_field)

# Inspect the components of the survey object
print(survey.components)

# Returns: ['gx', 'gz']
# Expected: 2
```

### Unintended output 2: Different components for different receivers

Here the same number of receivers are defined for each receiver, but different types of component are defined at each receiver. Here the output is the component(s) of the first receiver. What should the expected result be here though? Should it be 1, as each receiver (in the list) has 1 component? Or should it be 2, as there are 2 components in the receiver list?

```python
from simpeg.potential_fields import gravity

rx = gravity.receivers.Point([0, 0, 0], ["gx"])
rz = gravity.receivers.Point([0, 0, 0], ["gz"])

source_field = gravity.sources.SourceField(receiver_list=[rx, rz])

survey = gravity.survey.Survey(source_field)

# Inspect the components of the survey object
print(survey.components)

# Returns: ['gx']
# Expected: 1 or 2
```

### Unintended output 3: Different numbers of components for each point receiver

One can define different numbers of components for different receivers. Here the value returned is the component(s) of the first receiver. But I think here the expected output is not well defined.

```python
r1 = gravity.receivers.Point([0, 0, 0], ["gx"])
r2 = gravity.receivers.Point([0, 0, 0], ["gx", "gz"])

source_field = gravity.sources.SourceField(receiver_list=[r1, r2])

survey = gravity.survey.Survey(source_field)

# Inspect the components of the survey object
print(survey.components)

# Returns: ['gx']
# Expected: ???
```

## Final thoughts

 - A fix for this will make the docstring and implementation consistent. However, it should be clear what the intended use is.
 - This inconsistency may be present in other survey classes. I haven't had a chance to look yet. I'll update here if I find more.

```python
from simpeg.potential_fields import gravity

r = gravity.receivers.Point([0, 0, 0], ["gx", "gz"])

source_field = gravity.sources.SourceField(receiver_list=[r])

survey = gravity.survey.Survey(source_field)

# Inspect the components of the survey object
print(survey.components)

# Returns: ['gx', 'gz']
# Expected: 2
```

### Error message:

```shell

```

### Runtime information:

--------------------------------------------------------------------------------
  Date: Thu Feb 20 16:57:00 2025 PST

                OS : Darwin (macOS 14.6.1)
            CPU(s) : 12
           Machine : arm64
      Architecture : 64bit
               RAM : 36.0 GiB
       Environment : Python
       File system : apfs

  Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:54:21)
  [Clang 16.0.6 ]

            simpeg : 0.23.0
        discretize : 0.11.1
       pymatsolver : 0.3.1
             numpy : 2.2.2
             scipy : 1.15.1
        matplotlib : 3.10.0
            geoana : 0.7.2
            libdlf : 0.3.0
           sklearn : 1.6.1
            pandas : 2.2.3
             sympy : 1.13.1
           IPython : 8.30.0
        ipywidgets : 8.1.5
            plotly : 5.24.1
               vtk : 9.3.1
--------------------------------------------------------------------------------

### Context for the issue:

_No response_

santisoler:
#### Summary

Deprecate the `components` properties in potential field surveys. These properties weren't returning the expected output. Since receivers within the same survey can have different components, returning the components of the first receiver is misleading.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fixes #1609

### DIFF
--- a/simpeg/potential_fields/magnetics/simulation.py
+++ b/simpeg/potential_fields/magnetics/simulation.py
@@ -1629,10 +1629,6 @@ class Simulation3DDifferential(BaseMagneticPDESimulation):
         components = receivers[0].components
         if not all(components == rx.components for rx in receivers):
             raise NotImplementedError()
-
-        # Cast to list so it always return a list of strings
-        if isinstance(components, str):
-            components = list(components)
         return components
 
     def projectFieldsAsVector(self, B):
--- a/simpeg/potential_fields/magnetics/simulation.py
+++ b/simpeg/potential_fields/magnetics/simulation.py
@@ -1524,8 +1524,16 @@ class Simulation3DDifferential(BaseMagneticPDESimulation):
             \text{TMI} = \vec{B}_s \cdot \hat{B}_0
 
         """
-        # TODO: There can be some different tyes of data like |B| or B
-        components = self.survey.components
+        # Get components for all receivers, assuming they all have the same components
+        try:
+            components = self._get_components()
+        except NotImplementedError as e:
+            e.args = (
+                "Found receivers with different set of components in the survey. "
+                "Calling `projectFields` with such type of survey is not currently "
+                "supported.",
+            )
+            raise e
 
         fields = {}
         if "bx" in components or "tmi" in components:
@@ -1561,7 +1569,16 @@ class Simulation3DDifferential(BaseMagneticPDESimulation):
         Especially, this function is for TMI data type
         """
 
-        components = self.survey.components
+        # Get components for all receivers, assuming they all have the same components
+        try:
+            components = self._get_components()
+        except NotImplementedError as e:
+            e.args = (
+                "Found receivers with different set of components in the survey. "
+                "Calling `projectFieldsDeriv` with such type of survey is not "
+                "currently supported.",
+            )
+            raise e
 
         fields = {}
         if "bx" in components or "tmi" in components:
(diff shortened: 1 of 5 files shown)


simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1524,8 +1524,16 @@ def projectFields(self, u):
...
+        # Get components for all receivers, assuming they all have the same components
+        try:
+            components = self._get_components()
+        except NotImplementedError as e:
+            e.args = (
+                "Found receivers with different set of components in the survey. "
+                "Calling `projectFields` with such type of survey is not currently "
+                "supported.",
+            )
+            raise e
```
jcapriot, line 1536: Can we issue an error on the survey class instead? If a survey has multiple receivers with different locations, accessing the `survey.components` will issue an error (in addition to it being deprecated)?

santisoler: I'd like to keep the error messages in the simulation methods. Mainly because after we remove the `survey.components` property the simulation method would still work without any other change.

We can also make `survey.components` to raise error in case that receivers have different components. What do you think?

jcapriot: I know it’s a little rough due to the current mutability of surveys, but would it better to validate the survey on assignment to this simulation? Just check if a survey exists with multiple sources or receivers and fail the validation. (I expect that this type of mutability is very rare to encounter, as once surveys are assigned to a simulation, I doubt users change the surveys.)

jcapriot: Only change would be to update it on the simulation’s survey setter, not on `__init__`.

tests/pf/test_components.py
```diff
@@ -0,0 +1,190 @@
...
+    @pytest.mark.parametrize("invalid_rx", ["no-rx", "different-components"])
+    @pytest.mark.parametrize("method", ["projectFields", "projectFieldsDeriv"])
+    def test_project_fields_invalid(
+        self, sample_simulation, receiver_locations, invalid_rx, method
+    ):
+        """
+        Test ``projectFields`` and ``projectFieldsDeriv`` on invalid surveys.
+        """
+        # Override receivers in simulation's survey
+        if invalid_rx is None:
```
santisoler, line 169:
```suggestion
        if invalid_rx == "no-rx":
```

Issue #1611, opened 2025-02-25
BUG: `time_domain.Simulation1DLayered.getJ` not including mapping derivative


santisoler:
### Describe the issue:

This bug was reported by @ghwilliams in Mattermost on 2025-02-23. Original text:

> I'm unsure if `simpeg.electromagnetics.time_domain.Simulation1DLayered.getJ` returns the sensitivity matrix post multiplied by the derivative of the mapping. I see this is done in the code for `getJ` in the DC resistivity simulation `simpeg/electromagnetics/static/resistivity/simulation_1d.py` (`Simulation1DLayers`, around line 324). 

I wrote a simple script that builds two 1D simulations: one with an identity map and another one with an exponential map. Then the script generates two models: one with conductivity and another one with log conductivity. Finally, it computes the `dpred` and `getJ` for each simulation and model, respectively.

The two results of the `dpred` should be equal: the physical properties of the cells end up being the same, so the predicted data should also be the same.

But the two J matrices should be different: these are derivatives of the simulation with respect to the model. In the first case the model are conductivities, while in the second one those are log conductivities, therefore the derivatives with respect these two should be different.

### Reproducible code example:

```python
import numpy as np
import simpeg.electromagnetics.time_domain as tdem
from simpeg import maps


def create_simulation_and_conductivities(identity_mapping: bool):
    # Create Survey
    # -------------
    # Source properties
    source_location = np.array([0.0, 0.0, 20.0])
    source_orientation = "z"  # "x", "y" or "z"
    source_current = 1.0  # maximum on-time current
    source_radius = 6.0  # source loop radius

    # Receiver properties
    receiver_location = np.array([0.0, 0.0, 20.0])
    receiver_orientation = "z"  # "x", "y" or "z"
    times = np.logspace(-5, -2, 31)  # time channels (s)

    # Define receiver list. In our case, we have only a single receiver for each source.
    # When simulating the response for multiple component and/or field orientations,
    # the list consists of multiple receiver objects.
    receiver_list = []
    receiver_list.append(
        tdem.receivers.PointMagneticFluxDensity(
            receiver_location, times, orientation=receiver_orientation
        )
    )

    # Define the source waveform. Here we define a unit step-off. The definition of
    # other waveform types is covered in a separate tutorial.
    waveform = tdem.sources.StepOffWaveform()

    # Define source list. In our case, we have only a single source.
    source_list = [
        tdem.sources.CircularLoop(
            receiver_list=receiver_list,
[...]

PR #1638, opened 2025-03-31
Fix `getJ` method in TDEM and FDEM 1D simulations

ghwilliams:
#### Summary
@santisoler  Here we're gonna work on a solution for issue #1611.
Initially, I've included a test in tests/em/tdem/test_TDEM_Simulation1DLayered_getJ.py.
In the following steps, we're gonna work on making the necessary changes to the relevant classes/methods as already discussed during some previous dev meetings. 

#### Squash and Merge summary

Standardize the `getJ` methods of TDEM and FDEM 1D simulations so they return a dense array instead of a dictionary with blocks of the `J` matrix. Make the old `getJ` method private by renaming it to `_getJ`. The new `getJ` method reuses the `_getJ` and builds the while `J` as a dense array before returning it. Update `Jvec` and `Jtvec` so they use the new private method. Add tests to check the behaviour of the new public method.

#### PR Checklist
      to a Pull Request


santisoler (commented):
Awesome, thanks @ghwilliams for starting this!

I just left a minor comment regarding the test class you wrote.

jcapriot:
Quick note on how this should be done:

1: change current `getJ` to `_getJ`
2: change internal calls inside `Jvec`, `JTvec` and  `getJtJdiag` from `getJ` to `_getJ`
3: add new `getJ` method that multiplies each object in the `_getJ` dictionary by the respective mapping derivative, then sums them all together.

ghwilliams:
> Quick note on how this should be done:
> 
> 1: change current `getJ` to `_getJ` 2: change internal calls inside `Jvec`, `JTvec` and `getJtJdiag` from `getJ` to `_getJ` 3: add new `getJ` method that multiplies each object in the `_getJ` dictionary by the respective mapping derivative, then sums them all together.

@jcapriot I've already made many of these modifications. Yesterday, during the dev meeting, we discussed (@santisoler and I) the implementation of the new getJ. We found out that there is a problem related to the order in which the full sensitivity matrix must be assembled from the individual blocks corresponding to the sensitivity related to all the different parameters composing the full model parameter vector, whose order is defined by the user when creating the simulation object. It is unclear how we can obtain this information from inside getJ. We know the user may have defined a composed mapping, for example, by multiplying an ExpMap by a Wire map. I'm looking for ways to get the specific order in which the user composed the full model parameter vector from this map (inside getJ). Please look at the code and let me know if you think getJ could be implemented differently, possibly avoiding this order issue altogether.

simpeg/electromagnetics/base_1d.py
```diff
@@ -335,29 +335,11 @@ def compute_complex_mu(self, frequencies):
 
     def Jvec(self, m, v, f=None):
         Js = self.getJ(m, f=f)
```
jcapriot, line 337: The only change to this function should be to change `getJ` to `_getJ`

simpeg/electromagnetics/base_1d.py
```diff
@@ -335,29 +335,11 @@ def compute_complex_mu(self, frequencies):
...
-            out = out + Js["ds"] @ (self.sigmaDeriv @ v)
-        if self.muMap is not None:
-            out = out + Js["dmu"] @ (self.muDeriv @ v)
-        if self.thicknessesMap is not None:
-            out = out + Js["dthick"] @ (self.thicknessesDeriv @ v)
-        return out
+        return Js @ v
 
     def Jtvec(self, m, v, f=None):
         Js = self.getJ(m, f=f)
```
jcapriot, line 341: Only change to this function should be `getJ` to `_getJ`

simpeg/electromagnetics/frequency_domain/simulation_1d.py
```diff
@@ -244,6 +245,28 @@ def getJ(self, m, f=None):
                     self._J["dthick"] = self._project_to_data(v_dthick)
         return self._J
 
+    def getJ(self, m, f=None):
```
jcapriot, line 248: All you have to do is sum them…

`Js[‘ds’] @ sigmaDeriv + Js[`dmu`] @ muDeriv +…`

Of course you’ll need to conditionally sum them if they’re there.

jcapriot (changes requested):

jcapriot:
> > Quick note on how this should be done:
> > 1: change current `getJ` to `_getJ` 2: change internal calls inside `Jvec`, `JTvec` and `getJtJdiag` from `getJ` to `_getJ` 3: add new `getJ` method that multiplies each object in the `_getJ` dictionary by the respective mapping derivative, then sums them all together.
> 
> @jcapriot I've already made many of these modifications. Yesterday, during the dev meeting, we discussed (@santisoler and I) the implementation of the new getJ. We found out that there is a problem related to the order in which the full sensitivity matrix must be assembled from the individual blocks corresponding to the sensitivity related to all the different parameters composing the full model parameter vector, whose order is defined by the user when creating the simulation object. It is unclear how we can obtain this information from inside getJ. We know the user may have defined a composed mapping, for example, by multiplying an ExpMap by a Wire map. I'm looking for ways to get the specific order in which the user composed the full model parameter vector from this map (inside getJ). Please look at the code and let me know if you think getJ could be implemented differently, possibly avoiding this order issue altogether.

You haven’t made those changes, and I’ve marked them directly on the review for you now. You’re overthinking it, if you implement it as I have requested, it is minimal changes overall.

simpeg/electromagnetics/frequency_domain/simulation_1d.py
ghwilliams: Ok. I'll do it right now and see if it works. Thanks @jcapriot .

ghwilliams:
@jcapriot @santisoler I updated getJ following Joseph's suggestions. I ran the tests in tests/em/em1d, and they all passed. Looks like it will work.

jcapriot:
Yeah, the getJ works as I expected. Just FYI the way mappings work is that they map from the entire model vector to a specific physical property. Their derivatives naturally include the necessary information about which model parameters the physical properties are derivatives of.

Also please implement my second requested point:

> 2: change internal calls inside `Jvec`, `Jtvec` and `getJtJdiag` from `getJ` to `_getJ`

and revert all your other changes to these functions.

I don’t want them to unnecessarily create matrices full of zeros.

simpeg/electromagnetics/frequency_domain/simulation_1d.py
```diff
@@ -244,6 +244,32 @@ def getJ(self, m, f=None):
...
+
+        if self.thicknessesMap is not None:
+            J = (
+                Js["dthick"] @ self.thicknessesDeriv
+                if J is None
+                else J + Js["dthick"] @ self.thicknessesDeriv
+            )
+
+        return J
+
```
santisoler, line 272: Nice! I'm glad to see that this works as easily as this! I'm just leaving a minor refactor suggestion to simplify the code. I think this is easier to read, but let me know if you disagree.

This code assumes that the `_getJ` method only adds pieces of the jacobian for the parameters that have a mapping assigned. That's why I'm not checking if mappings are None here.

```python
    def getJ(self, m, f=None):
        # Get Js matrices (we need to set the model before accessing derivatives)
        Js = self._getJ(m, f=f)
        # Map parameters with their corresponding derivatives
        param_and_derivs = {
            "dh": self.hDeriv,
            "ds": self.sigmaDeriv,
            "dmu": self.muDeriv,
            "dthick": self.thicknessesDeriv,
        }
        # Compute J matrix
        J = sum(Js[param] @ param_and_derivs[param] for param in Js)
        return J
```

santisoler (commented):
Thanks @ghwilliams! I'm glad to see that the ordering of the J rows is easily solved. I left a comment below regarding that implementation. Let me know what do you think.

Thanks also for adding the tests in `tests/em/em1d/test_EM1D_FD_jac_layers2.py`. If you have the time, it would be nice to switch those tests from using `unittest` to using `pytest`. It's not a huge blocker, but such change would align with the current way of writing tests in the project. I can totally do it afterwards if you don't have the time or interest, no worries.

simpeg/electromagnetics/frequency_domain/simulation_1d.py
ghwilliams: @santisoler Ok. I'll try those suggestions. Regarding rewriting the tests using pytest, I could open another PR for that because all the tests under tests\em\em1d use unittest. Let's discuss this further during today's dev meeting.

santisoler: Sure! I didn't mean to change _all_ tests in `em1d` to use pytest in this PR. I meant to write _new_ tests using pytest and avoid using unittest in those.

santisoler:
**TODO Before merging**:

- [ ] Include a docstring to the `getJ` method

ghwilliams:
> @ghwilliams, I pushed some docstrings to the `getJ` and `_getJ` methods of the TDEM and FDEM simulations. I also changed the title of the PR and wrote a small summary I'll use when merging it.
> 
> The CIs pass, so I think I'll merge this tomorrow morning (local time). If you have the chance, could you take another look to my changes?
> 
> Thanks again for the contributions!

Everything is fine. The documentation for both _getJ and getJ is ok.

jcapriot:
Should add a test for the JtJdiag method

PR #1656, opened 2025-04-25
CG Minimizer Updates

jcapriot:
#### Summary
Contains several updates for the CG based minimizer routines and their options to bring in line with more recent versions of SciPy. Also contains a bit of common code movement by using a `Bounded`  and an `InexactCG` mix in classes for the minimizers.

#### PR Checklist
      to a Pull Request

Our CG based minimizers did not support specifying both a relative tolerance and an absolute tolerance to the CG routines. Moreover, the option `tolCG` did not control the same behavior amongst the Inexact minimizer routines. For these reasons I've added the following options, (in line with pep8):

* `cg_rtol`
* `cg_atol`

For the deprecations and translations:
* On `ProjectedGradient`, and `InexactGaussNewton`, the `tolCG` parameter is now `cg_rtol`.
* On `ProjectedGNCG`, `tolCG` was used as an absolute tolerance (thus the variable behavior of this class for certain problems). I have currently set the default parameters of the class to mimic the current behavior, but included a notice that the defaults will change in a future release. I would recommend updating them to `cg_rtol=1E-3` and `cg_atol=0.0`, for better results throughout.

There are also a few QOL improvements for when using `InexactGaussNewton` and `ProjectedGNCG` in an inversion:
 * Biggest is that they no longer do unneccessary operation involving the BFGS minimizer when when using an `UpdatePreconditioner` directive (a useless and large memory and time sink in the current implementation that immediately gets discarded.)
   * `BaseInvProblem` now has an `init_bfgs` option, and will also only initialize BFGS on `BFGS` minimizers.
   * `Inversion` will check if there is an `UpdatePrecondtioner` directive with `InexactGaussNewton` minimizers and turn off bfgs intialization. 
   * `DirectiveList` is now iterable (to make the previous check easier).
 * There are more `debug` statements in `ProjectedGNCG` available during the CG iterations.
 * `ProjectedGNCG` adds more information to the inversion log about the convergence of the CG iterations.

#### Additional information
I've added a few more detailed notes to the minimization routines that are "correct" in that they were previously loose with their uses of "Active set" and "Inactive set" for bounded problems. Literature states that the "Active Set" are values on the bounds, and "Inactive set" are values not on the bounds. 

I'll add tests for these changes shortly...

### DIFF
--- a/.ci/environment_test.yml
+++ b/.ci/environment_test.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - numpy>=1.22
-  - scipy>=1.8
+  - scipy>=1.12
   - pymatsolver>=0.3
   - matplotlib-base
   - discretize>=0.11
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 # dependencies
   - python=3.11
   - numpy>=1.22
-  - scipy>=1.8
+  - scipy>=1.12
   - pymatsolver>=0.3
   - matplotlib-base
   - discretize>=0.11


simpeg/optimization.py
```diff
@@ -248,6 +237,22 @@ class IterationPrinters(object):
...
+    }
+
+    iteration_CG_abs_residual = {
+        "title": "CG |Ax-b|",
+        "value": lambda M: M.cg_abs_resid,
+        "width": 11,
+        "format": "%1.2e",
+        # "format": lambda v: f"{v:1.2e}",
+    }
+
```
jcapriot, line 255: Open to suggestions on the title of these two writers... some thoughts of mine:
* `CG rel resid`, `CG abs resid`
* `CG residual_r`, `CG residual_a`
* ... ?

simpeg/optimization.py
```diff
@@ -436,7 +441,11 @@ def startup(self, x0):
         self.iterLS = 0
         self.stopNextIteration = False
 
-        x0 = self.projection(x0)  # ensure that we start of feasible.
+        try:
+            x0 = self.projection(x0)  # ensure that we start of feasible.
+        except Exception as err:
+            raise RuntimeError("Initial model is not projectable") from err
+
```
jcapriot, line 448: This is to add a slightly improved message for `Bounded` problems, will be thrown if initial model does not broadcast with the upper or lower bounds.

simpeg/optimization.py
```diff
@@ -774,45 +783,50 @@ def _doEndIterationRemember(self, *args):
                 self._rememberList[param[0]].append(param[1](self))
 
 
-class ProjectedGradient(Minimize, Remember):
-    name = "Projected Gradient"
-
-    maxIterCG = 5
-    tolCG = 1e-1
+class Bounded(object):
```
jcapriot, line 786: There were multiple classes that had the same "bound" options, now it's unified

simpeg/optimization.py
```diff
@@ -821,7 +835,11 @@ def projection(self, x):
         Make sure we are feasible.
 
         """
-        return np.median(np.c_[self.lower, x, self.upper], axis=1)
+        if self.lower is not None:
+            x = np.maximum(x, self.lower)
+        if self.upper is not None:
+            x = np.minimum(x, self.upper)
+        return x
```
jcapriot, line 842: `lower` and `upper` are now both optional arguments (even though they are defaulted to `-np.inf` and `np.inf` in the concrete implementations, they could be set to `None` for the same effect and avoid an operation.

simpeg/optimization.py
```diff
@@ -851,9 +874,120 @@ def bindingSet(self, x):
...
-        return np.logical_or(bind_up, bind_low)
+        out = np.zeros(x.shape, dtype=bool)
+        if self.lower is not None:
+            out |= (x <= self.lower) & (self.g >= 0)
+        if self.upper is not None:
+            out |= (x >= self.upper) & (self.g <= 0)
+        return out
+
+
+class InexactCG(object):
```
jcapriot, line 885: Used to hold the CG options... used by `ProjectedGradient`, `InexactGaussNewton` and `ProjectedGNCG` minimizers.

simpeg/optimization.py
```diff
@@ -954,9 +1090,6 @@ class BFGS(Minimize, Remember):
     name = "BFGS"
     nbfgs = 10
 
-    def __init__(self, **kwargs):
-        Minimize.__init__(self, **kwargs)
-
```
jcapriot, line 959: This was effectively doing nothing

simpeg/optimization.py
```diff
@@ -1038,7 +1171,7 @@ def findSearchDirection(self):
         return Solver(self.H) * (-self.g)
 
 
-class InexactGaussNewton(BFGS, Minimize, Remember):
+class InexactGaussNewton(InexactCG, BFGS):
```
jcapriot, line 1174: In the future, I'd like to remove the BFGS Inheritance here...
Currently it is only used as a default preconditioner to the inner CG iterations, which is not necessarily a bad thing if the objective function doesn't change, but it's questionable if it is appropriate to continue using it when a directive changes the objective function, as the BFGS memory is not related to the current objective function. I'm not sure of the best approach just yet, but I don't think inheriting from BFGS just to get access to it's behavior for default options is a good approach.

simpeg/optimization.py
```diff
@@ -1055,13 +1188,14 @@ class InexactGaussNewton(BFGS, Minimize, Remember):
...
 
-    name = "Inexact Gauss Newton"
+        self._was_default_hinv = False
 
-    maxIterCG = 5
-    tolCG = 1e-1
+    name = "Inexact Gauss Newton"
 
     @property
     def approxHinv(self):
```
jcapriot, line 1201: This should likely be moved to the `InexactCG` class and then used within all of the `Inexact` step direction minimizers. It's left here due to the default BFGS preconditioner.

simpeg/optimization.py
```diff
@@ -1077,7 +1211,9 @@ def approxHinv(self):
             M = sp.linalg.LinearOperator(
                 (self.xc.size, self.xc.size), self.bfgs, dtype=self.xc.dtype
             )
+            self._was_default_hinv = True
             return M
+        self._was_default_hinv = False
```
jcapriot, line 1216: used to conditionally do the `BFGS` iteration end routines,

simpeg/optimization.py
```diff
@@ -1196,88 +1339,127 @@ def evalFunction(x, return_g=False):
...
+        self.printers.extend(
+            [
+                IterationPrinters.iterationCG,
+                IterationPrinters.iteration_CG_rel_residual,
+                IterationPrinters.iteration_CG_abs_residual,
+            ]
+        )
 
-    @count
-    def projection(self, x):
```
jcapriot, line 1221: All functionality for the box bound constrains is moved to the `Bounded` mixin, which is also re-used by `ProjectedGradient`

simpeg/optimization.py
```diff
@@ -1196,88 +1339,127 @@ def evalFunction(x, return_g=False):
...
+        r_norm0 = norm(r)
 
-        while np.all([np.linalg.norm(r) > self.tolCG, count < self.maxIterCG]):
+        atol = max(self.cg_rtol * norm(r_norm0), self.cg_atol)
+        if self.debug:
+            print(f"CG Target tolerance: {atol}")
+        r_norm = r_norm0
+        while r_norm > atol and count < self.cg_maxiter:
+            if self.debug:
+                print(f"CG Iteration: {count}, residual norm: {r_norm}")
```
jcapriot, line 1462: Previous tolerance was always an absolute tolerance, making the behavior of this minimizer very problem dependent... This updated check is exactly equivalent to `scipy.sparse.linalg.cg`.

simpeg/optimization.py
```diff
@@ -1287,33 +1469,57 @@ def findSearchDirection(self):
...
+        # if that gradient would move them away from the boundary.
+        if self.step_active_set and sum(Inactive) != self.xc.size:
+            rhs_a = Active * -self.g
+
+            # reasonable guess at the step length for the gradient on the
+            # active cell boundaries. Basically scale it to have the same
+            # maximum as the cg step on the cells that are not on the
+            # boundary.
             dm_i = max(abs(step))
             dm_a = max(abs(rhs_a))
```
jcapriot, line 1496: should consider adding a stabilizer to this if `rhs_a` are all zeros.

simpeg/directives/directives.py
```diff
@@ -283,6 +283,9 @@ def __init__(self, *directives, inversion=None, debug=False, **kwargs):
         self.inversion = inversion
         self.verbose = debug
 
+    def __iter__(self):
+        return self.dList.__iter__()
```
jcapriot, line 287:
```suggestion
        return iter(self.dList)
```

simpeg/optimization.py
domfournier: How can this happen? 
Should we not validate the type of x0 instead?

simpeg/optimization.py
```diff
@@ -830,7 +848,12 @@ def activeSet(self, x):
         If we are on a bound
 
         """
-        return np.logical_or(x == self.lower, x == self.upper)
+        out = np.zeros(x.shape, dtype=bool)
+        if self.lower is not None:
+            out |= x <= self.lower
+        if self.upper is not None:
+            out |= x >= self.upper
+        return out
```
domfournier, line 874: That's good, don't need to expect that the projection has happened already.

simpeg/optimization.py
domfournier: Would be descent docstrings to be added ;)

simpeg/optimization.py
```diff
@@ -851,9 +874,120 @@ def bindingSet(self, x):
...
+class InexactCG(object):
+
+    def __init__(self, *, cg_rtol=1e-1, cg_atol=0, cg_maxiter=5, **kwargs):
+        super().__init__(**kwargs)
+        self.cg_rtol = cg_rtol
+        self.cg_atol = cg_atol
+        self.cg_maxiter = cg_maxiter
+
+    @property
+    def cg_atol(self):
```
domfournier:
```suggestion
    def cg_atol(self) -> float:
```
Any objection if we start switching to typing? I think it will pay off down the road if we start bringing those in as we retouch the classes.

simpeg/optimization.py
```diff
@@ -851,9 +874,120 @@ def bindingSet(self, x):
...
+        """Absolute tolerance for inner CG iterations.
+
+        CG iterations are terminated if:
+        >>> norm(A @ x_k - b) <= max(cg_rtol * norm(A @ x_0 - b), cg_atol)
+
+        or if the maximum number of CG iterations is reached.
+
+        Returns
+        -------
+        float
```
domfournier, line 960:
```suggestion
```
Won't need all those if we just type the return.

simpeg/optimization.py
```diff
@@ -1196,88 +1339,132 @@ def evalFunction(x, return_g=False):
...
+        # "inactive" are cells with values inside the limits.
+
+        # The basic logic of this method is to do CG iterations only
+        # on the inactive set, then also add a scaled gradient for the
+        # active set, (if that gradient points away from the limits.)
+
         self.cg_count = 0
         Active = self.activeSet(self.xc)
-        temp = sum((np.ones_like(self.xc.size) - Active))
+        Inactive = 1 - Active
```
domfournier:
```suggestion
        inactive = 1 - active
```
Good time to lower case those variables.

simpeg/optimization.py
domfournier: So can we just remove the CG while loop here? Feels like we could just multiple the `self.approxHinv` with the `inactive` and pass it to scipy?

domfournier (commented):
Looks great. Feels like it's a good time to combine `ProjectedGradient` and `ProjectedGNCG`. We just need to make the default `approxHinv` to be identity, and bring back the step off bound part of the other.

simpeg/optimization.py
santisoler: No objection at all. We started adding them in new pieces of code. So type hints are welcomed!

simpeg/optimization.py
santisoler: I would still leave the Returns in the docstrings. The main reason is because Sphinx can be configured to hide type hints in API docs, to not convolute the signature. In those cases, having the Returns section documents the outputs.

simpeg/optimization.py
```diff
@@ -911,11 +1054,13 @@ def reduceHess(v):
...
-
-            inp = {tol_key: self.tolCG, "maxiter": self.maxIterCG}
-            p, info = sp.linalg.cg(operator, -Z.T * self.g, **inp)
+            p, info = sp.linalg.cg(
+                operator,
+                -Z.T * self.g,
+                rtol=self.cg_rtol,
+                atol=self.cg_atol,
+                maxiter=self.cg_maxiter,
+            )
```
santisoler, line 1108: One minor thought here. What if instead of defining `operator` as a `LinearOperator` with the `Z` acting as a preconditioner (and having to include `Z.T` on the right hand side too), we pass `Z` as a preconditioner to `cg`? Something like:

```suggestion
            p, info = sp.linalg.cg(
                self.H,
                -self.g,
                M=Z,  # use Z as preconditioner
                rtol=self.cg_rtol,
                atol=self.cg_atol,
                maxiter=self.cg_maxiter,
            )
```

simpeg/optimization.py
santisoler: What about `CG residual (abs)` and `CG residual (rel)`?

santisoler:
I like these changes @jcapriot! Feel free to start adding tests so we can aim to include this PR in the next release.

simpeg/optimization.py
jcapriot: `Z` is not a preconditioner here, it is a matrix that reduces the problem to the inactive set (the model values not equal to bounds)

simpeg/optimization.py
jcapriot: I know late but, it could fail if a bound is not the correct shape relative to the model (length mismatches, dimension mismatches, etc.)

santisoler (approved):
Looks good to me, @jcapriot.

After we merge this, tests and doc builds will get a lot of FutureWarnings coming from setting `tolCG` mostly. Are you thinking we should update those in a separate PR and add filterwarnings to `pyproject.toml` for those warnings?

PR #1657, opened 2025-04-25
Standardize output directives and make the more reliable

jcapriot:
#### Summary
Fixes and standardizes the `SaveEveryIteration` child directives. They now reliably use the passed `directory` argument, and this also adds a few more safeguards to the file writes.

#### PR Checklist
      to a Pull Request

The `SaveEveryIteration` directives were not very reliable... Some always output the file to the current working directory, some had odd file names, They all had different parameter names which effectively controlled the same thing. Also there were potentially erroring write outs (if a user deleted an output directory in between iterations).

* This unifies the different parameter names that were controlling if the directive saved information to disk, now `on_disk` instead of `save_txt` and `saveOnDisk`.
* Uses the more portable `pathlib.Path` to describe the output directory and full file path.
* Notifies the user of the absolute file path where the output will be stored.
* Uses a bit more inheritance to reduce duplicated code.

Will add tests soon

### DIFF
--- a/simpeg/directives/directives.py
+++ b/simpeg/directives/directives.py
@@ -1839,7 +1839,7 @@ class SaveModelEveryIteration(SaveEveryIteration):
 
     @property
     def file_abs_path(self):
-        return self.directory / self._time_iter_file_name.with_suffix("npy")
+        return self.directory / self._time_iter_file_name.with_suffix(".npy")
 
     def endIter(self):
         self._mkdir_and_check_output_file(should_exist=False)
@@ -2101,7 +2101,7 @@ class SaveOutputDictEveryIteration(SaveEveryIteration):
     @property
     def file_abs_path(self):
         if self.on_disk:
-            return self.directory / self._time_iter_file_name.with_suffix("npz")
+            return self.directory / self._time_iter_file_name.with_suffix(".npz")
 
     def initialize(self):
         super(
--- a/simpeg/directives/directives.py
+++ b/simpeg/directives/directives.py
@@ -1,8 +1,11 @@
+from abc import ABCMeta, abstractmethod
 from typing import TYPE_CHECKING
+
+from datetime import datetime
+import pathlib
 import numpy as np
 import matplotlib.pyplot as plt
 import warnings
-import os
 import scipy.sparse as sp
 from ..typing import RandomSeed
 from ..data_misfit import BaseDataMisfit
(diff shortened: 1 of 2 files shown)


simpeg/directives/_directives.py
```diff
@@ -1678,74 +1706,108 @@ def name(self, value):
...
         )
 
+    @property
+    def on_disk(self):
+        """This class always saves to disk."""
+        return True
+
+    @property
+    def file_abs_path(self):
+        return self.directory / self._time_iter_file_name.with_suffix(".npy")
```
santisoler: Nit picking here: since the property is called `file_abs_path`, wouldn't it be better to resolve the `directory` (or use `directory.absolute()`)?
```suggestion
    @property
    def file_abs_path(self):
        return self.directory.resolve() / self._time_iter_file_name.with_suffix(".npy")
```

simpeg/directives/_directives.py
```diff
@@ -1782,40 +1856,48 @@ def endIter(self):
...
+                        self.opt.iter,
+                        self.beta[-1],
+                        self.phi_d[-1],
+                        self.phi_m[-1],
+                        self.phi_m_small[-1],
+                        self.phi_m_smooth_x[-1],
+                        self.phi_m_smooth_y[-1],
+                        self.phi_m_smooth_z[-1],
+                        self.phi[-1],
+                    )
```
santisoler, line 1941: Style suggestion to use f-strings instead. I'm respecting the double spaces leading some of the fields, but not sure if those are wanted or not.
```suggestion
                f.write(
                    f" {self.opt.iter:3d}"
                    f" {self.beta[-1]:1.4e}"
                    f" {self.phi_d[-1]:1.4e}"
                    f" {self.phi_m[-1]:1.4e}"
                    f" {self.phi_m_small[-1]:1.4e}"
                    f" {self.phi_m_smooth_x[-1]:1.4e}"
                    f" {self.phi_m_smooth_y[-1]:1.4e}"
                    f"  {self.phi_m_smooth_z[-1]:1.4e}"
                    f"  {self.phi[-1]:1.4e}\n"
```

simpeg/directives/_sim_directives.py
```diff
@@ -168,30 +154,39 @@ def endIter(self):
...
+                    " {0:2d}  {1}  {2}  {3}  {4}  {5:1.4e}  {6:d}  {7:1.4e}\n".format(
+                        self.opt.iter,
+                        self.betas[-1],
+                        self.lambd[-1],
+                        self.phi_d[-1],
+                        self.phi_m[-1],
+                        self.phi_sim[-1],
+                        self.opt.cg_count,
+                        self.phi[-1],
+                    )
```
santisoler:
```suggestion
                    f.write(
                        f" {self.opt.iter:2d}"
                        f"  {self.betas[-1]}"
                        f"  {self.lambd[-1]}"
                        f"  {self.phi_d[-1]}"
                        f"  {self.phi_m[-1]}"
                        f"  {self.phi_sim[-1]:1.4e}"
                        f"  {self.opt.cg_count:d}"
                        f"  {self.phi[-1]:1.4e}"
                        f"\n"
```

santisoler (approved):
Looks good to me, @jcapriot! I just left some minor comments, mostly nit picking. Do yo want to add some tests to it? Let me know if you need any help with those.

simpeg/directives/_directives.py
jcapriot: it probably would be better, but was just leaning towards not modifying the current method of writing out, just to make it a bit more reliable on the location they write to :)

simpeg/directives/_directives.py
jcapriot: `resolve()` is already called on the setter for `self.directory`.

jcapriot:
> Looks good to me, @jcapriot! I just left some minor comments, mostly nit picking. Do yo want to add some tests to it? Let me know if you need any help with those.

Adding tests would be great, I won't be able to get to it until later this week so if you want to do it yourself before that, feel free.

simpeg/directives/_directives.py
santisoler: Sounds good.

simpeg/directives/_directives.py
```diff
@@ -1678,74 +1706,108 @@ def name(self, value):
...
 
     This directive saves the model as a numpy array at each iteration. The
-    default directory is the current directoy and the models are saved as
-    ``InversionModel-YYYY-MM-DD-HH-MM-iter.npy``
+    default directory is the current directory and the models are saved as
+    ``InversionModel_YYYY-MM-DD-HH-MM_iter.npy``
     """
 
+    def __init__(self, **kwargs):
+        kwargs.pop("on_disk", None)
```
santisoler: Should we warn users that we are ignoring the `on_disk` argument?

jcapriot: It’s mostly just a guard because it is explicitly true for this class, and cannot be changed. It’s not part of the signature for the class and has not been. It could be changed to throw a `TypeError` instead of just popping it off though.

simpeg/directives/_directives.py
```diff
@@ -1755,6 +1817,18 @@ def initialize(self):
         self.phi_m_smooth_z = []
         self.phi = []
 
+    @property
+    def file_abs_path(self):
+        if self.save_txt:
```
santisoler: I guess we should use `on_disk` since `save_txt` is deprecated.
```suggestion
        if self.on_disk:
```

PR #1661, opened 2025-05-08
Improved right-hand side for NSEM primary secondary 

dccowan:
#### Summary
This PR is meant to improve the 3D NSEM primary secondary formulation. More suitable boundary conditions are implemented when solving the 1D problem that is used as the primary solution on the 3D mesh.

#### PR Checklist
      to a Pull Request

The boundary conditions that are currently used to solve the 1D problem for the 3D primary-secondary simulation are sub-optimal. As a result, one needs to pad out excessively when using the 3D primary secondary formulation to obtain numerically accurate results. Here, more suitable boundary conditions are implemented to solve the 1D problem.

### DIFF
--- a/simpeg/electromagnetics/natural_source/utils/source_utils.py
+++ b/simpeg/electromagnetics/natural_source/utils/source_utils.py
@@ -190,7 +190,7 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
 
 
 def primary_e_1d_solution(mesh, sigma_1d, freq, n_pad=2000):
-    """Compute 1D electric field solution. 
+    r"""Compute 1D electric field solution. 
 
     Parameters
     ----------
@@ -212,7 +212,125 @@ def primary_e_1d_solution(mesh, sigma_1d, freq, n_pad=2000):
 
     Notes
     -----
-    ADD DERIVATION HERE.
+    For the 1D electric field solution, Maxwell's equations take the form:
+
+    .. math::
+        \begin{align}
+        &\frac{\partial e_x}{\partial z} + i\omega b_y = 0 \\
+        &\frac{\partial h_y}{\partial z} + \sigma e_x = 0
+        \end{align}
+
+    **Boundary Conditions:**
+
+    At the top boundary, we set $h_y^{(top)} = 1$. Therefore from Faraday's law:
+
+    .. math::
+        \frac{\partial e_x^{(top)}}{\partial z} = - i\omega\mu_0
+
+    At the bottom boundary, there is only a downgoing wave of the form:
+
+    .. math::
+        e_x = E^- \exp (ikz)
+
+    where
+
+    .. math::
+        k = \sqrt{-i\omega\mu_0\sigma} = (1 - i)\sqrt{\frac{\omega \mu_0\sigma}{2}}
+
+    So if $\Delta z$ is negative, the downgoing wave decays. From Faraday's law:
+
+    .. math::
+        \begin{align}
+        & ik E^- \exp (ikz) + i\omega b = 0 \\
+        \implies & ik e_x + i \omega b = 0 \\
+        \implies & k e_x + \omega b = 0
+        \end{align}
+
+    **Discrete system:**
+
+    We take the inner product of Faraday's law with a test function $\mathbf{u}$, and the
+    inner product of Ampere's law with a test function $\mathbf{f}$.
+    .. math::
+        \begin{align}
+        &\langle u , \partial_z e_x \rangle + i\omega \langle u , b \rangle = 0 \\
+        &\langle f , \partial_z h_y \rangle + \langle f, \sigma e \rangle = 0
+        \end{align}
+
+    The inner-product with Ampere's law is Integrated by parts:
+    
+    .. math::
+        \begin{align}
+        &\langle u , \partial_z e_x \rangle + i\omega \langle u , b_y \rangle = 0 \\
+        &- \langle \partial_z f , h_y \rangle + f h_y \bigg |_{bot}^{top} + \langle f, \sigma e_x \rangle = 0
+        \end{align}
+
+    In discrete form, the above equations are approximated by:
+    .. math::
+        \begin{align}
+        &\mathbf{G_n e} = - i\omega \mathbf{b} \\
+        &-i \omega \mathbf{G_n^T M_{\mu} b} + i \omega\mathbf{M_\sigma e} + i\omega h_y \bigg |_{bot}^{top} = 0
+        \end{align}
+
+    Combining these equations, we obtain the following system:
+    
+    .. math::
+        \big [ \mathbf{G_n^T M_{\mu} G_n} + i \omega\mathbf{M_\sigma} \big ] \mathbf{e} + i\omega h_y \bigg |_{bot}^{top} = 0
+
+    When $\mu = \mu_0$, we can multiply through an obtain:
+
+    .. math::
+        \big [ \mathbf{G_n^T G_n} + i \omega \mu_0 \, diag (\sigma) \big ] \mathbf{e} + i\omega \mu_0 h_y \bigg |_{bot}^{top} = 0
+    
+    **Implementing Discrete Boundary Conditions:**
+
+    At the top node of the mesn, we know that:
+
+    .. math::
+    \frac{-e_{n-1} + 2 e_n - e_{n+1}}{h^2} + i \omega \mu_0 \sigma_n e_n
+    = \bigg ( \frac{-e_{n-1} + e_n}{h^2} \bigg ) + \bigg ( \frac{e_n - e_{n+1}}{h^2} \bigg ) + i \omega \mu_0 \sigma_n e_n = 0
+
+    Using Taylor expansion:
+
+    .. math::
+        e_{n+1} - e_n = \frac{\partial e_n}{\partial z} h = - i\omega\mu_0 h
+
+    Thus:
+
+    .. math::
+        \frac{-e_{n-1} + 2 e_n - e_{n+1}}{h^2} + i \omega \mu_0 \sigma_n e_n
+        = \bigg ( \frac{-e_{n-1} + e_n}{h^2} \bigg ) + \bigg ( \frac{i\omega \mu_0}{h} \bigg ) + i \omega \mu_0 \sigma_n e_n = 0
+
+    And we set
+    
+    .. math::
+        q_n = - \frac{i \omega \mu_0}{h}
+
+    At the bottom of the mesh, we know that:
+
+    .. math::
+    \frac{-e_{-1} + 2 e_0 - e_{1}}{h^2} + i \omega \mu_0 \sigma e_0
+    = \frac{-e_{-1} + e_0}{h^2} + \frac{e_0 - e_{1}}{h^2} + i \omega \mu_0 \sigma e_0 = 0
+
+    But we know at the bottom, from Taylor expansion:
+
+    .. math::
+    e_{-1} - e_0 = -\frac{\partial e_0}{\partial z} h = i\omega b_0 h
+
+    Thus:
+
+    .. math::
+        \frac{e_0 - e_{1}}{h^2} + i \omega \mu_0 \sigma e_0 - \frac{i \omega b_0}{h} = 0
+
+    But
+
+    .. math::
+        k e_0 + \omega b_0 = 0
+
+    So
+
+    .. math::
+        \frac{e_0 - e_{1}}{h^2} + i \omega \mu_0 \sigma e_0 + \frac{ik}{h} e_0 = 0
+
     """
 
     # Extract vertical discretization
@@ -253,6 +371,20 @@ def primary_e_1d_solution(mesh, sigma_1d, freq, n_pad=2000):
 
 
 def project_e_1d_to_e_primary(mesh, e_1d):
+    """Project 1D electric field solution on nodes to edges of a mesh. 
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseTensorMesh
+        A 1d, 2d or 3d tensor mesh or tree mesh.
+    e_1d : np.ndarray
+        1D electric field solution along the vertical discretization.
+
+    Returns
+    -------
+    numpy.ndarray (n_edges, n_polarization)
+        Electric fields on the edges of the mesh for each polarization.
+    """
 
     if mesh.dim == 1:
         return e_1d
@@ -260,25 +392,27 @@ def project_e_1d_to_e_primary(mesh, e_1d):
     hz = mesh.h[-1]
     mesh_1d = TensorMesh([hz], origin=[mesh.origin[-1]])
 
-    if mesh.dim == 2:
-        raise NotImplementedError("The 'project_e_1d_to_e_primary' function has not been implemented for 2D meshes.")
+    # Incident E-field polarized along x-direction
+    ep_x = (
+        mesh_1d.get_interpolation_matrix(
+            mesh.edges_x[:, -1], location_type="nodes"
+        ) @ e_1d
+    )
 
-    if mesh.dim == 3:
-        # Incident E-field polarized along x-direction
-        fields_x = (
-            mesh_1d.get_interpolation_matrix(
-                mesh.edges_x[:, 2], location_type="nodes"
-            ) @ e_1d
-        )
-        ep_x = np.r_[fields_x, np.zeros(mesh.n_edges_y + mesh.n_edges_z)]
+    if mesh.dim == 2:
+        return np.r_[ep_x, np.zeros(mesh.n_edges_y)]
+
+    elif mesh.dim == 3:
+        
+        ep_x = np.r_[ep_x, np.zeros(mesh.n_edges_y + mesh.n_edges_z)]
 
         # Incident E-field polarized along y-direction
-        fields_y = (
+        ep_y = (
             mesh_1d.get_interpolation_matrix(
                 mesh.edges_y[:, 2], location_type="nodes"
             ) @ e_1d
         )
-        ep_y = np.r_[np.zeros(mesh.n_edges_x), fields_y, np.zeros(mesh.n_edges_z)]
+        ep_y = np.r_[np.zeros(mesh.n_edges_x), ep_y, np.zeros(mesh.n_edges_z)]
 
         return np.c_[ep_x, ep_y]
 
--- a/simpeg/electromagnetics/natural_source/utils/source_utils.py
+++ b/simpeg/electromagnetics/natural_source/utils/source_utils.py
@@ -1,8 +1,11 @@
 import numpy as np
-import discretize
+from scipy.constants import mu_0
+from discretize import TensorMesh
+from discretize.utils import sdiag
 
 from ....utils import mkvc
 from .solutions_1d import get1DEfields
+from pymatsolver import Solver
 
 
 def homo1DModelSource(mesh, freq, sigma_1d):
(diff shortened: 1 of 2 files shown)


dccowan:
> Looking through the code, the only thing this really does different to the current implementation is that it solves the primary problem on a mesh that is padded beyond the current mesh correct? Is there any reason we don't just use the analytic response for this in general?

I wasn't quite sure what boundary conditions were being imposed originally because the underlying math wasn't documented. I only knew that they weren't correct. One thing this PR is working towards is better organization of the NSEM module and setting things up to add the fictitious sources approach.

jcapriot:
It would also appear there's some errors in the bottom boundary condition in the implementation as well, Some quick comparison's on my end:

image

As you can see, there is a large phase error at the bottom boundary in this proposal.

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -183,3 +186,237 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
+    A[0, 1] = -1 / hz[0] ** 2
+
+    q = np.zeros(mesh_1d_ext.n_faces, dtype=np.complex128)
+    q[-1] = -1j * w * mu_0 / hz[-1]
+
+    Ainv = Solver(A)
+    e_1d = Ainv * q
+
+    # Return solution along original vertical discretization
+    return e_1d[n_pad:-n_pad]
```
jcapriot:
```suggestion
def primary_e_1d_solution(mesh, sigma_1d, freq, n_pad=2000):

    # Extract vertical discretizations
    hz = mesh.h[-1]

    if len(hz) != len(sigma_1d):
        raise ValueError(
            "Number of cells in vertical direction must match length of 'sigma_1d'. Here hz has length {} and sigma_1d has length {}".format(
                len(hz), len(sigma_1d)
            )
        )

    # Generate extended 1D mesh and model to solve 1D problem
    hz = np.pad(hz, (n_pad, n_pad), mode='edge')
    mesh = discretize.TensorMesh([hz], origin=[mesh.origin[-1] - hz[0] * n_pad])

    sigma_1d = np.pad(sigma_1d, (n_pad, n_pad), mode='edge')
    mu = np.full(mesh.n_cells, mu_0)
    # Solve the 1D problem for electric fields on nodes/faces
    G = mesh.nodal_gradient
    M_e_mui = mesh.get_edge_inner_product(mu, invert_model=True)
    M_f_sigma = mesh.get_face_inner_product(sigma_1d)
    
    omega = 2 * np.pi * freq
    k_bot = np.sqrt(-1.0j * omega * mu[0] * sigma[0])
    
    q = np.zeros(mesh.n_nodes, dtype=np.complex128)
    q[-1] = -1j * omega
    
    A = G.T @ M_e_mui @ G + 1j * omega * M_f_sigma
    A[0, 0] += 1j * k_bot/mu[0]

    Ainv = get_default_solver()(A)
    e_1d = Ainv @ q

    # Return solution along original vertical discretization
    if n_pad != 0:
        e_1d = e_1d[n_pad:-n_pad]
    return e_1d
```

This is a bit cleaner code to understand imo, and also allows for potentially including spatially variable $\mu$ if you wanted.

jcapriot:
Went through my own derivation to double check the boundary conditions, see my notes below, but I do get to the same result as your code, but I'm just not convinced it is a better boundary condition than just using the analytic as dirichlet conditions, here are the real and imaginary parts for a simple layered model:

image

Derivation
-----------
First the PDE you are solving, with boundary conditions clearly stated.

$$
\begin{align}
\frac{\partial e}{\partial z} + i\omega b = 0\\
\frac{\partial h}{\partial z} + \sigma e = 0
\end{align}
$$

s.t.

$$
\begin{align}
\frac{\partial e}{\partial z} = i \omega \mu \hspace{10pt}\textrm{for } z = z_{top}\\
\frac{\partial e}{\partial z}  - i k e = 0 \hspace{10pt}\textrm{for } z = z_{bot}
\end{align}
$$

and in a weak form,

$$
\begin{align}
< u, \partial_z e > + i \omega < u, b > = 0\\
< v, \partial_z \mu^{-1} b > + < v, \sigma e > = 0
\end{align}
$$

Discretizing this with an E-B formulation with $e$ on faces (nodes in 1D) and $b$ on edges (cell centers in 1D),

$$
\begin{align}
u^T M_{e} G e + i \omega u^T M_{e} b = 0\\
-v^T G^T M_{e, \mu^{-1}} b + (v \mu^{-1} b)|^{top - bot} + v^T M_{f, \sigma} e = 0
\end{align}
$$

Using the two boundary conditions to solve for $b_{top}$ and $b_{bot}$ from Faraday's Law:

$$
\begin{align}
i\omega b_{top} = -\partial_z e |^{top} = i\omega \mu\\
b_{top} = \mu
\end{align}
$$

$$
\begin{align}
i \omega b_{bot} = -\partial_z e |^{bot} = -ike\\
b_{bot} = -\frac{k}{\omega} e
\end{align}
$$

Which gives

$$
-v^T G^T M_{e,\mu^{-1}} b + v_{top} + v_{bot}\frac{k}{\mu\omega} e + v^T M_{f, \sigma} e = 0
$$

multiply by $i\omega$

$$
v^T G^T M_{e,\mu^{-1}} (-i \omega b) + i \omega (v_{top} + v_{bot}\frac{k}{\mu\omega} e) + i\omega v^T M_{f, \sigma} e = 0
$$

Substitute in discretized Faraday's Law,

$$
v^T G^T M_{e,\mu^{-1}} G e + i \omega (v_{top} + v_{bot}\frac{k}{\mu\omega} e) + i\omega v^T M_{f, \sigma} e = 0
$$

Re-arrange

$$
v^T G^T M_{e,\mu^{-1}} G e + i(v \frac{k}{\mu} e)|^{bot} + i\omega v^T M_{f, \sigma} e = -i \omega (v_{top})
$$

So then system of equations is then:

$$
(G^T M_{e,\mu^{-1}} G  +i\omega M_{f, \sigma}   + i B_{c, bot} \frac{k_{bot}}{\mu_{bot}} ) e = -i \omega b_{c, top}
$$

Where $B_{c, bot}$ is a matrix of zeros, with a single 1 along the diagonal at the bottom node index, and $b_{c, top}$ is a vector of 0's with a single 1 at the top node index.

dccowan:
> I think I see more now looking at it. Currently the Primary field solution uses a Dirichlet Boundary condition on the top and bottom nodes with values taken from the analytic solution.
> 
> e s o l ( z b o t t o m ) = e a n a ( z b o t t o m ) e s o l ( z t o p ) = e a n a ( z t o p )
> 
> It would appear that in this proposal that you're using a Neumann boundary on e at the top,
> 
> h t o p = 1 → ∂ e ∂ z | t o p = − i ω μ 0 ,
> 
> and at the bottom you're using a Robin boundary condition (or at least something close to one),
> 
> ∂ e ∂ z − i k e b o t = 0
> 
> If so, then it is indeed different, and we can likely simplify up this implementation a decent amount using some of `discretize`'s pre-built operators and be a little more rigorous with the derivation. It would be good to explicitly state these boundary conditions more clearly in the docstring though.

Ya, basically. I wanted to use the prebuilt operators in discretize for boundary conditions but couldn't quite figure it out.

jcapriot:
> > Went through my own derivation to double check the boundary conditions, see my notes below, but I do get to the same result as your code, but I'm just not convinced it is a better boundary condition than just using the analytic as dirichlet conditions, here are the real and imaginary parts for a simple layered model:
> 
> My derivation is almost the identical. However, I'm pretty sure you have a sign missing on the boundary conditions at the top. There are two negatives that cancel out.

Ah yep, missed that on the copy over.

jcapriot:
The current formulation sets Dirichlet conditions on E at the top and bottom boundary. Their values come from the analytic solution of a layered earth, normalized to 1 at the top.

dccowan:
I think everything necessary has been addressed prior to merging. Does anyone have any final issues with this PR as is? @jcapriot ?

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -1,27 +1,37 @@
+# noqa: D100
```
santisoler: Just adding some docstring in public module.
```suggestion
"""
Utility functions for NSEM sources.
"""
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -135,51 +152,403 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
+#     # Assign the source to ey_py
+#     for i in np.arange(mesh.vnEy[0]):
+#         for j in np.arange(mesh.vnEy[1]):
+#             ey_py[i, j, :] = e0_1d
+#     # ey_py[1:-1, 1:-1, 1:-1] = 0
+#     eBG_py = np.vstack((ex_py, mkvc(ey_py, 2), ez_py))
 
 #     # Return the electric fields
 #     eBG_bp = np.hstack((eBG_px, eBG_py))
 #     return eBG_bp
```
santisoler: I'm not a fan of leaving commented code. It never gets run, it's not tested. I see you updated this commented function. Any reason to do so?

I'd be inclined to remove these commented lines all together. In case you want to keep a record of this function for a future idea, an issue would be the best place to paste it.

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -135,51 +152,403 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
+def primary_e_1d_solution(
+    mesh, sigma_1d, freq, top_bc="dirichlet", bot_bc="robin", n_pad=500
+):
+    r"""Compute 1D electric field solution on nodes.
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseTensorMesh
+        A 1d, 2d or 3d tensor mesh or tree mesh.
+    sigma_1d :
```
santisoler: I guess the `sigma_1d` should be an array, right?
```suggestion
    sigma_1d : array
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -135,51 +152,403 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
+def primary_h_1d_solution(
+    mesh, sigma_1d, freq, top_bc="dirichlet", bot_bc="dirichlet", n_pad=500
+):
+    r"""Compute 1D magnetic field solution on nodes.
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseTensorMesh
+        A 1d, 2d or 3d tensor mesh or tree mesh.
+    sigma_1d :
```
santisoler:
```suggestion
    sigma_1d : array
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -70,25 +80,32 @@ def homo1DModelSource(mesh, freq, sigma_1d):
...
+        1D conductivity model defined along the vertical discretization.
+        Conductivities are defined from the bottom cell upwards.
 
+    Returns
+    -------
+    numpy.ndarray (n_edges, 2)
+        E fields for the background model at both polarizations with shape.
     """
     from simpeg.NSEM.Utils import getEHfields
```
santisoler, line 103: I'm noticing that this function is not being tested. There's no `simpeg.NSEM` module, so a test running this function should be failing. I wouldn't add a test for it here, but something to keep in mind for the future.

simpeg/electromagnetics/natural_source/utils/test_utils.py
```diff
@@ -571,3 +572,28 @@ def twoLayer(conds):
...
+        -------
+        numpy.ndarray
+            Primary electric field.
+        """
+        if self._ePrimary is None:
+            sigma_1d, _ = self._get_sigmas(simulation)
+            self._ePrimary = homo1DModelSource(
+                simulation.mesh, self.frequency, sigma_1d
+            )
+        return self._ePrimary
```
santisoler: I don't understand the idea behind introducing a new class that it's already marked as deprecated. I understand that this is implementing the same `ePrimary` method that the `PlanewaveXYPrimary` has on `main`. But as far as I can see, this class is only used in a test, and therefore we would be testing code that is not part of the API.

Is there a reason why we should keep the old implementation? Isn't the new implementation better?

tests/em/nsem/forward/test_fields_1d_vs_propagator_pytest.py
```diff
@@ -0,0 +1,95 @@
...
+        elif top_bc == "neumann":
+            u0 = Ed + Eu
+            u0 /= u0[-1]
+            u0 = mesh.average_node_to_cell @ u0
+
+            u1 = mesh.nodal_gradient @ u1
+            u1 /= sig_1d
+
+    assert np.real(u1) == pytest.approx(np.real(u0), abs=ABS_TOLERANCE)
+    assert np.imag(u1) == pytest.approx(np.imag(u0), abs=ABS_TOLERANCE)
```
santisoler: Let's better use `np.testing.assert_allclose` for this:

```suggestion
    np.testing.assert_allclose(np.real(u1), np.real(u0))
```

Nonetheless, I think that assert will fail. I wouldn't recommend using such high absolute tolerance. Basically, all values below 3e-3 will not be compared because they fall below it. Maybe choosing a reasonable `rtol`, plus some `atol` to skip very low values.

tests/em/nsem/forward/test_Simulation3D_vs_Analytic_pytest.py
```diff
@@ -197,6 +205,42 @@ def test_analytic_halfspace_solution(
...
+        ]
+    )
+    analytic_solution = np.repeat(analytic_solution, n_locations)
+
+    # # Error
+    err = np.abs(
+        (numeric_solution - analytic_solution) / (analytic_solution + ABS_TOLERANCE)
+    )
+
+    assert np.all(err < REL_TOLERANCE)
```
santisoler, line 197: Similarly here, it's better to use `np.testing.assert_allclose` for this:

```suggestion
    np.testing.assert_allclose(
        numeric_solution, analytic_solution, rtol=REL_TOLERANCE, atol=ABS_TOLERANCE
    )
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -135,51 +152,403 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
+    impose boundary conditions.
+    """
+    # Extract vertical discretization
+    hz = mesh.h[-1]
+
+    if len(hz) != len(sigma_1d):
+        raise ValueError(
+            "Number of cells in vertical direction must match length of "
+            "'sigma_1d'. Here hz has length {} and sigma_1d has length "
+            "{}".format(len(hz), len(sigma_1d))
```
santisoler: Small style improvement here, to use f-strings instead of `.format()`:
```suggestion
            "Number of cells in vertical direction must match length of "
            f"'sigma_1d'. Here hz has length {len(hz)} and sigma_1d has "
            f"length {len(sigma_1d)}"
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -135,51 +152,403 @@ def analytic1DModelSource(mesh, freq, sigma_1d):
...
```
santisoler:
```suggestion
            "Number of cells in vertical direction must match length of "
            f"'sigma_1d'. Here hz has length {len(hz)} and sigma_1d has "
            f"length {len(sigma_1d)}"
```

tests/em/nsem/forward/test_Simulation3D_vs_Analytic_pytest.py
```diff
@@ -197,6 +205,42 @@ def test_analytic_halfspace_solution(
     assert np.all(err < REL_TOLERANCE)
 
 
+# Source term was generated using the function `homo1DModelSource` to solve the
+# 1D problem. This function has since been deprecated and will be removed in
+# SimPEG v0.27.
```
santisoler: I think this is related to one of the previous comments I made. I don't fully understand which function will be removed. Is it `homo1DModelSource`?

tests/em/nsem/forward/test_fields_1d_vs_propagator_pytest.py
```diff
@@ -0,0 +1,95 @@
+# noqa: D100
```
santisoler:
```suggestion
"""
Test NSEM solutions against propagator solution.
"""
```

simpeg/electromagnetics/natural_source/utils/test_utils.py
dccowan: I agree with you completely. But this was required by a previous reviewer.

tests/em/nsem/forward/test_fields_1d_vs_propagator_pytest.py
dccowan: Done. Made a 5% relative error and 1e-5 absolute error.

tests/em/nsem/forward/test_Simulation3D_vs_Analytic_pytest.py
dccowan: For now, it's best to leave this. It will be part of a clean up project for the module later.

tests/em/nsem/inversion/test_Problem3D_Derivs.py
```diff
@@ -62,7 +62,7 @@ def test_Jmatrix(model_simulation_tuple):
     # compare to JTvec function
     jtvec = simulation.Jtvec(model, v=vec)
 
-    np.testing.assert_allclose(Jmatrix_vec, jtvec)
+    np.testing.assert_allclose(Jmatrix_vec, jtvec, atol=1e-10)
```
santisoler, line 65: I'm curious why we now need to add an `atol` here. Is this new implementation making this test to fail if we set `atol=0`?

jcapriot:
> Thanks @dccowan for addressing my comments. I'm still hesitant to include that `PlanewaveXYPrimaryDeprecated` deprecated class into `main`. It won't be part of the public API (and even if it is, we are telling users not to use it because we'll remove it), and it's only being used to run a test that compares its numerical solutions against the analytical one. It's not doing any harm, but we would be adding code that doesn't have a purpose.
> 
> @jcapriot, do you have any thoughts on this? Is there any reason to add it?

There's no reason to add the Deprecated class into the main namespace. You're correct it's only used for a test, and to verify the behavior is somewhat consistent with the previous implementation. In that case, just move the class into the test file where it is being used.

tests/em/nsem/inversion/test_Problem3D_Derivs.py
dccowan: We will create a separate issue for this.

dccowan:
We have moved away from using 'homo1DModelSource' and 'analytic1DModelSource' within source_utils.py to generate RHS, opting to use 'primary_e_1d_solution' instead. Consequently, lines in 'homo1DModelSource' and 'analytic1DModelSource' are no longer being tested. I believe this was the motivation for making a test for 'PlanewaveXYPrimaryDeprecated'. Do we want to bring that test back in? Or are we okay with the current level of coverage?

santisoler:
I confirm that `homo1DModelSource` and `analytic1DModelSource` are not being used anywhere else in the code. I would therefore suggest to remove them. But don't worry about that in this PR. I'll open a separate one so it's clean in the git log where that happened (in case someone wants to retrieve them in the future).

I just pushed a few extra tests to increase coverage in some of the source utils, mostly related to errors that weren't being covered.

I'll merge this after CI passes, and open a separate PR to remove those unused functions.

Thanks @dccowan for the work!

PR #1668, opened 2025-05-16
Add key navigation to docs

prisae:
#### Summary

This PR enables moving forward/backward in the docs with the right/left arrow keys on the keyboard. I quite enjoy this feature in documentations that have enabled it.

If you want to see how it works, test it over at, e.g., https://empymod.emsig.xyz.

#### PR Checklist
* Just a docs-configuration change, the PR checklist does mostly not apply.

@simpeg/simpeg-developers - ready for review.


prisae:
Oh. I realized that the default was `True`, but it changed recently:
https://github.com/pydata/pydata-sphinx-theme/issues/1100

So it apparently is against the "gold standard" of keyboard navigation. Too bad, I love it.

Anyway, I leave it for a short while open, curious to see what you think.

santisoler:
😮  I didn't know about this feature. Let's merge it!

PR #1670, opened 2025-05-19
Use logging while setting default solver in PDE simulations

### LINKED ISSUE #1660: BUG: DefaultSolver warning message
### Describe the issue:

Calling a simulation without setting the solver raises a warning, stating
```
simpeg/base/pde_simulation.py#line=489: DefaultSolverWarning: Using the default solver: Mumps. 

If you would like to suppress this notification, add 
warnings.filterwarnings('ignore', simpeg.utils.solver_utils.DefaultSolverWarning)
 to your script.
  return get_default_solver(warn=True)
```

This contains a bug (for sure) and a misleading message (just in my opinion).

Bug: Setting
```
warnings.filterwarnings('ignore', simpeg.utils.solver_utils.DefaultSolverWarning)
```
results in a Type error message, `TypeError: message must be a string`. It should state:
```
warnings.filterwarnings('ignore', category=simpeg.utils.solver_utils.DefaultSolverWarning)
```

However, I personally think that it is not good to make users suppress warnings. Rather, there should be an advice to switch it off intrinsically. Either by setting the default solver, or make a SimPEG possibility to switch the warning off.

```python
# Dummy code, I am afraid I have no simple code to reproduce - if you can't reproduce it then I'll create one.
sim_te = nsem.simulation.Simulation2DMagneticField(
    mesh,
    survey=survey_te,
    sigma=sigma,
)
```

### Error message:

```shell
# Given in the description already.
```

### Runtime information:

```shell
--------------------------------------------------------------------------------
  Date: Wed May 07 18:38:45 2025 CEST

                OS : Darwin (macOS 15.4.1)
            CPU(s) : 14
           Machine : arm64
      Architecture : 64bit
               RAM : 48.0 GiB
       Environment : Jupyter
       File system : apfs

  Python 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:30)
  [Clang 18.1.8 ]

            simpeg : 0.24.0
        discretize : 0.11.2
       pymatsolver : 0.3.1
             numpy : 2.2.5
             scipy : 1.15.2
        matplotlib : 3.10.1
            geoana : 0.7.2
            libdlf : 0.3.0
             numba : 0.61.2
            pandas : 2.2.3
           IPython : 9.2.0
        ipywidgets : 8.1.7
--------------------------------------------------------------------------------
```

### Context for the issue:

To the issue I came through this site: https://mtgeophysics.github.io/

In all those examples of SimPEG, it is set in the beginngin:
```
import warnings
warnings.filterwarnings("ignore")
```

I removed this immediately, as I find it dangerous (missing potential important warnings), and then the above messages appeared.

santisoler:
#### Summary

Ditch the `DefaultSolverWarning` in `get_default_solver` and make the base PDE simulation class to handle the information message to users about using a default solver. The information message is sent through a new custom logger defined using `logging`. The information is sent from the constructor after obtaining the solver through `get_default_solver`. The solver class gets cached on instantiation, so the simulation will use the same solver even if the default solver changes later on. Warn users regarding poor performance when setting `SolverLU` or `Solver` as the solver. Adjust and extend tests. Also, add a `simpeg.utils.get_logger` function that returns the logger object used across the project, and a new `PerformanceWarning` class in a new `simpeg.utils.warnings` submodule.

#### PR Checklist
      to a Pull Request

#### Reference issue

Fixes #1660


simpeg/utils/logger.py
```diff
@@ -0,0 +1,23 @@
...
+import logging
+
+__all__ = ["LOGGER"]
+
+
+def _create_logger():
+    """
+    Create logger for SimPEG.
+    """
+    logger = logging.getLogger(__name__)
```
jcapriot, line 14: This will make it look like every logged line comes from this file. It would make more sense to instantiate a logger for the simulation in my opinion.

jcapriot:
I’m still very much in favor of it being a warning. The amount of times we’ve had people run simulations, asking why the runtime is so different, without knowing what solver they used has drastically decreased.

simpeg/utils/logger.py
santisoler: Make sense. What about a global logger for SimPEG?

santisoler:
> I’m still very much in favor of it being a warning. The amount of times we’ve had people run simulations, asking why the runtime is so different, without knowing what solver they used has drastically decreased.

Right. The idea is not to make that information to disappear, but to avoid warning users, and inform them instead. I think @prisae had a good point by saying that not choosing a solver is the default, so should we warn users if they are using the default choice? An information sounds better suited for that case.

Last week we were also discussing two relevant points:
1. Moving the warning from the `solver` setter to the constructor. So, users get the information when the simulation is instantiated and not when the solver gets accessed.
2. Making it simpler to silent that information/warning. Instructing users to filterwarnings can be quite dangerous. A `verbose` flag (like the one in this PR) seemed a simpler choice.

prisae:
I like the approach. And the `StreamHandler` will make sure it is also printed to screen, not just to a potential logging file, so the user does get informed ('warned'). Starting with the logging module is probably good anyway, I do not use it enough myself...

santisoler:
I've just applied a few changes to this PR, inline with the conversation we had yesterday. In a nutshell:

* We stream an INFO message through the logger instead of a warning every time the PDE simulation uses the `get_default_solver` to set a solver.
* The solver class gets cached, so the simulation will use the solver set in the constructor, even if later on the default solver changes. This makes the process _less magical_.
* I removed the `verbose` flag. We agreed that it's better to instruct users to pass their own solver if the don't want to see the message again.
* If the `SolverLU` is chosen, users will get a warning about its poor performance, recommending them to change to Pardiso or MUMPS.
* I added logging capabilities by exposing a `get_logger` function in `simpeg.utils`. We can think about using further in the future.
* I also added a `simpeg.utils.warnings` submodule where we can add custom warnings.

simpeg/base/pde_simulation.py
```diff
@@ -433,12 +434,24 @@ class BasePDESimulation(BaseSimulation):
...
         super().__init__(**kwargs)
+        if solver is None:
+            solver = get_default_solver()
+            get_logger().info(
+                f"Setting the default solver '{solver.__name__}' for the "
+                f"'{type(self).__name__}'.\n"
+                "To avoid receiving this message, pass a solver to the simulation. "
+                "For example:"
+                "\n\n"
+                "  from simpeg.utils.solver_utils import get_default_solver\n"
```
santisoler:
```suggestion
                "  from simpeg.utils import get_default_solver\n"
```

prisae:
Great, thanks @santisoler!

PR #1679, opened 2025-06-06
Fix bug in phase for recursive 1d NSEM simulation

dccowan:
The recursive solution exclusively computes the impedance, apparent resistivity and/or phase of the Zxy impedance at all frequencies. In the convention adopted by SimPEG, Zxy is in the lower-left quadrant of the complex plane. Using np.atan(imag/real) results in phase values between -90 and +90 which is wrong.

Suggest we use ``np.angle`` instead.

image

#### PR Checklist
      to a Pull Request


lheagy (approved):
Thanks for catching this @dccowan! If you are interested in adding Zyx as well, I suggest we do that in a separate PR, but go for it if you like!

PR #1682, opened 2025-06-10
Differential Magnetic Code Updates

johnweis0480:
#### Summary
Updates differential magnetics code to model in terms of susceptibility and remanent magnetization
Adds effective susceptibility mapping for consistency with integral formulation

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/maps/_property_maps.py
+++ b/simpeg/maps/_property_maps.py
@@ -554,6 +554,51 @@ class ChiMap(IdentityMap):
         return m / mu_0 - 1
 
 
+class Effective_Susceptibility_Map(IdentityMap):
+    r"""Effective susceptibility Map
+
+    Convert Effective Susceptibility (SI) to magnetic polarization (nT).
+
+
+    .. math::
+
+        \mu_0 M = esus * ||B_0||
+
+    Parameters
+    ----------
+    mesh : discretize.BaseMesh
+        The number of parameters accepted by the mapping is set to equal the number
+        of mesh cells.
+    nP : int
+        Set the number of parameters accepted by the mapping directly. Used if the
+        number of parameters is known. Used generally when the number of parameters
+        is not equal to the number of cells in a mesh.
+    ambient_field_magnitude : float
+        The magnitude of the ambient geomagnetic field.
+    """
+
+    def __init__(self, mesh=None, nP=None, ambient_field_magnitude=None, **kwargs):
+        super().__init__(mesh=mesh, nP=nP, **kwargs)
+        if ambient_field_magnitude is None or not isinstance(
+            ambient_field_magnitude, (float, int)
+        ):
+            raise TypeError(
+                "ambient_field_magnitude must be a float (or int convertible to float) and cannot be None"
+            )
+        self.ambient_field_magnitude = ambient_field_magnitude
+
+    def _transform(self, m):
+        return m * self.ambient_field_magnitude
+
+    def deriv(self, m, v=None):
+        if v is not None:
+            return self.ambient_field_magnitude * v
+        return self.ambient_field_magnitude * sp.eye(self.nP)
+
+    def inverse(self, m):
+        return m / self.ambient_field_magnitude
+
+
 class MuRelative(IdentityMap):
     r"""Mapping that computes the magnetic permeability given a set of relative permeabilities.
 
--- a/simpeg/potential_fields/magnetics/simulation.py
+++ b/simpeg/potential_fields/magnetics/simulation.py
@@ -3,6 +3,7 @@ import warnings
 import numpy as np
 from numpy.typing import NDArray
 import scipy.sparse as sp
+from functools import cached_property
 from geoana.kernels import (
     prism_fxxy,
     prism_fxxz,
@@ -18,11 +19,9 @@ from scipy.sparse.linalg import LinearOperator, aslinearoperator
 from simpeg import props, utils
 from simpeg.utils import mat_utils, mkvc, sdiag
 from simpeg.utils.code_utils import deprecate_property, validate_string, validate_type
-from simpeg.utils.solver_utils import get_default_solver
 
 from ...base import BaseMagneticPDESimulation
 from ..base import BaseEquivalentSourceLayerSimulation, BasePFSimulation
-from .analytics import CongruousMagBC
 from .survey import Survey
 
 from ._numba import choclo, NUMBA_FUNCTIONS_3D, NUMBA_FUNCTIONS_2D
@@ -1665,63 +1664,134 @@ class SimulationEquivalentSourceLayer(
 
 
 class Simulation3DDifferential(BaseMagneticPDESimulation):
+    r"""A secondary field simulation for magnetic data.
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseMesh
+    survey : magnetics.survey.Survey
+    mu : float, array_like
+        Magnetic Permeability Model (H/ m). Set this for forward
+        modeling or to fix while inverting for remanence. This is used if
+        muMap == None
+    rem : float, array_like
+        Magnetic Polarization \mu_0*M (nT). Set this for forward
+        modeling or to fix remanent magnetization while inverting for permeability.
+        This is used if remMap == None
+    muMap : SimPEG.maps.IdentityMap, optional
+        The mapping used to go from the simulation model to `mu`. Set this
+        to invert for `mu`.
+    remMap : SimPEG.maps.IdentityMap, optional
+        The mapping used to go from the simulation model to `mu_0*M`. Set this
+        to invert for `mu_0*M`.
+    storeJ: bool
+        Whether to store the sensitivity matrix. If set to True
+    use_float32_solver: bool
+        Whether to solve Ainv*rhs using float32 precision.
+
+
+    Notes
+    -----
+    This simulation solves for the magnetostatic PDE:
+    \nabla \cdot \Vec{B} = 0
+
+    where the constitutive relation is specified as:
+    \Vec{B} = \mu\Vec{H} + \mu_0\Vec{M_r}
+
+    where \Vec{M_r} is a fixed magnetization unaffected by the inducing field
+    and \mu\Vec{H} is the induced magnetization
     """
-    Secondary field approach using differential equations!
-    """
 
-    def __init__(self, mesh, survey=None, **kwargs):
-        super().__init__(mesh, survey=survey, **kwargs)
+    _Ainv = None
 
-        Pbc, Pin, self._Pout = self.mesh.get_BC_projections(
-            "neumann", discretization="CC"
-        )
+    rem, remMap, remDeriv = props.Invertible(
+        "Magnetic Polarization (nT)", optional=True
+    )
 
-        Dface = self.mesh.face_divergence
-        Mc = sdiag(self.mesh.cell_volumes)
-        self._Div = Mc * Dface * Pin.T.tocsr() * Pin
+    def __init__(
+        self,
+        mesh,
+        survey=None,
+        mu=None,
+        rem=None,
+        remMap=None,
+        muMap=None,
+        storeJ=False,
+        use_float32_solver=False,
+        **kwargs,
+    ):
+        if mu is None:
+            mu = mu_0
+
+        super().__init__(mesh=mesh, survey=survey, mu=mu, muMap=muMap, **kwargs)
+
+        self.rem = rem
+        self.remMap = remMap
+
+        self.storeJ = storeJ
+        self.use_float32_solver = use_float32_solver
+
+        self._MfMu0i = self.mesh.get_face_inner_product(1.0 / mu_0)
+        self._Div = self.Mcc * self.mesh.face_divergence
+        self._DivT = self._Div.T.tocsr()
+        self._Mf_vec_deriv = self.mesh.get_face_inner_product_deriv(
+            np.ones(self.mesh.n_cells * 3)
+        )(np.ones(self.mesh.n_faces))
+
+        self.solver_opts = {"is_symmetric": True, "is_positive_definite": True}
+
+        self._Jmatrix = None
+        self._stored_fields = None
 
     @property
     def survey(self):
-        """The survey for this simulation.
+        """The magnetic survey object.
 
         Returns
         -------
-        simpeg.potential_fields.magnetics.survey.Survey
+        SimPEG.potential_fields.magnetics.Survey
         """
         if self._survey is None:
             raise AttributeError("Simulation must have a survey")
         return self._survey
 
     @survey.setter
-    def survey(self, obj):
-        if obj is not None:
-            obj = validate_type("survey", obj, Survey, cast=False)
-            self._validate_survey(obj)
-        self._survey = obj
+    def survey(self, value):
+        if value is not None:
+            value = validate_type("survey", value, Survey, cast=False)
+        self._survey = value
 
     @property
-    def MfMuI(self):
-        return self._MfMuI
+    def storeJ(self):
+        """Whether to store the sensitivity matrix
+
+        Returns
+        -------
+        bool
+        """
+        return self._storeJ
+
+    @storeJ.setter
+    def storeJ(self, value):
+        self._storeJ = validate_type("storeJ", value, bool)
 
     @property
-    def MfMui(self):
-        return self._MfMui
+    def use_float32_solver(self):
+        """Whether to solve Ainv*rhs using float32 precision.
 
-    @property
-    def MfMu0(self):
-        return self._MfMu0
+        Returns
+        -------
+        bool
+        """
+        return self._use_float32_solver
 
-    def makeMassMatrices(self, m):
-        mu = self.muMap * m
-        self._MfMui = self.mesh.get_face_inner_product(1.0 / mu) / self.mesh.dim
-        # self._MfMui = self.mesh.get_face_inner_product(1./mu)
-        # TODO: this will break if tensor mu
-        self._MfMuI = sdiag(1.0 / self._MfMui.diagonal())
-        self._MfMu0 = self.mesh.get_face_inner_product(1.0 / mu_0) / self.mesh.dim
-        # self._MfMu0 = self.mesh.get_face_inner_product(1/mu_0)
+    @use_float32_solver.setter
+    def use_float32_solver(self, value):
+        self._use_float32_solver = validate_type("use_float32_solver", value, bool)
 
+    @cached_property
     @utils.requires("survey")
-    def getB0(self):
+    def _B0(self):
         b0 = self.survey.source_field.b0
         B0 = np.r_[
             b0[0] * np.ones(self.mesh.nFx),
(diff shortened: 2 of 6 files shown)


simpeg/maps/_property_maps.py
```diff
@@ -554,6 +554,51 @@ def inverse(self, m):
         return m / mu_0 - 1
 
 
+class Effective_Susceptibility_Map(IdentityMap):
```
domfournier:
```suggestion
class EffectiveSusceptibilityMap(IdentityMap):
```
Class names should be CamelCase

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,63 +1664,134 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    )
+
+    def __init__(
+        self,
+        mesh,
+        survey=None,
+        mu=None,
+        rem=None,
+        remMap=None,
+        muMap=None,
```
domfournier:
```suggestion
        mu=None,
        muMap=None,
        rem=None,
        remMap=None,
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
-                \frac{\partial \mathbf{\mu}}{\partial \mathbf{m}}
-                \left[
-                    \Div \diag(\M^f_{\mu_{0}^{-1}}\mathbf{B}_0) \dMfMuI
-                \right]
-                - \diag(\mathbf{v}) \mathbf{D} \mathbf{P}_{out}^T
-                    \frac{\partial B_{sBC}}{\partial \mathbf{m}}
+        return dpred
 
-        In the end,
+    def Magnetic_Polarization(self, m):
```
domfournier:
```suggestion
    def magnetic_polarization(self, m):
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
-                 - \diag (\Div^T\mathbf{u})\dMfMuI
-            \right ]
+        self._Ainv = self.solver(self.getA(m), **self.solver_opts)
 
-             -  (\MfMui)^{-1}\Div^T\frac{\delta\mathbf{u}}{\delta \mathbf{m}}
+        rhs = self._getRHS(m)
 
-        Finally we evaluate the above, but we should remember that
+        u = self._Ainv * rhs
+        B = -self.MfMuiI * self._DivT * u
```
domfournier:
```suggestion
        b_field = -self.MfMuiI * self._DivT * u
```
I know that we don't really enforce snake_case for variables in simpeg, but this is a 1D vector, so at least lower case that one.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
-        .. math ::
+    def getJ(self, m, f=None):
+        self.model = m
+        if self._Jmatrix is None:
+            if f is None:
+                f = self.fields(m)
+            if m.size < self.survey.nD:
+                J = self._Jvec(m, v=None, f=f)
+            else:
+                J = self._Jtvec(m, v=None, f=f).T
```
domfournier:
```suggestion
        if self._Jmatrix:
            return self._Jmatrix
        if f is None:
            f = self.fields(m)
        if m.size < self.survey.nD:
            J = self._Jvec(m, v=None, f=f)
        else:
            J = self._Jtvec(m, v=None, f=f).T
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
 
-            \mathbf{P}_{derv} = \frac{\partial \mathbf{P}}{\partial\mathbf{B}}
+        else:
+            J = self._Jmatrix
```
domfournier:
```suggestion
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
+            DivTatsol_p_QT = (
+                self._DivT * (self._Ainv * ((-self._Div * (self.MfMuiI.T * (Q.T * v)))))
+                + Q.T * v
+            )
 
-        B, u = u["B"], u["u"]
-        mu = self.mapping * (m)
-        dmu_dm = self.mapping.deriv(m)
-        # dchidmu = sdiag(1 / mu_0 * np.ones(self.mesh.nC))
+        del Q
```
domfournier: Is this required? Q looks sparse.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
-                \mathbf{J}^{T}\mathbf{v} =
-                (\frac{\delta \mathbf{P}\mathbf{B}} {\delta \mathbf{m}})^{T} \mathbf{v}
+        Q = self._projectFieldsDeriv(B)
 
-        """
-        if u is None:
-            u = self.fields(m)
+        if v is None:
+            v = np.eye(Q.shape[0])
+            DivTatsol_p_QT = (
```
domfournier:
```suggestion
            div_solve_q = (
```
Just a suggestion. Those names are getting a bit nuts IMO

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1730,497 +1800,387 @@ def getB0(self):
...
-            msg = "Found invalid survey without receivers."
-            raise ValueError(msg)
-        components = receivers[0].components
-        if not all(components == rx.components for rx in receivers):
-            msg = (
-                "Found invalid survey with receivers that have mixed components. "
-                f"Surveys for the {type(self).__name__} class must contain receivers "
-                "with the same components."
+            dpred = (
+                np.sqrt((bx + B0[0]) ** 2 + (by + B0[1]) ** 2 + (bz + B0[2]) ** 2) - Bot
```
domfournier: Code duplication with _projectFields. Suggest creating a private method to extract components giving fields.
Something along those lines
```
def _get_components(self, fields) -> dict[str, np.ndarray]:
    """
    Extract components from magnetic fields.
    """
    B0 = self.survey.source_field.b0
    comps = {
    "bx": self._Qfx * fields,
    "by": self._Qfy * fields,
    "bz": self._Qfz * fields,
    "b_total": np.linalg.norm(B0)
     }
     
     comps["tmi"] =  np.sqrt(
                (comps["bx"] + B0[0]) ** 2 + (comps["by"] + B0[1]) ** 2 + (comps["bz"] + B0[2]) ** 2
            ) - comps["b_total"]
     return comps
```
so you can just call it in both methods

tests/pf/test_forward_Mag_differential.py
domfournier, line 1: Snake case the name

tests/pf/test_forward_Mag_differential.py
```diff
@@ -0,0 +1,823 @@
+import pytest
+import discretize
+import simpeg.potential_fields as PF
+from simpeg import utils, maps
+from discretize.utils import mkvc, refine_tree_xyz
+import numpy as np
+
+
+class ProlateEllispse:
```
domfournier: Seems out of place. Should go to `geoana` or at least `utils`

domfournier (commented):
Did a quick first pass not really looking at the math. Main points are:
- Class and function/method names should be `CamelCase` and `snake_case` respectively.
- Variable names are getting a bit crazy. I get that simpeg has those everywhere, but we could simplify those a bit to improve readability.
- The `ProlateEllipse` class looks general and probably belongs somewhere else, not in tests.

I can start testing though.

simpeg/potential_fields/magnetics/simulation.py
johnweis0480: Tried to simplify the names throughout a bit

simpeg/potential_fields/magnetics/simulation.py
johnweis0480: On this one doing Q * fields is not required unless tmi data is used for the deriv, so I have left it separately for now

simpeg/maps/_property_maps.py
```diff
@@ -554,6 +554,51 @@ def inverse(self, m):
...
 
+class EffectiveSusceptibilityMap(IdentityMap):
+    r"""Effective susceptibility Map
+
+    Convert Effective Susceptibility (SI) to magnetic polarization (nT).
+
+
+    .. math::
+
+        \mu_0 M = esus * ||B_0||
```
santisoler: I would improve this math description a bit:


```suggestion
    This map converts effective susceptibility values (:math:`\chi_\text{eff}`) into magnetic
    polarization (:math:`\mathbf{I}`):
    
    .. math::

        \mathbf{I} = \mu_0 \mathbf{M} = \chi_\text{eff} \lVert \mathbf{B}_0 \rVert
    
    where :math:`\mathbf{M}` is the magnetization vector, and 
    :math:`\lVert \mathbf{B}_0 \rVert` is the magnitude of the ambient field.
```

Also, it would be nice to move this mathematical description to a new `Notes` section below `Parameters`. Like:

```
    Notes
    -----
    This map converts effective [...]
```

simpeg/maps/_property_maps.py
```diff
@@ -554,6 +554,51 @@ def inverse(self, m):
...
+    ----------
+    mesh : discretize.BaseMesh
+        The number of parameters accepted by the mapping is set to equal the number
+        of mesh cells.
+    nP : int
+        Set the number of parameters accepted by the mapping directly. Used if the
+        number of parameters is known. Used generally when the number of parameters
+        is not equal to the number of cells in a mesh.
+    ambient_field_magnitude : float
+        The magnitude of the ambient geomagnetic field.
```
santisoler: The description of the map states that the output is in nT. This only happens if the ambient field magnitude is in nT too, so it's better to clarify it in the description of `ambient_field_magnitude`:

```suggestion
    ambient_field_magnitude : float
        The magnitude of the ambient geomagnetic field in nT.
```

simpeg/maps/_property_maps.py
```diff
@@ -554,6 +554,51 @@ def inverse(self, m):
...
+    """
+
+    def __init__(self, mesh=None, nP=None, ambient_field_magnitude=None, **kwargs):
+        super().__init__(mesh=mesh, nP=nP, **kwargs)
+        if ambient_field_magnitude is None or not isinstance(
+            ambient_field_magnitude, (float, int)
+        ):
```
santisoler: It's better to check if a given float is a `Real` number (`from numbers import Real`) rather than checking if it's an instance of `float` or `int`.

```suggestion
        if ambient_field_magnitude is None or not isinstance(
            ambient_field_magnitude, Real
        ):
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseMesh
+    survey : magnetics.survey.Survey
+    mu : float, array_like
+        Magnetic Permeability Model (H/ m). Set this for forward
+        modeling or to fix while inverting for remanence. This is used if
+        muMap == None
+    muMap : SimPEG.maps.IdentityMap, optional
```
santisoler: Minor thing here: use `simpeg` in lowercase.  While doing so, we allow Sphinx to generate links to those objects.
```suggestion
    muMap : simpeg.maps.IdentityMap, optional
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        The mapping used to go from the simulation model to `mu`. Set this
+        to invert for `mu`.
+    rem : float, array_like
+        Magnetic Polarization \mu_0*M (nT). Set this for forward
+        modeling or to fix remanent magnetization while inverting for permeability.
+        This is used if remMap == None
+    remMap : SimPEG.maps.IdentityMap, optional
```
santisoler:
```suggestion
    remMap : simpeg.maps.IdentityMap, optional
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    ----------
+    mesh : discretize.base.BaseMesh
+    survey : magnetics.survey.Survey
+    mu : float, array_like
+        Magnetic Permeability Model (H/ m). Set this for forward
+        modeling or to fix while inverting for remanence. This is used if
+        muMap == None
+    muMap : SimPEG.maps.IdentityMap, optional
+        The mapping used to go from the simulation model to `mu`. Set this
+        to invert for `mu`.
```
santisoler: Minor format fix. In RST, two backticks make it to render as code.
```suggestion
        The mapping used to go from the simulation model to ``mu``. Set this
        to invert for ``mu``.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    rem : float, array_like
+        Magnetic Polarization \mu_0*M (nT). Set this for forward
```
santisoler: Add some LaTeX math here:
```suggestion
        Magnetic Polarization :math:`\mu_0 \mathbf{M}` (nT). Set this for forward
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        modeling or to fix remanent magnetization while inverting for permeability.
+        This is used if remMap == None
```
santisoler: Minor correction here to make it more verbose, and Pythonic.
```suggestion
        This is used if ``remMap`` is None.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    remMap : SimPEG.maps.IdentityMap, optional
+        The mapping used to go from the simulation model to `mu_0*M`. Set this
+        to invert for `mu_0*M`.
```
santisoler:
```suggestion
        The mapping used to go from the simulation model to :math:`\mu_0 \mathbf{M}`. 
        Set this to invert for :math:`\mu_0 \mathbf{M}`.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    storeJ: bool
+        Whether to store the sensitivity matrix. If set to True
+    use_float32_solver: bool
+        Whether to solve Ainv*rhs using float32 precision.
```
santisoler:
```suggestion
        Whether to solve ``Ainv @ rhs`` using float32 precision.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    Notes
+    -----
+    This simulation solves for the magnetostatic PDE:
+    \nabla \cdot \Vec{B} = 0
+
+    where the constitutive relation is specified as:
+    \Vec{B} = \mu\Vec{H} + \mu_0\Vec{M_r}
+
+    where \Vec{M_r} is a fixed magnetization unaffected by the inducing field
+    and \mu\Vec{H} is the induced magnetization
```
santisoler: Just improving the math here:
```suggestion
    This simulation solves for the magnetostatic PDE:
    
    .. math::
        \nabla \cdot \Vec{B} = 0

    where the constitutive relation is specified as:

    .. math::
        \Vec{B} = \mu\Vec{H} + \mu_0\Vec{M_r}

    where :math:`\Vec{M_r}` is a fixed magnetization unaffected by the inducing field
    and :math:`\mu\Vec{H}` is the induced magnetization.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    def __init__(
+        self,
+        mesh,
+        survey=None,
+        mu=None,
+        muMap=None,
+        rem=None,
+        remMap=None,
+        storeJ=False,
+        use_float32_solver=False,
```
santisoler: I'm wondering if it would be better to have a `solver_dtype` argument (feel free to rename it to something better) for this option that defaults to `np.float64`. If users want to use a float32 they can pass `np.float32` instead. Such argument is quite standard in Numpy's API to set dtypes.

For example:
```
   solver_dtype=np.float64,
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    r"""A secondary field simulation for magnetic data.
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseMesh
+    survey : magnetics.survey.Survey
+    mu : float, array_like
+        Magnetic Permeability Model (H/ m). Set this for forward
+        modeling or to fix while inverting for remanence. This is used if
+        muMap == None
```
santisoler:
```suggestion
        ``muMap`` is None.
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        if f is None:
+            f = self.fields(m)
+        if m.size < self.survey.nD:
+            J = self._Jvec(m, v=None, f=f)
+        else:
+            J = self._Jtvec(m, v=None, f=f).T
 
-                \mathbf{J}^{T}\mathbf{v} =
-                (\frac{\delta \mathbf{P}\mathbf{B}} {\delta \mathbf{m}})^{T} \mathbf{v}
+        if self.storeJ is True:
```
santisoler: No need to compare with `True`:
```suggestion
        if self.storeJ:
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        Qfz = self.mesh.get_interpolation_matrix(self.survey.receiver_locations, "Fz")
+        return Qfz
 
-            \text{TMI} = \vec{B}_s \cdot \hat{B}_0
+    def _projectFields(self, f):
 
-        """
-        # Get components for all receivers, assuming they all have the same components
-        components = self._get_components()
+        components = self.survey.components
```
santisoler: The `Survey.components` property has been deprecated. You might be getting a `FutureWarning` when running this line. In #1633 I added two private methods that work together to get the components of the receivers, checking first that all receivers have the same components: https://github.com/simpeg/simpeg/pull/1633/files#diff-8c81388b629bf1d754d878413a37d04c2a8f248f221153c34aa1625feb3f2707

I think it would be better to restore those methods and use them instead of `self.survey.components`.

Let me know if you have any questions regarding them.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
-        fields = {}
-        if "bx" in components or "tmi" in components:
-            fields["bx"] = self.Qfx
-        if "by" in components or "tmi" in components:
-            fields["by"] = self.Qfy
-        if "bz" in components or "tmi" in components:
-            fields["bz"] = self.Qfz
+    @utils.count
+    def _projectFieldsDeriv(self, bs):
+        components = self.survey.components
```
santisoler: Same here. Check previous comment on `Survey.components` being deprecated.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
 
     @property
     def survey(self):
-        """The survey for this simulation.
+        """The magnetic survey object.
 
         Returns
         -------
-        simpeg.potential_fields.magnetics.survey.Survey
+        SimPEG.potential_fields.magnetics.Survey
```
santisoler:
```suggestion
        simpeg.potential_fields.magnetics.Survey
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+    def _b0(self):
         b0 = self.survey.source_field.b0
-        B0 = np.r_[
+        b0 = np.r_[
             b0[0] * np.ones(self.mesh.nFx),
             b0[1] * np.ones(self.mesh.nFy),
             b0[2] * np.ones(self.mesh.nFz),
         ]
-        return B0
+        return b0
```
santisoler: Is there a specific reason why we need to allocate these three arrays full of constants? These arrays can be moderately large, specially on large meshes. It's not likely that they will fill the entire memory, but I think there's some room for optimization here.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
-        The A matrix has the form:
+        return rhs
 
-        .. math ::
+    def _getA(self):
+        A = self._Div * self.MfMuiI * self._DivT
 
-            \mathbf{A} =  \Div(\MfMui)^{-1}\Div^{T}
+        if self.use_float32_solver:
+            A = A.astype(np.float32)
```
santisoler: If we decide to move to a `solver_dtype` argument (instead of the `use_float32_solver` flag), we just need to replace these two lines by (no need for an if statement):

```python
A = A.astype(self.solver_dtype)
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        m : (n_param,) numpy.ndarray
+            The model parameters.
 
-             -  (\MfMui)^{-1}\Div^T\frac{\delta\mathbf{u}}{\delta \mathbf{m}}
+        Returns
+        -------
+        mu0_m : np.ndarray
+            The magnetic polarization μ₀ * M in nanoteslas (nT), defined on the mesh faces.
+            The result is ordered as a concatenation of the x, y, and z face components
+            (i.e., [Mx_faces, My_faces, Mz_faces]).
```
santisoler: Minor format improvements here:
```suggestion
            The magnetic polarization :math:`\mu_0 \mathbf{M}` in nanoteslas (nT), defined on the mesh faces.
            The result is ordered as a concatenation of the x, y, and z face components
            (i.e., ``[Mx_faces, My_faces, Mz_faces]``).
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
 
-        Finally we evaluate the above, but we should remember that
 
-        .. note ::
+        """
+        self.model = m
 
-            We only want to evaluate
+        self._Ainv = self.solver(self.getA(m), **self.solver_opts)
```
santisoler: The `self.getA` method is not defined. I guess we want to use `self._getA`, right?

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
-                - \frac{\partial \mathbf{rhs}(\mathbf{m})}{\partial \mathbf{m}}
-            \right]
+        dpred = self._projectFields(f)
 
-        A little tricky point here is we are not interested in potential (u), but interested in magnetic flux (B).
-        Thus, we need sensitivity for B. Now we take derivative of B w.r.t m and have
+        return dpred
 
-        .. math ::
+    def magnetic_polarization(self, m):
```
santisoler: This method is currently not being tested. I think we should add some test for it.

Moreover, I noticed that it implements a computation of the B fields, in a similar way as the `fields` method does. Would it be possible to reuse the `fields` method here?

tests/pf/test_forward_mag_differential.py
```diff
@@ -0,0 +1,308 @@
...
+        rem=u0_Mr_model,
+    )
+
+    simulation_integral = PF.magnetics.simulation.Simulation3DIntegral(
+        survey=survey,
+        mesh=mesh,
+        chi=eff_sus_model,
+        model_type="vector",
+        store_sensitivities="forward_only",
+        ind_active=ind_ellipsoid,
```
santisoler: Using `active_cells`, since `ind_active` is deprecated:
```suggestion
        active_cells=ind_ellipsoid,
```

tests/utils/shared_helpers.py
```diff
@@ -0,0 +1,518 @@
+from simpeg import utils
```
santisoler: @johnweis0480, did you reuse Takahashi's code in this class, or is this a new implementation of yours based on the paper?

If you reused some of their code, that's perfectly fine. We just need to add a license notice. I would rename this file to `ellipsoids.py` (so we only have this code here), and include a string in the top of it with something like:

```
"""
The code for forward modelling magnetic field of ellipsoids present in this
file is based on the implementation made by Diego Takahashi Tomazella and
Vanderlei C. Oliveira Jr., which is available in
https://github.com/pinga-lab/magnetic-ellipsoid, and has been released under
the BSD 3-Clause license:

> Copyright (c) Diego Takahashi Tomazella and Vanderlei C. Oliveira Jr.
> All rights reserved.
>
> Redistribution and use in source and binary forms, with or without
> modification, are permitted provided that the following conditions are met:
> 
> * Redistributions of source code must retain the above copyright notice, this
>   list of conditions and the following disclaimer.
> * Redistributions in binary form must reproduce the above copyright notice,
>   this list of conditions and the following disclaimer in the documentation
>   and/or other materials provided with the distribution.
> * Neither the names of the copyright holders nor the names of any
>   contributors may be used to endorse or promote products derived from this
>   software without specific prior written permission.
> 
> THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
> AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
> IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
> ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
> LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
> CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
> SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
> INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
> CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
> ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
> POSSIBILITY OF SUCH DAMAGE.
"""
```

tests/utils/shared_helpers.py
```diff
@@ -0,0 +1,518 @@
...
+import numpy as np
+
+
+class ProlateEllispse:
```
santisoler: Fixing a typo here, but also making a proposal for renaming this class:
```suggestion
class ProlateEllipsoid:
```

santisoler (changes requested):
Nice to see this getting in good shape! I left a few comments below, most of them regarding format and style, specially for the docs.

I agree with the changes you applied after @domfournier's review. And I like the tests you wrote! I think keeping the ellipsoid class in `tests.utils` is the right move. Hopefully we'll have a code like that somewhere else soon.

I saw there are a few untested lines. You can check them out through the "Files changed" tab in GitHub. Codecov reports back on which lines are currently being missed. It would be nice to add tests for most (if not all) of those, specially critical ones where a bug could lead to a wrong inversion or forward.

Feel free to ask any question. I'm also keen on meeting in person if you want to chat about this.

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1665,562 +1664,531 @@ def _gtg_diagonal_without_building_g(self, weights):
...
-    """
-    Secondary field approach using differential equations!
+    r"""A secondary field simulation for magnetic data.
+
+    Parameters
+    ----------
+    mesh : discretize.base.BaseMesh
+    survey : magnetics.survey.Survey
+    mu : float, array_like
+        Magnetic Permeability Model (H/ m). Set this for forward
```
domfournier, line 1664: Why not invert for `susceptibility` instead, and have mu_0 multiplied under the hood? This is generally the quantity of interest.

simpeg/potential_fields/magnetics/simulation.py
domfournier: Similar comment here, `\mu_0` could be baked in the computation, and user could invert for an "effective susceptibility" where M = H_0 * k_effective. This would bring both quantities to invert for be dimensionally equivalent.

domfournier (approved):
I think the code itself is in great shape (see @santisoler comments for final touch up).
However, I would like to propose a change in units. It is currently a bit cumbersum to have units of H/m (very small values) for the induced, and nT (very large and not SI) for the remanent vector. This could potentially also be an issue when inverting for both, as they are very far apart unit-wise.

Could it be possible to describe both quantities in terms of susceptiblity-like quantities? The induced component is just susceptibility, and the remanence described in terms of "effective susc" (scaled by the inducing field strength)? That would only require to have mu_0 multiplied everywhere under the hood, and a scaling of H0 for the remanent component (I think?).
Just a thought, and we don't need to act on it here.

simpeg/potential_fields/magnetics/simulation.py
santisoler: Maybe add a comment like this for now:
```
# TODO: experiment with avoiding having to define these three large arrays with constant values
```

simpeg/potential_fields/magnetics/simulation.py
```diff
@@ -1654,562 +1654,532 @@ def _gtg_diagonal_without_building_g(self, weights):
...
+        Magnetic Polarization :math:`\mu_0 \mathbf{M}` (nT). Set this for forward
+        modeling or to fix remanent magnetization while inverting for permeability.
+        This is used if ``remMap`` is None.
+    remMap : simpeg.maps.IdentityMap, optional
+        The mapping used to go from the simulation model to :math:`\mu_0 \mathbf{M}`.
+        Set this to invert for :math:`\mu_0 \mathbf{M}`.
+    storeJ: bool
+        Whether to store the sensitivity matrix. If set to True
+    use_float32_solver: bool
+        Whether to solve ``Ainv @ rhs`` using float32 precision.
```
santisoler: Updating docs after the change on the `solver_dtype` argument.
```suggestion
    solver_dtype: dtype, optional
        Data type to use for the matrix that gets passed to the ``solver``.
        Default to `numpy.float64`.
```

santisoler (approved):
Thanks @johnweis0480 for addressing all my comments.

I just left a minor suggestion to update the docs after the introduction of the `solver_dtype` argument. I will also push three commits:
* One provides a more explicit error message for invalid `solver_dtype`, and it adds a test for that error.
* The second one avoids using the `Survey.components` property on the `survey.setter`, since it's deprecated. It also adds a test for the error message upon unsupported components.
* The last one adds a test for `getJ`, since that method wasn't being tested. The test compares `getJ(model) @ vector` against `Jvec(model, vector)`, and `getJ(model).T @ vector` against `Jtvec(model, vector)`.

I think this PR is ready to be merged. If there's anything in particular we want to improve, I think we can do it in a separate PR once we merge this one.

simpeg/potential_fields/magnetics/simulation.py
johnweis0480: Yep I missed those two, thanks for looking at it in detail

PR #1683, opened 2025-06-10
Add shift to discrete topography for NSEM

dccowan:
#### Summary
The PR adds a utility for shifting locations relative to discrete surface topography.

#### PR Checklist
      to a Pull Request

We measure electric fields at the Earth's surface when performing MT surveys. Like DC/IP, the original measurement locations of the electric fields can end up in air cells when we define discrete surface topography. This utility allows the user to shift locations relative to discrete topography on Tensor and Tree meshes. For Airborne NSEM, this utility also allows the user to preserve the original flight heights.

#### Additional information
Unlike the DC/IP utility that alters a survey object, this utility is inputs the original locations and outputs the shifted ones. I could see us needing similar functionality for NSEM survey objects if they are created directly by loading some type of data file (like DC/IP).

### DIFF
--- a/simpeg/electromagnetics/natural_source/utils/__init__.py
+++ b/simpeg/electromagnetics/natural_source/utils/__init__.py
@@ -31,3 +31,4 @@ from .test_utils import (
     blockInhalfSpace,
     twoLayer,
 )
+from .nsem_utils import shift_to_discrete_topography
--- a/simpeg/electromagnetics/natural_source/utils/nsem_utils.py
+++ b/simpeg/electromagnetics/natural_source/utils/nsem_utils.py
@@ -1,0 +1,37 @@
+from ...static.utils import drapeTopotoLoc
+
+def shift_to_discrete_topography(
+    locations,
+    mesh,
+    active_cells,
+    option="top",
+    height=0.
+):
+    """Shift locations relative to discrete surface topography.
+
+    Parameters
+    ----------
+    locations : (n, dim) numpy.ndarray
+        The original locations.
+    mesh : discretize.TensorMesh or discretize.TreeMesh
+        The mesh defining the discrete domain.
+    active_cells : numpy.ndarray of int or bool
+        Active topography cells.
+    option : {"top", "center"}
+        Define discrete topography at tops of cells or cell centers.
+    height : float or (n,) numpy.ndarray
+        Location height relative to surface topography.
+
+    Returns
+    -------
+    (n, dim) numpy.ndarray
+        The shifted locations.
+    """
+
+    locations_shifted = drapeTopotoLoc(
+        mesh, locations, active_cells=active_cells, option=option
+    )
+
+    locations_shifted[:, -1] += height
+
+    return locations_shifted
--- a/simpeg/electromagnetics/natural_source/utils/nsem_utils.py
+++ b/simpeg/electromagnetics/natural_source/utils/nsem_utils.py
@@ -1,11 +1,8 @@
 from ...static.utils import drapeTopotoLoc
 
+
 def shift_to_discrete_topography(
-    locations,
-    mesh,
-    active_cells,
-    option="top",
-    height=0.
+    locations, mesh, active_cells, option="top", height=0.0
 ):
     """Shift locations relative to discrete surface topography.
 


dccowan:
I noticed something important about these projections functions. When originally developed for DC/IP, the projection ensured the final electrode locations didn't end up on an 'exposed' edge or corner of a mesh cell. This was accomplished by shifting the horizontal position of the original locations to lie over the nearest cell centers. This is somewhat necessary for electric field measurements but not necessarily for magnetic field measurements.

Does anyone have any opinions on this?

jcapriot:
> I noticed something important about these projections functions. When originally developed for DC/IP, the projection ensured the final electrode locations didn't end up on an 'exposed' edge or corner of a mesh cell. This was accomplished by shifting the horizontal position of the original locations to lie over the nearest cell centers. This is somewhat necessary for electric field measurements but not necessarily for magnetic field measurements.
> 
> Does anyone have any opinions on this?

It’s definitely useful for the CC dc simulation, as shifting it to the centers is consistent with the 0 Neumann BC at the top. For the nodal formulation, this is less of an issue as the nodes (which are where the interpolated values live) lay directly on the boundary faces.

dccowan:
> > I noticed something important about these projections functions. When originally developed for DC/IP, the projection ensured the final electrode locations didn't end up on an 'exposed' edge or corner of a mesh cell. This was accomplished by shifting the horizontal position of the original locations to lie over the nearest cell centers. This is somewhat necessary for electric field measurements but not necessarily for magnetic field measurements.
> > Does anyone have any opinions on this?
> 
> It’s definitely useful for the CC dc simulation, as shifting it to the centers is consistent with the 0 Neumann BC at the top. For the nodal formulation, this is less of an issue as the nodes (which are where the interpolated values live) lay directly on the boundary faces.

In the original implementation, forcing shifted locations over cell centers only happened for tree meshes. I added a "shift_horizontal" option to the new utilities. By default it is set to True since we mostly use this for electric dipole receivers.

dccowan:
I'm not sure I like the kwarg 'option' for setting discrete surface topography base on the top or center of cells. In the original utilities it was fine because the number of input arguments was small. We may want to choose something more descriptive? Although it might mean I need to change the deprecation of the original functionality. Thoughts?

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    """Return indices of closest gridded points for a set of input points.
+
+    Parameters
+    ----------
+    grid : (n, dim) numpy.ndarray
+        A gridded set of points.
+    pts : (m, dim) numpy.ndarray
+        Points being projected to gridded locations.
+    dim : int, default=2
+        Dimension of the points.
```
santisoler: I guess we can get rid of this parameter and get the `dim` from the shape of `grid` and `pts`. Am I right?

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+            ACTIND = active_cells.reshape(
+                (mesh.vnC[0] * mesh.vnC[1], mesh.vnC[2]), order="F"
+            )
+            ZC = zc.reshape((mesh.vnC[0] * mesh.vnC[1], mesh.vnC[2]), order="F")
+            topoCC = np.zeros(ZC.shape[0])
+
+            for i in range(ZC.shape[0]):
+                ind = np.argmax(ZC[i, :][ACTIND[i, :]])
+                if option == "top":
+                    dz = mesh.h[2][ACTIND[i, :]][ind] * 0.5
```
santisoler: Minor improvement here. The `ind` variable is only used for the `option="top"` case, so we can avoid running an `argmax` for the other cases.
```suggestion
                if option == "top":
                    ind = np.argmax(ZC[i, :][ACTIND[i, :]])
                    dz = mesh.h[2][ACTIND[i, :]][ind] * 0.5
```

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+                elif option == "center":
+                    dz = 0.0
+                else:
+                    raise Exception()
```
santisoler: Usually it's not desirable to raise an `Exception`, but to use an error instead. In this case, we can raise a `ValueError`, since in this branch of the `if` statement, the `option` argument would have a value different than `"center"` or "`top"`.

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+            YC = yc.reshape((mesh.vnC[0], mesh.vnC[1]), order="F")
+            topoCC = np.zeros(YC.shape[0])
+            for i in range(YC.shape[0]):
+                ind = np.argmax(YC[i, :][ACTIND[i, :]])
+                if option == "top":
+                    dy = mesh.h[1][ACTIND[i, :]][ind] * 0.5
+                elif option == "center":
+                    dy = 0.0
+                else:
+                    raise Exception()
```
santisoler: Same for this exception.

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+                topoCC[i] = YC[i, :][ACTIND[i, :]].max() + dy
+            return mesh1D, topoCC
+
+    elif mesh._meshType == "TREE":
+        inds = mesh.get_boundary_cells(active_cells, direction="zu")[0]
+
+        if option == "top":
+            dz = mesh.h_gridded[inds, -1] * 0.5
+        elif option == "center":
+            dz = 0.0
```
santisoler: Here we could also raise a `ValueError` if `option` is not `"top"` or `"center"`. Currently, if that happens it'll make the `dz` to be undefined, and users will get a `NameError` because `dz` is not defined, which isn't very informative.

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    if mesh._meshType != "TENSOR" and mesh._meshType != "TREE":
+        raise NotImplementedError(
+            "shift_to_discrete_topography only supported for TensorMesh and TreeMesh'."
+        )
+
+    if not isinstance(heights, (int, float)):
+        if len(pts) != len(heights):
+            raise ValueError(
+                "If supplied as a `numpy.ndarray`, the number of heights must equal the number of points."
+            )
```
santisoler: We can combine the two if statements here to reduce nesting:
```suggestion
    if not isinstance(heights, (int, float)) and len(pts) != len(heights):
        raise ValueError(
            "If supplied as a `numpy.ndarray`, the number of heights must equal the number of points."
        )
```

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+        the topography.The topography is defined using the 'topo' input parameter.
+    heights : float or (n,) numpy.ndarray
+        Height(s) relative to the true surface topography. Used to preserve flight
+        heights or borehole depths.
+    shift_horizontal : bool
+        When True, locations are shifted horizontally to lie vertically over cell
+        centers. When False, the original horizontal locations are preserved.
+    topo : (n, dim) numpy.ndarray
+        Surface topography. Can be used if an active indices array cannot be
+        provided for the input parameter 'active_cells'.
```
santisoler: Adding a few `optional` to these docs:
```suggestion
    option : {"top", "center"}, optional
        Define whether the cell center or entire cell of actice cells must be below
        the topography.The topography is defined using the 'topo' input parameter.
    heights : float or (n,) numpy.ndarray, optional
        Height(s) relative to the true surface topography. Used to preserve flight
        heights or borehole depths.
    shift_horizontal : bool, optional
        When True, locations are shifted horizontally to lie vertically over cell
        centers. When False, the original horizontal locations are preserved.
    topo : (n, dim) numpy.ndarray, optional
        Surface topography. Can be used if an active indices array cannot be
        provided for the input parameter 'active_cells'.
```

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,200 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    active_cells : numpy.ndarray of bool or int
+        Active cells index; i.e. indices of cells below surface
+    option : {"top", "center"}
+        Use string to specify if the surface passes through the
+        tops or cell centers of surface cells.
+
+    Returns
+    -------
+    (n, dim) numpy.ndarray
+        xy[z] topography.
```
santisoler: By reading the code I noticed that for a `TensorMesh`, the function returns a mesh and an array. While for the `TreeMesh` case, it returns two arrays.

Would it be possible to stay with only one type of output? Also, we should update the docstrings so the returns match the output of the function.

tests/utils/test_mesh_utils.py
```diff
@@ -0,0 +1,139 @@
...
+    if option == "center":
+        correct_elevations -= 0.5 * DH
+
+    if shift_horizontal:
+        correct_locations = np.round(pts) + 0.5 * DH
+    else:
+        correct_locations = pts.copy()
+    correct_locations[:, -1] = correct_elevations
+
+    assert np.all(np.isclose(pts_shifted, correct_locations))
```
santisoler: Minor thing here. We could use `numpy.testing.assert_allclose` here instead of a simple `assert`. The former provides more information in case the assertion fails, which is very useful when debugging.

simpeg/utils/mesh_utils.py
dccowan: I think so. I used squeeze() in case the grid and pts were (n, 1).

simpeg/utils/mesh_utils.py
dccowan: For the TensorMesh, it looks like we are leveraging the closest_points_index method. But the TreeMesh class also appears to have the method closest_points_index. I'll need to think a little bit more about what we want the outputs to be for functionality that users will interact with.

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,192 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+
+
+def shift_to_discrete_topography(
+    mesh,
+    pts,
+    active_cells=None,
+    option="top",
+    shift_horizontal=True,
+    heights=0.0,
+    topo=None,
```
jcapriot: Is this argument necessary, or can we just force the user to provide `active_cells`?

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,192 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+            [np.abs(pt - grid).argmin() for pt in pts.tolist()], dtype=int
+        )
+    else:
+        tree = cKDTree(grid)
+        _, grid_inds = tree.query(pts)
+
+    return grid_inds
+
+
+def get_discrete_topography(mesh, active_cells, option="top"):
```
jcapriot: I think having this method return a N-1D mesh, and an array in all cases would simplify the `shift_to_discrete_topography` function, and make it so you don't have to reimplement all of the `_closest_grid_indeces` functionality that is already built into discretize.

simpeg/utils/mesh_utils.py
```diff
@@ -97,15 +100,192 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+            return mesh1D.cell_centers, topoCC
+
+    elif mesh._meshType == "TREE":
+        inds = mesh.get_boundary_cells(active_cells, direction="zu")[0]
+
+        if option == "top":
+            dz = mesh.h_gridded[inds, -1] * 0.5
+        elif option == "center":
+            dz = 0.0
+        return mesh.cell_centers[inds, :-1], mesh.cell_centers[inds, -1] + dz
```
jcapriot: building a new mesh for the tree mesh would actually be pretty simple, from this array of indices. Just form a new mesh in D-1, grab the level of each cell from the old mesh, then insert each cell into the new mesh at the same level.

jcapriot: The logic should mimic how discretize plot's a slice through the `TreeMesh`:
https://github.com/simpeg/discretize/blob/4690ff5c9b44c1f99d0a12cca926fdfdbe2bf162/discretize/mixins/mpl_mod.py#L2152-L2163

simpeg/utils/mesh_utils.py
jcapriot: Also, be aware that using a KDTree to lookup the closest cell center is **not** the same thing as finding which cell *contains* the point.

dccowan: For this project, I intended to preserve the outputs of the original functionality when the original inputs are used. The goals were to:
1. add functionality
2. have the functionality named and located within SimPEG appropriately.

If we make changes to the underlying methods or outputs, then we would be making new functions. Is that what you would like to do? In that case, the original functionality would remain in static_utils.py, but be deprecated.

dccowan:
Just for an update on this:

- the original utility functions in electromagnetics/static/utils/static_utils.py will remain for now an be deprecated
- new utility functions are in utils/mesh_utils.py
- the new functionality relies on TensorMesh.point2index, which will not be available until discretize 0.11.4(?).
- I am not overly concerned about using the _closest_grid_indices function. Locations shouldn't be defined immediately at cell size increases
- the shift_to_discrete_topography function allows the user to provide or not provide the elevations. I did this in the event we have surface DC/IP locations.
- the shift electrodes method for DC/IP surveys calls the new functionality

simpeg/electromagnetics/static/utils/static_utils.py
```diff
@@ -1521,15 +1524,23 @@ def r_unit(p1, p2):
     return vec, r
 
 
-def gettopoCC(mesh, ind_active, option="top"):
+@deprecated(
+    "The `gettopoCC` function is deprecated, "
+    "and will be removed in SimPEG v0.26.0. "
```
santisoler: We remove bits of code after two minor releases since the deprecation warning as minimum, so we better remove this in v0.27.0.
```suggestion
    "and will be removed in SimPEG v0.27.0. "
```

simpeg/utils/mesh_utils.py
```diff
@@ -94,3 +98,187 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    pts : (m, dim) numpy.ndarray
+        Points being projected to gridded locations.
+    dim : int
+        The `dim` input is deprecated and will be removed in SimPEG v0.24.0.
+
+    Returns
+    -------
+    (n,) numpy.ndarray
+        Indices of the closest gridded points for all *pts* supplied.
+    """
```
santisoler: I think we can remove the `dim` argument, right?
```suggestion
def _closest_grid_indices(grid, pts):
    """Return indices of closest gridded points for a set of input points.

    Parameters
    ----------
    grid : (n, dim) numpy.ndarray
        A gridded set of points.
    pts : (m, dim) numpy.ndarray
        Points being projected to gridded locations.

    Returns
    -------
    (n,) numpy.ndarray
        Indices of the closest gridded points for all *pts* supplied.
    """
```

simpeg/utils/mesh_utils.py
```diff
@@ -94,3 +98,187 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    mesh : TensorMesh or discretize.TreeMesh
+        A tensor or tree mesh.
+    active_cells : numpy.ndarray of bool or int
+        Active cells index; i.e. indices of cells below surface
+    topo_cell_cutoff : {"top", "center"}
+        String to specify the cutoff for ground cells and the locations of
+        the discrete topography. For "top", ground cells lie entirely below
+        the surface topography and the discrete topography is defined on the
+        top faces of surface cells. For "center", only the cell centers must
+        lie below the surface topograpy and the discrete topography is defined
```
santisoler: Fix minor typo here:
```suggestion
        lie below the surface topography and the discrete topography is defined
```

simpeg/utils/mesh_utils.py
```diff
@@ -94,3 +98,187 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    pts : (n, dim) numpy.ndarray
+        The original set of points being shifted relative to the discretize
+        surface topography.
+    active_cells : numpy.ndarray of int or bool, optional
+        Index array for all cells lying below the surface topography.
+    topo_cell_cutoff : {"top", "center"}
+        String to specify the cutoff for ground cells and the locations of the discrete
+        topography. For "top", ground cells lie entirely below the surface topography
+        and the discrete topography is defined on the top faces of surface cells.
+        For "center", only the cell centers must lie below the surface topograpy and
```
santisoler: Fix minor typo here:
```suggestion
        For "center", only the cell centers must lie below the surface topography and
```

simpeg/utils/mesh_utils.py
```diff
@@ -94,3 +98,187 @@ def surface2inds(vrtx, trgl, mesh, boundaries=True, internal=True):
...
+    topo_cell_cutoff="top",
+    shift_horizontal=True,
+    heights=0.0,
+):
+    """
+    Shift locations relative to discrete surface topography.
+
+    Parameters
+    ----------
+    mesh : TensorMesh or discretize.TreeMesh
```
santisoler: Use discretize.TensorMesh so sphinx can render the intersphinx link:
```suggestion
    mesh : discretize.TensorMesh or discretize.TreeMesh
```

simpeg/electromagnetics/static/resistivity/survey.py
```diff
@@ -239,9 +240,11 @@ def drape_electrodes_on_topography(
         self,
         mesh,
         active_cells,
-        option="top",
+        topo_cell_cutoff="top",
         topography=None,
         force=False,
```
santisoler, line 245: The `topography` and `force` arguments are not being used in the function. I noticed it here, but this is something that comes from `main`. Shall we deprecated them?

We could add something like this in the body of the method:

```python
        if topography is not None:
            msg = (
                "The `topography` argument is not used in the "
                "`drape_electrodes_on_topography` and will be removed in "
                "SimPEG v0.27.0."
            )
            warnings.warn(msg, category=FutureWarning, stacklevel=2)
        if force is not False:
            msg = (
                "The `force` argument is not used in the "
                "`drape_electrodes_on_topography` and will be removed in "
                "SimPEG v0.27.0."
            )
            warnings.warn(msg, category=FutureWarning, stacklevel=2)
```

Would be also nice to filter those warnings when testing by adding the following to the `filterwarnings` list in `pyproject.toml`:
```
    'error:The `topography` argument is not used in the `drape_electrodes_on_topography`:FutureWarning',
    'error:The `force` argument is not used in the `drape_electrodes_on_topography`:FutureWarning',
```

I can add this later on in a different PR, no worries.

santisoler (approved):

simpeg/electromagnetics/static/resistivity/survey.py
santisoler: Opened #1723 for that

Issue #1685, opened 2025-07-03
BUG: ComplexMap - bug in deriv when in combo - sparse matrix and linear operator error


lheagy:
### Describe the issue:

When using the ComplexMap in a combo, we run into an error in the derivatives.
> TypeError: Unable to multiply a LinearOperator with a sparse matrix. Wrap the matrix in aslinearoperator first.

### Reproducible code example:

```python
import numpy as np
import discretize
from simpeg import maps

mesh_test = discretize.TensorMesh([4]) 
active_cells = mesh_test.cell_centers < 0.5
mapping = maps.InjectActiveCells(mesh_test, active_cells=active_cells, value_inactive=0) * maps.ComplexMap(nP=np.sum(active_cells)*2)
mapping.deriv(np.ones(4))
```

### Error message:

```shell
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/miniforge3/envs/sp-csem-inv/lib/python3.11/site-packages/scipy/sparse/linalg/_interface.py:366, in LinearOperator.matmat(self, X)
    365 try:
--> 366     Y = self._matmat(X)
    367 except Exception as e:

File ~/miniforge3/envs/sp-csem-inv/lib/python3.11/site-packages/scipy/sparse/linalg/_interface.py:613, in _CustomLinearOperator._matmat(self, X)
    612 else:
--> 613     return super()._matmat(X)

File ~/miniforge3/envs/sp-csem-inv/lib/python3.11/site-packages/scipy/sparse/linalg/_interface.py:213, in LinearOperator._matmat(self, X)
    207 """Default matrix-matrix multiplication handler.
    208 
    209 Falls back on the user-defined _matvec method, so defining that will
    210 define matrix multiplication (though in a very suboptimal way).
    211 """
--> 213 return np.hstack([self.matvec(col.reshape(-1,1)) for col in X.T])

File ~/miniforge3/envs/sp-csem-inv/lib/python3.11/site-packages/scipy/sparse/linalg/_interface.py:213, in <listcomp>(.0)
    207 """Default matrix-matrix multiplication handler.
    208 
    209 Falls back on the user-defined _matvec method, so defining that will
    210 define matrix multiplication (though in a very suboptimal way).
    211 """
--> 213 return np.hstack([self.matvec(col.reshape(-1,1)) for col in X.T])

[...]

PR #1686, opened 2025-07-03
LinearOperators in mapping derivs 

lheagy:
This is a first cut at resolving #1685. This is a lightweight patch-fix. We may want to unify all maps so that they take a `v`, and rather than doing the multiplication like this, we pass the v through and have the individual map cast things if it needs to

### DIFF
--- a/simpeg/maps/_base.py
+++ b/simpeg/maps/_base.py
@@ -7,6 +7,7 @@ import discretize
 import numpy as np
 import scipy.sparse as sp
 from scipy.sparse import csr_matrix as csr
+from scipy.sparse.linalg import aslinearoperator, LinearOperator
 from discretize.tests import check_derivative
 from discretize.utils import Zero, Identity, mkvc, speye, sdiag
 import uuid
@@ -529,7 +530,10 @@ class ComboMap(IdentityMap):
 
         mi = m
         for map_i in reversed(self.maps):
-            deriv = map_i.deriv(mi) * deriv
+            if isinstance(deriv, LinearOperator):
+                deriv = aslinearoperator(map_i.deriv(mi)) * deriv
+            else:
+                deriv = map_i.deriv(mi) * deriv
             mi = map_i * mi
         return deriv
 


simpeg/maps/_property_maps.py
```diff
@@ -963,7 +963,8 @@ def adj(v):
 
         if v is not None:
             return LinearOperator(shp, matvec=fwd, rmatvec=adj) * v
-        return LinearOperator(shp, matvec=fwd, rmatvec=adj)
+        # return LinearOperator(shp, matvec=fwd, rmatvec=adj)
+        return sp.hstack([sdiag(np.ones(nC)), 1j * sdiag(np.ones(nC))])
```
jcapriot: This is not technically the same as the previous one, the transpose is not the same :/.

jcapriot (changes requested):

simpeg/maps/_property_maps.py
jcapriot:
```suggestion
        return sp.diags([1, 1j], [0, nC], [nC, 2 * nC])
```

santisoler:
@lheagy, since we are not returning the `LinearOperator` anymore, what if we simplify the method to something like:

```python
def deriv(self, m, v=None):
            nC = self.shape[0]
        if v is not None:
            return v[:nC] + v[nC:] * 1j
        return sp.diags([1, 1j], [0, nC], [nC, 2 * nC])
```

We should also include a comment stating that this method is not returning the adjoint.

lheagy:
That sounds good to me @santisoler!

santisoler:
Done! Please, double check if the admonition is correct. Feel free to fix it if not.

simpeg/maps/_property_maps.py
```diff
@@ -950,6 +949,18 @@ def deriv(self, m, v=None):
...
+
+            Calculating the transpose of the derivative of the
+            :class:`~simpeg.maps.ComplexMap` as follows doesn't return the adjoint of
+            the matrix, but its transpose:
+
+            .. code:: python
+
+                complex_map = ComplexMap(...)
+                derivative = complex_map.deriv(m)
+                derivative.T  # this is not the adjoint
```
lheagy:
```suggestion
                derivative.T  # this is not the complex adjoint
```

lheagy: minor suggestion just to specify that the complex part is where this isn't truly the adjoint

PR #1689, opened 2025-08-06
Allow for list of int. Only project active times

domfournier:
(cherry picked from commit 1dfafa5b013d610fdc1d67589400b8fcc5ad8c5f)

#### Summary
This proposed change streamlines the compute of predicted data from time domain fields. Instead of looping over all dts, only loop over the times that needed by the projection of the receivers.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/electromagnetics/time_domain/receivers.py
+++ b/simpeg/electromagnetics/time_domain/receivers.py
@@ -1,5 +1,5 @@
 import scipy.sparse as sp
-
+import numpy as np
 from ...utils import mkvc, validate_type, validate_direction
 from discretize.utils import Zero
 from ...survey import BaseTimeRx
@@ -120,6 +120,20 @@ class BaseRx(BaseTimeRx):
         projected_time_grid = f._TLoc(self.projField)
         return time_mesh.get_interpolation_matrix(self.times, projected_time_grid)
 
+    def active_times(self, projection):
+        """Get active times for the receiver.
+
+        Parameters
+        ----------
+        projection : Sparse matrix
+
+        Returns
+        -------
+        numpy.ndarray
+            Active times for the receiver.
+        """
+        return np.unique(sp.find(projection)[1])
+
     def getP(self, mesh, time_mesh, f):
         """Returns projection matrices as a list for all components collected by the receivers.
 
@@ -145,6 +159,7 @@ class BaseRx(BaseTimeRx):
 
         Ps = self.getSpatialP(mesh, f)
         Pt = self.getTimeP(time_mesh, f)
+        Pt = Pt[:, self.active_times(Pt)]
         P = sp.kron(Pt, Ps)
 
         if self.storeProjections:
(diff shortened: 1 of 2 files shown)


lheagy (approved):

lheagy (commented):
Thanks for this @domfournier! This is a nice update. I ran it for a simple example and am seeing about a 5-10% speed-up, which is meaningful! 

**Before: ** 
image

**After:**
image

It looks like some of the EM tests are failing. I can take a look. I see the pr is still a draft. Is there something else you would like to update before bringing it in? or just the tests?

simpeg/electromagnetics/time_domain/receivers.py
```diff
@@ -120,6 +120,20 @@ def getTimeP(self, time_mesh, f):
...
+        Parameters
+        ----------
+        projection : Sparse matrix
+
+        Returns
+        -------
+        numpy.ndarray
+            Active times for the receiver.
+        """
+        return np.unique(sp.find(projection)[1])
```
lheagy, line 135: I think we should be able to do the following since it is a sparse matrix 
```suggestion
        return projection.indices
```

simpeg/electromagnetics/time_domain/receivers.py
```diff
@@ -145,6 +159,7 @@ def getP(self, mesh, time_mesh, f):
 
         Ps = self.getSpatialP(mesh, f)
         Pt = self.getTimeP(time_mesh, f)
+        Pt = Pt[:, self.active_times(Pt)]
```
santisoler, line 162: The one issue I see by adding this line in `getP` is that any time we want to compute the dot product between `P` and any vector, the vector should also be "reduced" to only the active times. This makes the `getP` method less _standalone_: we need to call `active_times` again if we want to compute dot products of `P` with any vector.

Would it be possible to apply the optimization directly in `eval` and `evalDeriv`, without modifying the output of `getP`? ~I guess I'm asking if filtering out non-active times in the `P` matrix (`P = sp.kron(Pt, Ps)`) is possible and would achieve same performance as doing so in `Pt`.~ **Update:** no, it wouldn't be possible because the output of `kron` is a `coo_matrix` that cannot be subscripted.

santisoler (commented):
Leaving a few comments below. I'll try to experiment with my second thought and share results.

simpeg/electromagnetics/time_domain/receivers.py
santisoler: I suspect that using `.indices` is going to be faster. Just two caveats:
* We still need to do `np.unique(projection.indices)`. If we don't get the unique ones, the `Pt` matrix after doing `Pt = Pt[:, self.active_times(Pt)]` will be actually larger than the original one (in case we have repeated column indices).
* Using `indices` or the `find` function is equivalent only if `projection` is a `csr_matrix`. For example, if it's a `csc_matrix`, then the `.indices` will be row indices instead of the column indices. Since `getP` always return a `csr_matrix`, this should not be a problem, But I would clarify this in the docstring, and maybe checking `isinstance(projection, csr_matrix)` would be a good idea.

PR #1699, opened 2025-09-16
Add How to Guide page on moving mesh to survey area

santisoler:
#### Summary

Add a new page to the How to Guide that explains how to move a mesh to a given survey area. This is handy when working with real-world data, whose coordinates are going to be given in projected plain coordinates.

#### PR Checklist
      to a Pull Request

### DIFF
--- a/docs/content/user-guide/how-to-guide/move-mesh-to-survey.rst
+++ b/docs/content/user-guide/how-to-guide/move-mesh-to-survey.rst
@@ -1,0 +1,202 @@
+============================
+Locating mesh on survey area
+============================
+
+The :mod:`discretize` package allows us to define 3D meshes that can be used
+for running SimPEG's forward and inverse problems.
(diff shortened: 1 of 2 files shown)


santisoler:
Do you mean moving sources and receivers to the default location of the mesh (around 0, 0, 0)?

jcapriot:
Yeah, or on a centered mesh with origin an ‘CCN’

santisoler:
That could be a good addition. I think the goal of this page is to address a common question that users have: "how to locate the mesh in the coordinates of the survey", from a more applied point of view. I'd recommend practitioners to move the mesh rather than the survey, so at least for now I think I'd prefer to keep only this guide, just to avoid any confusion of what is the preferred method.

My point is having the mesh in the survey coordinates ultimately simplify workflows: mesh slices and plots will be already in survey coordinates, exported meshes (that sometimes are opened with other software like Geoscience ANALYST) are already in those coordinates.

PR #1709, opened 2025-10-09
Start debugging failing test in EM

santisoler:
#### Summary

TBD

#### PR Checklist
      to a Pull Request

### DIFF
--- a/.ci/azure/docs.yml
+++ b/.ci/azure/docs.yml
@@ -5,7 +5,7 @@ jobs:
     pool:
       vmImage: ubuntu-latest
     variables:
-      python.version: "3.10"
+      python.version: "3.11"
     timeoutInMinutes: 240
     steps:
       # Checkout simpeg repo.
--- a/.ci/azure/pypi.yml
+++ b/.ci/azure/pypi.yml
@@ -12,7 +12,7 @@ jobs:
 
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.10"
+          versionSpec: "3.11"
         displayName: "Setup Python"
 
       - bash: |
@@ -57,7 +57,7 @@ jobs:
 
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.10"
+          versionSpec: "3.11"
         displayName: "Setup Python"
 
       - bash: |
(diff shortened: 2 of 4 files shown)


tests/em/fdem/forward/test_FDEM_dipolar_sources.py
```diff
@@ -104,4 +104,6 @@ def test_dipolar_fields(simulation_type, field_test, mur, orientation="Z"):
     elif field_test == "hPrimary":
         analytic = projection(dipole.magnetic_field(grid))
 
+    print(f"{numeric=}")
+    print(f"{analytic=}")
     assert np.abs(np.mean((numeric / analytic)) - 1) < TOL
```
jcapriot, line 109: One thing is that is wrong with this test is that it is not actually testing for the mean relative error:

it should be:
```
assert np.mean(np.abs(numeric/analytic - 1)) < TOL
```
but then it would fail at this tolerance level.

It appears to be testing that the absolute value of the mean ratio of the fields  - 1 is below a tolerance, which is not actually a good test. For example, consider these two numeric and analytic arrays that would pass this test:

```python
>>> numeric = np.r_[1, 3]
>>> analytic = np.r_[-1, 1]
>>> numeric / analytic
array([-1.,  3.])
>>> np.mean(numeric/analytic)
np.float64(1.0)
>>> np.mean(numeric/analytic) - 1
np.float64(0.0)
>>> np.abs(np.mean(numeric/analytic) - 1)
np.float64(0.0)
>>> np.abs(np.mean(numeric/analytic) - 1) < 1E-18
np.True_
```

@lheagy Can you redesign these tests?

jcapriot: Note this same issue is apparent on all of the tests added in #1575

santisoler: I'd suggest to use `np.testing.assert_allclose` instead for a couple of reasons:

* It tests that `|a - b| / |b| < tol`, but element-wise.
* It's more verbose and easy to read: `np.testing.assert_allclose(numeric, analytic, rtol=rtol)`.
* And when it fails, it provides way more information than just an `AssertionError`.

santisoler: Alternatively, if we don't want to test element-wise, I would test against a root-mean square instead.

PR #1714, opened 2025-10-15
Allow for specifying ramp start and ramp end in `RampOffWaveform`

jcapriot:
#### Summary
Allows for specifying ramp start and end times for the ramp off waveform

#### Squash and Merge summary

Allow passing `ramp_start` and `ramp_end` to the `RampOffWaveform` TDEM source and fix issue with waveform being wrong on times before `-epsilon`. Deprecate the `off_time` argument.

#### PR Checklist
      to a Pull Request

#### Fixes
Corrects waveform for times less than `epsilon`, Previously the waveform would be zero if evaluated at times before `-epsilon`, when it should be `1` (on)

### DIFF
--- a/simpeg/electromagnetics/time_domain/sources.py
+++ b/simpeg/electromagnetics/time_domain/sources.py
@@ -219,14 +219,13 @@ class RampOffWaveform(BaseWaveform):
                 raise TypeError(
                     "Can not specify both `off_time` and a `ramp_end` value."
                 )
-            else:
-                ramp_end = off_time
-                warnings.warn(
-                    "`off_time` keyword arg has been deprecated and will be removed in "
-                    "SimPEG v0.26.0, pass the ramp end time as the last positional argument.`",
-                    DeprecationWarning,
-                    stacklevel=2,
-                )
+            ramp_end = off_time
+            warnings.warn(
+                "`off_time` keyword arg has been deprecated and will be removed in "
+                "SimPEG v0.27.0, pass the ramp end time as the last positional argument.`",
+                DeprecationWarning,
+                stacklevel=2,
+            )
 
         nargs = len(args)
         if nargs == 0:


simpeg/electromagnetics/time_domain/sources.py
```diff
@@ -199,31 +199,82 @@ class RampOffWaveform(BaseWaveform):
 
     """
 
-    def __init__(self, off_time=0.0, **kwargs):
-        super().__init__(off_time=off_time, has_initial_fields=True, **kwargs)
+    def __init__(self, ramp_end=0.0, ramp_start=0.0, **kwargs):
+        if (off_time := kwargs.pop("off_time", None)) is not None:
+            ramp_end = off_time
```
santisoler: There's something here that it's making my brain itch. If a user passes `RampOffWavefor(ramp_start=0.0, ramp_end=1.0, off_time=1.2)`, then the `ramp_end` will get assigned to `1.2` without any notice. Should we deprecate `off_time` directly? Also, we should update the docstrings and explain what `ramp_end` and `ram_start` are.

simpeg/electromagnetics/time_domain/sources.py
```diff
@@ -199,31 +199,82 @@ class RampOffWaveform(BaseWaveform):
...
         else:
             return 0.0
 
     def eval_deriv(self, time):
         t = np.asarray(time, dtype=float)
         out = np.zeros_like(t)
 
-        if self.off_time > 0:
-            out[(t < self.off_time) & (t >= self.epsilon)] = -1.0 / self.off_time
+        ramp_width = self.ramp_end - self.ramp_start
```
santisoler, line 325: Nit picking here. Should we add a check for `ramp_width` to be positive? Maybe raise a warning if not...

simpeg/electromagnetics/time_domain/sources.py
santisoler: And what about changing the order of `ramp_end` and `ramp_start`? I know that would break backward compatibility though... but there's something in that order that bugs me 🙃 (nit picking here too).

simpeg/electromagnetics/time_domain/sources.py
jcapriot: The end time is already enforced to always be larger the start time.

simpeg/electromagnetics/time_domain/sources.py
jcapriot: Yeah, I was originally going for backward compatibility and keyword arg support, but I think the cleanest is to just do it with variable length positional args to allow both:
* `RampOffWaveform(0.1)` to have an implicit start time of 0 (also backwards compatible for positional argument passing).
* `RampOffWaveform(0.01, 0.1)` for specifying both start and end times
* `RampOffWaveform(off_time=0.1)` same as first, but issues a deprecated warning.
* `RampOffWaveform(0.01, off_time=0.1)` raises a `TypeError`
* `RampOffWaveform()` raises a `TypeError` (as well as when calling with too many positional arguments.)

simpeg/electromagnetics/time_domain/sources.py
```diff
@@ -197,37 +206,135 @@ class RampOffWaveform(BaseWaveform):
...
+            if ramp_start is not None:
+                raise TypeError(
+                    "argument for RampOffWaveform() given by name ('ramp_start') and position (position 0)"
+                )
+            if ramp_end is not None:
+                ramp_start = args[0]
+            else:
+                ramp_start = 0
+                ramp_end = args[0]
+        elif len(args) == 2:
```
santisoler:
```suggestion
        elif nargs == 2:
```

simpeg/electromagnetics/time_domain/sources.py
```diff
@@ -197,37 +206,135 @@ class RampOffWaveform(BaseWaveform):
...
+        if off_time is not None:
+            if len(args) > 1 or ramp_end is not None:
+                raise TypeError(
+                    "Can not specify both `off_time` and a `ramp_end` value."
+                )
+            else:
+                ramp_end = off_time
+                warnings.warn(
+                    "`off_time` keyword arg has been deprecated and will be removed in "
+                    "SimPEG v0.26.0, pass the ramp end time as the last positional argument.`",
```
santisoler: Shall we remove it in two minor versions?
```suggestion
                    "SimPEG v0.27.0, pass the ramp end time as the last positional argument.`",
```

simpeg/electromagnetics/time_domain/sources.py
```diff
@@ -197,37 +206,135 @@ class RampOffWaveform(BaseWaveform):
...
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
santisoler: Minor suggestion here. We don't need the `else` (I also included the change to v0.27.0 here):
```suggestion
            if len(args) > 1 or ramp_end is not None:
                raise TypeError(
                    "Can not specify both `off_time` and a `ramp_end` value."
                )
            ramp_end = off_time
            warnings.warn(
                "`off_time` keyword arg has been deprecated and will be removed in "
                "SimPEG v0.27.0, pass the ramp end time as the last positional argument.`",
                DeprecationWarning,
                stacklevel=2,
            )
```

PR #1717, opened 2025-10-16
Simple fix for pymatsolver 0.4.0 

jcapriot:
#### PR Checklist
      to a Pull Request

Squeeze these two locations in the SimulationDC operation (to be consistent on output of `__inner_mat_mul `operation)

#### Additional information
The `__inner_mat_mul` function will "squeeze" out single length second axes, so this fix causes it to be consistent for this operation in CG. This function had to perform this way because of the way pymatsoler previously handled single length array inputs.


jcapriot:
A more robust solution will be to fix the shape of items returned by accessing a field object with a single source,
* Instead of returning a (n, 1) array always, make it return an (n,) array if only passing a single source object (similar to returning a slice from a 2d array) this will likely require many changes throughout the code where we've specifically written things to undo this...
* Update NSEM primary/secondary fields to use an array shaped like (2, n, 1), as that is more consistent with normal broadcasting rules. I.e. it would naturally broadcast against many right hand side vectors shaped like (2, n, 10).
* This would likely allow the _inner_mat_mul_op function to be simplified.

PR #1736, opened 2025-11-25
Fix bug in apparent conductivity

dccowan:
#### Summary

This PR is meant to fix a bug in the NSEM ApparentConductivity receiver class. The current implementation adds the fields for x and y polarizations before taking the absolute value. This is incorrect. As you can see, the anomalies do not align with 3D structures. This was missed in current pytesting since the forward simulation is being validated for a 1D model.

image

#### PR Checklist
      to a Pull Request

### DIFF
--- a/simpeg/utils/pgi_utils.py
+++ b/simpeg/utils/pgi_utils.py
@@ -256,11 +256,34 @@ class WeightedGaussianMixture(GaussianMixture if sklearn else object):
 
         return weights
 
-    def _check_parameters(self, X):
+    def _warn_xp_not_numpy(self, xp):
+        """
+        Raise warning if the passed array API is not Numpy.
+
+        SimPEG's Gaussian Mixture Models don't currently support other array APIs beside
+        Numpy, so it's better to warn users that are intending to use another API.
+        """
+        if xp is None or xp is np:
+            return
+        try:
+            module = xp.__array_namespace_info__.__module__
+        except AttributeError:
+            module = "unknown"
+        if module.lower() != "numpy":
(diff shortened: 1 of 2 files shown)


santisoler (changes requested):
Nice catch, @dccowan! Would you add a test that passes with the bugfix, but fails without it?

I also see a few changes to the docstrings. Are those intended for this PR or do you want to open another one for those?

lheagy:
Thanks @dccowan and @santisoler! This is a good catch. 

Designing a test for this one is subtle.... There isn't an easy analytic solution to compare against. However, we might be able to test for symmetry. Would the following work: 
- use an example of a single, elongated block in a halfspace (e.g. only the conductor or the resistor) and have it centered in x and y.
- run a simulation and check that the data on the left-hand side is a mirror image of the data on the right hand side, and similarly for the top half and bottom half. e.g. something like
```
inds_left = rx_locs[:, 0] < 0
inds_right = rx_locs[:, 0] > 0
inds_bottom = rx_locs[:, 1] < 0
inds_top = rx_locs[:, 1] > 0

dobs_left = dobs[inds_left]
dobs_right = dobs[inds_right]
assert np.allclose(np.flipud(dobs_left), dobs_right)  # I can't remember if we should use flipud or fliplr here?? 

dobs_bottom = dobs[inds_bottom]
dobs_top = dobs[inds_top]
assert np.allclose(np.flipud(dobs_bottom), dobs_top) 
```  

This should test for symmetry, which would have failed before but should now be fixed

santisoler (approved):

PR #1729, opened 2025-10-27
Fix bug in IP effects on EM1D simulations

GrindingTheophile:
Hi there, @simpeg/simpeg-developers

Please, refer to #1706 for details!

#### Summary
This PR fixes a typo in [EM `compute_complex_sigma` method](https://github.com/simpeg/simpeg/blob/249e69a880c33a1554f7368397b7bc0051a4fef8/simpeg/electromagnetics/base_1d.py#L236-L281) and completes its docstring.

#### PR Checklist
* [x] Added [tests](https://docs.simpeg.xyz/latest/content/getting_started/contributing/testing.html) to verify changes to the code.
* [x] Added necessary documentation to any new functions/classes following the
      expect [style](https://docs.simpeg.xyz/latest/content/getting_started/contributing/documentation.html).
* [x] Marked as ready for review (if this is was a draft PR), and converted 
      to a Pull Request
* [x] Tagged ``@simpeg/simpeg-developers`` when ready for review.

#### Reference issue
Fixes #1706

### Squash and Merge summary

Fix bug in EM1D simulations through which if at least one element of the chargeability array was zero, the IP effects weren't taken into account when computing the complex conductivity matrix. Improve documentation, specifying that the Pelton-Cole equation that is implemented makes use of the conductivity at infinity frequency (and not zero frequency). Add tests for the bugfix.

PR #1745, opened 2026-01-19
DOC: Fix typo in docs Pelton/Cole-Cole

prisae:
Tiny typo in the docs of the Pelton (Cole-Cole) model (now it is also consistent with the actual code).


prisae:
Indeed, my bad @santisoler , didn't see this (would have saved me some time had I come across that PR ;-) ) - I see #1729 is very likely to be merged, so I close this one.

PR #1766, opened 2026-04-16
Nsem recursive add zyx

dccowan:
#### Summary
Adds data types corresponding to the Zyx impedance to and better testing of receiver types for Simulation1DRecursion

#### PR Checklist
      to a Pull Request

As of right now, any source type and orientation can be used in the survey for the Simulation1DRecursive class. No matter what is used, the simulation outputs data corresponding to the Zxy impedance. This is a problem orientation is set to xx, yx or yy. If orientation is set to xx or yy, an error should be thrown. If orientation is set to yx, the sign of the output data is wrong.

This PR adds the functionality required to properly output data associated with the Zyx impedance. More thorough validation of sources and receivers upon setting the survey object for the simulation are added.


simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -80,14 +84,33 @@ def survey(self):
...
+                if type(src) is not Planewave:
+                    raise NotImplementedError(
+                        (
+                            "Simulation1DRecursive defines sources using the "
+                            f"Planewave class, got {type(src)} instead."
+                        )
+                    )
                 for rx in src.receiver_list:
-                    if not isinstance(rx, Impedance):
+                    if type(rx) is not Impedance:
```
jcapriot:
```suggestion
                    if not isinstance(rx, Impedance):
```

I would leave this as is, the `type(rx)` check does not allow for inheritance of `Impedance` class receivers.

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -80,14 +84,33 @@ def survey(self):
...
+                            )
                         )
+                    if (rx.orientation != "xy") and (rx.orientation != "yx"):
+                        raise NotImplementedError(
+                            (
+                                "Simulation1DRecursive only allows 'xy' or 'yx' for the "
+                                f"orientation property of Impedance receivers, got {rx.orientation}."
+                            )
+                        )
+
```
jcapriot, line 113: I guess we could also technically support `xx` and `yy` components for completeness, but just fill it with zeros.

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -238,17 +256,25 @@ def dpred(self, m, f=None):
         for src in self.survey.source_list:
             i_freq = np.searchsorted(self.survey.frequencies, src.frequency)
             for rx in src.receiver_list:
+
+                if rx.orientation == "xy":
+                    pm = 1
+                elif rx.orientation == "yx":
+                    pm = -1
+
```
jcapriot, line 272:
```suggestion
                elif rx.orientation == "yx":
                    pm = -1
                else:
                    d.append(np.zeros(rx.nD))
                    continue
```

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -299,7 +331,7 @@ def getJ(self, m, f=None):
                     bot = real**2 + imag**2
                     d_real_dm = np.real(Js_row)
                     d_imag_dm = np.imag(Js_row)
-                    Jrows = C * (-imag / bot * d_real_dm + real / bot * d_imag_dm)
+                    Jrows = pm * C * (-imag / bot * d_real_dm + real / bot * d_imag_dm)
```
jcapriot:
```suggestion
                if rx.orientation in ["xx", "yy"]:
                    Jrows = np.zeros((rx.nD, Js.shape[1]))
                else:
                    if rx.orientation == "yx":
                        pm = -1
                    else:
                        pm = 1

                    if rx.component == "real":
                        Jrows = pm * np.real(Js_row)
                    elif rx.component == "imag":
                        Jrows = pm * np.imag(Js_row)
                    elif rx.component == "apparent_resistivity":
                        Jrows = (np.pi * src.frequency * mu_0) ** -1 * (
                                np.real(Z[i_freq]) * np.real(Js_row)
                                + np.imag(Z[i_freq]) * np.imag(Js_row)
                        )
                    elif rx.component == "phase":
                        C = 180 / np.pi
                        real = np.real(Z[i_freq])
                        imag = np.imag(Z[i_freq])
                        bot = real ** 2 + imag ** 2
                        d_real_dm = np.real(Js_row)
                        d_imag_dm = np.imag(Js_row)
                        Jrows = pm * C * (-imag / bot * d_real_dm + real / bot * d_imag_dm)
```

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -80,14 +84,33 @@ def survey(self):
 
     @survey.setter
     def survey(self, value):
+        """Define the survey."""
         if value is not None:
             value = validate_type("survey", value, Survey, cast=False)
             for src in value.source_list:
+                if type(src) is not Planewave:
```
jcapriot:
```suggestion
                if not isinstance(src, Planewave):
```

I know you're trying to limit it to just the `Planewave` class, but leaving it as an `isinstance` check is still better as you have no idea what end users have done. Maybe they have defined their own class that inherits from it? It does technically work just fine using our `PlanewaveXYPrimary` class, there's just unused information from that subclaass.

jcapriot (changes requested):
A few changes to support inheritance, but also why not just support `xx` and `yy` components for completeness sake!

dccowan:
> A few changes to support inheritance, but also why not just support `xx` and `yy` components for completeness sake!

Because I don't think it provides anything of value.
* If you have sufficient knowledge of NSEM methods, you would never bother to simulate it since you know the value is zero
* If you have insufficient knowledge, you might waste time trying to figure out why the simulated data are always zero
* This might encourage users to actually try and invert Zxx and Zyy data componenents.

While there is a certain satisfaction in completeness, I feel like in this case it would be detrimental.

simpeg/electromagnetics/natural_source/simulation_1d.py
dccowan: The recursive simulation class simulates only impedance data. So it should only allow impedance receivers. For now, there are no receivers that inherit the Impedance class so it doesn't exactly matter at the moment.

simpeg/electromagnetics/natural_source/simulation_1d.py
dccowan: > I know you're trying to limit it to just the `Planewave` class, but leaving it as an `isinstance` check is still better as you have no idea what end users have done. Maybe they have defined their own class that inherits from it? It does technically work just fine using our `PlanewaveXYPrimary` class, there's just unused information from that subclaass.

This is a very good point. My intent is to guide users towards using appropriate classes when performing NSEM simulations. Although I see the value in allowing inheritance so that users can create their own classes.

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -283,10 +309,16 @@ def getJ(self, m, f=None):
             i_freq = np.searchsorted(self.survey.frequencies, src.frequency)
             Js_row = Js[i_freq]
             for rx in src.receiver_list:
+
+                if rx.orientation == "yx":
+                    pm = -1
+                else:
+                    pm = 1
```
santisoler: Minor suggestion, just to mimic the same logic in `dpred`.

```suggestion
                if rx.orientation == "xy":
                    pm = 1
                elif rx.orientation == "yx" :
                    pm = -1
```

It's very unlikely that a receiver with another orientation will be in the survey due to the checks performed by the `survey` setter. But still, receivers can be "replaced" within the `receiver_list`.

With this logic, `pm` will not be defined if the orientation is not xy or yx and code will error out later on, instead of returning values corresponding to the xy orientation.

santisoler (approved):
Looks good to me! I just left a minor suggestion below.

Makes sense to leave the xx and yy not implemented. My only note on that would be to extend the error message (and probably add some admonition in the docstring of the simulation class) explaining why it's not implemented. Saying that for the 1D problem the diagonal elements are always zero should be enough.

simpeg/electromagnetics/natural_source/simulation_1d.py
jcapriot: It won’t always error out, ‘pm’ would only be undefined if the first receiver in the list is not xy/yx. If the first is okay, and the second is not, ‘pm’ will just have the value it had previously.

santisoler: True, it's inside the for loop! Sorry, my bad. In that case, I think we should actually raise an error if `pm` is not either xy or yx, just to avoid any potential unintended result that might be hard to track down.

santisoler:
```suggestion
                if rx.orientation == "xy":
                    pm = 1
                elif rx.orientation == "yx":
                    pm = -1
                else:
                    msg = (
                        f"Invalid receiver orientation '{rx.orientation}' "
                        f"for receiver '{rx}'. "
                        "Only 'xy' and 'yx' orientations are supported for "
                        f"the {type(self).__name__} class."
                    )
                    raise ValueError(msg)
```

simpeg/electromagnetics/natural_source/simulation_1d.py
```diff
@@ -238,17 +256,25 @@ def dpred(self, m, f=None):
         for src in self.survey.source_list:
             i_freq = np.searchsorted(self.survey.frequencies, src.frequency)
             for rx in src.receiver_list:
+
+                if rx.orientation == "xy":
+                    pm = 1
+                elif rx.orientation == "yx":
+                    pm = -1
```
santisoler, line 263:
```suggestion
                if rx.orientation == "xy":
                    pm = 1
                elif rx.orientation == "yx":
                    pm = -1
                else:
                    msg = (
                        f"Invalid receiver orientation '{rx.orientation}' "
                        f"for receiver '{rx}'. "
                        "Only 'xy' and 'yx' orientations are supported for "
                        f"the {type(self).__name__} class."
                    )
                    raise ValueError(msg)
```

santisoler (approved):
Updating my suggestions based on @jcapriot's comment regarding the `pm` variable.

jcapriot (approved):
LGTM

PR #1769, opened 2026-04-23
Stable padding for 1D MT solutions

dccowan:
#### Summary
This PR ensures better stability of the 1D solutions used to generate the RHSs of NSEM simulations.

#### PR Checklist
      to a Pull Request

In natural_source.utils.source_utils.py, we have two functions 'primary_e_1d_solution' and 'primary_h_1d_solution' which solve the 1D problem along the vertical discretization used for the 3D simulation. We add additional padding cells to the bottom to ensure the boundary conditions at the bottom are respected. There is a problem with the current implementation.

The extent of the padding for the 1D problem is determined by setting the number of padding cells. For a tree mesh, the base cell widths is used and the extent of the padding is reasonable compared to the scale of the problem. For a tensor mesh, we add an extreme number of very large cells. If the Dirichlet condition is set as the bottom BC, the analytic propagator matrix solution is used to solve the 1D problem. For a tensor mesh with the default number of padding cells, this solution is unstable.

It would be better to set the extend of the padding for the 1D problem based on the additional number of skin depths.
Since the Robin condition is more stable, it should be set as the default option for the bottom BC in both the e_1d and h_1d solution. This pull request implements this.


dccowan:
@santisoler Instead of setting the padding using a set number of padding cells, we are now setting it based on the number of additional skin depths to be added. So, we are making a change to the input arguments of the 1d field solution functions.

Do we want to allow the padding to be set with n_pad for the time being and throw a warning? Or can we just make the switch right away since the n_pad arg was only briefly available in a single simpeg patch release?

jcapriot (changes requested):
How about we make it so that both `n_skin_depths=None` and `n_pad=None`, but disallow setting both at the same time, then inside the functions you could do:

```python
if n_pad is None:
    n_skin_depths = 3 if n_skin_depths is None else n_skin_depths
    # do stuff to set the n_pad
    n_pad = ...
elif n_skin_depths is not None:
    raise TypeError("Cannot pass both `n_skin_depths` and `n_pad`")
```

Then update the `n_skin_depths` doc portion to say that its default is 3.0.

Another question: `n_skin_depths` is more like a multiplier considering it is described as a `float`, but `n_` portion would imply it is an integer. How about, `skin_depth_factor` or `extension_factor`?

dccowan:
> How about we make it so that both `n_skin_depths=None` and `n_pad=None`, but disallow setting both at the same time, then inside the functions you could do:
> 
> ```python
> if n_pad is None:
>     n_skin_depths = 3 if n_skin_depths is None else n_skin_depths
>     # do stuff to set the n_pad
>     n_pad = ...
> elif n_skin_depths is not None:
>     raise TypeError("Cannot pass both `n_skin_depths` and `n_pad`")
> ```
> 
> Then update the `n_skin_depths` doc portion to say that its default is 3.0.
> 
> Another question: `n_skin_depths` is more like a multiplier considering it is described as a `float`, but `n_` portion would imply it is an integer. How about, `skin_depth_factor` or `extension_factor`?

If `n_skin_depths` is confusing, I don't mind `skin_depth_factor`. Otherwise, I'm happy with the current implementation. I was under the impression we wanted to set default values when defining the function handle.

santisoler:
Glad to come back at this one!

Since these functions have not been included in the API Reference, I wouldn't mind introducing breaking changes to their signatures.

If we would like to keep both options (padding with number of cells, or with number of skin depths), I'd avoid setting them both to `None` and then defaulting to use one of them. If users read `n_skin_depths=None` and `n_pad=None` as default values, it's reasonable to expect that no padding is being applied.

In that sense I'd prefer having `n_skin_depths=3` and `n_pad=None` as defaults. And if anyone wants to use `n_pad` only, they'll have to explicitly pass `n_skin_depths=None`. So:

```python
# This will not apply any padding
primary_e_1d_solution(
    mesh, sigma_1d, freq, skin_depth_factor=None, n_pad=None,
)

# This will apply padding using 3 skin depths
primary_e_1d_solution(
    mesh, sigma_1d, freq, skin_depth_factor=3, n_pad=None,
)

# This will apply padding using 100 cells
primary_e_1d_solution(
    mesh, sigma_1d, freq, skin_depth_factor=None, n_pad=100,
)

# This will error out
primary_e_1d_solution(
    mesh, sigma_1d, freq, skin_depth_factor=3, n_pad=100,
)
```

Another idea would be to have a `padding_factor` argument for the number of either cells or skin depths, and then a `padding_strategy` that can be `"cells"` or `"skin_depths"`. Something like:

```python
def primary_e_1d_solution(
    mesh,
    sigma_1d,
    freq,
    top_bc="dirichlet",
    bot_bc="robin",
    padding_factor=3.0,
    padding_strategy="skin_depths",
):
```

What do you think?

jcapriot:
> Another idea would be to have a `padding_factor` argument for the number of either cells or skin depths, and then a `padding_strategy` that can be `"cells"` or `"skin_depths"`. Something like:
> 
> ```python
> def primary_e_1d_solution(
>     mesh,
>     sigma_1d,
>     freq,
>     top_bc="dirichlet",
>     bot_bc="robin",
>     padding_factor=3.0,
>     padding_strategy="skin_depths",
> ):
> ```
> 
> What do you think?

I like this set of inputs a lot, it's pretty clean and clear what is going on.

dccowan:
It's a utility function that users aren't expected to interact with directly. I think we should pick the optimum way to set the padding to keep things simple. I was fine with ``skin_depth_factor``. I don't necessarily like ``padding_factor`` because of how it is used in mesh creation.

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -7,12 +7,19 @@
...
 
 def primary_e_1d_solution(
-    mesh, sigma_1d, freq, top_bc="dirichlet", bot_bc="robin", n_pad=500
+    mesh,
+    sigma_1d,
+    freq,
+    top_bc="dirichlet",
+    bot_bc="robin",
+    skin_depth_factor=3,
+    n_pad=None,
```
jcapriot: For here and below, these functions just silently ignore any value passed to `n_pad`. It would be better to either 
- add a notice `if n_pad is not None` that the functionality has changed and they should instead use `skin_depth_factor`. 
- or, remove it completely so it errors.

I'm happy with either route to take, as long as the silently ignoring aspect is changed.

jcapriot (changes requested):

simpeg/electromagnetics/natural_source/utils/source_utils.py
dccowan: Good catch. I forgot to remove the kwarg from the function

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -7,12 +7,18 @@
...
 
 
 def primary_e_1d_solution(
-    mesh, sigma_1d, freq, top_bc="dirichlet", bot_bc="robin", n_pad=500
+    mesh,
+    sigma_1d,
+    freq,
+    top_bc="dirichlet",
+    bot_bc="robin",
+    skin_depth_factor=3,
```
santisoler: Just matching the type of the default value with the type in the description. This also helps type checkers to find out the expected type of the argument even if it's not type hinted.

```suggestion
    skin_depth_factor=3.0,
```

simpeg/electromagnetics/natural_source/utils/source_utils.py
```diff
@@ -168,7 +176,12 @@ def primary_e_1d_solution(
 
 
 def primary_h_1d_solution(
-    mesh, sigma_1d, freq, top_bc="dirichlet", bot_bc="dirichlet", n_pad=500
+    mesh,
+    sigma_1d,
+    freq,
+    top_bc="dirichlet",
+    bot_bc="robin",
+    skin_depth_factor=3,
```
santisoler: Same here: 

```suggestion
    skin_depth_factor=3.0,
```

simpeg/electromagnetics/natural_source/utils/test_utils.py
```diff
@@ -236,7 +236,7 @@ def setupSimpegNSEM_tests_location_assign_list(
     # discretize.TensorMesh.write_UBC(mesh,'Mesh-pre.msh', models={'Sigma-pre.dat': np.exp(model_true)})
     # create background conductivity model
     sigma_back = 1e-2
-    sigBG = np.zeros(mesh.nC) * sigma_back
+    sigBG = np.zeros(mesh.nC) + sigma_back
```
santisoler: I'd suggest using `np.full` for this, to avoid potential bugs like the one you spotted here. Also, `np.full` performs only two operations: allocate memory and fill the memory with the `fill_value`, instead of `np.zeros() + sigma_back` that allocates, fills with zeros and then adds values to the array.

```suggestion
    sigBG = np.full(mesh.nC, fill_value=sigma_back)
```

simpeg/electromagnetics/natural_source/utils/test_utils.py
```diff
@@ -357,7 +357,7 @@ def setupSimpegNSEM_PrimarySecondary(inputSetup, freqs, comp="Imp", singleFreq=F
...
```
santisoler: Same here:

```suggestion
    sigBG = np.full(mesh.nC, fill_value=sigma_back)
```

santisoler (approved):
Looking good to me! I left a few minor suggestions below.

Before merging, I would rephrase the PR name to something more descriptive, like: "Pad mesh in NSEM 1D solutions by a factor of skin depth". In the description we could explain that this improves stability of the solutions regardless of the type of mesh we are using for the same default value. Feel free to tweak my words!

jcapriot (approved):
LGTM

PR #1770, opened 2026-06-01
TileMap --> use the discretize volume averaging

lheagy:
#### Summary
use the discretize volume averaging in the tile map so that it is robust to the scenario where the local mesh is finer than the global one

#### PR Checklist
      to a Pull Request

Right now, if you ended up with a local mesh that had a higher level of refinement than the global mesh, the tile map will run, but produce nonsensical results. If we instead use the discretize volume averaging, it is more stable (test added to show this)

### DIFF
--- a/simpeg/maps/_base.py
+++ b/simpeg/maps/_base.py
@@ -8,7 +8,7 @@ import numpy as np
 import scipy.sparse as sp
 from scipy.sparse import csr_matrix as csr
 from discretize.tests import check_derivative
-from discretize.utils import Zero, Identity, mkvc, speye, sdiag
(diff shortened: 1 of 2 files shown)


lheagy:
@simpeg/simpeg-developers: this is ready fir review whenever you have a moment. Thanks!

jcapriot:
I’ll take a closer look at this soon, but just dropping a note that some of the simulations getJtJdiag functions rely on maps being surjective.

jcapriot (approved):
Looks good to me, I know the `volume_average` function is more robust (and therefore slightly slower in general) but it does also have the faster shortcut when the two meshes have the same underlying mesh (which is what was implicitly assumed before).

PR #1775, opened 2026-08-21
Transfer functions

jcapriot:
#### Summary
Adds many transfer functions I've been using in some of my own experiments regarding level set inversions. Basically these are all slightly ways of limiting the values, just with all slightly different shapes.

#### PR Checklist
      to a Pull Request

#### Additional information
In general, I'm referring to transfer functions as mapping the full number line from -inf to inf, to be between two numbers in a smooth way (think the arctan function). These functions allow a bit more control over some of the previous in terms of how *fast* they transfer between the extreme values. I found the natural way for me to think about this was by using the "half-width" of the derivative. I.e. the width of the derivative at half it's maximum amplitude. They also generally allow you to set a `deriv_half_width` parameter if you want to separately control the derivative and the transformation, (e.g. the mapping itself is very sharp, but you want to use a "smoothed" version of it when calculating derivatives).

### DIFF
--- a/simpeg/__init__.py
+++ b/simpeg/__init__.py
@@ -68,10 +68,12 @@ Mappings
 .. autosummary::
   :toctree: generated/
 
+  maps.ArctanMap
   maps.BaseParametric
   maps.ChiMap
   maps.ComboMap
   maps.ComplexMap
+  maps.ErfMap
   maps.ExpMap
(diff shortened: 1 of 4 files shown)


jcapriot:
I don't know why codecov is showing the function definition lines as uncovered yet there bodies are covered.

PR #1776, opened 2026-08-22
Enhance robustness and error handling in multiprocessing simulations

jcapriot:
#### Summary
Refactors inter-process communication with an enum for clarity and safety. Corrects chunking in `MultiprocessingSumMetaSimulation` to use `SumMetaSimulation` for proper accumulation. Adds comprehensive worker error propagation and prevents `join()` deadlocks by draining unread results. Ensures graceful shutdown of worker processes with `atexit` fallback cleanup.

#### PR Checklist
      to a Pull Request

We've always had a few hanging errors related to the multiprocessing meta simulation that I was never able to identify on my own. I took the opportunity to let Claude Code analyze it to look for possible deadlocks and race conditions.

#### Additional information
I've also replaced the (easy to get wrong) communication messages between the main process and the child processes with enums instead of specific values.

### DIFF
--- a/simpeg/meta/multiprocessing.py
+++ b/simpeg/meta/multiprocessing.py
@@ -1,8 +1,32 @@
+import atexit
+import threading
+from enum import Enum, auto
+from queue import Empty
 from multiprocessing import Process, Queue, cpu_count
 from simpeg.meta import MetaSimulation, SumMetaSimulation, RepeatedSimulation
 from simpeg.props import HasModel
 import uuid
 import numpy as np
+
+
+class _Op(Enum):
+    """The set of operations a `_SimulationProcess` understands.
+
+    Sent as the first element of each task tuple put on a process's
+    `task_queue`; members pickle by reference (name), so this only needs
+    to be importable from the same module path in both processes, which
+    it is since a spawned child re-imports this module.
+    """
+
+    SET_SIM = auto()
+    GET_ITEM = auto()
+    DEL_ITEM = auto()
+    STORE_MODEL = auto()
+    CREATE_FIELDS = auto()
+    DPRED = auto()
+    JVEC = auto()
+    JTVEC = auto()
+    JTJ_DIAG = auto()
 
 
 class SimpleFuture:
@@ -17,7 +41,7 @@ class SimpleFuture:
     # classes stash the simulation, so this requires serializing
     # the simulation (something we explicitly want to avoid in all cases).
     # def result(self):
-    #     self.t_queue.put(("get_item", (self.item_id,)))
+    #     self.t_queue.put((_Op.GET_ITEM, (self.item_id,)))
     #     item = self.r_queue.get()
     #     if isinstance(item, Exception):
     #         raise item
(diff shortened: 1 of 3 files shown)


jcapriot:
The uncovered lines are either un-reportable lines (the ones inside the `while True` of the `_SimulationProcess`), or items that are called at the end of the session (after coverage finalizes).

simpeg/meta/multiprocessing.py
```diff
@@ -241,22 +331,40 @@ def __init__(self, simulations, mappings, n_processes=None):
...
+                i_end = i_start + chunk
+                sim_chunk = self._chunk_sim_class(
+                    self.simulations[i_start:i_end], self.mappings[i_start:i_end]
+                )
+                chunk_nd.append(sim_chunk.survey.nD)
+                p = _SimulationProcess()
+                processes.append(p)
+                p.start()
+                p.set_sim(sim_chunk)
+                i_start = i_end
```
santisoler: Leaving a suggestion to move the chunk code into a generator function to avoid code repetition and micro management of the chunking operation:

```python
def split_in_chunks(*args, n_chunks: int):
    """
    Split a sequence or sequences into chunks of approximately equal size.

    If the number of elements is not multiple of ``n_chunks``, the first chunks will
    contain one extra element than the last ones.

    Parameters
    ----------
    *args : Sequence
        Sequences of equal amount of elements.
    n_chunks : int
        Number of chunks to split the sequences in.
        If ``n_chunks`` is larger than the total number of elements in the sequences,
        each chunk will contain a single element.

    Yields
    ------
    chunks : Sequence
        Chunk(s) of the passed sequence(s).

    Raises
    ------
    ValueError :
        If the sequences passed through ``*args`` don't have the same length.
    """
    # Get number of elements in sequences and check if they are all equal.
    n_elements = {len(sequence) for sequence in args}
    if len(n_elements) != 1:
        msg = (
            "Invalid sequences with different number of elements. "
            "All sequences should have the same lenght."
        )
        raise ValueError(msg)
    (n_elements,) = n_elements

    # Clip n_chunks to the minimum value between it and n_elements, so if n_chunks is
    # larger than n_elements, each chunk will contain a single element.
    n_chunks = min(n_chunks, n_elements)

    # Estimate the size of each chunk.
    chunk_size = n_elements // n_chunks

    # Get number of elements that remain when dividing the seqs in chunks of equal size
    offset = n_elements % chunk_size

    # Yield chunks.
    index = 0
    while index < n_elements:
        # Get the size of this particular chunk. If there's still an offset, add one.
        _chunk_size = chunk_size if offset == 0 else chunk_size + 1
        chunk = tuple(sequence[index : index + _chunk_size] for sequence in args)
        yield _unpack_sequence(chunk)
        # Increase index
        index += _chunk_size
        # Lower offset by one
        if offset > 0:
            offset -= 1


def _unpack_sequence(sequence):
    """
    Unpack a sequence if it contains a single element.
    """
    if len(sequence) == 1:
        (sequence,) = sequence
        return sequence
    return sequence
```

So, for example, these bits could be replaced for something like:

```python
chunks = split_in_chunks(self.simulations, self.mappings, n_chunks=n_processes)
try:
    for simulations_chunk, mappings_chunk in chunks:
        sim_chunk = self._chunk_sim_class(simulations_chunk, mappings_chunk)
        chunk_nd.append(sim_chunk.survey.nD)
        p = _SimulationProcess()
        processes.append(p)
        p.start()
        p.set_sim(sim_chunk)
```

jcapriot: I factored out the common logic here, making use of a closure for the internal per-sim logic that is different between the two.

PR #1777, opened 2026-08-23
Make `simpeg.utils.download` respect the `verbose` argument

### LINKED ISSUE #629: SimPEG.Utils.download "verbose" switch does not control all the output
I am using SimPEG.Utils.download for gpgLabs. The download messages are distracting in notebooks, so I wanted to suppress them. The option is controlled by "verbose=false", but it does not stop writing "Downloading ... Saved to:.." message.
Can verbose=false swtich kill all the print output?

image

manduinca:
#### Summary
`simpeg.utils.download` printed its progress messages regardless of the `verbose` argument.

#### PR Checklist
* [x] Linted my code according to the style guides.
* [x] Added tests to verify changes to the code.

#### Reference issue
Closes #629

Only the messages about preexisting files were guarded by `verbose`. The "Downloading ...", "   saved to: ..." and "Download completed!" messages were printed unconditionally, so `verbose=False` did not silence the function. I put the three remaining prints behind the same check.

#### Additional information
The new tests mock `urllib.request.urlretrieve` so they don't hit the network, and use `capsys` to check the output. They live in a separate pytest class because the existing `TestDownload` is a `unittest.TestCase`, which can't take the `capsys` fixture. Both fail on main and pass with the change.

I kept the `verbose` flag instead of moving these messages to the SimPEG logger, to keep the change scoped to the issue. Happy to do that instead if you prefer.

### Squash and Merge Summary

Make `simpeg.utils.download` to only print out messages to stdout if `verbose` is True. Test the bugfix by mocking `urllib` to avoid requiring network connection.

### DIFF
--- a/simpeg/utils/io_utils/io_utils_general.py
+++ b/simpeg/utils/io_utils/io_utils_general.py
@@ -146,10 +146,10 @@ def download(url, folder=".", overwrite=False, verbose=True):
     urllist = url if isinstance(url, list) else [url]
     for u, f in zip(urllist, downloadpath):
         if verbose:
-            print("Downloading {}".format(u))
+            print(f"Downloading {u}")
         urlretrieve(u, f)
         if verbose:
-            print("   saved to: " + f)
+            print(f"   saved to: {f}")
 
     if verbose:
         print("Download completed!")
--- a/simpeg/utils/io_utils/io_utils_general.py
+++ b/simpeg/utils/io_utils/io_utils_general.py
@@ -145,9 +145,12 @@ def download(url, folder=".", overwrite=False, verbose=True):
     # download files
     urllist = url if isinstance(url, list) else [url]
     for u, f in zip(urllist, downloadpath):
-        print("Downloading {}".format(u))
+        if verbose:
+            print("Downloading {}".format(u))
         urlretrieve(u, f)
-        print("   saved to: " + f)
+        if verbose:
+            print("   saved to: " + f)
 
-    print("Download completed!")
+    if verbose:
+        print("Download completed!")
     return downloadpath if isinstance(url, list) else downloadpath[0]
--- a/tests/base/test_utils.py
+++ b/tests/base/test_utils.py
@@ -4,6 +4,7 @@ import numpy as np
 import scipy.sparse as sp
 import os
 import shutil
+import urllib.request
 from simpeg.utils import (
     sdiag,
     sub2ind,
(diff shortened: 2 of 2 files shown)


simpeg/utils/io_utils/io_utils_general.py
```diff
@@ -145,9 +145,12 @@ def rename_path(downloadpath):
     # download files
     urllist = url if isinstance(url, list) else [url]
     for u, f in zip(urllist, downloadpath):
-        print("Downloading {}".format(u))
+        if verbose:
+            print("Downloading {}".format(u))
```
santisoler: Since we are changing this line, let's use f-strings instead:
```suggestion
            print(f"Downloading {u}")
```

simpeg/utils/io_utils/io_utils_general.py
```diff
@@ -145,9 +145,12 @@ def rename_path(downloadpath):
...
         urlretrieve(u, f)
-        print("   saved to: " + f)
+        if verbose:
+            print("   saved to: " + f)
```
santisoler: Same with this one:
```suggestion
            print(f"   saved to: {f}")
```

tests/base/test_utils.py
```diff
@@ -320,6 +321,38 @@ def test_downloads(self):
...
 
 
+class TestDownloadVerbose:
+    """
+    Tests for the verbose argument of the download function
+    """
+
+    URL = "https://example.com/simpeg/data.txt"
+
+    @pytest.fixture(autouse=True)
```
santisoler: I think I would prefer using this fixture explicitly, just to ensure we are not using a mocked `urlretrieve` if it's not intended (nothing wrong here, just adding a safeguard for the future):
```suggestion
    @pytest.fixture
```

tests/base/test_utils.py
```diff
@@ -320,6 +321,38 @@ def test_downloads(self):
...
+    def mock_urlretrieve(self, monkeypatch):
+        def urlretrieve(url, filename):
+            with open(filename, "w") as fid:
+                fid.write("data")
+
+        monkeypatch.setattr(urllib.request, "urlretrieve", urlretrieve)
+
+    def test_silent(self, tmp_path, capsys):
```
santisoler: In line with the previous suggestion:
```suggestion
    def test_silent(self, tmp_path, capsys, mock_urlretrieve):
```

tests/base/test_utils.py
```diff
@@ -320,6 +321,38 @@ def test_downloads(self):
...
+        download(self.URL, folder=str(tmp_path), verbose=False)
+        assert capsys.readouterr().out == ""
+
+    def test_silent_on_existing_file(self, tmp_path, capsys):
```
santisoler:
```suggestion
    def test_silent_on_existing_file(self, tmp_path, capsys, mock_urlretrieve):
```

tests/base/test_utils.py
```diff
@@ -320,6 +321,38 @@ def test_downloads(self):
...
+        download(self.URL, folder=str(tmp_path), overwrite=True, verbose=False)
+        download(self.URL, folder=str(tmp_path), overwrite=True, verbose=False)
+        assert capsys.readouterr().out == ""
+
+    def test_verbose(self, tmp_path, capsys):
```
santisoler:
```suggestion
    def test_verbose(self, tmp_path, capsys, mock_urlretrieve):
```

santisoler (changes requested):
Thanks @manduinca a lot for taking the time to work on this PR. I left a few comments below. Let me know what do you think!

manduinca:
Thanks for the review @santisoler! I applied all your suggestions: f-strings in the two prints and the explicit `mock_urlretrieve` fixture in the three tests. I ran `tests/base/test_utils.py` locally (29 passed) and black/flake8 come back clean.

santisoler:
Thanks a lot @manduinca! I'm merging this! 🚀

santisoler (approved):

PR #1802, opened 2026-09-10
Drop support for Python 3.11

santisoler:
#### Summary

Since Python 3.11 will reach its EOL on October 2026, we are dropping support for it. Require Python 3.12 as the minimum compatible version for SimPEG. Update Python version in `environment.yml` and CI configurations and scripts. Add entry to version compatibility table: list SimPEG v0.25.2 as the last version compatible with Python 3.11. Go back to using `python3` as Python version in `pre-commit` config.

#### PR Checklist
      to a Pull Request

#### Reference issue

In this PR I'm restoring the `pre-commit` configuration to use any Python version available. Such change was introduced in c12b527d21d1b11d515531f60990d86384fe7b3e, but explicit Python versions were restored in c37536ce0757ea9dfe38ec470f6ea2d79ba2a297.

### DIFF
--- a/.ci/azure/docs.yml
+++ b/.ci/azure/docs.yml
@@ -5,7 +5,7 @@ jobs:
     pool:
       vmImage: ubuntu-latest
     variables:
-      python.version: "3.11"
+      python.version: "3.12"
     timeoutInMinutes: 240
     steps:
       # Checkout simpeg repo.
--- a/.ci/azure/pypi.yml
+++ b/.ci/azure/pypi.yml
@@ -12,7 +12,7 @@ jobs:
 
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.11"
+          versionSpec: "3.12"
         displayName: "Setup Python"
 
       - bash: |
@@ -57,7 +57,7 @@ jobs:
 
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.11"
+          versionSpec: "3.12"
         displayName: "Setup Python"
 
       - bash: |
--- a/.ci/azure/style.yml
+++ b/.ci/azure/style.yml
@@ -6,7 +6,7 @@ jobs:
     steps:
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.11"
+          versionSpec: "3.12"
       - bash: .ci/install_style.sh
         displayName: "Install dependencies to run the checks"
       - script: make black
@@ -19,7 +19,7 @@ jobs:
     steps:
       - task: UsePythonVersion@0
         inputs:
-          versionSpec: "3.11"
+          versionSpec: "3.12"
       - bash: .ci/install_style.sh
         displayName: "Install dependencies to run the checks"
       - script: make flake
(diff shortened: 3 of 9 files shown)


simpeg/regularization/cross_gradient.py
```diff
@@ -220,12 +220,13 @@ def calculate_cross_gradient(self, model, normalized=False, rtol=1e-6):
...
+        # Expand the vectors with a z coordinate equal to zero if they are 2D
+        if self.regularization_mesh.dim == 2:
+            grad_m1, grad_m2 = _expand_2d_vector(grad_m1), _expand_2d_vector(grad_m2)
+
         # for each model cell, compute the cross product of the gradient vectors.
         cross_prod = np.cross(grad_m1, grad_m2)
-        if self.regularization_mesh.dim == 3:
-            cross_prod = np.linalg.norm(cross_prod, axis=-1)
-
-        return cross_prod
```
santisoler, line 228: I suspect there was actually a bug here... If the problem is 2D, then the returned values would just be the third coordinate(s) of the cross product between the two gradients, but not its absolute values. This could lead to negative values returned by this method.

The patch would fix that, since we are always computing the norm of the cross products (for 2D and 3D vectors).

jcapriot: We just need two separate pathways, one for 2D that does:
```
if self.regularization_mesh.dim == 2:
    return np.abs(grad_m1[..., 0] * grad_m2[..., 1] - grad_m1[..., 1] * grad_m2[..., 0])
else: # or elif
   cross_prod = np.cross(grad_m1, grad_m2)
   return np.linalg.norm(cross_prod, axis=-1)
```

santisoler: Update: the new patch will still fix that, but because it will return the `np.abs(...)` of the cross product between the 2D vectors (only the third coordinate of the cross product vector).

santisoler: > We just need two separate pathways, one for 2D that does:
> 
> ```
> if self.regularization_mesh.dim == 2:
>     return np.abs(grad_m1[..., 0] * grad_m2[..., 1] - grad_m1[..., 1] * grad_m2[..., 0])
> else: # or elif
>    cross_prod = np.cross(grad_m1, grad_m2)
>    return np.linalg.norm(cross_prod, axis=-1)
> ```

Yes, just saw this! I ended up doing that, but with a private function and some checks (plus tests).

simpeg/regularization/cross_gradient.py
```diff
@@ -391,3 +393,64 @@ def deriv2(self, model, v=None):
             return (
                 2 * np.r_[p1, p2]
             )  # factor of 2 from derviative of | grad m1 x grad m2 | ^2
+
+
+def _cross_product_2d(a, b):
```
jcapriot, line 398: It would be worth an audit to see where else we might be using it. If I remember correctly, `geoana` does something similar in a few places. But this type of thing could be useful as a utility on its own.

santisoler: Sounds good. I checked within SimPEG and I found that `np.cross` is only used in the `define_plane_from_points` function. But it works with 3D vectors, so it's fine.

If geoana needs such function we could move it there and use it from SimPEG when needed.

In the meantime, I'm inclined to merge this PR for now, what do you think?

